PR #1043, opened 2016-07-25
DDF-2313 SourceConfigurationHandler doesn't handle multiple bindings with same fpid but different bindingTypes

gordocanchola:
#### What does this PR do?
Updating SourceConfigurationHandler to update ID information for stale configurations. Updating to handle the case where the same factory pid may have multiple binding types.
@mcalcote @clockard @ryeats 
@figliold 
@shaundmorris 
Verify the build
DDF-2313
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

Fixing SourceConfigurationHandler to account for registry entry with multiple bindings with the same fpid but different bindingTypes
Fixing SourceConfigurationHandler to update ID for stale configuration.
Updating tests

### DIFF
--- a/catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
+++ b/catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
@@ -218,7 +218,7 @@ public class SourceConfigurationHandler implements EventHandler {
             bindingTypeToActivate = getBindingTypeToActivate(bindingTypes);
             if (StringUtils.isNotEmpty(bindingTypeToActivate)) {
                 String fPidToActivate = bindingTypeToFactoryPidMap.get(bindingTypeToActivate);
-                activateDeactivateExistingConfiguration(fPidToActivate, fpidToConfigurationMap);
+                activateDeactivateExistingConfiguration(fPidToActivate, fpidToConfigurationMap, bindingTypeToActivate);
             }
         }
 
@@ -236,10 +236,11 @@ public class SourceConfigurationHandler implements EventHandler {
             }
             Configuration curConfig = findOrCreateConfig(factoryPid,
                     fpidToConfigurationMap,
+                    factoryPidMask,
                     (autoActivateConfigurations && factoryPidMask.equals(bindingTypeToActivate)));
 
             Hashtable<String, Object> serviceConfigurationProperties = new Hashtable<>();
-            if (fpidToConfigurationMap.containsKey(curConfig.getFactoryPid())) {
+            if (fpidToConfigurationMap.containsKey(curConfig.getFactoryPid() + curConfig.getProperties().get(BINDING_TYPE))) {
                 serviceConfigurationProperties.putAll(getConfigurationsFromDictionary(curConfig.getProperties()));
             } else {
                 serviceConfigurationProperties.putAll(getMetatypeDefaults(factoryPid));
@@ -251,7 +252,23 @@ public class SourceConfigurationHandler implements EventHandler {
             serviceConfigurationProperties.put(RegistryObjectMetacardType.REGISTRY_ID, registryId);
 
             curConfig.update(serviceConfigurationProperties);
+            fpidToConfigurationMap.remove(curConfig.getFactoryPid().concat(factoryPidMask));
         }
+
+        //if a binding was removed the configuration could still exist, If the registry
+        //entry gets renamed the non-binding configs will not get renamed appropriately so we
+        //go through them here and make sure the name is set correctly.
+        fpidToConfigurationMap.values().stream().forEach(config -> {
+            try {
+                Dictionary<String, Object> properties = config.getProperties();
+                properties.put(ID, configId);
+                properties.put(SHORTNAME, configId);
+                config.update(properties);
+            } catch (IOException e) {
+                LOGGER.error("Could not remove configuration for {}:{}",
+                        config.getProperties().get(ID), config.getFactoryPid());
+            }
+        });
 
     }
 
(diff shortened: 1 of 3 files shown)


catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
```diff
@@ -297,15 +315,18 @@ private Configuration findOrCreateConfig(String factoryPid,
      * @throws IOException
      */
     private void activateDeactivateExistingConfiguration(String fPidToActivate,
-            Map<String, Configuration> fpidToConfigurationMap) throws IOException {
+            Map<String, Configuration> fpidToConfigurationMap, String bindingType) throws IOException {
         String fpid;
+        String fpidToActivateWithBinding = fPidToActivate + bindingType;
+        String disabledFpidToActivateWithBinding =
+                fPidToActivate.concat(DISABLED_CONFIGURATION_SUFFIX) + bindingType;
```
shaundmorris: Since you are already using `concat()` you might as well use it again for `bindingType` or use `String.format`

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
```diff
@@ -315,14 +336,22 @@ private void activateDeactivateExistingConfiguration(String fPidToActivate,
         //Order of disable/enable important. Can't have two enabled configurations with the same
         //id so if there is one to disable, disable it before enabling another one.
         if (configToDisable != null) {
-            fpidToConfigurationMap.remove(configToDisable.getFactoryPid());
+            fpidToConfigurationMap
+                    .remove(configToDisable.getFactoryPid() + configToDisable.getProperties()
```
shaundmorris: You may be able to reduce the String `+` here with `format` as well

shaundmorris:
👍

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
gordocanchola: Changed to use concat again.

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
gordocanchola: I'll leave it with the + concat, if that's cool. It does the automatic type conversion for the getProperties property (which is why I'm not using concat here)  and is more clear than the String.format. String.format doesn't bring any performance boost or anything.

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
```diff
@@ -236,10 +236,11 @@ private synchronized void updateRegistryConfigurations(Metacard metacard, boolea
             }
             Configuration curConfig = findOrCreateConfig(factoryPid,
                     fpidToConfigurationMap,
+                    factoryPidMask,
                     (autoActivateConfigurations && factoryPidMask.equals(bindingTypeToActivate)));
 
             Hashtable<String, Object> serviceConfigurationProperties = new Hashtable<>();
-            if (fpidToConfigurationMap.containsKey(curConfig.getFactoryPid())) {
+            if (fpidToConfigurationMap.containsKey(curConfig.getFactoryPid() + curConfig.getProperties().get(BINDING_TYPE))) {
```
figliold: Will this always have a bindingType property?

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
```diff
@@ -312,14 +333,22 @@ private void activateDeactivateExistingConfiguration(String fPidToActivate,
         //Order of disable/enable important. Can't have two enabled configurations with the same
         //id so if there is one to disable, disable it before enabling another one.
         if (configToDisable != null) {
-            fpidToConfigurationMap.remove(configToDisable.getFactoryPid());
+            fpidToConfigurationMap
+                    .remove(configToDisable.getFactoryPid() + configToDisable.getProperties()
+                            .get(BINDING_TYPE));
```
figliold: `configToDisable.getProperties().get(BINDING_TYPE))` should never return `null`...right?

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
gordocanchola: There will not always be a bindingType property. In the case where there isn't a bindingType, the String "null" will be appended to the factoryPid. These values aren't outward facing, just used as the key for the fpidToConfigurationMap.

catalog/spatial/registry/registry-source-configuration-handler/src/main/java/org/codice/ddf/registry/source/configuration/SourceConfigurationHandler.java
gordocanchola: Config.getProperties should never return null. The get(BINDING_TYPE) can return null if the bindingType does not exist. In that case it will append the String "null" to the factoryPid as the key.

mcalcote:
Hero build passed 👍

shaundmorris:
@gordocanchola also submit a backport PR for 2.9.x

PR #1053, opened 2016-07-26
DDF-2251 Refactor Registry Report Builder to be able to generate 'summaries' of parts of the overall report

ani6gup:
#### What does this PR do?

In general, the registry report now utilizes the maps generated from the `registry-schema-bindings` and through `RegistryReportBuilder`, the values in the maps populate a string of html that is returned and compiled through`RegistryRestEndpoint`.  

The new reports can be accessed through the endpoints: `.../services/registries/{registryID}/report` , `.../report/summary` , and `.../report/organizations` . Further description is presented in background context. 

@gordocanchola @clockard @mcalcote @vinamartin @brianfelix 

@pklinef @shaundmorris @figliold

Start DDF Registry 
Go to DDF Catalog 
Click the Identity Node - Must add a service, add attributes to the identity node 
Go to sources tab , click corresponding identity node 
Click Registry Source Report 
Ensure that all fields you entered have been filled 
For Summary and Organizations Report, insert the corresponding URL (/report/summary and /report/organizations) and add Organization attributes 

There are three different types of html strings that can be generated which are determined based on the endpoints—two new endpoints (summary and section) were added in `RegistryRestEndpoint`. Descriptions of each endpoint are below: 

1.  `/report`: Utilizes the `report.hbt` handlebar template where all entities (Node Info, Services, Organizations, Contacts, and Collections) of the registry node are displayed based on the map values generated from the class in the schema-bindings `RegistryObjectListConverter`

2. `/report/summary`:  Utilizes the `summary.hbt` handlebar template where one the entities of an organization (Point of Contact, Name, Address, Email, Phone Number) is displayed based on the map values extracted from the metacard. 

3. `/report/{section}`: {section} gives freedom to extract certain sections of the report, however specific handlebar templates and additional methods must be made to accommodate for this. Currently Organizations has been extracted from the report.  

- `/report/organizations`: Utilizes the `organizations.hbt` handlebar template where all of the organizations and their entities (ids, Addresses, Emails, Phone Numbers. etc. ) are  displayed based on the map values generated from the class in the schema-bindings `OrganizationWebConverter`

[DDF-2251](https://codice.atlassian.net/browse/DDF-2251)
[DDF-2336 ](https://codice.atlassian.net/browse/DDF-2336)(blocked by this ticket, will provide an action provider to allow the summary report to be shown directly on the UI)
[DDF-2234](https://codice.atlassian.net/browse/DDF-2334) (Shows null values in summary report when certain optional fields are not filled in) 

**Full Report** 

screen shot 2016-07-26 at 1 56 46 pm

**Summary Report** 

screen shot 2016-07-26 at 1 56 29 pm

When null values are present, should be fixed with DDF-2235: 

screen shot 2016-07-26 at 2 17 21 pm

**Organizations Report** 

screen shot 2016-07-26 at 1 56 15 pm

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests


catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/HandlebarsHelper.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.codice.ddf.registry.rest.endpoint;
+
+import java.io.IOException;
+
+import com.github.jknack.handlebars.Options;
+
+public class HandlebarsHelper {
+
+    public CharSequence ifeq(String value, Options options) throws IOException {
```
mcalcote: Why was this method put into its own class?

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
```diff
@@ -0,0 +1,169 @@
...
+
+        h3 {
+            margin-left: 25px
+        }
+
+        h4 {
+            margin-left: 50px
+        }
+
+
```
mcalcote: Could remove some of these newlines.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+        return summaryMap;
+    }
+
+    private Map<String, Object> getFullRegistryMap(RegistryPackageType registryPackage) {
+        Map<String, Object> registryMap;
+        registryObjectListWebConverter = new RegistryObjectListWebConverter();
+        registryMap =
+                registryObjectListWebConverter.convert(registryPackage.getRegistryObjectList());
+
+        return registryMap;
```
mcalcote: Can avoid instantiation and initialization if you just return registryObjectListWebConverter.convert(registryPackage.getRegistryObjectList())

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+        } catch (NullPointerException e) {
+            LOGGER.debug("Null value not put into summary map" + e.toString());
+        }
+
+        return summaryMap;
+    }
+
+    private Map<String, Object> getFullRegistryMap(RegistryPackageType registryPackage) {
+        Map<String, Object> registryMap;
+        registryObjectListWebConverter = new RegistryObjectListWebConverter();
```
mcalcote: This is a private global variable that only seems to be used here (and initialized twice). Maybe move the declaration into this private method.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+ * More specifically within this class, maps are generated through the classes contained
+ * in registry-schema-bindings.The details contained in the maps populate a specific handlebar
+ * template and result in a String html that is relayed to RegistryRestEndpoint.
+ */
+public class RegistryReportBuilder {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RegistryReportBuilder.class);
+
+    private RegistryObjectListWebConverter registryObjectListWebConverter =
+            new RegistryObjectListWebConverter();
```
mcalcote: Does this need to be global?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public String getHTMLFromMetacard(Metacard metacard, ClassPathTemplateLoader templateLoader,
+            String handlebarTemplate) throws IOException {
+
+        Map<String, Object> reportMap = getSummaryMap(metacard);
+        Handlebars handlebars = getHandlebarsFromTemplate(templateLoader);
+        Template template = handlebars.compile(handlebarTemplate);
+        String html = template.apply(reportMap);
+        html = html.replaceAll("\n", "");
```
mcalcote: Did you still need to pull out all the newlines?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+        return html;
+    }
+
+    public String getHTMLFromRegistryPackage(RegistryPackageType registryPackage,
+            ClassPathTemplateLoader templateLoader, String handlebarTemplate) throws IOException {
+
+        Map<String, Object> reportMap = new HashMap<>();
+        if (handlebarTemplate.equals("organizations")) {
+            reportMap = getOrganizationMap(registryPackage);
+        } else if (handlebarTemplate.equals("report")) {
```
mcalcote: I feel like you could make "report", "summary" and "organizations" public static final strings in this class and then use RegistryRestEndpoint.REPORT (etc) in the method call(s), but not sure how much better that is.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+                    null);
+
+            summaryMap.put("Point of Contact", pointOfContact);
+            summaryMap.put("Organization Name", organizationName);
+            summaryMap.put("Address", organizationAddress);
+            summaryMap.put("Email Address", organizationEmail);
+            summaryMap.put("Phone Number", organizationPhoneNumber);
+        } catch (NullPointerException e) {
+            LOGGER.debug("Null value not put into summary map" + e.toString());
+        }
```
mcalcote: Maybe there is another way we can handle the missing values here, perhaps by making the default value "not provided" or something. Then you would avoid NPE and partially formed summaryMaps altogether.

Too bad you couldn't just use the constants as the labels also... then you could just stream a list of desired fields instead.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+        if (handlebarTemplate.equals("organizations")) {
+            reportMap = getOrganizationMap(registryPackage);
+        } else if (handlebarTemplate.equals("report")) {
+            reportMap = getFullRegistryMap(registryPackage);
+        }
+        Handlebars handlebars = getHandlebarsFromTemplate(templateLoader);
+        Template template = handlebars.compile(handlebarTemplate);
+        String html = template.apply(reportMap);
+
+        return html;
```
mcalcote: return template.apply(reportMap);

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+        return registryMap;
+    }
+
+    private Map<String, Object> getOrganizationMap(RegistryPackageType registryPackage) {
+        Map<String, Object> organizationMap = new HashMap<>();
+
+        List<Map<String, Object>> organizations = new ArrayList<>();
+        RegistryPackageTypeHelper rpt = new RegistryPackageTypeHelper(registryPackage);
+        OrganizationWebConverter owc = new OrganizationWebConverter();
+        for (OrganizationType org : rpt.getOrganizations()) {
```
mcalcote: organizations.addAll(rpt.getOrganizations()
                .stream()
                .map(owc::convert)
                .collect(Collectors.toList()));

catalog/spatial/registry/registry-rest-endpoint/pom.xml
```diff
@@ -114,22 +124,22 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.90</minimum>
+                                            <minimum>0.98</minimum>
```
mcalcote: NICE!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/HandlebarsHelper.java
```diff
@@ -0,0 +1,31 @@
...
+        if (value == null || options.param(0) == null) {
```
stustison: Shouldn't null == null be true as well?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,150 @@
...
+            String organizationPhoneNumber = RegistryUtility.getStringAttribute(metacard,
+                    RegistryObjectMetacardType.ORGANIZATION_PHONE_NUMBER,
+                    null);
+
+            summaryMap.put("Point of Contact", pointOfContact);
+            summaryMap.put("Organization Name", organizationName);
+            summaryMap.put("Address", organizationAddress);
+            summaryMap.put("Email Address", organizationEmail);
+            summaryMap.put("Phone Number", organizationPhoneNumber);
+        } catch (NullPointerException e) {
```
stustison: It isn't usually desirable to catch runtime exceptions like this in our own code. Sometimes it is unavoidable when using libraries, but that isn't the case here.

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/report.hbt
```diff
@@ -81,126 +80,274 @@
...
+            {{#each this}}
+                <table class="table table-striped ">
+                    <tr>
+                        <td>{{#if areaCode}}({{areaCode}})
+                        {{/if}}{{#if number}}{{number}}{{/if}}{{#if extension}}
+                            ext. {{extension}}{{/if}}</td>
+                    </tr>
+                </table>
+            {{/each}}
+        {{else}}
```
stustison: Why nest each of these as "if else if"? It'd be less code to just do "if if if" etc.

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
```diff
@@ -0,0 +1,111 @@
...
+<html>
+<head>
+    <meta content="text/html; charset=windows-1252" http-equiv="content-type"/>
+    <!-- styles -->
+    <link href="/services/registries/assets/css/bootstrap.min.css" rel="stylesheet"/>
+    <style type="text/css">
+        body {
+        background-color: {
+        {
+            background
```
stustison: This isn't valid CSS.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/HandlebarsHelper.java
ani6gup: The jacoco report showed significant drops because the if statements were not tested when it was contained with the handlebars.  Creating a separate class allowed to check if the helper through `HandlebarsHelperTest.java`. Also, this allows for more handlebar helpers to be added if needed.

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
ani6gup: Thanks for the catch, will do!

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/report.hbt
ani6gup: The main reason the "if else if" was added is because there is a case to generate the values dynamically at the end and I did not want duplicate values to be shown.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Ok, I will switch this!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Thank you, modified it!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Nope it did not need to be, just changed it

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Thank you for the catch, I think initially I was using it in two places which caused for the overlap. Just changed it!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Thank you, will keep this for future reference too!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,14 +149,114 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
         }
 
-        Map<String, Object> registryMap = reportMapBuilder.buildRegistryMap(registryPackage);
+        try {
+            html = registryReportBuilder.getHTMLFromRegistryPackage(registryPackage,
+                    templateLoader,
+                    "report");
+        } catch (IOException e) {
+            String message = "Error in compiling applying report template.";
+            LOGGER.error("Error in compiling applying report template");
```
gordocanchola: You can log message

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -55,7 +54,7 @@
 
     private RegistryPublicationService registryPublicationService;
 
-    private RegistryReportMapBuilder reportMapBuilder;
+    private RegistryReportBuilder registryReportBuilder;
 
     public RegistryRestEndpoint() {
         templateLoader = new ClassPathTemplateLoader();
```
gordocanchola: Is there a reason that the templateLoader is in the RegistryRestEndpoint if we're just passing it in the the RegistryReportBuilder?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,14 +149,114 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
 
         try {
-            Handlebars handlebars = new Handlebars(templateLoader);
-            Template template = handlebars.compile("report");
-            html = template.apply(registryMap);
+            html = registryReportBuilder.getHTMLFromMetacard(metacard, templateLoader, "summary");
         } catch (IOException e) {
-            LOGGER.error("Error compiling and applying report template.");
+            String message = "Error in compiling applying report template.";
+            LOGGER.error("Error in compiling applying report template");
```
gordocanchola: Do we want to log message. There was a similar error that read 'compiling and applying' was the 'and' removed on purpose? Do we want to add the registry id to the LOGGER message and say that it is the 'summary' report?

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
ani6gup: Checked and background-color and color do not really affect the report, modified accordingly

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,14 +149,114 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+        return Response.ok(html)
+                .build();
+    }
+
+    @Path("/{registryId}/report/{section}")
+    @Produces(MediaType.TEXT_HTML)
+    @GET
+    public Response viewSectionInfoHtml(@PathParam("registryId") final String registryId,
+            @QueryParam("sourceId") final List<String> sourceIds,
+            @PathParam("section") final String section) {
```
gordocanchola: We're not using section. I think the idea is to validate section and then use it below instead of the hard coded 'organizations'.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,14 +149,114 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+                    .entity(message)
+                    .build();
+        }
+
+        try {
+            html = registryReportBuilder.getHTMLFromRegistryPackage(registryPackage,
+                    templateLoader,
+                    "organizations");
+        } catch (IOException e) {
+            String message = "Error in compiling applying report template.";
```
gordocanchola: Same comments as the other two, about the logging stuff.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: I will log message, I think the 'and ' was accidentally removed. I can add the registry id to the logger message as well

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: It was there originally, I can move it over to the RegistryReportBuilder now since it is not really being utilized anywhere else in RegistryRestEndpoint

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Yeah, still trying to investigate why it is happening. I believe it's because of the loops at the moment

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
ani6gup: Since they are strings that will not change, I think constants would be the best

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,128 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+                    message,
+                    registryId,
+                    sourceIds);
+            return Response.status(Response.Status.NOT_FOUND)
+                    .entity(message)
+                    .build();
+        }
+
+        if(!isValidSection(section)) {
+            String message = section + " is an unknown section name";
```
shaundmorris: recommend using `String.format()` to combine strings

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpointTest.java
```diff
@@ -103,6 +106,128 @@ public void testReportWithRegistryPackage() throws ParserException, FederationAd
...
+        Response response = restEndpoint.viewSectionInfoHtml("metacardId",
+                Arrays.asList("sourceIds"),
+                "organizations");
+        assertThat(response.getStatusInfo()
+                .getStatusCode(), is(Response.Status.OK.getStatusCode()));
+
+    }
+
+    @Test
+    public void testSectionReportWithIOException() throws IOException, FederationAdminException {
```
shaundmorris: Shouldn't you expect an IOException on this test rather than throw one?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: modified it to `Error in compiling and applying report template` and logged the message instead of the string literal again

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: modified it to take in any section rather than just 'organizations', thank you!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
vinamartin: I'd recommend suggesting the available sections in the error message.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilder.java
```diff
@@ -0,0 +1,143 @@
...
+    public void setup() {
+        templateLoader = new ClassPathTemplateLoader();
+        templateLoader.setPrefix("/templates");
+        templateLoader.setSuffix(".hbt");
+        handlebars = new Handlebars(templateLoader);
+        handlebars.registerHelpers(new HandlebarsHelper());
+        handlebars.registerHelpers(StringHelpers.class);
+    }
+
+    public String getHTMLFromMetacard(Metacard metacard) throws IOException {
```
vinamartin: Might just need your explanation, but I'm confused about the naming of this method vs. the naming of the method below. 
This one is named, `getHTMLFromMetacard` but then it looks like it only generates the summary. Why is that?

ani6gup: The summary is a unique case and is generated just from certain metacard attributes. I could specify in the name that it is generating the summaryHTML for clarification though! Do `getSummaryHtmlFromMetacard` and `getSectionOrFullHtmlFromRegistryPackage` help eliminate confusion?

vinamartin: `getSummaryHtmlFromMetacard` would help clear up some of the confusion!

ani6gup: Awesome, thank you for the feedback!

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpointTest.java
ani6gup: I think that would be a better way to check. I will look over the unit tests that throw exceptions, thank you!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: Thank you for the feedback, I will be modifying the code to include all of the above!

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpointTest.java
ani6gup: After looking through the unit tests, the main reason the exceptions are being thrown are because some`RegistryRestEndpoint` methods call methods from mocked classes which contain these exceptions.  Some of the tests already were written and I used that as a foundation, let me know if you have any suggestions for alternative ways or if I should keep it as is.

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilderTest.java
```diff
@@ -0,0 +1,455 @@
...
+                .item(68)
+                .getTextContent();
+
+        assertThat(collectionName, is("Another Name"));
+        assertThat(versionInfoName, is("VersionInfo"));
+        assertThat(versionInfoValue, is("2.10.x"));
+    }
+
+    private Document xmlFileToDocument(String html) throws Exception {
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
```
shaundmorris: You may want to consider using try with resources in here to account for Closable implementations that need to get cleaned-up.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,130 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+
+        if (StringUtils.isBlank(registryId)) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity("Registry Id cannot be blank")
+                    .build();
+        }
+        try {
+            metacard =
+                    (federationAdminService.getRegistryMetacardsByRegistryIds((Collections.singletonList(
+                            registryId)))
```
figliold: Is it safe to assume that there is always going to be an element in the list?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,130 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
                     .build();
         }
 
-        Map<String, Object> registryMap = reportMapBuilder.buildRegistryMap(registryPackage);
+        try {
```
figliold: It looks like you have a constant for "report" declared in another file.  Can you use that here instead of a hard-coded string?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: As long as there is a value in registryId, there should be an element. I will double check with the Blue team though!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: Yes, thanks for the catch!

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilderTest.java
```diff
@@ -0,0 +1,455 @@
...
+    public void testFullReportWithSimplifiedRegistryPackage() throws Exception {
+        RegistryPackageType registryPackage = getRegistryPackageFromResource(
+                "/csw-simplified-registry-package.xml");
+        String html = reportHelper.getHtmlFromRegistryPackage(registryPackage, reportHelper.REPORT);
+        Document htmlDoc = xmlFileToDocument(html);
+        assertFullReportValues(htmlDoc);
+    }
+
+    private void assertSummaryReportValues(Document htmlDoc) {
+        String organizationName = htmlDoc.getElementsByTagName("strong")
```
shaundmorris: Not sure if I can think of a great way around this, but any concerns about the report changing fields, values, and having to re-writing a lot of the testing as a result?

ani6gup: The report values are all generated from the maps, if the structure of the maps change then another refactor of the report would have to change. 

As far as the tests go, they are testing the values in specific locations of the html to make sure that the title matches with the value, if the handlebars are modified to an extent then the tests would have to be modified as well. 

The additions to the report however can take place by adding separate sections, the organizations section was added as an example.

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilderTest.java
ani6gup: Implemented try with resources when creating the `ByteArrayInputStream`

catalog/spatial/registry/registry-rest-endpoint/pom.xml
```diff
@@ -56,6 +56,12 @@
             <artifactId>commons-lang3</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.codice.ddf.registry</groupId>
+            <artifactId>registry-ebrim-transformer</artifactId>
+            <version>${project.version}</version>
```
mcalcote: Do we need this version tag here?

ani6gup: I used the same foundation that was in the `FederationAdminServiceImpl` pom. Would it be better to remove it?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,144 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity("Registry Id cannot be blank")
+                    .build();
+        }
+        try {
+            List<Metacard> metacardList =
+                    federationAdminService.getRegistryMetacardsByRegistryIds((Collections.singletonList(
+                            registryId)));
+            if (CollectionUtils.isNotEmpty(metacardList)) {
+                metacard =
```
gordocanchola: You already have the metacardList, use that instead of using the federationAdminService again.

ani6gup: Thank you for the catch, changed it!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,144 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+                        (federationAdminService.getRegistryMetacardsByRegistryIds((Collections.singletonList(
+                                registryId)))
+                                .get(0));
+            } else {
+                String message = "Could not retrieve Metacard";
+                LOGGER.error("{} For registry id: '{}'", message, registryId);
+                return Response.status(Response.Status.BAD_REQUEST)
```
gordocanchola: Maybe return NOT_FOUND here.

brianfelix:
LGTM :+1:

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: modified it and added a unit test!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,141 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+            LOGGER.error("{} For registry id: '{}', optional source ids: {}.",
+                    message,
+                    registryId,
+                    sourceIds);
+            return Response.status(Response.Status.NOT_FOUND)
+                    .entity(message)
+                    .build();
+        }
+
+        List<String> validSections = Arrays.asList(registryReportBuilder.ORGANIZATIONS);
```
gordocanchola: Do we want to recreate this list every time this is called?

ani6gup: It is not necessary, I made it a private global variable in the class

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,141 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+        if (!isValidSection(validSections, section)) {
```
gordocanchola: We should probably move this validation up next to the other request validation (the one checking registryId) so bad requests can be tested before any calls to federation admin service.

ani6gup: I moved it to be right after the registryID check, thank you for the feedback!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +140,141 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+            LOGGER.error("{} For registry id: '{}'", message, registryId);
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity(message)
+                    .build();
+        }
+        try {
+            html = registryReportBuilder.getHtmlFromRegistryPackage(registryPackage, section);
+        } catch (IOException e) {
+            String message = "Error when compiling and applying report template.";
+            LOGGER.error("{} For registry id: '{}'", message, registryId);
```
gordocanchola: Maybe add the section to the error message.

ani6gup: Will do!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/HandlebarsHelper.java
ani6gup: Currently modifying handlebars helper to allow this case. I noticed some unexpected outputs in the unit test so I am working to create a solution.

shaundmorris:
👍 , however I defer to @pklinef, @stustison, or @andrewkfiedler  for the HandleBars specifics

emanns95:
Passed build and manual test :+1:

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
```diff
@@ -0,0 +1,139 @@
...
+<html>
+<head>
+    <meta content="text/html; charset=windows-1252" http-equiv="content-type"/>
+    <!-- styles -->
+    <link href="/services/registries/assets/css/bootstrap.min.css" rel="stylesheet"/>
+    <style type="text/css">
+       body {
+            margin-left: 15px;
+        }
+        .scrollable-list {
```
andrewkfiedler: This section of css can be removed (from .scrollable-list to .panel).
```
       .scrollable-list {
            height: auto;
            max-height: 600px;
            overflow: auto;
        }
        .spinner {
            height: 100px;
            position: relative;
        }
        .banner-container {
            text-align: center;
            height: 20px;
        }
        .header-layout,
        .footer-layout,
        #content {
            position: fixed;
            right: 0;
            left: 0;
        }
        .header-layout {
            top: 0;
        }
        .footer-layout {
            bottom: 0;
        }
        #content {
            top: 0px;
            bottom: 0px;
            overflow: auto;
        }
        .jumbotron {
            background-color: rgba(0, 0, 0, 0.25);
        }
        .panel {
            color: black;
        }
```

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/report.hbt
```diff
@@ -1,206 +1,326 @@
...
-        {
-            foreground
-        }
-        }
-        ;
-        margin-left: 15px;
+            margin-left: 15px;
         }
-
         .scrollable-list {
```
andrewkfiedler: This section of css can be removed (from .scrollable-list to .panel).
```
       .scrollable-list {
            height: auto;
            max-height: 600px;
            overflow: auto;
        }
        .spinner {
            height: 100px;
            position: relative;
        }
        .banner-container {
            text-align: center;
            height: 20px;
        }
        .header-layout,
        .footer-layout,
        #content {
            position: fixed;
            right: 0;
            left: 0;
        }
        .header-layout {
            top: 0;
        }
        .footer-layout {
            bottom: 0;
        }
        #content {
            top: 0px;
            bottom: 0px;
            overflow: auto;
        }
        .jumbotron {
            background-color: rgba(0, 0, 0, 0.25);
        }
        .panel {
            color: black;
        }
```

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
```diff
@@ -0,0 +1,76 @@
...
+<html>
+<head>
+    <meta content="text/html; charset=windows-1252" http-equiv="content-type"/>
+    <!-- styles -->
+    <link href="/services/registries/assets/css/bootstrap.min.css" rel="stylesheet"/>
+    <style type="text/css">
+        body {
+            margin-left: 15px;
+        }
+        .scrollable-list {
```
andrewkfiedler: This section of css can be removed (from .scrollable-list to .panel).
```
       .scrollable-list {
            height: auto;
            max-height: 600px;
            overflow: auto;
        }
        .spinner {
            height: 100px;
            position: relative;
        }
        .banner-container {
            text-align: center;
            height: 20px;
        }
        .header-layout,
        .footer-layout,
        #content {
            position: fixed;
            right: 0;
            left: 0;
        }
        .header-layout {
            top: 0;
        }
        .footer-layout {
            bottom: 0;
        }
        #content {
            top: 0px;
            bottom: 0px;
            overflow: auto;
        }
        .jumbotron {
            background-color: rgba(0, 0, 0, 0.25);
        }
        .panel {
            color: black;
        }
```

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
ani6gup: removed most of the css, took out the h3 and h4 css as well in the summary.hbt since they were not used.

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -150,20 +142,141 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
 
-        Map<String, Object> registryMap = reportMapBuilder.buildRegistryMap(registryPackage);
+        try {
+            html = registryReportBuilder.getHtmlFromRegistryPackage(registryPackage,
+                    registryReportBuilder.REPORT);
+        } catch (IOException e) {
+            String message = "Error in compiling and applying report template.";
+            LOGGER.error("{} For registry id: '{}'", message, registryId);
+            return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+                    .entity(message)
```
pklinef: Your error messages should match the Produces type.  I would recommend making an error html template that takes a message and return HTML instead of a string.  What you have now will work in browsers but it is not following the standard and could break another system if they try to parse the error message as HTML.

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
```diff
@@ -0,0 +1,109 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta content="text/html; charset=windows-1252" http-equiv="content-type"/>
```
pklinef: I would expect the charset to be UTF-8.  Is there a reason it is `windows-1252`?

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
```diff
@@ -0,0 +1,109 @@
...
+        .panel {
+            color: black;
+            height: auto;
+            max-height: 600px;
+            overflow: auto;
+        }
+        .table {
+            margin-left: 50px;
+            border: 1px solid #ddd;
+            max-width: 2000px;
```
pklinef: Why is this `max-width` needed?  Seems very high to be useful as it is.

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/report.hbt
```diff
@@ -1,206 +1,295 @@
...
         <table class="table table-striped ">
             <tbody>
             {{#each this}}
-                <tr>
-                    <td>{{this}}</td>
-                </tr>
+                {{#ifeq @key 'Slot'}}
+                    {{#each this}}
+                        <tr>
+                            <td width="250 px">
```
pklinef: Instead of inline style, I would recommend making a css class for this.

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta content="text/html; charset=windows-1252" http-equiv="content-type"/>
+    <!-- styles -->
+    <link href="/services/registries/assets/css/bootstrap.min.css" rel="stylesheet"/>
+    <style type="text/css">
```
pklinef: It looks like all of the report handlebar templates have duplicated css.  Can we put all of the duplicated css under /services/registries/assets/css/report.css or something along those lines?

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
```diff
@@ -0,0 +1,35 @@
...
+    </style>
+    <link href="/services/registries/assets/css/bootstrap-theme.min.css" rel="stylesheet"/>
+</head>
+<body>
+<h2>Summary</h2>
+<table class="table table-striped">
+    <tbody>
+    {{#each this}}
+        <tr>
+        <td width="250 px"><strong>{{@key}}</strong></td>
```
pklinef: Another inline width that should be a css class.

catalog/spatial/registry/registry-rest-endpoint/src/test/java/org/codice/ddf/registry/rest/endpoint/report/RegistryReportBuilderTest.java
pklinef: This looks too brittle for any changes to the template.  The tests are too coupled to the exact template.  It might be useful to use an HTML parser with more predictable selectors.  For example, you could add an organization css class for the element containing organization information and another css class for the organization name.  Then you could use something like jsoup to select on `doc.select("organization name")`.  

https://jsoup.org/cookbook/extracting-data/selector-syntax

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
ani6gup: `windows-1252` was originally in the report handlebars before, I have changed it to `UTF-8` though

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
ani6gup: Yes, moved everything into a report.css class and linked it in each handlebar file

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/summary.hbt
ani6gup: made a css class called column that sets the width

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/report.hbt
ani6gup: thank you, made a css class accordingly

catalog/spatial/registry/registry-rest-endpoint/src/main/resources/templates/organizations.hbt
ani6gup: max-width was not needed, removed it!

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
ani6gup: does this apply to all errors within `RegistryRestEndpoint`?

catalog/spatial/registry/registry-rest-endpoint/src/main/java/org/codice/ddf/registry/rest/endpoint/RegistryRestEndpoint.java
```diff
@@ -146,24 +143,150 @@ public Response viewRegInfoHtml(@PathParam("registryId") final String registryId
...
+                    .entity(getErrorHtml(message))
+                    .build();
         }
 
         return Response.ok(html)
                 .build();
     }
 
+    private String getErrorHtml(String error) {
+        return String.format("<html>%s</html>", error);
```
pklinef: Though this will probably work with most parsers, it is still a long way off from what is needed at a minimum to be valid.

http://www.brucelawson.co.uk/2010/a-minimal-html5-document/
https://www.sitepoint.com/a-minimal-html-document-html5-edition/

pklinef:
👍

PR #1054, opened 2016-07-27
DDF-2333 Add a command for seeding the metacard and product caches

jrnorth:
#### What does this PR do?
Adds a `catalog:seed` command that allows the user to seed the local metacard and product caches from the enterprise or from specific federated sources.

@bdeining 
@glenhein 
@rzwiefel 

@jlcsmith 
@coyotesqrl 

Set up two DDFs. Ingest some products on DDF-1. On DDF-2, add DDF-1 as a federated source.

Run the `catalog:seed` command on DDF-2 with different values for the `--cql`, `--product-limit`, and `--source` options and verify that the expected metacards and products are added to the local caches.

[DDF-2333](https://codice.atlassian.net/browse/DDF-2333)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
@@ -1,0 +1,192 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.commands.catalog;
+
+import static java.util.AbstractMap.SimpleEntry;
+import static java.util.Map.Entry;
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
+import static ddf.catalog.data.Metacard.ANY_TEXT;
+import static ddf.catalog.filter.FilterDelegate.WILDCARD_CHAR;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+
+import org.apache.felix.gogo.commands.Command;
+import org.apache.felix.gogo.commands.Option;
+import org.fusesource.jansi.Ansi;
+import org.geotools.filter.text.cql2.CQL;
+import org.opengis.filter.Filter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Attribute;
+import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
+import ddf.catalog.filter.FilterBuilder;
+import ddf.catalog.operation.QueryResponse;
+import ddf.catalog.operation.ResourceRequest;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.operation.impl.ResourceRequestById;
+import ddf.catalog.resource.ResourceNotFoundException;
+import ddf.catalog.resource.ResourceNotSupportedException;
+
+@Command(scope = CatalogCommands.NAMESPACE, name = "seed", description =
+        "Seeds the local metacard and product caches from the enterprise or from specific federated sources.\n\n"
+                + "\tThis command will query the enterprise or the specified sources for metacards in increments "
+                + "the size of the `--product-limit` argument (default 20) until that number of product downloads "
+                + "have started or until there are no more results. Local products will not be added to the cache. "
+                + "This command will not re-download products that are up-to-date in the cache, so subsequent runs "
+                + "of the command will attempt to cache metacards and products that have not already been cached.\n\n"
+                + "\tNote that this command will begin the product downloads and some may still be in progress "
+                + "by the time it completes. Also, product caching must be enabled in the catalog framework for "
+                + "this command to seed the product cache.")
+public class SeedCommand extends SubjectCommands {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SeedCommand.class);
+
+    private static final Map<String, Serializable> CACHE_UPDATE_PROPERTIES =
+            Collections.singletonMap("mode", "update");
+
+    private static final String RESOURCE_CACHE_STATUS = "internal.local-resource";
+
+    @Option(name = "--source", multiValued = true, aliases = {
+            "-s"}, description = "Source(s) to query (e.g., -s source1 -s source2). Default is to query the entire enterprise.")
+    List<String> sources;
+
+    @Option(name = "--cql", description = "CQL expression specifying the metacards to seed.\n"
+            + "CQL Examples:\n" + "\tTextual:   seed --cql \"title like 'some text'\"\n"
+            + "\tTemporal:  seed --cql \"modified before 2012-09-01T12:30:00Z\"\n"
+            + "\tSpatial:   seed --cql \"DWITHIN(location, POINT (1 2), 10, kilometers)\"\n"
+            + "\tComplex:   seed --cql \"title like 'some text' AND modified before 2012-09-01T12:30:00Z\"")
+    String cql;
+
+    @Option(name = "--product-limit", aliases = {"-pl"}, description =
+            "The maximum number of products to download to the cache. The number of metacards cached "
+                    + "might not be equal to this number because some metacards may not have associated "
+                    + "products.")
+    int productLimit = 20;
+
+    private final Predicate<Metacard> isNonCachedResource = metacard -> {
+        Attribute cached = metacard.getAttribute(RESOURCE_CACHE_STATUS);
+        return cached == null || !((Boolean) cached.getValue());
+    };
+
+    private Filter createFilter() throws Exception {
+        if (isNotBlank(cql)) {
+            return CQL.toFilter(cql);
+        } else {
+            final FilterBuilder builder = getService(FilterBuilder.class);
+            if (builder != null) {
+                return builder.attribute(ANY_TEXT)
+                        .is()
+                        .like()
+                        .text(WILDCARD_CHAR);
+            } else {
+                throw new RuntimeException("Could not retrieve the FilterBuilder service.");
+            }
+        }
+    }
+
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        if (productLimit <= 0) {
+            printErrorMessage("A limit of " + productLimit
+                    + " was supplied. The limit must be greater than 0.");
+            return null;
+        }
+
+        final Filter filter = createFilter();
+
+        final long start = System.currentTimeMillis();
+        int productDownloads = 0;
+        int downloadErrors = 0;
+        int pageCount = 0;
+
+        while (productDownloads < productLimit) {
+            final QueryImpl query = new QueryImpl(filter);
+            query.setPageSize(productLimit);
+            query.setStartIndex(pageCount * productLimit + 1);
+            ++pageCount;
+
+            final QueryRequestImpl queryRequest;
+            if (sources != null && !sources.isEmpty()) {
+                queryRequest = new QueryRequestImpl(query, sources);
+            } else {
+                queryRequest = new QueryRequestImpl(query, true);
+            }
+            queryRequest.setProperties(new HashMap<>(CACHE_UPDATE_PROPERTIES));
+
+            final CatalogFramework framework = getService(CatalogFramework.class);
+
+            final QueryResponse queryResponse;
+            if (framework != null) {
+                queryResponse = framework.query(queryRequest);
+            } else {
+                throw new RuntimeException("Could not retrieve the CatalogFramework service.");
+            }
+
+            if (queryResponse.getResults()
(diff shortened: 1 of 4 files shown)


jlcsmith:
I realize that we are inconsistent internally, but I'd prefer "resource" over "product".  Some consider a product to be a finished piece of data while resource is more generic and can include raw data.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,195 @@
...
+                            .map(Result::getMetacard)
+                            .filter(isNonCachedResource)
+                            .map(metacard -> new SimpleEntry<>(new ResourceRequestById(metacard.getId()),
+                                    metacard.getSourceId()))
+                            .collect(toList());
+
+            for (Entry<? extends ResourceRequest, String> requestAndSourceId : resourceRequests) {
+                final ResourceRequest request = requestAndSourceId.getKey();
+                try {
+                    framework.getResource(request, requestAndSourceId.getValue());
```
stustison: Seems like you need to close all of the input streams returned from these resource requests.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,195 @@
...
+        int pageCount = 0;
+
+        while (productDownloads < productLimit) {
+            final QueryImpl query = new QueryImpl(filter);
+            query.setPageSize(productLimit);
+            query.setStartIndex(pageCount * productLimit + 1);
+            ++pageCount;
+
+            final QueryRequestImpl queryRequest;
+            if (sources != null && !sources.isEmpty()) {
```
bdeining: I think you could also use `CollectionUtils.isNotEmpty(sources)`

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TestSeedCommand.java
```diff
@@ -0,0 +1,280 @@
...
+        QueryRequest request = queryCaptor.getValue();
+        queryRequestAssertions.accept(request);
+    }
+
+    @Test
+    public void testProductLimit() throws Exception {
+        final int limit = 2;
+        seedCommand.productLimit = limit;
+
+        final String id1 = "1";
```
bdeining: These IDs could be made constants

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TestSeedCommand.java
```diff
@@ -0,0 +1,280 @@
...
+                        .getStartIndex() < stopReturningResultsAtIndex)));
+
+        QueryResponse noResults = mock(QueryResponse.class);
+        when(noResults.getResults()).thenReturn(Collections.emptyList());
+        doReturn(noResults).when(framework)
+                .query(argThat(isQueryWithStartIndex(request -> request.getQuery()
+                        .getStartIndex() >= stopReturningResultsAtIndex)));
+    }
+
+    private IsQueryWithStartIndex isQueryWithStartIndex(Predicate<QueryRequest> test) {
```
bdeining: `isQueryWithStartIndex` makes this seem like it should return a boolean; it might be more clear as `createIsQueryWithStartIndex`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,195 @@
...
+
+@Service
+@Command(scope = CatalogCommands.NAMESPACE, name = "seed", description = "Seeds the local metacard "
+        + "and product caches from the enterprise or from specific federated sources.", detailedDescription =
+        "This command will query the enterprise or the specified sources for metacards in increments "
+                + "the size of the `--product-limit` argument (default 20) until that number of product downloads "
+                + "have started or until there are no more results. Local products will not be added to the cache. "
+                + "This command will not re-download products that are up-to-date in the cache, so subsequent runs "
+                + "of the command will attempt to cache metacards and products that have not already been cached.\n\n"
+                + "Note that this command will begin the product downloads and some may still be in progress "
```
coyotesqrl: This sentence isn't very clear for describing the async nature of the command. Suggest something like...
`Note: this command will trigger product downloads in the background. Those downloads may continue after control is returned to the console.`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,195 @@
...
+                + "this command to seed the product cache.")
+public class SeedCommand extends SubjectCommands implements Action {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SeedCommand.class);
+
+    private static final Map<String, Serializable> CACHE_UPDATE_PROPERTIES =
+            Collections.singletonMap("mode", "update");
+
+    private static final String RESOURCE_CACHE_STATUS = "internal.local-resource";
+
+    @Option(name = "--source", multiValued = true, aliases = {
```
coyotesqrl: If this annotation is really formatting this way, it should probably be wrapped in a `@formatter:off`/`@formatter:on` block so it can be better formatted manually. I suggest that one attribute per line would read much better.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,195 @@
...
+            final List<Entry<? extends ResourceRequest, String>> resourceRequests =
+                    queryResponse.getResults()
+                            .stream()
+                            .map(Result::getMetacard)
+                            .filter(isNonCachedResource)
+                            .map(metacard -> new SimpleEntry<>(new ResourceRequestById(metacard.getId()),
+                                    metacard.getSourceId()))
+                            .collect(toList());
+
+            for (Entry<? extends ResourceRequest, String> requestAndSourceId : resourceRequests) {
```
coyotesqrl: Instead of collecting to a list above and then iterating the list here, you might consider adding an exec service to this class and then sending tasks off to it, collecting the Futures to a list. Might be a wash.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,195 @@
...
+                    break;
+                }
+            }
+        }
+
+        printProgressAndFlush(start, productDownloads, productDownloads);
+
+        console.println();
+        if (downloadErrors > 0) {
+            printColor(Ansi.Color.YELLOW,
```
coyotesqrl: Is there a reason you don't want to use `printErrorMessage()` here? That will print this in RED, but this seems like it is an error condition to me.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TestSeedCommand.java
jrnorth: Will refactor `IsQueryWithStartIndex` to `QueryWithStartIndex` and `isQueryWithStartIndex` to `queryWithStartIndex`. Then the mockito call will look like: `doReturn(noResults).when(framework).query(argThat(is(queryWithStartIndex(...))));`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
jrnorth: So to safely close the input stream, we have to know when the download is finished, right? How can we tell when it's finished?

stustison: I believe the reliable resource reader will continue to cache regardless of what the client actually does. That was the way it used to work, anyways. You should be able to just simply close the streams.

jrnorth: Unfortunately I think it will only work that way if `Always Cache Product` has been set in the resource download settings: https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/resources/OSGI-INF/metatype/downloadManager.xml#L43

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,241 @@
...
+import ddf.catalog.operation.QueryResponse;
+import ddf.catalog.operation.ResourceRequest;
+import ddf.catalog.operation.ResourceResponse;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.operation.impl.ResourceRequestById;
+import ddf.catalog.resource.ResourceNotFoundException;
+import ddf.catalog.resource.ResourceNotSupportedException;
+
+//@formatter:off
```
jlcsmith: Remove if not being used?  Or is this an indication to future developers to not format this section?

coyotesqrl: It's a directive to Intellij to leave the formatting in a block of code as-is. Ends on line 47.

jlcsmith: ah, thanks

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,241 @@
...
+                + "control is returned to the console. Also, resource caching must be enabled in the catalog framework "
+                + "for this command to seed the resource cache.")
+//@formatter:on
+public class SeedCommand extends SubjectCommands implements Action {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SeedCommand.class);
+
+    private static final Map<String, Serializable> CACHE_UPDATE_PROPERTIES =
+            Collections.singletonMap("mode", "update");
+
+    private static final String RESOURCE_CACHE_STATUS = "internal.local-resource";
```
jlcsmith: Since both this and the catalog-core-resourcestatusplugin use this constant, maybe it should be moved to a common area.  If that constant changes in the plugin, then this command breaks.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,241 @@
...
+                20,
+                30,
+                TimeUnit.SECONDS,
+                new LinkedBlockingQueue<>());
+        EXECUTOR.allowCoreThreadTimeOut(true);
+    }
+
+    @Option(name = "--source", multiValued = true, aliases = {
+            "-s"}, description = "Source(s) to query (e.g., -s source1 -s source2). Default is to query the entire enterprise.")
+    List<String> sources;
```
jlcsmith: can this be a Set?

jrnorth: The Karaf documentation says the field must be a `List`: https://github.com/apache/karaf/blob/master/shell/core/src/main/java/org/apache/karaf/shell/api/action/Option.java#L86

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,241 @@
...
+
+    private static final Map<String, Serializable> CACHE_UPDATE_PROPERTIES =
+            Collections.singletonMap("mode", "update");
+
+    private static final String RESOURCE_CACHE_STATUS = "internal.local-resource";
+
+    private static final ThreadPoolExecutor EXECUTOR;
+
+    static {
+        EXECUTOR = new ThreadPoolExecutor(20,
```
coyotesqrl: That's a lot of threads to devote to this operation. Granted, it won't be run by multiple users concurrently, but you're still using a lot of resources.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
```diff
@@ -0,0 +1,241 @@
...
+        private ResourceCloseHandler(ResourceResponse response) {
+            resourceStream = response.getResource()
+                    .getInputStream();
+        }
+
+        @Override
+        public void run() {
+            byte[] buffer = new byte[4096];
+            try {
+                while (resourceStream.read(buffer, 0, buffer.length) != -1) {
```
coyotesqrl: You're just reading the stream until its end, dropping its contents on the floor? Because the caching of the metacard/resource is happening in another thread and all you care about is the completion of this stream? Am I understanding that correctly? If so, I take back what I said earlier about the size of the execpool as it's probably okay to be that large to close these out quickly.

coyotesqrl:
Assuming I've correctly understood the close handler, 👍

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
jrnorth: There was some discussion about this in [this PR](https://github.com/codice/ddf/pull/1003). I'll ask about what the decision was regarding this value.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java
jrnorth: Yes, these threads are only being used to close the resources' input streams after the downloads are complete. Also, there won't be any threads in the pool if there aren't any downloads running since the core threads will time out after 30 seconds.

jlcsmith:
Built and tested - LGTM 👍

PR #1056, opened 2016-07-27
DDF-2199 Added null metacard check to RegistryPolicyPlugin

mcalcote:
#### What does this PR do?

While resolving merge conflicts from the previous 2199 PR, a null check was lost in RegistryPolicyPlugin.getWritePolicy. This PR fixes that.

@harrison-tarr @gordocanchola @clockard @spearskw 

@coyotesqrl
@shaundmorris
Build

https://github.com/codice/ddf/pull/1049

DDF-2199

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
+++ b/catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
@@ -159,7 +159,7 @@ public class RegistryPolicyPlugin implements PolicyPlugin {
 
     private PolicyResponse getWritePolicy(Metacard input, Map<String, Serializable> properties) {
         HashMap<String, Set<String>> operationPolicy = new HashMap<>();
-        if (Requests.isLocal(properties) && RegistryUtility.isRegistryMetacard(input)) {
+        if (Requests.isLocal(properties) && input != null && RegistryUtility.isRegistryMetacard(input)) {
             String registryBaseUrl = RegistryUtility.getStringAttribute(input,
                     RegistryObjectMetacardType.REGISTRY_BASE_URL,
                     null);


catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
```diff
@@ -159,7 +159,7 @@ private void parsePermissionsFromString(List<String> permStrings,
 
     private PolicyResponse getWritePolicy(Metacard input, Map<String, Serializable> properties) {
         HashMap<String, Set<String>> operationPolicy = new HashMap<>();
-        if (Requests.isLocal(properties) && RegistryUtility.isRegistryMetacard(input)) {
+        if (Requests.isLocal(properties) && input != null && RegistryUtility.isRegistryMetacard(input)) {
```
coyotesqrl: It might be better to update `RegistryUtility.isRegistryMetacard()` to return false for a null metacard.

coyotesqrl: Then again, none of your other methods in that utility class are performing null checks so it would be asymmetric. Ignore.

coyotesqrl:
👍

catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
shaundmorris: Yeah @coyotesqrl good point. I am little confused as to why a null metacard would be getting passed in here. Handling null is good, but I'm wondering if we need to look at the bigger flow.

harrison-tarr: @shaundmorris When an update occurs, the `CatalogFrameworkImpl` tries to map the new metacard to the old metacard. In the case of an `id` update, it tries to use the `id`. In the case that the `id` is null (which for a new metacard it could be), it fails to do the matching and passes a `null` metacard to the policy plugins. Somehow the updates still work, but I'm not well-versed in the update logic.

shaundmorris: Ah thanks @harrison-tarr I was missing the update case.

mcalcote: 😂

shaundmorris:
👍

harrison-tarr:
Successfully heroed

PR #1057, opened 2016-07-27
DDF-2337 changed the filter plugin logs for create to use metacard.getTitle instead of metacard.getId

stustison:
@coyotesqrl 
@pklinef
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests


coyotesqrl:
👍

pklinef:
Should we fall back to the ID if the title is blank?  Should we truncate the title if it is too long?  Are we concerned with duplicated titles?

stustison:
Chances are pretty high that the ID is going to be null. It is only non-null when ingesting files as opposed to metacards. I could still use the ID if it is filled out and then fallback to the title, I just thought that might be a little confusing if you ended up with some records with titles and some with IDs in the same log message. I, personally, am not concerned with duplicates as this is just meant to be an aid to an integrator as to why a particular ingest failed.

pklinef:
👍

PR #1064, opened 2016-07-28
DDF-2300: Range Headers are now supported for Product Retrieval

garrettfreibott:
#### What does this PR do?
Adds range header support for product retrieval retries. When the ReliableResourceDownloader attempts to retry retrieval of a product, it will no longer start over from the beginning of the download every time. There was also a major bug where DDF would properly use range headers to retrieve a product during a retry, but it would skip bytes in the input stream regardless of whether or not the server responding supported it or not, meaning unread data would be skipped and the file would become corrupted.

#### Who is reviewing it?
@pklinef 
@oconnormi 
@jrnorth 

#### Choose 2 committers to review/merge the PR:
@coyotesqrl
@figliold 

itests were added to check that the CSW retrieval works properly and to test retries via simulated network disconnects. There were also a few itests for simple opensearch downloads. There is a ticket out that we will be picking up soon to add an opensearch stub server similar to the csw stub server, but for now, the best way to test is to spin up an instance and test that product retrieval still works for all the various sources WITHOUT the data becoming corrupted. Ingest some sort of plain text file so that it can be easily double checked for missing characters. It would also probably be best to re-download the file immediately to check that the returned cache copy is not corrupted either.

Since I needed to change ReliableResourceDownloader, it was necessary to change the URLResourceReader to also support Partial Content responses, so that had the side-effect of (hopefully) making all the sources support range headers. Hooray!

#### Commit Messages:
Product Retrievals will now properly request range headers and handle
Partial Content responses so that subsequent retries will not have
to redownload the entire product. Previously the endpoints would send
requests with range headers properly, but would not handle the response
correctly, potentially resulting in corrupted data.

After modifying the ReliableResourceDownloader[RRD], changes were needed
in the URLResourceReader to adjust for the fact that the RRD no longer
skipped ahead bytes when re-retrieving a resource. These changes also cause
all of the other endpoints that use it to support range header retrieval.
Hooray!

### DIFF
--- a/catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
+++ b/catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
@@ -38,12 +38,12 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.cxf.jaxrs.client.WebClient;
 import org.apache.cxf.jaxrs.ext.multipart.ContentDisposition;
 import org.apache.tika.Tika;
-import org.apache.tika.metadata.HttpHeaders;
 import org.codice.ddf.security.common.jaxrs.RestSecurity;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.net.HttpHeaders;
 
 import ddf.catalog.data.Metacard;
 import ddf.catalog.operation.ResourceResponse;
@@ -260,6 +260,8 @@ public class URLResourceReader implements ResourceReader {
             bytesToSkip = properties.get(BYTES_TO_SKIP)
                     .toString();
             LOGGER.debug("bytesToSkip: {}", bytesToSkip);
+        } else {
+            bytesToSkip = "0";
         }
 
         if (resourceURI.getScheme()
@@ -356,7 +358,8 @@ public class URLResourceReader implements ResourceReader {
             Object entityObj = clientResponse.getEntity();
             if (entityObj instanceof InputStream) {
                 is = (InputStream) entityObj;
-                if (Response.Status.OK.getStatusCode() != clientResponse.getStatus()) {
+                if (Response.Status.OK.getStatusCode() != clientResponse.getStatus()
+                        && Response.Status.PARTIAL_CONTENT.getStatusCode() != clientResponse.getStatus()) {
                     String error = null;
                     try {
                         if (is != null) {
@@ -376,7 +379,16 @@ public class URLResourceReader implements ResourceReader {
                         "Received null response while retrieving resource.");
             }
 
-            skipBytes(is, bytesToSkip);
+            long responseBytesSkipped = 0L;
+            if (headers.getFirst(HttpHeaders.CONTENT_RANGE) != null) {
+                String contentRangeHeader =
+                        String.valueOf(headers.getFirst(HttpHeaders.CONTENT_RANGE));
+                responseBytesSkipped = Long.parseLong(StringUtils.substringBetween(
+                        contentRangeHeader.toLowerCase(),
+                        "bytes ",
+                        "-"));
+            }
+            alignStream(is, Long.parseLong(bytesToSkip), responseBytesSkipped);
 
             return new ResourceResponseImpl(new ResourceImpl(new BufferedInputStream(is),
                     mimeType,
@@ -462,6 +474,36 @@ public class URLResourceReader implements ResourceReader {
                 LOGGER.debug("Did not skip specified bytes while retrieving resource."
                         + " Bytes to skip: {} -- Skipped Bytes: {}", bytesToSkip, bytesSkipped);
             }
+        }
+    }
+
+    private void alignStream(InputStream in, long requestedBytesToSkip, long responseBytesSkipped)
+            throws IOException {
+        long misalignment = requestedBytesToSkip - responseBytesSkipped;
+
+        if (misalignment == 0) {
+            LOGGER.trace("Server responded with the correct byte range.");
+            return;
+        }
+
+        try {
+            if (requestedBytesToSkip > responseBytesSkipped) {
+                LOGGER.warn("Server returned incorrect byte range, skipping first [{}] bytes",
+                        misalignment);
+                if (in.skip(misalignment) != misalignment) {
+                    throw new IOException(String.format(
+                            "Input Stream could not be skipped %d bytes.",
+                            misalignment));
+                }
+
+            } else {
+                throw new IOException(
+                        "Server skipped more bytes than requested in the range header.");
+            }
+        } catch (IOException e) {
+            throw new IOException(String.format(
+                    "Unable to align input stream with the requested byteOffset of %d",
+                    requestedBytesToSkip), e);
         }
     }
 
--- a/catalog/core/catalog-core-urlresourcereader/src/test/java/ddf/catalog/resource/impl/ResourceReaderTest.java
+++ b/catalog/core/catalog-core-urlresourcereader/src/test/java/ddf/catalog/resource/impl/ResourceReaderTest.java
@@ -47,7 +47,6 @@ import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
 
 import org.apache.cxf.jaxrs.client.WebClient;
-import org.apache.tika.metadata.HttpHeaders;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -58,6 +57,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.ImmutableSet;
+import com.google.common.net.HttpHeaders;
 
 import ddf.catalog.operation.ResourceResponse;
 import ddf.catalog.resource.Resource;
@@ -139,6 +139,15 @@ public class ResourceReaderTest {
         byte[] sampleBytes = {65, 66, 67, 68, 69};
 
         return new ByteArrayInputStream(sampleBytes);
+    }
+
+    private static InputStream getBinaryDataWithOffset(int offset) throws IOException {
+
+        byte[] sampleBytes = {65, 66, 67, 68, 69};
+        InputStream is = new ByteArrayInputStream(sampleBytes);
+        is.skip(offset);
+
+        return is;
     }
 
     @Before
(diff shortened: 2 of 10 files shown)


catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -84,6 +84,8 @@
 
     private static final String BYTES_TO_SKIP = "BytesToSkip";
 
+    private static final String CONTENT_RANGE_HEADER = "Content-Range";
```
figliold: You might be able to use this constant from Guava: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/net/HttpHeaders.html#CONTENT_RANGE

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -465,6 +474,23 @@ private void skipBytes(InputStream is, String bytesToSkip) throws IOException {
         }
     }
 
+    private void alignStream(InputStream in, long requestedBytesToSkip, long responseBytesSkipped)
+            throws IOException {
```
figliold: Can the requestedBytesToSkip and the responseBytesSkipped really be different?

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -465,6 +474,23 @@ private void skipBytes(InputStream is, String bytesToSkip) throws IOException {
...
+        try {
+            if (requestedBytesToSkip > responseBytesSkipped) {
+                LOGGER.error(
+                        "Source incorrect byte range, skipping ahead [{}] bytes in the input stream",
+                        requestedBytesToSkip - responseBytesSkipped);
+                long skipped = in.skip(requestedBytesToSkip - responseBytesSkipped);
+            } else if (requestedBytesToSkip < responseBytesSkipped) {
+                throw new IOException(
```
figliold: Should we log the exception message here?

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -465,6 +474,23 @@ private void skipBytes(InputStream is, String bytesToSkip) throws IOException {
...
+                        "Source server skipped more bytes than requested in the range header.");
+            }
+        } catch (IOException e) {
```
figliold: Should we log the exception message here?

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/CswRecordCollection.java
```diff
@@ -61,6 +64,8 @@
 
     private Resource resource;
```
figliold: Can this be an ImmutableMap?

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
```diff
@@ -947,6 +959,23 @@ public ResourceResponse retrieveResource(URI resourceUri,
         return resourceReader.retrieveResource(resourceUri, requestProperties);
     }
```
figliold: Will each source type  need to have the alignStream code?

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/reader/GetRecordsMessageBodyReader.java
```diff
@@ -59,6 +62,10 @@
 public class GetRecordsMessageBodyReader implements MessageBodyReader<CswRecordCollection> {
     private static final Logger LOGGER = LoggerFactory.getLogger(GetRecordsMessageBodyReader.class);
 
+    public static final String BYTES_SKIPPED = "bytes-skipped";
+
```
figliold: May be able to use the Guava constant.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
garrettfreibott: Awesome, thanks!

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
garrettfreibott: Unfortunately, yes. If Partial Content responses are not supported by the server, then it will return the entire content, in which case we need to skip ahead in the input stream to avoid re-reading bytes. It is also possible for the server to support Partial Content responses and return a different byte range than what was requested. If we ask for bytes 5-10, for whatever reason, the server _is_ allowed to return a partial content response with the range 4-10, and we would have to skip over the 1 overlapping byte. The server *shouldn't* return a byte range higher than 5-10 (like 6-10), but I added that situation just in case to be thorough.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
garrettfreibott: Ya, I can add a log message as well. 

The URLResourceReader's retrieveResource method will throw this IOException and it should be handled as a failed download attempt in the ReliableResourceDownloader, retrying if it still has retries left, or failing if the amount of retries has been passed.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
garrettfreibott: Same comment as previous. I'll log an error as well, but it should be ok to throw an IOException and be treated as a failed download attempt.

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
garrettfreibott: For the most part, yes. The ReliableResourceDownloader was modified to expect `retrieveResource(bytesSkipped)` to return a resource that has an input stream which begins with the offset of `bytesSkipped`, and does not care whether or not the server it's requesting the product from supports range-headers. The URLResourceReader (and `GetRecordsMessageBodyReader`) now handles all the logic for sending range requests and modifying the input stream based on whether Partial Content responses are supported or not.

Aside from `GetRecordsMessageBodyReader` though, I believe that everything else delegates to the `URLResourceReader` to actually make the get request and open the input stream for the resource.

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/reader/GetRecordsMessageBodyReader.java
garrettfreibott: Thanks! I couldn't find it in any of the other HttpHeader classes we use for some reason. :+1:

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/CswRecordCollection.java
garrettfreibott: Ya, right now it can be immutable. I made it a HashMap in case any additional properties were needed later on, though.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -465,6 +477,23 @@ private void skipBytes(InputStream is, String bytesToSkip) throws IOException {
         }
     }
 
+    private void alignStream(InputStream in, long requestedBytesToSkip, long responseBytesSkipped)
+            throws IOException {
+        try {
+            if (requestedBytesToSkip > responseBytesSkipped) {
+                LOGGER.error(
```
coyotesqrl: As this is an expected case as you said in your earlier comment and you're able to handle it by skipping N more bytes in the stream, I think this should be a WARN log.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -465,6 +477,23 @@ private void skipBytes(InputStream is, String bytesToSkip) throws IOException {
...
+                        "Source incorrect byte range, skipping ahead [{}] bytes in the input stream",
+                        requestedBytesToSkip - responseBytesSkipped);
+                long skipped = in.skip(requestedBytesToSkip - responseBytesSkipped);
```
coyotesqrl: You're not doing anything with the result of the `skip` call. Might as well not get it.

catalog/core/catalog-core-urlresourcereader/src/test/java/ddf/catalog/resource/impl/ResourceReaderTest.java
```diff
@@ -379,15 +388,116 @@ public void testNameInContentDisposition() throws Exception {
...
+        when(mockResponse.getEntity()).thenReturn(getBinaryDataWithOffset(3));
+
+        String bytesToSkip = "2";
+
+        // this should throw an IOException since more bytes were skipped than requested
+        verifyFileFromURLResourceReader(uri, JPEG_FILE_NAME_1, JPEG_MIME_TYPE, bytesToSkip);
+    }
+
+    /**
+     * Tests that a Partial Content response that has a higher byte offset as what was requested
```
coyotesqrl: Copy paste error on this javadoc.

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
coyotesqrl: It's unfortunately to need to duplicate this sort of utility code everywhere. Is there nowhere we could centralize it?

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/reader/GetRecordsMessageBodyReader.java
```diff
@@ -98,6 +103,14 @@ public CswRecordCollection readFrom(Class<CswRecordCollection> type, Type generi
             throws IOException, WebApplicationException {
 
         CswRecordCollection cswRecords = null;
+        Map<String, Serializable> resourceProperties = new HashMap<>();
+        // Check if the server returned a Partial Content response (hopefully in response to a range header)
+        String contentRangeHeader = httpHeaders.getFirst(HttpHeaders.CONTENT_RANGE);
+        if (StringUtils.isNotBlank(contentRangeHeader)) {
+            contentRangeHeader = StringUtils.substringBetween(contentRangeHeader.toLowerCase(), "bytes ", "-");
```
coyotesqrl: Same regarding duplicate code.

Might be worth creating a small service for content range processing with two methods on it: one for retrieving the `bytesSkipped` and one for aligning input streams.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
garrettfreibott: Good point - switched it to a warn.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
garrettfreibott: Removed

coyotesqrl:
Fine as it stands, but if it's not too much extra work, I'd like to see that service extracted as I mentioned. Otherwise, minus cleanup: 👍

catalog/core/catalog-core-urlresourcereader/src/test/java/ddf/catalog/resource/impl/ResourceReaderTest.java
garrettfreibott: Nice catch - fixed.

garrettfreibott:
Talked with @coyotesqrl and we agreed that the `alignStreams` method was a bit small to warrant its own service, and we couldn't find a good place to throw it in a library. For now I'll leave it as-is. It may also be refactored a bit when we re-work the download code.

pklinef:
Due to the major bug mentioned in the description, I think this should be backported to 2.9.x.

coyotesqrl:
Heroed successfully. 👍

PR #1066, opened 2016-07-28
DDF-2344 Pull out functionality from CatalogFrameworkImpl to logical subcomponents.

coyotesqrl:
#### What does this PR do?
This is the first phase of a change to `CatalogFrameworkImpl` which turns it into a facade over several logical delegate classes. These delegates silo logically like functionality in order to facilitate further refactoring, cleanup, and future maintenance.

@rzwiefel @kcwire @ryeats @tbatie 

@pklinef
@shaundmorris
@stustison

As this breaks apart one of the core components of DDF, it will require not only a full build with itests, but an extensive run through manual use of the system to ensure that no functionality has been broken.

The current version of `CatalogFrameworkImpl` in master is 3397 lines long and contains much duplicative, redundant code. It is difficult to maintain and rather fragile. By breaking it up, we can isolate logical silos of work that are easier to reason about, test, and maintain. They will also be more easily refactored to remove duplication and inefficient code.

DDF-2344

N/A

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
@@ -1,0 +1,577 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p/>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p/>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.impl;
+
+import static ddf.catalog.Constants.CONTENT_PATHS;
+
+import java.io.Serializable;
+import java.nio.charset.Charset;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import javax.xml.bind.DatatypeConverter;
+
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import ddf.catalog.Constants;
+import ddf.catalog.content.StorageException;
+import ddf.catalog.content.StorageProvider;
+import ddf.catalog.content.data.ContentItem;
+import ddf.catalog.content.operation.CreateStorageRequest;
+import ddf.catalog.content.operation.CreateStorageResponse;
+import ddf.catalog.content.operation.StorageRequest;
+import ddf.catalog.content.operation.impl.CreateStorageRequestImpl;
+import ddf.catalog.content.plugin.PostCreateStoragePlugin;
+import ddf.catalog.content.plugin.PreCreateStoragePlugin;
+import ddf.catalog.data.AttributeDescriptor;
+import ddf.catalog.data.Metacard;
+import ddf.catalog.data.impl.AttributeImpl;
+import ddf.catalog.history.Historian;
+import ddf.catalog.operation.CreateRequest;
+import ddf.catalog.operation.CreateResponse;
+import ddf.catalog.operation.OperationTransaction;
+import ddf.catalog.operation.ProcessingDetails;
+import ddf.catalog.operation.impl.CreateRequestImpl;
+import ddf.catalog.operation.impl.CreateResponseImpl;
+import ddf.catalog.operation.impl.OperationTransactionImpl;
+import ddf.catalog.operation.impl.ProcessingDetailsImpl;
+import ddf.catalog.plugin.AccessPlugin;
+import ddf.catalog.plugin.PluginExecutionException;
+import ddf.catalog.plugin.PolicyPlugin;
+import ddf.catalog.plugin.PolicyResponse;
+import ddf.catalog.plugin.PostIngestPlugin;
+import ddf.catalog.plugin.PreIngestPlugin;
+import ddf.catalog.plugin.StopProcessingException;
+import ddf.catalog.source.CatalogProvider;
+import ddf.catalog.source.CatalogStore;
+import ddf.catalog.source.IngestException;
+import ddf.catalog.source.InternalIngestException;
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.util.impl.Requests;
+
+public class CreateOperations {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CreateOperations.class);
+
+    private static final Logger INGEST_LOGGER =
+            LoggerFactory.getLogger(Constants.INGEST_LOGGER_NAME);
+
+    private static final String PRE_INGEST_ERROR =
+            "Error during pre-ingest:\n\n";
+
+    private Supplier<CatalogProvider> catalogSupplier;
+
+    private Supplier<StorageProvider> storageSupplier;
+
+    // Inject properties
+    private FrameworkProperties frameworkProperties;
+
+    private QueryOperations queryOperations;
+
+    private SourceOperations sourceOperations;
+
+    private OperationsSecuritySupport opsSecuritySupport;
+
+    private OperationsMetacardSupport opsMetacardSupport;
+
+    private OperationsCrudSupport opsCrudSupport;
+
+    private Historian historian;
+
+    public CreateOperations(FrameworkProperties frameworkProperties,
+            QueryOperations queryOperations, SourceOperations sourceOperations,
+            OperationsSecuritySupport opsSecuritySupport,
+            OperationsMetacardSupport opsMetacardSupport, OperationsCrudSupport opsCrudSupport) {
+        this.frameworkProperties = frameworkProperties;
+        this.queryOperations = queryOperations;
+        this.sourceOperations = sourceOperations;
+        this.opsSecuritySupport = opsSecuritySupport;
+        this.opsMetacardSupport = opsMetacardSupport;
+        this.opsCrudSupport = opsCrudSupport;
+    }
+
+    void setCatalogSupplier(Supplier<CatalogProvider> catalogSupplier) {
+        this.catalogSupplier = catalogSupplier;
+    }
+
+    void setStorageSupplier(Supplier<StorageProvider> storageSupplier) {
+        this.storageSupplier = storageSupplier;
+    }
+
+    public void setHistorian(Historian historian) {
+        this.historian = historian;
+    }
+
+    //
+    // Delegate methods
+    //
+    CreateResponse create(CreateRequest createRequest)
+            throws IngestException, SourceUnavailableException {
+        boolean catalogStoreRequest = opsCrudSupport.isCatalogStoreRequest(createRequest);
+        queryOperations.setFlagsOnRequest(createRequest);
+
+        validateCreateRequest(createRequest);
+
+        if (Requests.isLocal(createRequest)
+                && !sourceOperations.isSourceAvailable(catalogSupplier.get())) {
+            SourceUnavailableException sourceUnavailableException = new SourceUnavailableException(
+                    "Local provider is not available, cannot perform create operation.");
+            if (INGEST_LOGGER.isWarnEnabled()) {
+                INGEST_LOGGER.warn(
+                        "Error on create operation, local provider not available. {} metacards failed to ingest. {}",
+                        createRequest.getMetacards()
+                                .size(),
+                        buildIngestLog(createRequest),
+                        sourceUnavailableException);
+            }
+            throw sourceUnavailableException;
+        }
+
+        CreateResponse createResponse = null;
+
+        Exception ingestError = null;
+        try {
+            createRequest = injectAttributes(createRequest);
+
+            setDefaultValues(createRequest);
+
+            Map<String, Serializable> unmodifiablePropertiesMap = Collections.unmodifiableMap(
+                    createRequest.getProperties());
+            HashMap<String, Set<String>> requestPolicyMap = new HashMap<>();
+            for (Metacard metacard : createRequest.getMetacards()) {
+                HashMap<String, Set<String>> itemPolicyMap = new HashMap<>();
+                for (PolicyPlugin plugin : frameworkProperties.getPolicyPlugins()) {
+                    PolicyResponse policyResponse = plugin.processPreCreate(metacard,
+                            unmodifiablePropertiesMap);
+                    opsSecuritySupport.buildPolicyMap(itemPolicyMap,
+                            policyResponse.itemPolicy()
+                                    .entrySet());
+                    opsSecuritySupport.buildPolicyMap(requestPolicyMap,
+                            policyResponse.operationPolicy()
+                                    .entrySet());
+                }
+
+                metacard.setAttribute(new AttributeImpl(Metacard.SECURITY, itemPolicyMap));
+            }
+            createRequest.getProperties()
+                    .put(PolicyPlugin.OPERATION_SECURITY, requestPolicyMap);
+
+            for (AccessPlugin plugin : frameworkProperties.getAccessPlugins()) {
+                createRequest = plugin.processPreCreate(createRequest);
+            }
+
+            createRequest.getProperties()
+                    .put(Constants.OPERATION_TRANSACTION_KEY,
+                            new OperationTransactionImpl(OperationTransaction.OperationType.CREATE,
+                                    Collections.emptyList()));
+
+            for (PreIngestPlugin plugin : frameworkProperties.getPreIngest()) {
+                try {
+                    createRequest = plugin.process(createRequest);
+                } catch (PluginExecutionException e) {
+                    LOGGER.info(
+                            "Plugin processing failed. This is allowable. Skipping to next plugin.",
+                            e);
+                }
+            }
+            validateCreateRequest(createRequest);
+
+            // Call the create on the catalog
+            LOGGER.debug("Calling catalog.create() with {} entries.",
+                    createRequest.getMetacards()
+                            .size());
+            if (Requests.isLocal(createRequest)) {
+                createResponse = catalogSupplier.get()
+                        .create(createRequest);
+                createResponse = historian.version(createResponse);
+            }
+
+            if (catalogStoreRequest) {
+                CreateResponse remoteCreateResponse = doRemoteCreate(createRequest);
+                if (createResponse == null) {
+                    createResponse = remoteCreateResponse;
+                } else {
+                    createResponse.getProperties()
+                            .putAll(remoteCreateResponse.getProperties());
+                    createResponse.getProcessingErrors()
+                            .addAll(remoteCreateResponse.getProcessingErrors());
+                }
+            }
+
+        } catch (IngestException iee) {
+            INGEST_LOGGER.warn("Ingest error", iee);
+            ingestError = iee;
+            throw iee;
+        } catch (StopProcessingException see) {
+            LOGGER.warn(PRE_INGEST_ERROR, see);
+            ingestError = see;
+            throw new IngestException(PRE_INGEST_ERROR + see.getMessage());
+        } catch (RuntimeException re) {
+            LOGGER.warn("Exception during runtime while performing create", re);
+            ingestError = re;
+            throw new InternalIngestException("Exception during runtime while performing create");
+        } finally {
+            if (ingestError != null && INGEST_LOGGER.isWarnEnabled()) {
+                INGEST_LOGGER.warn("Error on create operation. {} metacards failed to ingest. {}",
+                        createRequest.getMetacards()
+                                .size(),
+                        buildIngestLog(createRequest),
+                        ingestError);
+            }
+        }
+
+        try {
+            createResponse = validateFixCreateResponse(createResponse, createRequest);
+            for (final PostIngestPlugin plugin : frameworkProperties.getPostIngest()) {
+                try {
+                    createResponse = plugin.process(createResponse);
+                } catch (PluginExecutionException e) {
+                    LOGGER.info(
+                            "Plugin processing failed. This is allowable. Skipping to next plugin.",
+                            e);
+                }
+            }
+        } catch (RuntimeException re) {
+            LOGGER.warn(
+                    "Exception during runtime while performing doing post create operations (plugins and pubsub)",
+                    re);
+
+        }
+
+        // if debug is enabled then catalog might take a significant performance hit w/r/t string
+        // building
+        if (INGEST_LOGGER.isDebugEnabled()) {
+            INGEST_LOGGER.debug("{} metacards were successfully ingested. {}",
+                    createRequest.getMetacards()
+                            .size(),
+                    buildIngestLog(createRequest));
+        }
+        return createResponse;
+    }
+
+    CreateResponse create(CreateStorageRequest streamCreateRequest)
+            throws IngestException, SourceUnavailableException {
+        opsCrudSupport.prepareStorageRequest(streamCreateRequest,
+                streamCreateRequest::getContentItems);
+
+        Optional<String> historianTransactionKey = Optional.empty();
+
+        Map<String, Metacard> metacardMap = new HashMap<>();
+        List<ContentItem> contentItems = new ArrayList<>(streamCreateRequest.getContentItems()
+                .size());
+        HashMap<String, Path> tmpContentPaths = new HashMap<>(streamCreateRequest.getContentItems()
+                .size());
+        opsCrudSupport.generateMetacardAndContentItems(streamCreateRequest,
+                streamCreateRequest.getContentItems(),
+                metacardMap,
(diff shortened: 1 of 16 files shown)


stustison:
I assume you're going to break up the CrudOperations class even further into its operations? That is where the meat of the framework code is.

shaundmorris:
Yeah I agree @stustison. The CrudOperations class is still 2k+ (understood more refactoring on the way). Might be worth looking at just splitting query/read from create, update, delete. That might be sufficient.

coyotesqrl:
Fully intend to yank query out into its own class as well, but that will take a bit more in-depth refactoring. It'll be one of the first things I do in the second phase.

catalog/core/catalog-core-standardframework/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -370,49 +372,77 @@
         </reference-listener>
     </reference-list>
 
-    <!-- create the ddf bean -->
-    <bean id="ddf" class="ddf.catalog.impl.CatalogFrameworkImpl">
+    <!--
+        Framework properties bean
+    -->
+    <bean id="frameworkProperties" class="ddf.catalog.impl.FrameworkProperties">
```
kcwire: I wonder if we can break up this thing as well.

coyotesqrl: I didn't look at that, but maybe. I know we've been cramming a lot into those properties to drive framework logic. It's a good target for more attention.

coyotesqrl:
More restructuring done today. `QueryOperations` have been separated out from `CrudOperations` and two very lightweight beans have been added for a few shared operations.

Next up will be general cleanup, de-duplication, and investigation on cleaning up the dreaded fanout boolean.

coyotesqrl:
Tightened up a bit, moved the binding of the local catalog provider back into the `CatalogFrameworkImpl` and pass it and the fanout boolean along as needed to components.

Fanout logic still needs to be refactored more if possible and code cleanup needs to begin.

coyotesqrl:
Massive rework today, breaking CrudOperations out into their own separate silos for create, update, and delete.

The reorg feels complete at this point.

ryeats:
CatalogFrameworkImpl constructor needs javadocs update, Operations classes can be final, line 243 can be isEmpty instead of >0

coyotesqrl:
Updated the javadoc for the ctor on `CatalogFrameworkImpl` and updated all the `size() > 0` tests.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
```diff
@@ -0,0 +1,589 @@
...
+
+    public void setHistorian(Historian historian) {
+        this.historian = historian;
+    }
+
+    public void setQueryOperations(QueryOperations queryOperations) {
+        this.queryOperations = queryOperations;
+    }
+
+    public void setSourceOperations(SourceOperations sourceOperations) {
```
ryeats: Should we worry about being consistent with how all the operations are set by constructor or setter? CatalogFrameworkImpl does it by constructor and this does it through setters

coyotesqrl: That's a good point. I think I can move the operations and ops-support injections to be through ctor throughout.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
```diff
@@ -0,0 +1,589 @@
...
+                    .put(PolicyPlugin.OPERATION_SECURITY, requestPolicyMap);
+
+            for (AccessPlugin plugin : frameworkProperties.getAccessPlugins()) {
+                createRequest = plugin.processPreCreate(createRequest);
+            }
+
+            createRequest.getProperties()
+                    .put(Constants.OPERATION_TRANSACTION_KEY,
+                            new OperationTransactionImpl(OperationTransaction.OperationType.CREATE,
+                                    new ArrayList<>()));
```
ryeats: Can this be Collections.emptyList() so we can cut down on needless memory allocation?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
```diff
@@ -0,0 +1,589 @@
...
+            }
+
+        } catch (IngestException iee) {
+            INGEST_LOGGER.warn("Ingest error", iee);
+            ingestError = iee;
+            throw iee;
+        } catch (StopProcessingException see) {
+            LOGGER.warn(PRE_INGEST_ERROR, see);
+            ingestError = see;
+            throw new IngestException(PRE_INGEST_ERROR + see.getMessage());
```
ryeats: can we pass the original exception or are we trying to hide it on purpose?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
```diff
@@ -0,0 +1,589 @@
...
+        } catch (RuntimeException re) {
+            LOGGER.warn("Exception during runtime while performing create", re);
+            ingestError = re;
+            throw new InternalIngestException("Exception during runtime while performing create");
```
ryeats: same as above

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
coyotesqrl: I'd have to investigate whether that list needs to be mutable or not.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
coyotesqrl: I assume it was hidden on purpose. I did not change this functionality.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
```diff
@@ -0,0 +1,589 @@
...
+        }
+    }
+
+    /**
+     * Helper method to build ingest log strings
+     */
+    private String buildIngestLog(CreateRequest createReq) {
+        StringBuilder strBuilder = new StringBuilder();
+        List<Metacard> metacards = createReq.getMetacards();
+        final String newLine = System.getProperty("line.separator");
```
ryeats: System.lineSeparator() instead

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/DeleteOperations.java
```diff
@@ -0,0 +1,420 @@
...
+                        || !opsCrudSupport.storageIsAvailable(storageSupplier.get()))) {
+            throw new SourceUnavailableException(
+                    "Local provider is not available, cannot perform delete operation.");
+        }
+
+        DeleteStorageRequest deleteStorageRequest = null;
+
+        DeleteResponse deleteResponse = null;
+        try {
+            List<Filter> idFilters = new ArrayList<>();
```
ryeats: can we set the initial size equal to getAttributeValues size

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/DeleteOperations.java
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    private DeleteResponse doRemoteDelete(DeleteRequest deleteRequest) {
+        HashSet<ProcessingDetails> exceptions = new HashSet<>();
+        Map<String, Serializable> properties = new HashMap<>();
+
+        List<CatalogStore> stores = opsCrudSupport.getCatalogStoresForRequest(deleteRequest,
+                exceptions);
+
+        List<Metacard> metacards = new ArrayList<>();
```
ryeats: we never add to this list just replace it so we could probably use Collections.emptyList()

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/OperationsCrudSupport.java
```diff
@@ -0,0 +1,503 @@
...
+                LOGGER.error("Unable to remove temporary content for id: " + storageRequest.getId(),
+                        e1);
+            }
+        }
+        throw new IngestException("Unable to store products for request: " + id, ex);
+
+    }
+
+    void commitAndCleanup(StorageRequest storageRequest, Optional<String> historianTransactionKey,
+            HashMap<String, Path> tmpContentPaths) {
```
ryeats: Can be a map

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/DeleteOperations.java
coyotesqrl: I'm just very tentative about changing collections that get passed around that much into immutable versions without a full analysis. We have getters returning the collections, not copies of them, in several places. This collection gets passed into a ctor for one object, it has a getter for the collection which gets passed into _other_ ctors in other places. I like the idea of swapping it out but in practice it concerns me.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/OperationsCrudSupport.java
```diff
@@ -0,0 +1,503 @@
...
+                metacardMap.put(metacard.getId(), metacard);
+
+                ContentItem generatedContentItem = new ContentItemImpl(metacard.getId(),
+                        com.google.common.io.Files.asByteSource(tmpPath.toFile()),
+                        mimeTypeRaw,
+                        fileName,
+                        size,
+                        metacard);
+                contentItems.add(generatedContentItem);
+            } catch (Exception e) {
```
ryeats: We are using IngestException to break out of the loop then wrapping it in another IngestException just to do cleanup, their has to be a better way.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/OperationsCrudSupport.java
```diff
@@ -0,0 +1,503 @@
...
+                        .forEach(path -> FileUtils.deleteQuietly(path.toFile()));
+                tmpContentPaths.clear();
+                throw new IngestException("Could not create metacard.", e);
+            }
+        }
+    }
+
+    List<CatalogStore> getCatalogStoresForRequest(Request request,
+            Set<ProcessingDetails> exceptions) {
+        List<CatalogStore> results = new ArrayList<>();
```
ryeats: can this be intilized?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/OperationsCrudSupport.java
```diff
@@ -0,0 +1,503 @@
...
+
+    boolean isCatalogStoreRequest(Request request) {
+        return request != null && CollectionUtils.isNotEmpty(request.getStoreIds()) && (
+                request.getStoreIds()
+                        .size() > 1 || catalogSupplier.get() == null || !request.getStoreIds()
+                        .contains(catalogSupplier.get()
+                                .getId()));
+    }
+
+    boolean storageIsAvailable(StorageProvider storageProvider) {
```
ryeats: nit pick but isStorageAvailable rather than storageIsAvailable

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/OperationsCrudSupport.java
```diff
@@ -0,0 +1,503 @@
...
+            throw new MetacardCreationException(causeMessage.toString());
+        }
+
+        if (id != null) {
+            generatedMetacard.setAttribute(new AttributeImpl(Metacard.ID, id));
+        } else {
+            generatedMetacard.setAttribute(new AttributeImpl(Metacard.ID,
+                    UUID.randomUUID()
+                            .toString()
+                            .replaceAll("-", "")));
```
ryeats: Change to regex with a static final pattern

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/QueryOperations.java
```diff
@@ -0,0 +1,622 @@
...
+    }
+
+    void setCatalogSupplier(Supplier<CatalogProvider> catalogSupplier) {
+        this.catalogSupplier = catalogSupplier;
+    }
+
+    //
+    // Delegate methods
+    //
+    public QueryResponse query(QueryRequest fedQueryRequest, boolean fanoutEnabled)
```
ryeats: can these be made private?

coyotesqrl: You mean package-private? Should be no problem with that.

ryeats: Yeah that would work, I should have explained that at a glance it just seemed like fanoutEnabled was only the concern of this class so it shouldn't be exposed if thats an invalid assumption then it can stay public also.

coyotesqrl: I updated all the delegate methods to be package-private.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
```diff
@@ -0,0 +1,577 @@
...
+import ddf.catalog.util.impl.Requests;
+
+public class CreateOperations {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CreateOperations.class);
+
+    private static final Logger INGEST_LOGGER =
+            LoggerFactory.getLogger(Constants.INGEST_LOGGER_NAME);
+
+    private static final String PRE_INGEST_ERROR =
+            "Error during pre-ingest service invocation:\n\n";
```
shaundmorris: We should probably update this message while we are at it. Before pre-ingest plugins, we had pre-ingest services, so this is a holdover from that.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/ResourceOperations.java
```diff
@@ -0,0 +1,750 @@
...
+    public ResourceResponse getResource(ResourceRequest resourceRequest, boolean isEnterprise,
+            String resourceSiteName, boolean fanoutEnabled)
+            throws IOException, ResourceNotFoundException, ResourceNotSupportedException {
+        ResourceResponse resourceResponse = null;
+        ResourceRequest resourceReq = resourceRequest;
+        String resourceSourceName = resourceSiteName;
+        ResourceRetriever retriever = null;
+
+        if (fanoutEnabled) {
+            isEnterprise = true;
```
ryeats: this is a little weird you don't actually need isEnterprise since it will always be equivalent to fanoutEnabled?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
coyotesqrl: I'm just going to cut off the `service invocation` part of the message if that's okay.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/ResourceOperations.java
```diff
@@ -0,0 +1,750 @@
...
+        LOGGER.debug("Returning resourceURI: {}", resourceUri);
+        if (resourceUri == null) {
+            throw new ResourceNotFoundException(DEFAULT_RESOURCE_NOT_FOUND_MESSAGE);
+        }
+
+        return new ResourceInfo(metacard, resourceUri);
+    }
+
+    protected static class ResourceInfo {
+        private Metacard metacard;
```
ryeats: These could be final

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/ResourceOperations.java
coyotesqrl: It is a little weird, but I can see why it was that way before (and why I'm going to keep it now). `isEnterprise` does get updated further down in the method - line 357 - so in order to be semantically consistent I think it's better to keep the value. I could update the line to just set `isEnterprise = fanoutEnabled`, however.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/SourceOperations.java
```diff
@@ -0,0 +1,335 @@
...
+    // Helper methods - check visibility
+    //
+
+    /**
+     * Checks that the specified source is valid and available.
+     *
+     * @param source the {@link Source} to check availability of
+     * @return true if the {@link Source} is available, false otherwise
+     */
+    boolean sourceIsAvailable(Source source) {
```
ryeats: another nitpick isSourceAvailable

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/SourceOperations.java
```diff
@@ -0,0 +1,335 @@
...
+    private SourceInfoResponse getFanoutSourceInfo(SourceInfoRequest sourceInfoRequest)
+            throws SourceUnavailableException {
+
+        SourceInfoResponse response;
+        SourceDescriptorImpl sourceDescriptor;
+        try {
+
+            // request
+            if (sourceInfoRequest == null) {
+                throw new IllegalArgumentException("SourceInfoRequest was null");
```
ryeats: Exception is being used for flow control, granted its on an error path but it could be cleaner.

coyotesqrl: Yeah. It's a little raw there. Might try to clean that up in phase 2.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/ResourceOperations.java
```diff
@@ -0,0 +1,750 @@
...
+
+                if (sourceIdOfResult != null && sourceIdOfResult.equals(getId())) {
+                    // found entry on local source
+                    supportedOptions = getOptionsFromLocalProvider(metacard);
+                } else if (sourceIdOfResult != null && !sourceIdOfResult.equals(getId())) {
+                    // found entry on federated source
+                    supportedOptions = getOptionsFromFederatedSource(metacard, sourceIdOfResult);
+                }
+            } else {
+                String message = "Unable to find metacard " + metacardId + " on enterprise.";
```
shaundmorris: I'm assuming you are tracking these, (maybe on the next round of the refactoring) to use something like `String.format()`.

coyotesqrl: Tracking, no. Expecting to cleanup, yes. 👍

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/TransformOperations.java
```diff
@@ -0,0 +1,94 @@
...
+    //
+    BinaryContent transform(Metacard metacard, String transformerId,
+            Map<String, Serializable> requestProperties) throws CatalogTransformerException {
+
+        ServiceReference[] refs;
+        try {
+            // TODO replace shortname with id
+            refs = frameworkProperties.getBundleContext()
+                    .getServiceReferences(MetacardTransformer.class.getName(),
+                            "(|" + "(" + Constants.SERVICE_SHORTNAME + "=" + transformerId + ")"
```
ryeats: could use String.format instead then we can use it below also

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/UpdateOperations.java
```diff
@@ -0,0 +1,478 @@
...
+                    "Local provider is not available, cannot perform update operation.");
+        }
+
+        UpdateResponse updateResponse = null;
+        try {
+            injectAttributes(updateRequest);
+
+            setDefaultValues(updateRequest);
+
+            List<Filter> idFilters = new ArrayList<>();
```
ryeats: Could set the initial capacity to mach the updates size

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/SourceOperations.java
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+            Set<SourceDescriptor> sourceDescriptors = new LinkedHashSet<>();
+            Set<String> ids = sourceInfoRequest.getSourceIds();
+
+            // Only return source descriptor information if this sourceId is
+            // specified
+            if (ids != null && !ids.isEmpty()) {
```
rzwiefel: Clean this up later.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/SourceOperations.java
```diff
@@ -0,0 +1,335 @@
...
+     * {@link Source}.
+     *
+     * @param sources {@link Collection} of {@link Source} to obtain descriptor information from
+     * @return new {@link Set} of {@link SourceDescriptor}
+     */
+    private Set<SourceDescriptor> getFederatedSourceDescriptors(Collection<FederatedSource> sources,
+            boolean addCatalogProviderDescriptor) {
+        SourceDescriptorImpl sourceDescriptor;
+        Set<SourceDescriptor> sourceDescriptors = new HashSet<>();
+        if (sources != null) {
```
rzwiefel: Clean up later. 268 onward could surely (hopefully, possibly, maybe..) be cleaned up by some good optionals or streaming.

rzwiefel:
:+1: :shipit:. Still more to do so i'm pro expediting this commit so we can continue moving forward.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
pklinef: It was intentional.  The exception is logged but not returned to the endpoint.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CreateOperations.java
pklinef: The transaction previous metacard state list will be immutable.

https://github.com/codice/ddf/blob/6723c7c87888afa1848d046d442345732139931e/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/OperationTransactionImpl.java#L37

pklinef:
I am not sure why `CatalogFrameworkImpl.registerBasicMetacard` was not done in blueprint.  Something to consider trying.

coyotesqrl:
@pklinef I'm going to hold off on moving the Metacard registration until phase 2. Should be simple enough, but it's enough of a change I don't want to block this merge for it.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/ResourceOperations.java
```diff
@@ -0,0 +1,750 @@
...
+import ddf.catalog.resource.download.DownloadException;
+import ddf.catalog.resourceretriever.LocalResourceRetriever;
+import ddf.catalog.resourceretriever.RemoteResourceRetriever;
+import ddf.catalog.resourceretriever.ResourceRetriever;
+import ddf.catalog.source.CatalogProvider;
+import ddf.catalog.source.FederatedSource;
+import ddf.catalog.source.UnsupportedQueryException;
+import ddf.catalog.util.impl.DescribableImpl;
+
+// TODO RAP 28 Jul 16: Really extend describableimpl?
```
pklinef: Please clean up the TODOs and include the ticket that it will be worked under.

catalog/core/catalog-core-standardframework/src/main/resources/OSGI-INF/blueprint/blueprint.xml
pklinef: I wonder if we can even break this blueprint file into multiple blueprint files.

pklinef:
Can we find a way to clean this up?  For common references like the catalog and storage providers, maybe we can create a composed object like we did with framework properties.  That should also remove a lot of boilerplate setters.

https://github.com/coyotesqrl/ddf/blob/f061accbe6d67b38aa1976f55e3897a14de3a7d3/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CatalogFrameworkImpl.java#L141-L165

pklinef:
Can we move the new *Operations classes into a new package?

pklinef:
We should be able to pull all of the fanout related logic into a new class.  It would be nice to have all of that in one place going forward.

pklinef:
👍

shaundmorris:
👍  I'm good for now on this one since we were looking primarily at the general structure for phase 1. I'm anxiously awaiting phase 2.

catalog/core/catalog-core-standardframework/src/main/resources/OSGI-INF/blueprint/blueprint.xml
stustison: My gut feel is that if you do try to break this thing apart, we'll just end up with the same properties in multiple objects as opposed to everything in this one object. It would also make it annoying down the road if you ever needed to add something to all of the operations.

stustison:
👍

ryeats:
👍  hero'd

PR #997, opened 2016-07-13
DDF-2298: Registry name field now auto-populates

brianfelix:
#### What does this PR do?
When connecting to a remote registry, the name field has been removed. A placeholder name is now determined by the hostname and port number of the service url until a connection to the remote registry can be made. Once the connection has been made, the name field is assigned the name of the remote registry.
#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged; if a component team is listed, at least one of its members needs to approve)?
@clockard @gordocanchola @vinamartin  @ani6gup @mcalcote 
#### How should this be tested?
You will need 2 DDF's with security or use loopback.
Setup a remote registry using the url of the other ddf. The name will populate with the hostname and port number of the remote registry.
Once the other DDF becomes available, the name field will change to be the remote registries name given by the admin.
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

When a user subscribes to a remote registry, the name is determined before connecting
using the hostname and port number of the given service url. Upon successful
connection to the remote registry, the remote name assigned to the registry is used
as it's name.

PR #1067, opened 2016-07-29
DDF-2298: Registry name field now auto populates

brianfelix:
#### What does this PR do?

When connecting to a remote registry, the name field has been removed. A placeholder name is now determined by the hostname and port number of the service url until a connection to the remote registry can be made. Once the connection has been made, the name field is assigned the name of the remote registry.

@clockard @gordocanchola @vinamartin @mcalcote

@pklinef 
@shaundmorris 

You will need 2 DDF's with security or use loopback.
Setup a remote registry using the url of the other ddf. The name will populate with the hostname:port of the remote registry and connection type.
Once the other DDF becomes available, the name field will change to be the remote registries name given by the admin + hostname:port + connection type.
Make sure that no duplicate values are entered and that the UI correctly allows you to edit the url information. Check that the UI disallows urls of the same connection type that are also on the same hostname:port (ex: the UI does not allow two csw connections over the same hostname:port).

This PR was mistakenly closed. Please see https://github.com/codice/ddf/pull/997 for previous conversations on the PR.

DDF-2298
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

When a user subscribes to a remote registry, the name is determined before connecting
using the hostname and port number of the given service url. Upon successful
connection to the remote registry, the remote name assigned to the registry is used
as part of it's name.


catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/Registry.view.js
```diff
@@ -70,9 +70,42 @@ function (ich,Marionette,_,$,Q,ModalConfig,ModalRegistry,EmptyView,wreqr,Utils,d
...
         updateRemoteStatus: function() {
-            if(this.remoteStatusModel.get('value')){
-                this.model.set('remoteName', this.remoteStatusModel.get('value').remoteName);
+            //collection of configurations which has all models representing every entry in the registryRows.
+            var allConfigurations = this.model.collection.models;
+
+            //ConfigIds has values that are all unique since they cant be duplicated locally (there is a dup check on creation locally)
+            // (only can be dupicated whena remote uses a name that is already used locally
+            var configIds = {};
+            for(var i =0; i<allConfigurations.length; i++){
```
stustison: you should be doing `this.model.collection.each(function(item) { ... });`

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/Registry.view.js
```diff
@@ -70,9 +70,42 @@ function (ich,Marionette,_,$,Q,ModalConfig,ModalRegistry,EmptyView,wreqr,Utils,d
...
+                if(currentConfig.id !== currentConfig.remoteName) {
+
+                    //if configIds already contains the current configs remoteName and it's id!== remoteName, it is a duplicate
+                    if (configIds.hasOwnProperty(currentConfig.remoteName)) {
+                        this.model.attributes.attributes.id = currentConfig.remoteName + " (" + currentConfig.shortname + ")";
+                    } else {
+                        //This indicates that id is being set for the first time (new connection) or its remotename is already taken by another config
+                        //When id !== "remoteName (shortname)", this config is being set for the first time
+                        if(currentConfig.id !== currentConfig.remoteName + " (" + currentConfig.shortname + ")") {
+                            this.model.attributes.attributes.remoteName =  currentConfig.remoteName;
```
stustison: do not hit the backbone models directly, you should be using `this.model.set('remoteName', val)`

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/Registry.view.js
stustison: Or you could just do `var configIds = _.uniq(this.model.collection.pluck('id')'` and do it all on one line...

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/templates/registryRow.handlebars
```diff
@@ -13,10 +13,10 @@
  --}}
 <td>
     <a href='#{{id}}' class='newLink' data-toggle="modal">
-        {{#if id}}
-            {{id}}
+        {{#if attributes.id}}
```
stustison: This doesn't look right to me. Attributes is an internal backbone construct, why do you need to call it out in the template? If you're doing a toJSON on the model, it will strip out the internal stuff like `attributes`

catalog/spatial/registry/registry-api-impl/pom.xml
```diff
@@ -246,22 +246,22 @@
...
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.85</minimum>
+                                            <minimum>0.75</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.72</minimum>
+                                            <minimum>0.58</minimum>
```
vinamartin: Ooof, this is a pretty big drop. Any reason why this got so low?

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -90,7 +95,7 @@
 
     private String registryId = "";
 
-    private String remoteName = "";
+    private String id = "";
```
clockard: Shouldn't need a private id string since this class inherits one from DescrbableImpl

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -250,8 +261,9 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
             if (registryId.equals(RegistryUtility.getRegistryId(singleResult.getMetacard()))) {
                 String metacardTitle = singleResult.getMetacard()
                         .getTitle();
-                if (metacardTitle != null && !remoteName.equals(metacardTitle)) {
-                    remoteName = metacardTitle;
+
+                if (metacardTitle != null && !id.equals(metacardTitle)) {
+                    id = metacardTitle;
```
clockard: should call setId(metacardTitle) instead of using local id

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -335,7 +343,7 @@ void registryInfoQuery() throws UnsupportedQueryException {
         SourceResponse identityMetacard = query(queryRequest);
         if (identityMetacard.getResults()
                 .size() > 0) {
-            remoteName = identityMetacard.getResults()
+            id = identityMetacard.getResults()
```
clockard: Same here

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -351,11 +359,53 @@ private void updateConfiguration() {
         try {
             Configuration currentConfig = configAdmin.getConfiguration(currentPid);
             Dictionary<String, Object> currentProperties = currentConfig.getProperties();
-            currentProperties.put(REMOTE_NAME, remoteName);
+            Configuration[] configsWithDupId = configAdmin.listConfigurations(String.format(
+                    "(id=%s)",
+                    (id)));
```
clockard: Basically everywhere you are using `id` switch it to using the getters/setters

catalog/spatial/registry/registry-api-impl/pom.xml
brianfelix: I agree and am currently working on raising this number. After removing 'remoteName' from registryStoreImpl, I had to remove a quite few tests that were only being used to test that remoteName was being set correctly. This is the main cause of the drop

clockard: Ya, we need to add some tests to cover the new functionality

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/ModalRegistry.view.js
```diff
@@ -15,264 +15,258 @@
...
-            this.mode = options.mode;
-            this.type = options.registryType;
-        },
-        onRender: function() {
+        ModalRegistry.View = Marionette.Layout.extend({
+            template: 'modalRegistry',
+            className: 'modal',
+            events: {
+                "click .submit-button": "submitData",
+                "click .cancel-button": "cancel",
```
mcalcote: It looks like "cancel" should be "closeAndUnbind"

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,81 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
             Dictionary<String, Object> currentProperties = currentConfig.getProperties();
-            currentProperties.put(REMOTE_NAME, remoteName);
-            currentProperties.put(RegistryObjectMetacardType.REGISTRY_ID, registryId);
-            currentConfig.update(currentProperties);
+            URI uri;
+            String registryUrl = "";
+            if (currentProperties.get("registryUrl") != null) {
+                registryUrl = (String) currentProperties.get("registryUrl");
+            } else if (currentProperties.get("url") != null) {
+                LOGGER.warn("current properties had url");
```
mcalcote: Maybe more of a debug level message? Also could provide more info on what is happening.

brianfelix: Nice catch. These were actually only supposed to be temporary log messages while I debugged a test that was passing locally but failing on bamboo. These log messages will be removed

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,81 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+                registryUrl = (String) currentProperties.get("url");
+            }
+            if (registryUrl != null) {
+                LOGGER.warn("registryUrl was {}", registryUrl);
```
mcalcote: Also shouldn't be warn

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,81 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+                try {
+                    uri = new URI(registryUrl);
+                } catch (URISyntaxException e) {
+                    LOGGER.error(
```
mcalcote: Warn?

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,81 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+                            "Unable to update registry configurations. Syntax error on registryUrl",
+                            e);
+                    return;
+                }
+                StringBuilder nameToSet = new StringBuilder(metacardTitle + " (");
+                StringBuilder hostnamePort = new StringBuilder(uri.getHost());
+                if (uri.getPort() != -1) {
+                    hostnamePort.append(":" + uri.getPort());
+                }
+                LOGGER.warn("hostnamePort is {}", hostnamePort.toString());
```
mcalcote: https://codice.atlassian.net/wiki/display/DDF/Log+Levels

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,76 @@ void registryInfoQuery() throws UnsupportedQueryException {
         SourceResponse identityMetacard = query(queryRequest);
         if (identityMetacard.getResults()
                 .size() > 0) {
-            remoteName = identityMetacard.getResults()
+            String metacardTitle = identityMetacard.getResults()
                     .get(0)
                     .getMetacard()
                     .getTitle();
+            setId(identityMetacard.getResults()
```
clockard: Just use the metacardTitle string here instead of pulling it out of the metacard.

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,76 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
             Configuration currentConfig = configAdmin.getConfiguration(currentPid);
             Dictionary<String, Object> currentProperties = currentConfig.getProperties();
-            currentProperties.put(REMOTE_NAME, remoteName);
-            currentProperties.put(RegistryObjectMetacardType.REGISTRY_ID, registryId);
-            currentConfig.update(currentProperties);
+            URI uri;
+            String registryUrl = "";
+            if (currentProperties.get("registryUrl") != null) {
+                registryUrl = (String) currentProperties.get("registryUrl");
+            } else if (currentProperties.get("url") != null) {
```
clockard: Why are we checking for this other property 'url'

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,76 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
         String currentPid = getConfigurationPid();
         try {
             Configuration currentConfig = configAdmin.getConfiguration(currentPid);
             Dictionary<String, Object> currentProperties = currentConfig.getProperties();
-            currentProperties.put(REMOTE_NAME, remoteName);
-            currentProperties.put(RegistryObjectMetacardType.REGISTRY_ID, registryId);
-            currentConfig.update(currentProperties);
+            URI uri;
+            String registryUrl = "";
+            if (currentProperties.get("registryUrl") != null) {
```
clockard: Should use REGISTRY_URL constant here and below

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,76 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+            if (registryUrl != null) {
+                try {
+                    uri = new URI(registryUrl);
+                } catch (URISyntaxException e) {
+                    LOGGER.error(
+                            "Unable to update registry configurations. Syntax error on registryUrl",
+                            e);
+                    return;
+                }
+                StringBuilder nameToSet = new StringBuilder(metacardTitle + " (");
```
clockard: I feel like this section could be simplified with a String.format and or some helper method.

clockard: String port = "";
if(uri.getPort() != -1){
   port = ":"+uri.getPort();
}
String newId = String.format("%s (%s%s) %s", metacardTitle, uri.getHost(), port, getConnectionTypeHelper())

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/ModalRegistry.view.js
```diff
@@ -15,264 +15,292 @@
...
+                var url = evt.target.value;
+                var hostname;
+                if (url) {
+                    hostname = this.extractHostnameFromUrl(url);
+                }
+                if (hostname && this.urlIsValid(url)) {
+                    var oldName;
+                    var nameToSet = hostname;
+                    var toCheck = "(" + hostname + ")";
+                    if (!this.type || this.type === "CSW Registry Store") {
```
clockard: Think it would be cleaner to just use the `type` here instead of converting one particular type to a special value.

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/ModalRegistry.view.js
```diff
@@ -15,264 +15,292 @@
...
+                        nameToSet += " (CSW)";
+                        toCheck += " (CSW)";
+                    } else {
+                        nameToSet += " (" + this.type + ")";
+                        toCheck += " (" + this.type + ")";
+                    }
+                    var count = 0;
+                    this.registry.get('collection').models.find(function (registry) {
+                        if (registry.id) {
+                            if (registry.id.endsWith(nameToSet) || registry.id.endsWith(nameToSet + ")") || registry.id.endsWith(toCheck) || registry.id.endsWith(toCheck + ")")) {
```
clockard: It feels like we only need one variable here nameToSet or toCheck. Since we are creating the name we should only need to check one case.

catalog/spatial/registry/registry-admin-modules/registry-admin-remote-ui/src/main/webapp/js/view/ModalRegistry.view.js
```diff
@@ -15,264 +15,292 @@
...
+                                count++;
+                            }
+                        }
+                    });
+                    oldName = this.model.get('id');
+                    if (count === 0 || (this.mode === 'edit' && ( oldName.endsWith(nameToSet) || oldName.endsWith(nameToSet + ")") || oldName.endsWith(toCheck) || oldName.endsWith(toCheck + ")") ) && count === 1)) {
```
clockard: See above comment.

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +298,76 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+                StringBuilder hostnamePort = new StringBuilder(uri.getHost());
+                if (uri.getPort() != -1) {
+                    hostnamePort.append(":" + uri.getPort());
+                }
+                String oldName = (String) currentProperties.get("id");
```
clockard: I know we talked about doing something like this to get the type name but now that I am looking at it I think it would be better just to just create a local constant variable and use it in this construction. It will greatly simplify this bit of code.

mcalcote:
Manually tested on IE & Chrome LGTM 👍

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -293,28 +305,88 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
                     .get(0)
                     .getMetacard());
+            updateConfiguration(metacardTitle);
         }
-        updateConfiguration();
+
     }
 
-    private void updateConfiguration() {
+    private String getObjectClassDefinition(Bundle bundle, String pid) {
```
clockard: This method should be renamed since it doesn't return the ObjectClassDefinition anymore

catalog/spatial/registry/registry-api-impl/src/test/resources/csw-full-registry-package.xml
```diff
@@ -240,7 +240,7 @@
             <rim:ServiceBinding id="urn:registry:federation:method:csw"
                                 service="urn:uuid:service:2014ca7f59ac46f495e32b4a67a51276">
                 <!--Slot defining the query url for this particular method of communicating with this instance-->
-                <rim:Slot name="cswUrl" slotType="xs:anyURI">
+                <rim:Slot name="registryUrl" slotType="xs:anyURI">
```
clockard: This isn't a registry store but properties for a csw source so its name should stay 'cswUrl'

vinamartin:
LGTM 👍

shaundmorris:
@brianfelix there are merge conflicts to address here

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -332,28 +344,88 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+        Locale locale = Locale.getDefault();
+        if (bundle != null) {
+            if (metaTypeService != null) {
+                MetaTypeInformation mti = metaTypeService.getMetaTypeInformation(bundle);
+                if (mti != null) {
+                    try {
+                        return mti.getObjectClassDefinition(pid, locale.toString())
+                                .getName();
+                    } catch (IllegalArgumentException e) {
+                        // MetaTypeProvider.getConnectionType might throw illegal
```
shaundmorris: Do we need to log at debug when this happens? Doesn't look like it currently and you still get the config showing up based on the comment

catalog/spatial/registry/registry-api-impl/src/test/java/org/codice/ddf/registry/api/impl/TestRegistryStore.java
```diff
@@ -153,17 +161,27 @@ protected SourceResponse query(QueryRequest queryRequest, ElementSetType element
...
         });
 
         registryStore.setFilterBuilder(filterBuilder);
         registryStore.setFilterAdapter(filterAdapter);
         registryStore.setConfigAdmin(configAdmin);
+        registryStore.setMetacardMarshaller(new MetacardMarshaller(parser));
         registryStore.setSchemaTransformerManager(transformer);
         registryStore.setAutoPush(true);
+        properties = new Hashtable<>();
+        properties.put("id", "registryId");
```
shaundmorris: You could use the new constants here from RegistryStoreImpl

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -332,28 +344,88 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
         try {
             Configuration currentConfig = configAdmin.getConfiguration(currentPid);
             Dictionary<String, Object> currentProperties = currentConfig.getProperties();
-            currentProperties.put(REMOTE_NAME, remoteName);
+            URI uri;
+            String registryUrl = (String) currentProperties.get(REGISTRY_URL);
+            try {
+                uri = new URI(registryUrl);
+            } catch (URISyntaxException e) {
+                LOGGER.error("Unable to update registry configurations. Syntax error on registryUrl",
```
brianfelix: Would the info level be more appropriate? Failing an update is an issue that might be important to the user but on the other hand, there is a stack trace to log

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -332,28 +344,88 @@ void registryInfoQuery() throws UnsupportedQueryException {
...
+                        e);
+                return;
+            }
+            String port = "";
+            if (uri.getPort() != -1) {
```
brianfelix: java.net.URI.getPort() returns -1 if the port number is not defined in the URL that was given. So if the registryUrl has a port, ":" + uri.getPort() is appended to the nameToSet. Otherwise, the nameToSet string building continues normally but without the port appended. Should I set a string constant with a descriptive name to -1 and check uri.getPort() against it instead?

shaundmorris: That was my suggestion, maybe it is more obvious to others. Probably would make it more readable.

brianfelix: I agree completely. Fixed

mcalcote:
Tested latest on IE, FF & Chrome :shipit:

brianfelix:
Hero build passed. Verified by @mcalcote

ricklarsen:
Should we include a note about this is in the Managing Registry documentation?

PR #1070, opened 2016-08-01
DDF-2328 Aligned expanded taxonomy in the GMD Transformer and added additional mappings

bdeining:
#### What does this PR do?
This PR aligns the GMD transformer to the expanded taxonomy.  It also adds additional mappings that were overlooked during initial development, and refactors some of the GMD code for clarity.
@glenhein @jrnorth 
@jlcsmith
@kcwire
Build and install DDF.  Configure a GMD federated source (http://csw.data.gov.uk/geonetwork/srv/en/csw), send out a query and investigate the metacards.  
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests


catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -146,28 +215,49 @@ public Metacard transform(InputStream inputStream)
...
 
-        String xml = null;
-        XstreamPathValueTracker pathValueTracker = null;
-        xml = IOUtils.toString(inputStream);
+    private Metacard handleTransform(InputStream inputStream, String id)
+            throws CatalogTransformerException {
+        String xml;
+        XstreamPathValueTracker pathValueTracker;
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+            readInputStreamIntoByteArrayOutputStream(inputStream, baos);
```
jrnorth: Could also use `IOUtils.copy(inputStream, baos)`.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/GmdConstants.java
```diff
@@ -0,0 +1,256 @@
...
+
+    public static final String REVISION = "revision";
+
+    public static final String CREATION = "creation";
+
+    public static final String EXPIRY = "expiry";
+
+    public static final String PUBLICATION = "publication";
+
+    public static final String RESOURCE_TYPE = "ext.resource-type";
```
jlcsmith: this should be removed in favor of dataset from the Core attribute group.  Also need to set the deprecated metadata-content-type attribute to match.

jlcsmith:
Built and tested.  Other than my few comments, this LGTM 👍

catalog/spatial/csw/spatial-csw-source-common/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/TestCswFilterDelegate.java
```diff
@@ -3400,4 +3407,12 @@ public String toString() {
         }
     }
 
+    private static MetacardType getGmdMetacardType() {
+        return new MetacardTypeImpl("gmdMetacardType", Arrays.asList(new CoreAttributes(),
```
glenhein: Isn't CoreAttributes already being added: https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardTypeImpl.java#L123

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/CswUnmarshallHelper.java
```diff
@@ -76,7 +77,7 @@ public static Date convertToDate(String value) {
         // the time zone and time are optional.
```
stustison: Sounds like this comment is no longer correct. Is UTC always the correct time zone? Do you need to parse first to determine if it is local and then fall back on UTC otherwise?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
stustison: We should really never be reading any externally obtained input stream to a byte array. What if someone passes in a XXGB stream? If you can guarantee the length, that might be different. However, in the case where it is ok to go InputStream -> Byte Array, then Joe is absolutely correct, use IOUtils.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+        List<Double> altitudes = covertStringListToDoubleList(serviceExAltitude);
+        if (CollectionUtils.isNotEmpty(altitudes)) {
+            metacard.setAttribute(Location.ALTITUDE, (Serializable) altitudes);
         }
+    }
 
+    private void addMetacardTemporalStartStop(final XstreamPathValueTracker pathValueTracker,
+            MetacardImpl metacard) {
+        List<String> temporalStartList =
+                pathValueTracker.getAllValues(toPath(GmdConstants.TEMPORAL_START_PATH));
```
glenhein: I know that datetime.start and .end are multivalued, but why?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+    public void addMetacardCrs(final XstreamPathValueTracker pathValueTracker,
             MetacardImpl metacard) {
-        String title = pathValueTracker.getFirstValue(toPath(GmdMetacardType.TITLE_PATH));
-        if (StringUtils.isNotEmpty(title)) {
-            metacard.setTitle(title);
+        List<String> crsCodes = new ArrayList<>();
+        List<String> types = pathValueTracker.getAllValues(toPath(GmdConstants.CRS_CODE_PATH));
+        List<String> codes = pathValueTracker.getAllValues(toPath(GmdConstants.CRS_AUTHORITY_PATH));
+        if (CollectionUtils.isNotEmpty(codes) && CollectionUtils.isNotEmpty(types)) {
+            for (int i = 0; i < codes.size(); i++) {
```
stustison: `i < codes.size() && i < types.size()`

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+            if (dataList != null && dataList.getLength() > 0) {
+                List<String> dateList = new ArrayList<>();
+                List<String> dateTypes =
+                        pathValueTracker.getAllValues(toPath(GmdConstants.CITATION_DATE_TYPE_PATH));
+
+                for (int i = 0; i < dataList.getLength(); i++) {
+                    Node node = dataList.item(i);
+                    String datestring = node.getTextContent()
+                            .trim()
+                            .replaceAll("(\\s|\\t)+", " ")
```
stustison: You should make these compiled regex for performance.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+                    e);
+            return;
         }
+    }
 
-        if (subjects.size() > 0) {
-            metacard.setAttribute(GmdMetacardType.GMD_SUBJECT, (Serializable) subjects);
+    private void setDates(List<String> dates, List<String> dateTypes, Metacard metacard) {
+        for (int i = 0; i < dates.size(); i++) {
+            if (i < dates.size() && i < dateTypes.size()) {
```
stustison: Just put this into the for statement instead of a separate if inside the for. Then you'll exit the loop early if this condition is unmet.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+                coordinates.add(new Coordinate(Double.parseDouble(stringArray[i]),
+                        Double.parseDouble(stringArray[i + 1])));
+            }
+            String wkt = WKT_WRITER_THREAD_LOCAL.get()
+                    .write(factory.createLineString(coordinates.toArray(new Coordinate[coordinates.size()])));
+            if (wkt != null) {
+                metacard.setAttribute(Core.LOCATION, wkt);
+            }
+        } catch (NumberFormatException nfe) {
+            LOGGER.debug("Unable to parse double in {}. Metacard location will not be set.",
```
glenhein: Should this be a warn with the exception?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+
+            LinearRing linearRing =
+                    factory.createLinearRing(coordinates.toArray(new Coordinate[coordinates.size()]));
+            String wkt = WKT_WRITER_THREAD_LOCAL.get()
+                    .write(factory.createPolygon(linearRing, null));
+            if (wkt != null) {
+                metacard.setAttribute(Core.LOCATION, wkt);
+            }
+        } catch (NumberFormatException nfe) {
+            LOGGER.debug("Unable to parse double in {}. Metacard location will not be set.",
```
glenhein: Should this be a warn with the exception?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
+                    if (wkt != null) {
+                        metacard.setAttribute(new AttributeImpl(Core.LOCATION, wkt));
+                    }
+                }
+            }
 
-        List<String> subjects = new ArrayList<>();
-        if (CollectionUtils.isNotEmpty(keywords)) {
-            subjects.addAll(keywords);
+        } catch (ParserConfigurationException | IOException | SAXException | XPathExpressionException e) {
```
glenhein: You aren't catching NumberFormatException

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
jlcsmith: I'd rather not see this sort of exception logged at warn. This could easily fill up the logs if we connect to a source with bad geometry data.  The result here is that location wouldn't be set on the metacard, but the transform still succeeds.  I'd prefer to leave this at debug.

bdeining: Yes, this was left as debug because it could technically happen 250 times on a query to a federated source and spam the logs

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
glenhein: \s already includes \t and \n: From java: \s => A whitespace character: [ \t\n\x0B\f\r]

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -210,187 +306,596 @@ private void copyArgumentsToContext(MarshallingContext context,
...
-            } catch (NumberFormatException nfe) {
-                LOGGER.info("Unable to parse double from GMD metadata {}, {}, {}, {}",
-                        westLon,
-                        eastLon,
-                        southLat,
-                        northLat);
+                Double altitudeDouble = Double.parseDouble(string);
+                doubles.add(altitudeDouble);
+            } catch (NumberFormatException e) {
+                LOGGER.debug("Unable to parse {} into double.  Skipping entry.", string);
```
glenhein: should this be a warn?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: Changed to use `IOUtils`.  While there is no guarantee of the size of the input stream, GMD files are xml and aren't likely to be XXGB.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/GmdConstants.java
bdeining: Removed this constant in favor of the new Core attribute.  `metadata-content-type` is already being set on the transformer.

catalog/spatial/csw/spatial-csw-source-common/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/TestCswFilterDelegate.java
bdeining: Yes, removed.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: Changed

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: Made these patterns and removed `\\t`

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: Done

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: Added catch for `NumberFormatException `

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
glenhein: Can't you remove the next line with the \n as well?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: There can be multiple dateTimes that can be mapped to this field from one resource (ex. image collection time, report date, etc)

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -146,28 +223,49 @@ public Metacard transform(InputStream inputStream)
...
             throw new CatalogTransformerException(TRANSFORM_EXCEPTION_MSG, e);
         } finally {
             IOUtils.closeQuietly(inputStream);
         }
+    }
 
-        MetacardImpl metacard = toMetacard(pathValueTracker, id);
-
-        metacard.setAttribute(Metacard.METADATA, xml);
+/*    private void readInputStreamIntoByteArrayOutputStream(InputStream inputStream,
```
kcwire: If this isn't needed - remove it.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -146,28 +223,49 @@ public Metacard transform(InputStream inputStream)
...
+        }
+        baos.flush();
+    }*/
 
-        return metacard;
+    public InputStream getSourceInputStream() {
+        return new ByteArrayInputStream(byteArray);
     }
 
     @Override
```
kcwire: I realize this wasn't modified here but these changes made me think of this - The metacard transform should check if the metadata is already GMD XML and if so simply return that instead of serializing from the attributes.  That will maintain full fidelity of the GMD XML through the system and should be a faster transformation (since there is nothing to transform).

Here is a reference of how this is done elsewhere: https://github.com/codice/ddf/blob/master/catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/CswRecordConverter.java#L232

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -75,24 +99,47 @@
...
+    private static MetacardType gmdMetacardType;
+
     private final XStream xstream;
 
     private final DataHolder argumentHolder;
 
     private final XMLInputFactory xmlFactory;
 
-    public GmdTransformer() {
+    private static final ThreadLocal<WKTWriter> WKT_WRITER_THREAD_LOCAL =
```
pklinef: It looks like you do not clean up your ThreadLocal.  This could cause a memory leak depending if the bundle is restarted or configuration causes the bean to reconstructed.

Are we sure WKTWriter is that expensive to require ThreadLocal?  WKTReader can be slow with the no arg constructor since it creates a new geometry factory but WKTWriter does not have any heavy weight construction that I am aware of.  We usually just create a new WKTWriter as needed.

Here is an example of a class that hooks up the destroy method in blueprint to avoid ThreadLocal memory leaks.

https://github.com/codice/ddf/blob/eaca0caec1f345cd96e800f2aeecc708054aecd1/platform/security/filter/security-filter-login/src/main/java/org/codice/ddf/security/filter/login/LoginFilter.java#L756-L760

bdeining: The main concern with using WKTWriter here is thread safety.  Added a destroy method in blueprint to avoid memory leaks.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
bdeining: Removed

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java
```diff
@@ -146,28 +228,39 @@ public Metacard transform(InputStream inputStream)
...
 
-        String xml = null;
-        XstreamPathValueTracker pathValueTracker = null;
-        xml = IOUtils.toString(inputStream);
+    private Metacard handleTransform(InputStream inputStream, String id)
+            throws CatalogTransformerException {
+        String xml;
+        XstreamPathValueTracker pathValueTracker;
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+            IOUtils.copy(inputStream, baos);
```
stustison: Are we guaranteed that this is never going to blow up the heap? What if someone passes in a 10GB XML document. Is this now a DOS attack against the system?

bdeining: There is no guarantee, but it seems unlikely that a 10GB XML document will be passed into the transformer.  Regardless, this was changed to a `TemporaryFileBackedOutputStream` to avoid any complications with large files.

stustison:
👍

PR #1073, opened 2016-08-02
DDF-2347: Updated catalog framework to only query resources by default

djblue:
#### What does this PR do?

Fixes issue with querying against the cache. Now only resources are queried by default.

#### Who is reviewing it?

@rzwiefel 
@andrewkfiedler  

@coyotesqrl 
@pklinef

A full build.

[DDF-2347](https://codice.atlassian.net/browse/DDF-2347)

#### Checklist:
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/source/SourceCache.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/source/SourceCache.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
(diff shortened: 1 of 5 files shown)


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/CatalogCacheSource.java
```diff
@@ -0,0 +1,24 @@
...
+package ddf.catalog.cache;
+
+import ddf.catalog.source.Source;
+
+/**
+ * Used by {@link ddf.catalog.plugin.PostFederatedQueryPlugin} and
+ * {@link ddf.catalog.plugin.PreFederatedQueryPlugin} to determine
+ * if a source is the cache.
+ */
+public interface CatalogCacheSource extends Source {
```
coyotesqrl: I think it would be better to move this interface into `catalog-core-api`. It's probably fine where it is, but is a bit out of place. 

I'm also concerned that on some bundle refreshes you might end up with a class casting issue which would present itself by always returning false when you test `instanceof`. I'd want to test that with the interface in this bundle.

pklinef: Since we expect federation plugins to use this interface, I could see it moving to `catalog-core-api` but we should mark as experimental for now.

coyotesqrl:
👍

PR #1077, opened 2016-08-02
DDF-2275 Return immutable attribute sets

brendan-hofmann:
#### What does this PR do?
Fixes new metacard types to return immutable sets of attributes instead of mutable sets.
@bdeining 
@glenhein 
@jlcsmith
@pklinef
https://codice.atlassian.net/browse/DDF-2275
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
@@ -14,6 +14,7 @@
 package ddf.catalog.data.impl;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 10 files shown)


bdeining:
:+1:

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/AssociationsAttributes.java
```diff
@@ -47,7 +48,7 @@
 
     @Override
     public Set<AttributeDescriptor> getAttributeDescriptors() {
-        return DESCRIPTORS;
+        return Collections.unmodifiableSet(DESCRIPTORS);
```
pklinef: Instead of creating an unmodifiable set on every `getAttributeDescriptors` call, you should create that unmodifiable set in the static block.

```
    private static final Set<AttributeDescriptor> DESCRIPTORS;

    static {
        HashSet<AttributeDescriptor> descriptors = new HashSet<>();
        descriptors.add(new AttributeDescriptorImpl(DERIVED,
                true /* indexed */,
                true /* stored */,
                true /* tokenized */,
                true /* multivalued */,
                BasicTypes.STRING_TYPE));
        descriptors.add(new AttributeDescriptorImpl(RELATED,
                true /* indexed */,
                true /* stored */,
                true /* tokenized */,
                true /* multivalued */,
                BasicTypes.STRING_TYPE));
        DESCRIPTORS = Collections.unmodifiableSet(descriptors);
    }
```

brendan-hofmann: In that case, should I make the same change in BasicTypes?

pklinef: I think so.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
```diff
@@ -311,7 +328,36 @@ public AttributeFormat getAttributeFormat() {
                 false /* tokenized */,
                 true /* multivalued */,
                 STRING_TYPE));
-        return descriptors;
+        DESCRIPTORS = Collections.unmodifiableSet(descriptors);
+
+        BASIC_METACARD = new MetacardTypeImpl(MetacardType.DEFAULT_METACARD_TYPE_NAME,
```
ryeats: So the basic metacard has a static final instance of itself. since their is only really a need for one immutable instance of any of these types can we propagate this pattern to the rest of the types.

brendan-hofmann: The other types already implement `MetacardType`. This is by design to allow for more declarative mix-in composition. At some point in the future we might consider adding a more convenient pattern for accessing them.

ryeats: Either way your going to have to create an instance copy all of its descriptors into your type in order to "mixin" lets just skip that first step.

brendan-hofmann: I'm not sure we're talking about the same thing. Look at how the MetacardType is being composited in this PR: https://github.com/glenhein/ddf/blob/beed46142680a63d29927fce10285c557ada9e0a/catalog/transformer/catalog-transformer-pdf/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L22 You can't do that with a static constant. Besides, since all the fields are static and unmodifiable there's practically no cost associated with instantiating these classes.

ryeats: that looks fine in blueprint but not in code if your adding one of those on the fly through a plugin its ugly.

brendan-hofmann: I don't agree that `new FooBar()` is any uglier than `FooBar.STATIC_FOO_BAR`. Besides, this a much more OO design that hides and encapsulates the implementation detail of where these attributes are populated from. Exposing the Type as a static object is an old design pattern that I'm no longer convinced is worth perpetuating.

ryeats: My apologies Ugly was not the correct word misleading is more accurate. If the accessor methods were static I would agree with you but our hands our tied in that regard so they belong to this instance and new implies a cost the underling descriptors could be created every time and then any modification would be local and it would be perfectly OO.

brendan-hofmann: From an OO standpoint, the implication of this API is that it is an immutable object, since it has no setters. The API is actually that it return an unmodifiable set, see: https://github.com/codice/ddf/blob/bb7177ea3dd800f5d267f8bf41070e704fc9b2bb/catalog/core/catalog-core-api/src/main/java/ddf/catalog/data/MetacardType.java#L41-L41

All this PR is doing is correctly enforcing API constraints that were missed in the initial implementation. I don't think there is anything misleading about this design, it is perfectly consistent with the javadocs.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/AssociationsAttributes.java
```diff
@@ -26,23 +27,25 @@
...
  */
 public class AssociationsAttributes implements Associations, MetacardType {
-    private static final Set<AttributeDescriptor> DESCRIPTORS = new HashSet<>();
+    private static final Set<AttributeDescriptor> DESCRIPTORS;
 
     private static final String NAME = "associations";
 
     static {
-        DESCRIPTORS.add(new AttributeDescriptorImpl(DERIVED,
+        HashSet<AttributeDescriptor> descriptors = new HashSet<>();
```
jlcsmith: Is there a specific reason you are declaring descriptors to be a HashSet and not a set?  Why not:
` Set<AttributeDescriptor> descriptors = new HashSet<>();`?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/DateTimeAttributes.java
```diff
@@ -23,35 +24,37 @@
 import ddf.catalog.data.types.DateTime;
 
 /**
- * This class provides attributes that reflect temporal aspects about the
+ * This class provides attributes that reflect descriptorsoral aspects about the
```
jlcsmith: Is descriptorsoral a word?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/VersionAttributes.java
```diff
@@ -26,53 +27,55 @@
...
                 true /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
-        DESCRIPTORS.add(new AttributeDescriptorImpl(TAGS,
+        descriptors.add(new AttributeDescriptorImpl(TAGS,
                 true /* indexed */,
                 true /* stored */,
                 false /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.DATE_TYPE));
```
jlcsmith: I know this wasn't introduced in this PR, but this shouldn't be a DATE_TYPE

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/VersionAttributes.java
```diff
@@ -26,53 +27,55 @@
...
-        DESCRIPTORS.add(new AttributeDescriptorImpl(TYPE,
+        descriptors.add(new AttributeDescriptorImpl(TYPE,
                 true /* indexed */,
                 true /* stored */,
                 false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.DATE_TYPE));
```
jlcsmith: `BasicTypes.DATE_TYPE `should be `BasicTypes.STRING_TYPE`

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/VersionAttributes.java
```diff
@@ -26,53 +27,55 @@
...
                 false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.BINARY_TYPE));
-        DESCRIPTORS.add(new AttributeDescriptorImpl(VERSIONED_BY,
+        descriptors.add(new AttributeDescriptorImpl(VERSIONED_BY,
                 true /* indexed */,
                 true /* stored */,
                 false /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.DATE_TYPE));
```
jlcsmith: `BasicTypes.DATE_TYPE `should be `BasicTypes.STRING_TYPE`

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/DateTimeAttributes.java
brendan-hofmann: Ha! Good catch. I think that was a rogue refactoring, I'll fix it. 😆

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/AssociationsAttributes.java
brendan-hofmann: No reason other than a personal preference for concrete types on locally-scoped references. You're right though, using the interface is best practice, I don't mind fixing it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/VersionAttributes.java
brendan-hofmann: No problem, I'll fix those types.

coyotesqrl:
👍

PR #1079, opened 2016-08-02
DDF-2321 Add Apache Artemis message bus application

roelens8:
#### What does this PR do?
Adds a highly available message bus (Apache Artemis) allowing message communication from JMS queues and topics.
@coyotesqrl 
@ryeats 
@pklinef 
@jaymcnallie
@kcwire

Install the artemis application, and make sure its features properly install.
Since the apache artemis' features file does not contain any feature with an auto install property, the application's "mainFeature" could not be determined, causing the application state to always be "ACTIVE" even if all of its feature were uninstalled. To fix this, I have added some logic to support recognizing the state of an application if it contains no features with an auto install property.

DDF-2321
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests


platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationServiceImpl.java
```diff
@@ -709,11 +729,25 @@ public void startApplication(String application) throws ApplicationServiceExcept
...
+            if (application.getAutoInstallFeatures()
+                    .isEmpty() && application.getMainFeature() != null) {
+                Feature mainFeature = application.getMainFeature();
+                if (featuresService.isInstalled(mainFeature)) {
+                    featuresService.uninstallFeature(mainFeature.getName(),
+                            mainFeature.getVersion(),
                             EnumSet.of(Option.NoAutoRefreshBundles));
                 }
+            } else {
+                // Otherwise install all application's feature.
```
brendan-hofmann: Typo in comment, I think? Believe your uninstalling, not installing, here.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
```diff
@@ -95,6 +99,19 @@ public ApplicationImpl(Repository repo) {
             name = mainFeature.getName();
             version = mainFeature.getVersion();
             description = mainFeature.getDescription();
+        } else if (autoInstallFeatures.size() == 0) {
+            // If the app's features file does not contain any auto install features, set
+            // mainFeature to the first feature found in the file.
```
brendan-hofmann: If we really wanted to be compatible with regular feature files, shouldn't we be attempting to calculate the root feature of a features file as a dependency graph? I'm wondering about the potential side-effects of making this assumption about the first feature, although perhaps the container does enough dependency traversal that it is safe.

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -0,0 +1,84 @@
...
+      <acceptors>
+         <!-- Default ActiveMQ Artemis Acceptor.  Multi-protocol adapter.  Currently supports Core, OpenWire, Stomp and AMQP. -->
+         <!-- performance tests have shown that openWire performs best with these buffer sizes -->
+         <acceptor name="artemis">tcp://0.0.0.0:61616?tcpSendBufferSize=1048576;tcpReceiveBufferSize=1048576</acceptor>
+
+         <!-- AMQP Acceptor.  Listens on default AMQP port for AMQP traffic.-->
+         <acceptor name="amqp">tcp://0.0.0.0:5672?protocols=AMQP</acceptor>
+
+         <!-- STOMP Acceptor. -->
+         <acceptor name="stomp">tcp://0.0.0.0:61613?protocols=STOMP</acceptor>
```
ryeats: should we disable the ones we aren't using?

brendan-hofmann:
I'd love to see some more unit tests in this PR for the new branches at least.

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -0,0 +1,84 @@
...
+
+      <address-settings>
+         <!--default for catch all-->
+         <address-setting match="#">
+            <dead-letter-address>jms.queue.DLQ</dead-letter-address>
+            <expiry-address>jms.queue.ExpiryQueue</expiry-address>
+            <redelivery-delay>0</redelivery-delay>
+            <max-size-bytes>10485760</max-size-bytes>
+            <message-counter-history-day-limit>10</message-counter-history-day-limit>
+            <address-full-policy>BLOCK</address-full-policy>
```
ryeats: how do we want this to behave by default if someone stops picking messages up off a queue then should we block or fail when it fills up. if we block it will consume connections from everything involved in that transaction until they run out which could hose other things, if we fail we may not be reliable during a spike of activity.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationServiceImpl.java
roelens8: Fixed typo.

distribution/ddf-common/src/main/resources/etc/artemis.xml
roelens8: Yes, fixed.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
roelens8: It's not necessarily the first defined feature in the features file that gets set as the main feature, it's the feature that equals the parsed name from the 'Repository' object. That feature should be the root feature of the features file.

roelens8:
Added unit tests for testing the new branches in ApplicationImpl and ApplicataionServiceImpl.

roelens8:
Because we only need a subset of the features from the artemis features file (artemis-core, artemis-amqp), I think it's best to just pull in those artemis features into platform-app instead of pulling in the artemis feature repository. 

Doing this will prevent the need to add the extra logic to support regular feature files and will allow Artemis to be enabled within DDF Platform instead of being enabled via its own application.

PR #1083, opened 2016-08-02
DDF-2346: Source Modal does not like single quoted names

gordocanchola:
#### What does this PR do?
This PR fixes the issue where the Source modal won't open up if there is a single quote in the source's name.
@mcalcote @clockard @vinamartin @ryeats @ani6gup @brianfelix @djblue 
@figliold 
@shaundmorris
Start up DDF with registry. 
Open the DDF Catalog app. 
Open the Node Information tab
Open the identity node and rename it to something containing a single quote
Open the Services tab.
Add a service with at least one binding. Save
Open the Sources tab.
Verify a source exists for you identity node, you may have to refresh the list.
Click on the source and verify the modal opens up.
DDF-2346
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

Updating the popover selector to remove the attribute check for 'data-title'.
An update was made to the backing library JQuery (or Sizzle) where the data-title attribute value containing a single quote is no longer recognized as an attribute.
Selecting on description class itself will add the popover.

### DIFF
--- a/catalog/admin/module/catalog-admin-module-sources/src/main/webapp/js/view/Utils.js
+++ b/catalog/admin/module/catalog-admin-module-sources/src/main/webapp/js/view/Utils.js
@@ -62,7 +62,7 @@ function ($,_, Marionette) {
          * Set up the popovers based on if the selector has a description.
          */
         setupPopOvers: function($popoverAnchor, id, title, description) {
-            var selector = ".description[data-title='" + id + "']",
+            var selector = ".description",
                 options = {
                     title: title,
                     content: description,


catalog/admin/module/catalog-admin-module-sources/src/main/webapp/js/view/Utils.js
```diff
@@ -62,7 +62,7 @@ function ($,_, Marionette) {
          * Set up the popovers based on if the selector has a description.
          */
         setupPopOvers: function($popoverAnchor, id, title, description) {
-            var selector = ".description[data-title='" + id + "']",
+            var selector = ".description",
```
ryeats: why did it have the data-title attribute filter in the selector in the first place is there going to be other unintended elements that will get a popover now

gordocanchola: I'm not sure why the data-title attribute filter was added. I checked the templates and the description class is only on the anchor tags that contain the question icon.

adimka:
passed hero build 👍

PR #1086, opened 2016-08-03
DDF-2321 Add artemis message bus features

roelens8:
#### What does this PR do?
Adds a message bus application (Apache Artemis) allowing message communication from JMS queues and topics.

@coyotesqrl 
@ryeats 
@pklinef 
@jaymcnallie
@kcwire
Install artemis-amqp, and make sure it installs correctly.
Once artemis is installed, do a port scan on port 5672 using localhost to verify the connection to the message bus is open.
DDF-2321
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/ddf-common/src/main/resources/etc/artemis.xml
+++ b/distribution/ddf-common/src/main/resources/etc/artemis.xml
@@ -1,0 +1,70 @@
+<!-- /**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/ -->
+<configuration xmlns="urn:activemq"
+               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+               xsi:schemaLocation="urn:activemq /schema/artemis-configuration.xsd">
+
+    <jms xmlns="urn:activemq:jms">
+        <queue name="DLQ"/>
+        <queue name="ExpiryQueue"/>
+    </jms>
+
+    <core xmlns="urn:activemq:core">
+
+        <!-- this could be ASYNCIO or NIO -->
+        <journal-type>NIO</journal-type>
+        <paging-directory>${ddf.home}/data/artemis/paging</paging-directory>
+        <bindings-directory>${ddf.home}/data/artemis/bindings</bindings-directory>
+        <journal-directory>${ddf.home}/data/artemis/journal</journal-directory>
+        <large-messages-directory>${ddf.home}/data/artemis/large-messages</large-messages-directory>
+        <journal-min-files>10</journal-min-files>
+
+        <!--
+         This value was determined through a calculation.
+         Your system could perform 0.63 writes per millisecond
+         on the current journal configuration.
+         That translates as a sync write every 1591999 nanoseconds
+        -->
+        <journal-buffer-timeout>1591999</journal-buffer-timeout>
+
+        <acceptors>
+            <!-- AMQP Acceptor.  Listens on default AMQP port for AMQP traffic.-->
+            <acceptor name="amqp">tcp://0.0.0.0:${artemis.amqp.port}?sslEnabled=true;keyStorePath=${javax.net.ssl.keyStore};keyStorePassword=${javax.net.ssl.keyStorePassword};enabledCipherSuites=${https.cipherSuites};enabledProtocols=${https.protocols}</acceptor>
+        </acceptors>
+
+        <security-settings>
+            <security-setting match="#">
+                <permission type="createNonDurableQueue" roles="manager"/>
+                <permission type="deleteNonDurableQueue" roles="manager"/>
+                <permission type="createDurableQueue" roles="manager"/>
+                <permission type="deleteDurableQueue" roles="manager"/>
+                <permission type="consume" roles="manager"/>
+                <permission type="send" roles="manager"/>
+                <!-- we need this otherwise ./artemis data imp wouldn't work -->
+                <permission type="manage" roles="manager"/>
+            </security-setting>
+        </security-settings>
+
+        <address-settings>
+            <!--default for catch all-->
+            <address-setting match="#">
+                <dead-letter-address>jms.queue.DLQ</dead-letter-address>
+                <expiry-address>jms.queue.ExpiryQueue</expiry-address>
+                <redelivery-delay>0</redelivery-delay>
+                <max-size-bytes>10485760</max-size-bytes>
+                <message-counter-history-day-limit>10</message-counter-history-day-limit>
+                <address-full-policy>BLOCK</address-full-policy>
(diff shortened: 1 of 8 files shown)


distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -0,0 +1,70 @@
...
+    <jms xmlns="urn:activemq:jms">
+        <queue name="DLQ"/>
+        <queue name="ExpiryQueue"/>
+    </jms>
+
+    <core xmlns="urn:activemq:core">
+
+        <!-- this could be ASYNCIO or NIO -->
+        <journal-type>NIO</journal-type>
+        <paging-directory>./data/paging</paging-directory>
```
coyotesqrl: Have these relative paths been tested on Windows?

coyotesqrl: Also, is this the $DDF_HOME/data directory or is it some other $ARTEMIS_HOME directory? If the former, could we move them all under a `./data/artemis` root?

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -0,0 +1,70 @@
...
+        <large-messages-directory>./data/large-messages</large-messages-directory>
+        <journal-min-files>10</journal-min-files>
+
+        <!--
+         This value was determined through a calculation.
+         Your system could perform 0.63 writes per millisecond
+         on the current journal configuration.
+         That translates as a sync write every 1591999 nanoseconds
+        -->
+        <journal-buffer-timeout>1591999</journal-buffer-timeout>
```
coyotesqrl: Is this machine-dependent? Is there a way to test on a production system and get a different empirical value for to set here?

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -0,0 +1,70 @@
...
+
+        <acceptors>
+            <!-- AMQP Acceptor.  Listens on default AMQP port for AMQP traffic.-->
+            <acceptor name="amqp">tcp://0.0.0.0:5672?protocols=AMQP</acceptor>
```
coyotesqrl: Is this port configurable? If not, how is that going to impact multiple concurrent DDF instances on a box at one time (the itest issue we addressed with dynamic port assignment)?

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -0,0 +1,70 @@
...
+            </security-setting>
+        </security-settings>
+
+        <address-settings>
+            <!--default for catch all-->
+            <address-setting match="#">
+                <dead-letter-address>jms.queue.DLQ</dead-letter-address>
+                <expiry-address>jms.queue.ExpiryQueue</expiry-address>
+                <redelivery-delay>0</redelivery-delay>
+                <max-size-bytes>10485760</max-size-bytes>
```
coyotesqrl: How was this value arrive at? What happens if a larger message is sent? Can it be configured?

platform/platform-app/src/main/resources/features.xml
```diff
@@ -1044,4 +1044,41 @@
...
 
+    <feature name="artemis-core" install="manual" version="${artemis.version}" description="ActiveMQ Artemis broker libraries">
+        <feature prerequisite="true">transaction</feature>
+        <feature prerequisite="true">netty-core</feature>
+        <feature prerequisite="true">scr</feature>
+        <configfile finalname="etc/org.apache.activemq.artemis.cfg">mvn:org.apache.activemq/artemis-features/${artemis.version}/cfg</configfile>
+        <configfile finalname="etc/artemis.xml">mvn:org.apache.activemq/artemis-features/${artemis.version}/xml/artemis</configfile>
+
+        <bundle>mvn:org.apache.geronimo.specs/geronimo-jms_2.0_spec/${geronimo-jms_2_spec.version}</bundle>
+        <bundle>mvn:com.google.guava/guava/${guava.version}</bundle>
```
coyotesqrl: This should be `<bundle dependency="true">` for guava at the very least.

distribution/ddf-common/src/main/resources/etc/artemis.xml
roelens8: These directories are actually created under $DDF_HOME/data/artemis/. And I verified these paths work on a windows VM.

tbatie: We should not be using relative paths at all. You can put it under a filter-resources folder and have maven replace it with a absolute path using ${ddf.home}

roelens8: I can replace the root of each directory with ${ddf.home}/artemis/...

tbatie: That sounds even better

distribution/ddf-common/src/main/resources/etc/artemis.xml
roelens8: The port can be edited in the file and then would require a restart of DDF in order for the new value to take effect, but it's not yet configurable through the Admin UI or integration tests. 

This will be an issue when we have integration tests using the artemis message bus when there are multiple instances of DDF running at the same time and we will have to create a System or Karaf property for the port number and dynamically set it in the itests similar to how we are using dynamic ports for the system port.

platform/platform-app/src/main/resources/features.xml
roelens8: Will fix.

distribution/ddf-common/src/main/resources/etc/artemis.xml
roelens8: Not that I know of, but I think it should be a value balanced for all machines.

distribution/ddf-common/src/main/resources/etc/artemis.xml
roelens8: It's the default value. If a message is sent to a queue with a size larger than the max size, it should fail with an exception returned to the client. The value can be changed in the file, but then requires a restart of DDF to take effect.

pklinef:
Please look at `Configuring Netty SSL`.  The current configuration appears to be in the clear.

https://activemq.apache.org/artemis/docs/1.3.0/configuring-transports.html 

@codice/security

stustison:
Going to have to 👎  this until it is secure by default. Check out the link that @pklinef suggested.

roelens8:
The message bus is now enabled with 1-way SSL by default, please re-review.

distribution/ddf-common/src/main/resources/etc/artemis.xml
roelens8: Added a system property for the port number, so it can now be edited from system.properties.

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -185,3 +185,8 @@ security.audit.roles=group,admin,manager,viewer,system-admin,system-history,syst
 # Disable Hazelcast version check
 #
 hazelcast.version.check.enabled=false
+
+#
+# Artemis Message Bus properties
+#
+artemis.amqp.port=5672
```
stustison: Does SSL use the same port as the normal amqp port? How does the connection distinguish between doing SSL or not? Are we always forcing SSL?

roelens8: SSL can be enabled on any port and the broker will enforce SSL when client connects to that port.

stustison:
👍

pklinef:
:+1:

kcwire:
👍

jaymcnallie:
Built and tested.  Verified dropping messages onto and getting messages off a topic over amqps.  👍

PR #1087, opened 2016-08-03
DDF-2358 added new GCM ciphers to list, added some updated java 8 system properties, and updated the pax web config

stustison:
@beyelerb @coyotesqrl @ahoffer 
@shaundmorris
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
+++ b/distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
@@ -38,3 +38,15 @@ org.ops4j.pax.web.ssl.keypassword=${javax.net.ssl.keyStorePassword}
 
 # Client authenticate at server is wanted
 org.ops4j.pax.web.ssl.clientauthwanted=true
+
+# Included SSL/TLS Cipher Suites comma-separated list of Regular Expressions
+org.ops4j.pax.web.ssl.ciphersuites.included=${https.cipherSuites}
+
+# Excluded SSL/TLS Cipher Suites comma-separated list of Regular Expressions
+org.ops4j.pax.web.ssl.ciphersuites.excluded=.*RC4.*,SSL.*,.*anon.*,.*DES.*,.*EXPORT.*,.*NULL.*,.*MD5.*,.*DSS.*
+
+# Included SSL/TLS Protocols
+org.ops4j.pax.web.ssl.protocols.included=${https.protocols}
+
+# Excluded SSL/TLS Protocols
+org.ops4j.pax.web.ssl.protocols.excluded=SSLv3,TLSv1
--- a/distribution/ddf-common/src/main/resources/etc/system.properties
+++ b/distribution/ddf-common/src/main/resources/etc/system.properties
@@ -50,10 +50,14 @@ org.codice.ddf.system.organization=${organization.name}
 # See "Configuring Thread Pools" under "Managing" documentation.
 org.codice.ddf.system.threadPoolSize=128
 
-# HTTPS Specific settings.  If making a Secure Connection not leveraging the HTTPS Jaav libraries and 
+# HTTPS Specific settings.  If making a Secure Connection not leveraging the HTTPS Java libraries and
 # classes (e.g. if you are using secure sockets directly) then you will have to set this directly
-https.cipherSuites=TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_DSS_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA
+https.cipherSuites=TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+# If using the JCE unlimited strength policy, uncomment this and use it instead.
+#https.cipherSuites=TLS_DHE_RSA_WITH_AES_256_GCM_SHA384,TLS_DHE_RSA_WITH_AES_256_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
 https.protocols=TLSv1.1,TLSv1.2
+jdk.tls.client.protocols=TLSv1.1,TLSv1.2
+jdk.tls.ephemeralDHKeySize=matched
 
 # File Upload settings. Files uploaded with these bad file extensions or names will have their file names sanitized before being saved
 bad.file.extensions=.exe,.jsp,.html,.js,.php,.phtml,.php3,.php4,.php5,.phps,.shtml,.jhtml,.pl,.py,.cgi,.msi,.com,.scr,.gadget,.application,.pif,.hta,.cpl,.msc,.jar,.kar,.bat,.cmd,.vb,.vbs,.vbe,.jse,.ws,.wsf,.wsc,.wsh,.ps1,.ps1xml,.ps2,.ps2xml,.psc1,.psc2,.msh,.msh1,.msh2,.mshxml,.msh1xml,.msh2xml,.scf,.lnk,.inf,.reg,.dll,.vxd,.cpl,.cfg,.config,.crt,.cert,.pem,.jks,.p12,.p7b,.key,.der,.csr,.jsb,.mhtml,.mht,.xhtml,.xht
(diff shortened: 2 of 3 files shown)


coyotesqrl:
👍

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
```diff
@@ -38,3 +38,15 @@ org.ops4j.pax.web.ssl.keypassword=${javax.net.ssl.keyStorePassword}
...
+org.ops4j.pax.web.ssl.ciphersuites.included=${https.cipherSuites}
+
+#Excluded SSL/TLS Cipher Suites Regular Expressions
+org.ops4j.pax.web.ssl.ciphersuites.excluded=.*RC4.*,SSL.*,.*anon.*,.*3DES.*
+
+#Included SSL/TLS Protocols
+org.ops4j.pax.web.ssl.protocols.included=${https.protocols}
+
+#Excluded SSL/TLS Protocols
+org.ops4j.pax.web.ssl.protocols.excluded=SSLv3,TLSv1
```
ahoffer: Should we exclude `SSLv2` as well? Or is that implied?

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
```diff
@@ -38,3 +38,15 @@ org.ops4j.pax.web.ssl.keypassword=${javax.net.ssl.keyStorePassword}
 
 # Client authenticate at server is wanted
 org.ops4j.pax.web.ssl.clientauthwanted=true
+
+#Included SSL/TLS Cipher Suites Regular Expressions
+org.ops4j.pax.web.ssl.ciphersuites.included=${https.cipherSuites}
+
+#Excluded SSL/TLS Cipher Suites Regular Expressions
+org.ops4j.pax.web.ssl.ciphersuites.excluded=.*RC4.*,SSL.*,.*anon.*,.*3DES.*
```
ahoffer: Can retire triple DES? It looks like Windows XP and Windows Server 2003 do not support AES.  Those OSes do not support TLS 1.1 or TLS 1.2.  I know the OSes are past end-of-life, but the military isn't known for upgrading systems in a timely manner.

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -52,8 +52,12 @@ org.codice.ddf.system.threadPoolSize=128
 
 # HTTPS Specific settings.  If making a Secure Connection not leveraging the HTTPS Jaav libraries and
```
lessarderic: `Java` instead of `Jaav`?

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
```diff
@@ -38,3 +38,15 @@ org.ops4j.pax.web.ssl.keypassword=${javax.net.ssl.keyStorePassword}
 
 # Client authenticate at server is wanted
 org.ops4j.pax.web.ssl.clientauthwanted=true
+
+#Included SSL/TLS Cipher Suites Regular Expressions
```
lessarderic: Should we mention that these are _comma-separated lists_ of regular expressions? Space between `#` and the first character in the comment also looks better IMHO ;-)

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
lessarderic: Having both, included and excluded, makes things a little confusing. What if a protocol is in both or in neither? Could we get this to work with either an exclusion or an inclusion list instead but not both?

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
pklinef: This configuration file can be edited as needed for compatibility.  Exclusions are partly based on NIST recommendations.

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -52,8 +52,12 @@ org.codice.ddf.system.threadPoolSize=128
 
 # HTTPS Specific settings.  If making a Secure Connection not leveraging the HTTPS Jaav libraries and 
 # classes (e.g. if you are using secure sockets directly) then you will have to set this directly
-https.cipherSuites=TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_DSS_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA
+https.cipherSuites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_DSS_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA
```
pklinef: https://www.owasp.org/index.php/Transport_Layer_Protection_Cheat_Sheet#Rule_-_Only_Support_Strong_Cryptographic_Ciphers

>Favor DHE over ECDHE
Did you sort ECDHE over DHE for performance?

> Use RSA-Keys (no DSA/DSS: they get very weak, if a bad entropy source is used during signing, cf. [5], [6])
Should we be concerned with TLS_DHE_DSS_WITH_AES_128_CBC_SHA and TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256?

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -52,8 +52,12 @@ org.codice.ddf.system.threadPoolSize=128
...
+# If using the JCE unlimited strength policy, uncomment this and use it instead.
+#https.cipherSuites=TLS_DHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_DHE_RSA_WITH_AES_256_CBC_SHA256,TLS_DHE_DSS_WITH_AES_256_GCM_SHA384,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_DSS_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA
 https.protocols=TLSv1.1,TLSv1.2
+jdk.tls.client.protocols=TLSv1.1,TLSv1.2
```
pklinef: Can we set this to `https.protocol` so they always stay in sync?

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
stustison: No, you need both in order to make this work, unfortunately. Blame Pax Web!

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
stustison: We can, I just copied the exact description of the fields out of their metatype, because that is all that you'll get in the Admin Console as well.

distribution/ddf-common/src/main/resources/etc/system.properties
stustison: I did sort it that way for performance.

We could remove the DSS once since that key isn't as secure as the RSA or ECDSA keys. I'm not concerned about the ECDSA one since that is more secure than RSA.

distribution/ddf-common/src/main/resources/etc/system.properties
stustison: Can you use a property in the file in which it was defined? My assumption was no and we don't do it anywhere else.

distribution/ddf-common/src/main/resources/etc/system.properties
stustison: A bigger question is whether or not we should get rid of the TLS_RSA_WITH_AES_128_CBC_SHA suite since this does not provide PFS. All modern web browsers support PFS at this point and the suite can always be added back in for compatibility reasons at install time. I think my vote is to remove it.

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.web.cfg
lessarderic: Yeah, didn't pay attention to the key names and didn't realize until later those weren't ours. Sorry about that.

distribution/ddf-common/src/main/resources/etc/system.properties
pklinef: I am in favor of removing TLS_RSA_WITH_AES_128_CBC_SHA and let administrators add that weaker cipher back in if they absolutely need it for compatibility.

distribution/ddf-common/src/main/resources/etc/system.properties
pklinef: I am not sure.  It would be nice if it works but understandable if it does not.

pklinef:
👍

adimka:
Hero build passed. 👍

shaundmorris:
👍  do we need to update any docs for this?

PR #1091, opened 2016-08-04
DDF-2307 Fix map preferences restore button

pklinef:
#### What does this PR do?
Fixes regressions caused by the original DDF-2307 commit.

- Fix restore button to use properties
- Sort views by label instead of alpha
- Fix render order to avoid model binder errors when sorting the
  collection
- Fix user model to not modify the original layer from properties

@djblue 
@andrewkfiedler 

@pklinef
@stustison

Use the restore defaults button in the Standard UI map preferences.

https://codice.atlassian.net/browse/DDF-2307

#### Checklist:
- [ n/a ] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/search-ui/standard/src/main/webapp/js/model/user.js
+++ b/catalog/ui/search-ui/standard/src/main/webapp/js/model/user.js
@@ -94,14 +94,15 @@ define([
             };
         },
         parse: function(resp) {
-            resp.label = 'Type: ' + resp.type;
-            if (resp.layer) {
-                resp.label += ' Layer: ' + resp.layer;
+            var layer = _.clone(resp);
+            layer.label = 'Type: ' + layer.type;
+            if (layer.layer) {
+                layer.label += ' Layer: ' + layer.layer;
             }
-            if (resp.layers) {
-                resp.label += ' Layers: ' + resp.layers.join(', ');
+            if (layer.layers) {
+                layer.label += ' Layers: ' + layer.layers.join(', ');
             }
-            return resp;
+            return layer;
         }
     });
 
--- a/catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
+++ b/catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
@@ -270,13 +270,21 @@ define([
             },
             resetDefaults: function () {
                 this.onEdit();
-                var model = this.model;
                 this.viewMapLayers.each(function (viewLayer) {
-                    var url = viewLayer.get('url');
-                    var defaultConfig = model.getMapLayerConfig(url);
-                    viewLayer.set('show', defaultConfig.show);
-                    viewLayer.set('alpha', defaultConfig.alpha);
-                    viewLayer.set('index', defaultConfig.index);
+                    var viewObj = viewLayer.toJSON();
+                    var defaultConfig = _.find(properties.imageryProviders, function (layerObj) {
+                        var keys = _.without(_.keys(layerObj), 'alpha');
+                        var viewProperties = _.pick(viewObj, keys);
+                        return _.isEqual(viewProperties, _.omit(layerObj, 'alpha'));
+                    });
+                    var alpha = 0;
+                    if (defaultConfig && _.isNumber(defaultConfig.alpha)) {
+                        alpha = defaultConfig.alpha;
+                    }
+                    viewLayer.set({
+                        show: true,
+                        alpha: alpha
+                    });
                 });
                 this.viewMapLayers.sort();
             }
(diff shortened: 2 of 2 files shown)


stustison:
👍

catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
```diff
@@ -270,13 +270,13 @@ define([
...
                 this.onEdit();
-                var model = this.model;
                 this.viewMapLayers.each(function (viewLayer) {
                     var url = viewLayer.get('url');
-                    var defaultConfig = model.getMapLayerConfig(url);
-                    viewLayer.set('show', defaultConfig.show);
+                    var defaultConfig = _.find(properties.imageryProviders, function (layerObj) {
+                        return url === layerObj.url;
+                    });
+                    viewLayer.set('show', true);
```
andrewkfiedler: You can do a single set here:
```
viewLayer.set({
   show: true,
   alpha: defaultConfig.alpha
});
```

catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
```diff
@@ -270,13 +270,13 @@ define([
             },
             resetDefaults: function () {
                 this.onEdit();
-                var model = this.model;
                 this.viewMapLayers.each(function (viewLayer) {
                     var url = viewLayer.get('url');
-                    var defaultConfig = model.getMapLayerConfig(url);
-                    viewLayer.set('show', defaultConfig.show);
+                    var defaultConfig = _.find(properties.imageryProviders, function (layerObj) {
```
andrewkfiedler: It's possible for this to be undefined if it can't find a match.  Not sure if it'll ever happen, but if so the ```viewLayer.set``` using ```defaultConfig.alpha``` will fail.

catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
```diff
@@ -320,9 +320,10 @@ define([
             ui: {
                 tbody: 'tbody'
             },
+            viewComparator: 'label',
             initialize: function() {
                 this.collection.each(function(model) {
```
andrewkfiedler: You don't need to loop over the models here, you can listen to the collection instead.
```
this.listenTo(this, 'change:alpha', this.updateSort)
```
Or you could bake it in like this:
```
{
   viewComparator: 'label',
   collectionEvents: {
    "change:alpha": "updateSort"
  }
}

File: catalog/ui/search-ui/standard/src/main/webapp/js/model/user.js [hunk 4]
```diff
@@ -94,14 +94,15 @@ define([
             };
         },
         parse: function(resp) {
-            resp.label = 'Type: ' + resp.type;
-            if (resp.layer) {
-                resp.label += ' Layer: ' + resp.layer;
+            var layer = _.clone(resp);
```
andrewkfiedler: This parse method looks like something that should go in the serializeData method of a view.

catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
stustison: It doesn't flow the change in the model up, which is why I looped over the models.

catalog/ui/search-ui/standard/src/main/webapp/js/model/user.js
pklinef: Eventually the layer name will be use in other views.  It would be nice to keep that logic in the model instead of duplicating it in all views that need it.

andrewkfiedler: Okay, as long as it's important to all the views using the model you should keep it here.

catalog/ui/search-ui/standard/src/main/webapp/js/view/preferences/PreferencesModal.view.js
andrewkfiedler: It should flow the change up.  

See here for an example of it: http://jsfiddle.net/GarethElms/aQnez/?utm_source=website&utm_medium=embed&utm_campaign=aQnez

andrewkfiedler: I also threw together a codepen example with a more recent backbone version: http://codepen.io/andrewkfiedler/pen/ZOmXON?editors=0010

andrewkfiedler: The latest Backbone (1.30-1.33) has a bug with this behavior, so I've opened an issue here: https://github.com/jashkenas/backbone/issues/4064

andrewkfiedler:
👍

PR #1096, opened 2016-08-04
DDF-2296: Fixed bug in Versioning

rzwiefel:
#### What does this PR do?
Fixes versioning. 
@jrnorth @djblue @kcwire 

@coyotesqrl
@pklinef

Ingest some non-resource metacard. Ensure in solr that all attributes were versioned. 
There was a bug originally where we were storing only the basic attributes of a metacard in the revisions. Once that was fixed it exposed another problem that was caused when updating by something other than metacard id. (Needed to be unique attribute when we go to search and update/delete it, but with history enabled there was now two of these 'unique' values on metacards)
DDF-2296
#### Checklist:
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
@@ -21,18 +21,23 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.opengis.filter.Filter;
+import org.opengis.filter.sort.SortBy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Sets;
 
 import ddf.catalog.Constants;
 import ddf.catalog.content.StorageException;
 import ddf.catalog.content.StorageProvider;
 import ddf.catalog.content.operation.DeleteStorageRequest;
 import ddf.catalog.content.operation.impl.DeleteStorageRequestImpl;
+import ddf.catalog.core.versioning.MetacardVersion;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
 import ddf.catalog.data.impl.AttributeImpl;
@@ -44,6 +49,8 @@ import ddf.catalog.operation.DeleteResponse;
 import ddf.catalog.operation.OperationTransaction;
 import ddf.catalog.operation.ProcessingDetails;
 import ddf.catalog.operation.QueryResponse;
+import ddf.catalog.operation.SourceResponse;
+import ddf.catalog.operation.impl.DeleteRequestImpl;
 import ddf.catalog.operation.impl.DeleteResponseImpl;
 import ddf.catalog.operation.impl.OperationTransactionImpl;
 import ddf.catalog.operation.impl.ProcessingDetailsImpl;
@@ -60,6 +67,7 @@ import ddf.catalog.source.CatalogStore;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.source.InternalIngestException;
 import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
 import ddf.catalog.util.impl.Requests;
 
 /**
@@ -252,6 +260,7 @@ public class DeleteOperations {
         }
 
         try {
+            deleteRequest = rewriteRequestToAvoidHistoryConflicts(deleteRequest);
             sourceOperations.getStorage()
                     .delete(deleteStorageRequest);
         } catch (StorageException e) {
@@ -259,12 +268,105 @@ public class DeleteOperations {
             throw new InternalIngestException(
                     "Unable to delete stored content items. Not removing stored metacards.",
                     e);
+        } catch (UnsupportedQueryException e) {
+            throw new IngestException("Could not rewrite request to avoid history conflicts", e);
         }
         DeleteResponse deleteResponse = sourceOperations.getCatalog()
                 .delete(deleteRequest);
         deleteResponse = injectAttributes(deleteResponse);
         historian.version(deleteResponse);
         return deleteResponse;
+    }
+
+    //
+    // Private helper methods
+    //
+
+    private DeleteRequest rewriteRequestToAvoidHistoryConflicts(DeleteRequest deleteRequest)
+            throws UnsupportedQueryException {
+        final String attributeName = deleteRequest.getAttributeName();
+        if (Metacard.ID.equals(attributeName)) {
+            return deleteRequest;
+        }
+
+        Filter filter = deleteRequest.getAttributeValues()
+                .stream()
+                .map(Object::toString)
+                .map((val) -> getNonhistoryFilter(attributeName, val))
+                .collect(Collectors.collectingAndThen(Collectors.toList(),
+                        frameworkProperties.getFilterBuilder()::anyOf));
+
+        SourceResponse response = sourceOperations.getCatalog()
+                .query(new QueryRequestImpl(new QueryImpl(filter,
+                        1,
+                        0,
+                        SortBy.NATURAL_ORDER,
+                        false,
+                        TimeUnit.SECONDS.toMillis(10))));
+
+        if (isAllAttributesMatched(deleteRequest, response, attributeName)) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug(
+                        "While rewriting the query, did not get a metacardId corresponding to every attribute."
+                                + "Original Delete By attribute was: " + attributeName);
+                LOGGER.debug("Metacards unable to get Metacard ID from are: "
+                        + deleteRequest.getAttributeValues()
+                        .stream()
+                        .map(Object::toString)
+                        .filter(s -> !response.getResults()
+                                .contains(s))
+                        .collect(Collectors.joining(", ", "[", "]")));
+            }
+            throw new UnsupportedQueryException("Could not find all metacards to delete.");
+        }
+
+        List<Serializable> updatedList = response.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .map(Metacard::getId)
+                .map(Serializable.class::cast)
+                .collect(Collectors.toList());
+
+        return new DeleteRequestImpl(updatedList,
+                Metacard.ID,
+                deleteRequest.getProperties(),
+                deleteRequest.getStoreIds());
+    }
+
+    private boolean isAllAttributesMatched(DeleteRequest deleteRequest, SourceResponse response,
+            String attributeName) {
+        Set<String> originalKeys = deleteRequest.getAttributeValues()
+                .stream()
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        Set<String> responseKeys = response.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .map(m -> m.getAttribute(attributeName))
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        boolean notSameSize = response.getResults()
+                .size() != deleteRequest.getAttributeValues()
+                .size();
+        return !Sets.difference(originalKeys, responseKeys)
+                .isEmpty() || notSameSize;
+    }
+
+    private Filter getNonhistoryFilter(String attributeName, String value) {
+        Filter attr = frameworkProperties.getFilterBuilder()
+                .attribute(attributeName)
+                .is()
+                .equalTo()
+                .text(value);
+        Filter notHistory = frameworkProperties.getFilterBuilder()
+                .not(frameworkProperties.getFilterBuilder()
+                        .attribute(Metacard.TAGS)
+                        .is()
+                        .like()
+                        .text(MetacardVersion.VERSION_TAG));
+
+        return frameworkProperties.getFilterBuilder()
+                .allOf(attr, notHistory);
     }
 
     private DeleteRequest processPreIngestPlugins(DeleteRequest deleteRequest)
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
@@ -20,8 +20,9 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
@@ -30,14 +31,12 @@ import org.opengis.filter.sort.SortBy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.Sets;
-
 import ddf.catalog.Constants;
 import ddf.catalog.content.StorageException;
 import ddf.catalog.content.StorageProvider;
 import ddf.catalog.content.operation.DeleteStorageRequest;
 import ddf.catalog.content.operation.impl.DeleteStorageRequestImpl;
-import ddf.catalog.core.versioning.MetacardVersion;
+import ddf.catalog.data.Attribute;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
 import ddf.catalog.data.impl.AttributeImpl;
@@ -46,6 +45,7 @@ import ddf.catalog.history.Historian;
 import ddf.catalog.impl.FrameworkProperties;
 import ddf.catalog.operation.DeleteRequest;
 import ddf.catalog.operation.DeleteResponse;
+import ddf.catalog.operation.Operation;
 import ddf.catalog.operation.OperationTransaction;
 import ddf.catalog.operation.ProcessingDetails;
 import ddf.catalog.operation.QueryResponse;
@@ -67,12 +67,11 @@ import ddf.catalog.source.CatalogStore;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.source.InternalIngestException;
 import ddf.catalog.source.SourceUnavailableException;
-import ddf.catalog.source.UnsupportedQueryException;
 import ddf.catalog.util.impl.Requests;
 
 /**
  * Support class for delete delegate operations for the {@code CatalogFrameworkImpl}.
- *
+ * <p>
  * This class contains one delegated delete method and methods to support it. No
  * operations/support methods should be added to this class except in support of CFI
  * delete operations.
@@ -131,18 +130,12 @@ public class DeleteOperations {
         deleteRequest = validateLocalSource(deleteRequest);
 
         try {
-            List<Metacard> metacards = populateMetacards(deleteRequest);
-
-            deleteStorageRequest = new DeleteStorageRequestImpl(metacards,
+            deleteRequest = populateMetacards(deleteRequest);
+            deleteStorageRequest = new DeleteStorageRequestImpl(getDeleteMetacards(deleteRequest),
                     deleteRequest.getProperties());
 
-            deleteRequest = processPreDeletePolicyPlugins(deleteRequest, metacards);
+            deleteRequest = processPreDeletePolicyPlugins(deleteRequest);
             deleteRequest = processPreDeleteAccessPlugins(deleteRequest);
-
-            deleteRequest.getProperties()
-                    .put(Constants.OPERATION_TRANSACTION_KEY,
-                            new OperationTransactionImpl(OperationTransaction.OperationType.DELETE,
-                                    metacards));
 
             deleteRequest = processPreIngestPlugins(deleteRequest);
             deleteRequest = validateDeleteRequest(deleteRequest);
@@ -183,6 +176,16 @@ public class DeleteOperations {
         }
 
         return deleteResponse;
+    }
+
+    private List<Metacard> getDeleteMetacards(DeleteRequest deleteRequest) {
+        return Optional.of(deleteRequest)
+                .map(Operation::getProperties)
+                .map(p -> p.get(Constants.OPERATION_TRANSACTION_KEY))
+                .filter(OperationTransaction.class::isInstance)
+                .map(OperationTransaction.class::cast)
+                .map(OperationTransaction::getPreviousStateMetacards)
+                .orElseGet(ArrayList::new);
     }
 
     //
@@ -260,7 +263,6 @@ public class DeleteOperations {
         }
 
         try {
-            deleteRequest = rewriteRequestToAvoidHistoryConflicts(deleteRequest);
             sourceOperations.getStorage()
                     .delete(deleteStorageRequest);
         } catch (StorageException e) {
@@ -268,8 +270,6 @@ public class DeleteOperations {
             throw new InternalIngestException(
                     "Unable to delete stored content items. Not removing stored metacards.",
                     e);
-        } catch (UnsupportedQueryException e) {
-            throw new IngestException("Could not rewrite request to avoid history conflicts", e);
         }
         DeleteResponse deleteResponse = sourceOperations.getCatalog()
                 .delete(deleteRequest);
@@ -282,42 +282,10 @@ public class DeleteOperations {
     // Private helper methods
     //
 
-    private DeleteRequest rewriteRequestToAvoidHistoryConflicts(DeleteRequest deleteRequest)
-            throws UnsupportedQueryException {
-        final String attributeName = deleteRequest.getAttributeName();
-        if (Metacard.ID.equals(attributeName)) {
+    private DeleteRequest rewriteRequestToAvoidHistoryConflicts(DeleteRequest deleteRequest,
+            SourceResponse response) {
+        if (Metacard.ID.equals(deleteRequest.getAttributeName())) {
             return deleteRequest;
-        }
-
-        Filter filter = deleteRequest.getAttributeValues()
-                .stream()
-                .map(Object::toString)
-                .map((val) -> getNonhistoryFilter(attributeName, val))
-                .collect(Collectors.collectingAndThen(Collectors.toList(),
-                        frameworkProperties.getFilterBuilder()::anyOf));
-
-        SourceResponse response = sourceOperations.getCatalog()
-                .query(new QueryRequestImpl(new QueryImpl(filter,
-                        1,
-                        0,
-                        SortBy.NATURAL_ORDER,
-                        false,
-                        TimeUnit.SECONDS.toMillis(10))));
-
-        if (isAllAttributesMatched(deleteRequest, response, attributeName)) {
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug(
-                        "While rewriting the query, did not get a metacardId corresponding to every attribute."
-                                + "Original Delete By attribute was: " + attributeName);
-                LOGGER.debug("Metacards unable to get Metacard ID from are: "
-                        + deleteRequest.getAttributeValues()
-                        .stream()
-                        .map(Object::toString)
-                        .filter(s -> !response.getResults()
-                                .contains(s))
-                        .collect(Collectors.joining(", ", "[", "]")));
-            }
-            throw new UnsupportedQueryException("Could not find all metacards to delete.");
         }
 
         List<Serializable> updatedList = response.getResults()
@@ -333,8 +301,8 @@ public class DeleteOperations {
                 deleteRequest.getStoreIds());
     }
 
-    private boolean isAllAttributesMatched(DeleteRequest deleteRequest, SourceResponse response,
-            String attributeName) {
+    private boolean foundAllDeleteRequestMetacards(DeleteRequest deleteRequest,
+            SourceResponse response) {
         Set<String> originalKeys = deleteRequest.getAttributeValues()
                 .stream()
                 .map(Object::toString)
@@ -342,31 +310,13 @@ public class DeleteOperations {
         Set<String> responseKeys = response.getResults()
                 .stream()
                 .map(Result::getMetacard)
-                .map(m -> m.getAttribute(attributeName))
+                .map(m -> m.getAttribute(deleteRequest.getAttributeName()))
+                .filter(Objects::nonNull)
+                .map(Attribute::getValue)
+                .filter(Objects::nonNull)
                 .map(Object::toString)
                 .collect(Collectors.toSet());
-        boolean notSameSize = response.getResults()
-                .size() != deleteRequest.getAttributeValues()
-                .size();
-        return !Sets.difference(originalKeys, responseKeys)
-                .isEmpty() || notSameSize;
-    }
-
-    private Filter getNonhistoryFilter(String attributeName, String value) {
-        Filter attr = frameworkProperties.getFilterBuilder()
-                .attribute(attributeName)
-                .is()
-                .equalTo()
-                .text(value);
-        Filter notHistory = frameworkProperties.getFilterBuilder()
-                .not(frameworkProperties.getFilterBuilder()
-                        .attribute(Metacard.TAGS)
-                        .is()
-                        .like()
-                        .text(MetacardVersion.VERSION_TAG));
-
-        return frameworkProperties.getFilterBuilder()
-                .allOf(attr, notHistory);
+        return originalKeys.equals(responseKeys);
     }
 
     private DeleteRequest processPreIngestPlugins(DeleteRequest deleteRequest)
@@ -390,8 +340,9 @@ public class DeleteOperations {
         return deleteRequest;
     }
 
-    private DeleteRequest processPreDeletePolicyPlugins(DeleteRequest deleteRequest,
-            List<Metacard> metacards) throws StopProcessingException {
+    private DeleteRequest processPreDeletePolicyPlugins(DeleteRequest deleteRequest)
+            throws StopProcessingException {
+        List<Metacard> metacards = getDeleteMetacards(deleteRequest);
         Map<String, Serializable> unmodifiableProperties =
                 Collections.unmodifiableMap(deleteRequest.getProperties());
 
@@ -409,35 +360,36 @@ public class DeleteOperations {
         return deleteRequest;
     }
 
-    private List<Metacard> populateMetacards(DeleteRequest deleteRequest)
+    private DeleteRequest populateMetacards(DeleteRequest deleteRequest)
             throws IngestException, StopProcessingException {
-        if (frameworkProperties.getPolicyPlugins()
-                .isEmpty()) {
-            return new ArrayList<>();
-        }
-
         QueryRequestImpl queryRequest = createQueryRequest(deleteRequest);
-
+        QueryResponse query;
         try {
-            QueryResponse query = queryOperations.doQuery(queryRequest,
+            query = queryOperations.doQuery(queryRequest,
                     frameworkProperties.getFederationStrategy(),
                     false);
-            List<Metacard> metacards = query.getResults()
-                    .stream()
-                    .map(Result::getMetacard)
-                    .collect(Collectors.toList());
-
-            if (metacards.size() < deleteRequest.getAttributeValues()
-                    .size()) {
-                throw new StopProcessingException(
-                        "Unable to remove all metacards contained in request.");
-            }
-
-            return metacards;
         } catch (FederationException e) {
             LOGGER.debug("Unable to complete query for updated metacards.", e);
             throw new IngestException("Exception during runtime while performing delete");
         }
+
+        List<Metacard> metacards = query.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .collect(Collectors.toList());
+
+        if (!foundAllDeleteRequestMetacards(deleteRequest, query)) {
+            logFailedQueryInfo(deleteRequest, query);
+            throw new StopProcessingException("Could not find all metacards specified in request");
+        }
+
+        deleteRequest = rewriteRequestToAvoidHistoryConflicts(deleteRequest, query);
+        deleteRequest.getProperties()
+                .put(Constants.OPERATION_TRANSACTION_KEY,
+                        new OperationTransactionImpl(OperationTransaction.OperationType.DELETE,
+                                metacards));
+        return deleteRequest;
+
     }
 
     private QueryRequestImpl createQueryRequest(DeleteRequest deleteRequest) {
@@ -451,10 +403,12 @@ public class DeleteOperations {
                 .collect(Collectors.toList());
 
         QueryImpl queryImpl =
-                new QueryImpl(queryOperations.getFilterWithAdditionalFilters(idFilters));
-        queryImpl.setStartIndex(1);
-        queryImpl.setPageSize(deleteRequest.getAttributeValues()
-                .size());
+                new QueryImpl(queryOperations.getFilterWithAdditionalFilters(idFilters),
+                        1,  /* start index */
+                        0,  /* page size */
+                        SortBy.NATURAL_ORDER,
+                        false, /* total result count */
+                        0   /* timeout */);
         return new QueryRequestImpl(queryImpl, deleteRequest.getStoreIds());
     }
 
@@ -574,5 +528,30 @@ public class DeleteOperations {
             return false;
         }
         return true;
+    }
+
+    private void logFailedQueryInfo(DeleteRequest deleteRequest, QueryResponse query) {
+        if (LOGGER.isDebugEnabled()) {
+            final String attributeName = deleteRequest.getAttributeName();
+            Set<String> queryResults = query.getResults()
+                    .stream()
+                    .map(Result::getMetacard)
+                    .map(m -> m.getAttribute(attributeName))
+                    .filter(Objects::nonNull)
+                    .map(Attribute::getValue)
+                    .filter(Objects::nonNull)
+                    .map(Object::toString)
+                    .collect(Collectors.toSet());
+
+            LOGGER.debug(
+                    "While rewriting the query, did not get a metacardId corresponding to every attribute.");
+            LOGGER.debug("Original Delete By attribute was: {}", attributeName);
+            LOGGER.debug("Metacards unable to get Metacard ID from are: "
+                    + deleteRequest.getAttributeValues()
+                    .stream()
+                    .map(Object::toString)
+                    .filter(s -> !queryResults.contains(s))
+                    .collect(Collectors.joining(", ", "[", "]")));
+        }
     }
 }
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
@@ -27,7 +27,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
@@ -37,7 +37,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Sets;
 
 import ddf.catalog.Constants;
 import ddf.catalog.content.StorageException;
@@ -47,7 +46,6 @@ import ddf.catalog.content.operation.UpdateStorageResponse;
 import ddf.catalog.content.operation.impl.UpdateStorageRequestImpl;
 import ddf.catalog.content.plugin.PostUpdateStoragePlugin;
 import ddf.catalog.content.plugin.PreUpdateStoragePlugin;
-import ddf.catalog.core.versioning.MetacardVersion;
 import ddf.catalog.data.Attribute;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
@@ -55,10 +53,10 @@ import ddf.catalog.data.impl.AttributeImpl;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.history.Historian;
 import ddf.catalog.impl.FrameworkProperties;
+import ddf.catalog.operation.Operation;
 import ddf.catalog.operation.OperationTransaction;
 import ddf.catalog.operation.ProcessingDetails;
 import ddf.catalog.operation.QueryResponse;
-import ddf.catalog.operation.SourceResponse;
 import ddf.catalog.operation.Update;
 import ddf.catalog.operation.UpdateRequest;
 import ddf.catalog.operation.UpdateResponse;
@@ -79,14 +77,13 @@ import ddf.catalog.source.CatalogStore;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.source.InternalIngestException;
 import ddf.catalog.source.SourceUnavailableException;
-import ddf.catalog.source.UnsupportedQueryException;
 import ddf.catalog.util.impl.Requests;
 import ddf.security.SecurityConstants;
 import ddf.security.Subject;
 
 /**
  * Support class for update delegate operations for the {@code CatalogFrameworkImpl}.
- *
+ * <p>
  * This class contains two delegated update methods and methods to support them. No
  * operations/support methods should be added to this class except in support of CFI
  * update operations.
@@ -148,15 +145,10 @@ public class UpdateOperations {
             updateRequest = injectAttributes(updateRequest);
             updateRequest = setDefaultValues(updateRequest);
 
-            Map<String, Metacard> metacardMap = populateMetacards(updateRequest);
+            updateRequest = populateMetacards(updateRequest);
 
-            updateRequest = populateUpdateRequestPolicyMap(updateRequest, metacardMap);
-            updateRequest = processPreUpdateAccessPlugins(updateRequest, metacardMap);
-
-            updateRequest.getProperties()
-                    .put(Constants.OPERATION_TRANSACTION_KEY,
-                            new OperationTransactionImpl(OperationTransaction.OperationType.UPDATE,
-                                    metacardMap.values()));
+            updateRequest = populateUpdateRequestPolicyMap(updateRequest);
+            updateRequest = processPreUpdateAccessPlugins(updateRequest);
 
             updateRequest = processPreIngestPlugins(updateRequest);
             updateRequest = validateUpdateRequest(updateRequest);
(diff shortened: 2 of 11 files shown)


stustison:
Does this need backporting to 2.9.x?

rzwiefel:
Bamboo is a dirty rotten liar (probably). Just did a full build locally and it was fine.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -311,6 +316,44 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
     //
     // Private helper methods
     //
+    private DeleteRequest rewriteQuery(DeleteRequest deleteRequest)
+            throws UnsupportedQueryException {
+        if (Metacard.ID.equals(deleteRequest.getAttributeName())) {
```
jrnorth: Might want to store the value of `deleteRequest.getAttributeName()` in a local variable since you use it in a few places.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -311,6 +316,44 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
...
+                .anyOf(deleteRequest.getAttributeValues()
+                        .stream()
+                        .map(e -> frameworkProperties.getFilterBuilder()
+                                .attribute(deleteRequest.getAttributeName())
+                                .is()
+                                .equalTo()
+                                .text(e.toString()))
+                        .collect(Collectors.toList()));
+        SourceResponse response = catalogSupplier.get().query(new QueryRequestImpl(new QueryImpl(filter), false));
+        Map<Serializable, Metacard> newUpdateList = response.getResults()
```
jrnorth: Maybe name this something other than `newUpdateList` since this is not a list?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -311,6 +316,44 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
...
+                .stream()
+                .filter(r -> MetacardVersion.isNotVersion(r.getMetacard()))
+                .collect(Collectors.toMap(r -> (Serializable) r.getMetacard().getId(),
+                        Result::getMetacard));
+
+        // confirm every item from original update list is present.
+        if (!newUpdateList.values()
```
jrnorth: The comment and the code disagree here. This code verifies that every metacard in `newUpdateList` was in the original delete request. It will return true even if some metacards from the original delete request are not present in `newUpdateList`.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -137,11 +148,16 @@ public static Action ofMetacard(Metacard metacard) {
 
     public static final String VERSION_TAGS = prefix.apply("tags");
 
+    public static final String VERSION_TYPE = prefix.apply("type");
+
+    public static final String VERSION_TYPE_BINARY = prefix.apply("type-binary");
+
     private static MetacardType metacardVersion;
 
+    private static Set<AttributeDescriptor> versionDescriptors =
```
jrnorth: Should this be `final`?

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -196,6 +233,20 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
...
+        Optional<MetacardType> type = types.stream()
+                .filter(mt -> sourceMetacard.getMetacardType()
+                        .getName()
+                        .equals(mt.getName()))
+                .findFirst();
+
+        this.setVersionType(sourceMetacard.getMetacardType()
+                .getName());
+        if (!type.isPresent()) {
+            this.setVersionType(sourceMetacard.getMetacardType()
```
jrnorth: Already being set before the if statement.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -209,31 +260,83 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
         this.setTags(Collections.singleton(VERSION_TAG));
     }
 
+    private byte[] getVersionType(Metacard sourceMetacard) {
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                ObjectOutputStream oos = new ObjectOutputStream(baos)) {
+            oos.writeObject(sourceMetacard.getMetacardType());
+            return baos.toByteArray();
+        } catch (IOException e) {
+            throw new RuntimeException("Could not serialize MetacardType");
```
jrnorth: Do we want to include the `IOException` as the cause?

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -209,31 +260,83 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
...
+        }
+    }
+
     /**
-     * Returns a {@link BasicTypes#BASIC_METACARD} version of the given
+     * Returns a {@link Metacard} version of the given
```
jrnorth: `Returns a ... version of this {@link MetacardVersion}`?

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -209,31 +260,83 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
...
         result.setTags(getVersionTags(source));
+
+        sanitizeVersionAttributes(result);
         return result;
     }
 
+    private static void sanitizeVersionAttributes(/*Mutable*/ Metacard source) {
+        Consumer<String> nullifySourceAttribute = (s) -> source.setAttribute(new AttributeImpl(s,
+                (Serializable) null));
+        versionDescriptors.stream()
```
jrnorth: What about this?
```
Sets.difference(versionDescriptors, BASIC_METACARD.getAttributeDescriptors())
    .stream()
    ...
```

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: Done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: fixed, good catch

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: done

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: fixed

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: done

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: fixed

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -311,6 +317,51 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
     //
     // Private helper methods
     //
+    private DeleteRequest rewriteQuery(DeleteRequest deleteRequest)
```
pklinef: This needs a better name.  It should be clear that this method is verifying that the attributes are unique but it will rewrite the query if they are not unique due to history matches.

rzwiefel: That sounds good. what can we use thats not too long? Only think i can think of is something like `verifyUnambigousIdentifier`

pklinef: `rewriteRequestToAvoidHistoryConflicts`?  If you want, you might be able to get it shorter by including helpful JavaDoc.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +267,54 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+                .stream()
+                .map(m -> m.getAttribute(attributeName))
+                .map(Attribute::getValue)
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+
+        if(!deleteRequest.getAttributeValues()
+                .stream()
+                .allMatch(updatedIds::contains)){
+            LOGGER.warn("Could not obtain metacard ID's for all items in Delete Request!");
```
stustison: debug

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -257,6 +262,58 @@ public UpdateResponse update(UpdateStorageRequest streamUpdateRequest)
...
+        // confirm every item from original update list is present.
+        Set<String> updatedKeys = updateList.stream()
+                .map(Map.Entry::getValue)
+                .map(m -> m.getAttribute(attributeName))
+                .map(Attribute::getValue)
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        if (!originalKeys.stream()
+                .allMatch(updatedKeys::contains)) {
+            LOGGER.warn("Could not obtain metacard ID's for all items in Update Request!");
```
stustison: debug

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -209,31 +296,82 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
...
+                .map(byte[].class::cast)
+                .orElseThrow(cannotDeserializeException);
+
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(typeBytes);
+                ObjectInputStream ois = new ObjectInputStream(bais)) {
+            return Optional.ofNullable(ois.readObject())
+                    .filter(MetacardType.class::isInstance)
+                    .map(MetacardType.class::cast);
+        } catch (IOException | ClassNotFoundException e) {
+            LOGGER.error("Error while processing metacard type", e);
```
stustison: info

pklinef:
Do the metacards returned from `populateMetacards` include history metacards?  If so, I think you can just use that instead of having to query a second time.  A lot of the code in `rewriteQuery` looks very similar to `populateMetacards` and its call chain.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +267,54 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+        Filter filter = frameworkProperties.getFilterBuilder()
+                .anyOf(deleteRequest.getAttributeValues()
+                        .stream()
+                        .map(e -> frameworkProperties.getFilterBuilder()
+                                .attribute(attributeName)
+                                .is()
+                                .equalTo()
+                                .text(e.toString()))
+                        .collect(Collectors.toList()));
+        SourceResponse response = sourceOperations.getCatalog().query(new QueryRequestImpl(new QueryImpl(filter), false));
```
pklinef: Did you test a scenario where the response would be larger than the default page size of 20?  I am afraid if you have more than 20 history and non-history results, you might throw the `StopProcessingException`.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +267,54 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+                .map(m -> m.getAttribute(attributeName))
+                .map(Attribute::getValue)
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+
+        if(!deleteRequest.getAttributeValues()
+                .stream()
+                .allMatch(updatedIds::contains)){
+            LOGGER.warn("Could not obtain metacard ID's for all items in Delete Request!");
+            throw new StopProcessingException("Unable to complete query.");
```
pklinef: Can we provide a better message here?  This will be returned to the endpoint and possibly to a user.  The message in the warning seems close to what we would want to return.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -186,7 +251,17 @@ public static Action ofMetacard(Metacard metacard) {
      * @throws IllegalArgumentException
      */
     public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject) {
-        super(sourceMetacard, metacardVersion);
+        this(sourceMetacard, action, subject, Collections.singletonList(BasicTypes.BASIC_METACARD));
+    }
+
+    public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject,
+            List<MetacardType> types) {
```
pklinef: Why did you add a constructor that takes a list of types?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done (removed this, just throwing exception as it gets logged up the call chain)

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: Changed to message from above

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: done

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: rewriting because i can do it smarter in general, and eliminate this problem all together.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: Added

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +269,80 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+
+        SourceResponse response = sourceOperations.getCatalog()
+                .query(new QueryRequestImpl(new QueryImpl(filter,
+                        1,
+                        0,
+                        SortBy.NATURAL_ORDER,
+                        false,
+                        TimeUnit.SECONDS.toMillis(10))));
+
+        if (response.getResults()
```
pklinef: Checking the result size might not catch all cases.  It is possible one or more identifiers do not match but there are duplicates for others so that the final size matches.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +269,80 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+    private DeleteRequest rewriteRequestToAvoidHistoryConflicts(DeleteRequest deleteRequest)
+            throws UnsupportedQueryException {
+        final String attributeName = deleteRequest.getAttributeName();
+        if (Metacard.ID.equals(attributeName)) {
+            return deleteRequest;
+        }
+
+        Filter filter = deleteRequest.getAttributeValues()
+                .stream()
+                .map(Object::toString)
```
pklinef: Will this actually work for all possible unique identifiers?  It would be odd to use but there might be issues if they use unique dates, floats, location, and binary attributes in their requests.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +273,65 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+    private boolean notFoundValidMetacards(DeleteRequest deleteRequest, SourceResponse response) {
+        Set<String> originalKeys = deleteRequest.getAttributeValues()
+                .stream()
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        Set<String> responseKeys = response.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .map(m -> m.getAttribute(deleteRequest.getAttributeName()))
+                .map(Attribute::getValue)
```
coyotesqrl: Do you need to filter out null attributes or empty strings?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -260,6 +273,65 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        boolean notSameSize = response.getResults()
+                .size() != deleteRequest.getAttributeValues()
+                .size();
+        return !Sets.difference(originalKeys, responseKeys)
```
coyotesqrl: Switch this conditional order around to short-circuit the call to `Sets.difference()`

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -466,4 +544,26 @@ private boolean isStorageAvailable(StorageProvider storageProvider) {
         }
         return true;
     }
+
+    private void logFailedQueryInfo(DeleteRequest deleteRequest, String attributeName,
+            QueryResponse query) {
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug(
```
coyotesqrl: Use param interpolation.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -466,4 +544,26 @@ private boolean isStorageAvailable(StorageProvider storageProvider) {
...
+                    "While rewriting the query, did not get a metacardId corresponding to every attribute."
+                            + "Original Delete By attribute was: " + attributeName);
+            LOGGER.debug("Metacards unable to get Metacard ID from are: "
+                    + deleteRequest.getAttributeValues()
```
coyotesqrl: Pull this out into a string and use param interpolation. The concatenation, especially with the inline stream processing, just looks terrible.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -466,4 +544,26 @@ private boolean isStorageAvailable(StorageProvider storageProvider) {
...
+                    .stream()
+                    .map(Object::toString)
+                    .filter(s -> !query.getResults()
```
coyotesqrl: You're recreating this set for each attributeValue. Extract the logic to create the set and change this predicated to 
`.filter(s -> !efficientlyCreatedSet.contains(s))`

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -342,10 +418,12 @@ private QueryRequestImpl createQueryRequest(DeleteRequest deleteRequest) {
...
-                new QueryImpl(queryOperations.getFilterWithAdditionalFilters(idFilters));
-        queryImpl.setStartIndex(1);
-        queryImpl.setPageSize(deleteRequest.getAttributeValues()
-                .size());
+                new QueryImpl(queryOperations.getFilterWithAdditionalFilters(idFilters),
+                        1,  /* start index */
+                        0,  /* page size */
+                        SortBy.NATURAL_ORDER,
+                        false, /* total result count */
+                        0   /* timeout */);
```
pklinef: Should we use the timeout of the original query?  Unlimited timeout does not seem safe here.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -255,6 +266,50 @@ public UpdateResponse update(UpdateStorageRequest streamUpdateRequest)
...
+                .stream()
+                .map(Result::getMetacard)
+                .map(m -> m.getAttribute(updateRequest.getAttributeName()))
+                .map(Attribute::getValue)
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        boolean notSameSize = response.getResults()
+                .size() != updateRequest.getUpdates()
+                .size();
+        return !Sets.difference(originalKeys, responseKeys)
```
pklinef: Can we remove `notSameSize`.  I think the set difference should always give the correct answer and short circuit before ORing `notSameSize`.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -460,34 +518,40 @@ private UpdateRequest populateUpdateRequestPolicyMap(UpdateRequest updateRequest
...
+        if (notFoundValidMetacards(updateRequest, query)) {
+            logFailedQueryInfo(updateRequest, query);
+            throw new IngestException("Unable to find all metacards in update request.");
+        }
+
+        updateRequest = rewriteRequestToAvoidHistoryConflicts(updateRequest, query);
+
+        HashMap<String, Metacard> metacardMap = new HashMap<>(results.stream()
+                .map(Result::getMetacard)
+                .collect(Collectors.toMap(metacard -> getAttributeStringValue(metacard,
```
pklinef: `toMap` will throw an exception if the request accidentally includes two metacards with the same value for the update attribute.  Are we verifying that before we get here?  It does not look like it.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: I'm going to reverse these conditions so we can short circuit out of it per @coyotesqrl comment.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: redid so i dont' have to param interp

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: no

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -144,6 +145,8 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
                     deleteRequest.getAttributeValues()
                             .size());
 
+            deleteStorageRequest = new DeleteStorageRequestImpl(getDeleteMetacards(deleteRequest),
```
coyotesqrl: One of my concerns during the recent refactor was the large number of side-effects. While I assume it's safe to create the `deleteStorageRequest` at this point, the inputs to the constructor are different than in the earlier invocation of the existing code. Are there any negative ramifications to that?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -177,6 +180,16 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
         return deleteResponse;
     }
 
+    private List<Metacard> getDeleteMetacards(DeleteRequest deleteRequest) {
+        return Optional.of(deleteRequest)
+                .map(Operation::getProperties)
+                .map(p -> p.get(Constants.OPERATION_TRANSACTION_KEY))
+                .filter(o -> o instanceof OperationTransaction)
```
coyotesqrl: `.filter(OperationTransaction.class::instanceOf)`

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -267,6 +280,50 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+        Set<String> originalKeys = deleteRequest.getAttributeValues()
+                .stream()
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        Set<String> responseKeys = response.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .map(m -> m.getAttribute(deleteRequest.getAttributeName()))
+                .filter(Objects::nonNull)
+                .map(Attribute::getValue)
```
coyotesqrl: Only dealing with single-valued attributes? Am I worrying about something that's irrelevant to this change?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -267,6 +280,50 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+                .filter(Objects::nonNull)
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        boolean notSameSize = response.getResults()
```
jrnorth: Why not just `return !originalKeys.equals(responseKeys)`?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: I think so, since I believe down further down at the solr level we are already have similar assumptions. maybe @pklinef has other thoughts

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -262,6 +273,50 @@ public UpdateResponse update(UpdateStorageRequest streamUpdateRequest)
...
+    private boolean notFoundValidMetacards(UpdateRequest updateRequest, QueryResponse response) {
+        Set<String> originalKeys = updateRequest.getUpdates()
+                .stream()
+                .map(Map.Entry::getKey)
+                .map(Object::toString)
+                .collect(Collectors.toSet());
+        Set<String> responseKeys = response.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .map(m -> m.getAttribute(updateRequest.getAttributeName()))
```
jrnorth: Do we need to filter out nulls?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: agreed, moved back to original spot.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -560,4 +626,30 @@ private String getAttributeStringValue(Metacard mcard, String attribute) {
...
+                    .map(m -> m.getAttribute(attributeName))
+                    .filter(Objects::nonNull)
+                    .map(Attribute::getValue)
+                    .filter(Objects::nonNull)
+                    .map(Object::toString)
+                    .collect(Collectors.toSet());
+
+            LOGGER.debug(
+                    "While rewriting the query, did not get a metacardId corresponding to every attribute.");
+            LOGGER.debug("Original Delete By attribute was: " + attributeName);
```
jrnorth: Update

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -467,34 +525,40 @@ private UpdateRequest populateUpdateRequestPolicyMap(UpdateRequest updateRequest
...
-                    .map(Result::getMetacard)
-                    .collect(Collectors.toMap(metacard -> getAttributeStringValue(metacard,
-                            updateRequest.getAttributeName()), m -> m));
         } catch (FederationException e) {
             LOGGER.warn("Unable to complete query for updated metacards.", e);
+            throw new IngestException("Exception during runtime while performing update");
         }
 
-        return new HashMap<>();
+        List<Result> results = query.getResults();
```
jrnorth: This could be moved to the line just above where `metacardMap` is declared.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -467,34 +525,40 @@ private UpdateRequest populateUpdateRequestPolicyMap(UpdateRequest updateRequest
...
+        if (notFoundValidMetacards(updateRequest, query)) {
+            logFailedQueryInfo(updateRequest, query);
+            throw new IngestException("Unable to find all metacards in update request.");
+        }
+
+        updateRequest = rewriteRequestToAvoidHistoryConflicts(updateRequest, query);
+
+        HashMap<String, Metacard> metacardMap = new HashMap<>(results.stream()
```
jrnorth: The result of your stream operations will be a map - you don't need to create a new `HashMap` from it (unless `HashMap` is the particular implementation you need).

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: It doesn't look like the timeout is exposed in the original request. Should we just set a hardcoded timeout or should we keep it at zero?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -267,6 +280,50 @@ private DeleteResponse performLocalDelete(DeleteRequest deleteRequest,
...
+                .map(Serializable.class::cast)
+                .collect(Collectors.toList());
+
+        return new DeleteRequestImpl(updatedList,
+                Metacard.ID,
+                deleteRequest.getProperties(),
+                deleteRequest.getStoreIds());
+    }
+
+    private boolean notFoundValidMetacards(DeleteRequest deleteRequest, SourceResponse response) {
```
jrnorth: This method name is a bit confusing - what does "valid" mean in this context? Maybe something like `foundAllDeleteRequestMetacards` (which would require inverting your conditionals)?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: Need the hashmap because it Map is not Serializable

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: I just got rid of it entirely and inlined it

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -307,35 +365,38 @@ private DeleteRequest processPreDeletePolicyPlugins(DeleteRequest deleteRequest,
...
         }
+
+        List<Metacard> metacards = query.getResults()
+                .stream()
+                .map(Result::getMetacard)
+                .collect(Collectors.toList());
+
+        if (notFoundValidMetacards(deleteRequest, query)) {
+            logFailedQueryInfo(deleteRequest, query);
+            throw new StopProcessingException("Unable to remove all metacards contained in request.");
```
jrnorth: Should this message say something similar to what you have in your log statements?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: agreed, updated this and update to new name

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: done

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -209,31 +309,82 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
...
+                .map(byte[].class::cast)
+                .orElseThrow(cannotDeserializeException);
+
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(typeBytes);
+                ObjectInputStream ois = new ObjectInputStream(bais)) {
+            return Optional.ofNullable(ois.readObject())
+                    .filter(MetacardType.class::isInstance)
+                    .map(MetacardType.class::cast);
+        } catch (IOException | ClassNotFoundException e) {
+            LOGGER.info("Error while processing metacard type", e);
```
jrnorth: debug?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
pklinef: 0 is fine.

pklinef: Query requests provide a timeout but update and delete requests do not.  We should provide a timeout for these operation in the catalog framework.  I am fine with making that a separate ticket and leaving it 0 for now.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: fixed

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -473,4 +529,29 @@ private boolean isStorageAvailable(StorageProvider storageProvider) {
...
+                    .stream()
+                    .map(Result::getMetacard)
+                    .map(m -> m.getAttribute(attributeName))
+                    .filter(Objects::nonNull)
+                    .map(Attribute::getValue)
+                    .filter(Objects::nonNull)
+                    .map(Object::toString)
+                    .collect(Collectors.toSet());
+
+            LOGGER.debug(
```
coyotesqrl: I'm not in love with the three separate debug messages. However, I do recognize that you want to provide a decent amount of information that's well formatted. @stustison do you have any thoughts on this?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -473,4 +529,29 @@ private boolean isStorageAvailable(StorageProvider storageProvider) {
...
+                    "While rewriting the query, did not get a metacardId corresponding to every attribute.");
+            LOGGER.debug("Original Delete By attribute was: " + attributeName);
```
coyotesqrl: Use param interpolation.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -473,4 +529,29 @@ private boolean isStorageAvailable(StorageProvider storageProvider) {
...
+            LOGGER.debug("Metacards unable to get Metacard ID from are: "
```
coyotesqrl: Use param interpolation.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
rzwiefel: ticket created, https://codice.atlassian.net/browse/DDF-2431

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
stustison: Well, we certainly try to avoid using multiple log statements at the same level at the same point. I'm not entirely opposed to it, in this case, because they seem like log messages with a lot of info in them. Unless someone has a better idea, I'd say it is fine.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -170,10 +169,21 @@ public UpdateResponse update(UpdateRequest updateRequest)
...
+                    re);
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private Map<String, Metacard> getUpdateMap(UpdateRequest updateRequest) {
+        return (Map<String, Metacard>) Optional.of(updateRequest)
+                .map(Operation::getProperties)
+                .map(p -> p.get(Constants.ATTRIBUTE_UPDATE_MAP_KEY))
+                .filter(m -> m instanceof Map)
```
coyotesqrl: `Map.class::isInstance` should be functionally equivalent.
From [the javadocs](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isInstance-java.lang.Object-):
> This method is the dynamic equivalent of the Java language instanceof operator. The method returns true if the specified Object argument is non-null and can be cast to the reference type represented by this Class object without raising a ClassCastException. It returns false otherwise.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -467,34 +523,40 @@ private UpdateRequest populateUpdateRequestPolicyMap(UpdateRequest updateRequest
...
-                    .size()) {
-                throw new IngestException("Unable to find all metacards in update request.");
-            }
-
-            return results.stream()
-                    .map(Result::getMetacard)
-                    .collect(Collectors.toMap(metacard -> getAttributeStringValue(metacard,
-                            updateRequest.getAttributeName()), m -> m));
         } catch (FederationException e) {
             LOGGER.warn("Unable to complete query for updated metacards.", e);
```
coyotesqrl: I think `info` or `debug` would be more appropriate.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -467,34 +523,40 @@ private UpdateRequest populateUpdateRequestPolicyMap(UpdateRequest updateRequest
...
+            throw new IngestException("Exception during runtime while performing update");
         }
 
-        return new HashMap<>();
+        if (foundAllUpdateRequestMetacards(updateRequest, query)) {
```
coyotesqrl: Should this be `!found`?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -560,4 +624,30 @@ private String getAttributeStringValue(Metacard mcard, String attribute) {
...
+                    .map(m -> m.getAttribute(attributeName))
+                    .filter(Objects::nonNull)
+                    .map(Attribute::getValue)
+                    .filter(Objects::nonNull)
+                    .map(Object::toString)
+                    .collect(Collectors.toSet());
+
+            LOGGER.debug(
+                    "While rewriting the query, did not get a metacardId corresponding to every attribute.");
+            LOGGER.debug("Original Update By attribute was: " + attributeName);
```
coyotesqrl: Param interpolation here and next line.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -46,134 +59,186 @@
...
+
+        public static Action ofMetacard(Metacard metacard) {
+            if (isNotVersion(metacard)) {
+                throw new IllegalArgumentException(
+                        "Cannot get action of a non version metacard [" + metacard.getId() + "]");
+            }
+            Serializable svalue = Optional.ofNullable(metacard.getAttribute(ACTION))
+                    .map(Attribute::getValue)
+                    .orElse(null);
+            if (!(svalue instanceof String)) {
```
coyotesqrl: Feels like you could probably do your test/cast in the optional processing and return a String. Then if the returned String is null, throw your `IllegalArgumentException`.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: updated for clarity

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -209,31 +309,82 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
...
-        MetacardImpl result = new MetacardImpl(source, BasicTypes.BASIC_METACARD);
+        String typeString = (String) source.getAttribute(VERSION_TYPE)
+                .getValue();
+        Optional<MetacardType> typeFromExisting = types.stream()
+                .filter(mt -> mt.getName()
+                        .equals(typeString))
+                .findFirst();
+
+        MetacardImpl result = new MetacardImpl(source,
+                typeFromExisting.orElseGet(() -> getMetacardTypeBinary(source).orElseThrow(
```
coyotesqrl: I don't like this inlined retrieval of type/fallback/throw. I'd rather you pulled this out to the stream above and change the `Optional<MetacardType>` to a `MetacardType`. I believe it will read more cleanly that way.

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
```diff
@@ -196,6 +284,18 @@ public MetacardVersion(Metacard sourceMetacard, Action action, Subject subject)
         this.setVersionOfId(sourceMetacard.getId());
         this.setVersionTags(sourceMetacard.getTags());
 
+        Optional<MetacardType> type = types.stream()
```
coyotesqrl: This block of code is currently duplicated in this file and is a good target for extraction to a method.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
rzwiefel: done

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: ticket to fix https://codice.atlassian.net/browse/DDF-2437

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/MetacardVersion.java
rzwiefel: ticket to fix https://codice.atlassian.net/browse/DDF-2437

coyotesqrl:
With the ticket open for my two issues, 👍

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
pklinef: Delete and update requests currently assume single value attributes are used.

PR #1105, opened 2016-08-05
DDF-2300: Range Headers supported for Product Retrieval & itests

garrettfreibott:
**BACKPORT OF https://github.com/codice/ddf/pull/1064**

**What does this backport do?**
Backport is identical to original PR but also brings in some additional download itests to make sure that future backports won't break range-header and download functionality

**Choose 2 committers to review/merge the PR:**

@coyotesqrl 
@shaundmorris 

**-- ORIGINAL PR MESSAGE --**

**What does this PR do?**

Adds range header support for product retrieval retries. When the ReliableResourceDownloader attempts to retry retrieval of a product, it will no longer start over from the beginning of the download every time. There was also a major bug where DDF would properly use range headers to retrieve a product during a retry, but it would skip bytes in the input stream regardless of whether or not the server responding supported it or not, meaning unread data would be skipped and the file would become corrupted.

**How should this be tested?**

itests were added to check that the CSW retrieval works properly and to test retries via simulated network disconnects. There were also a few itests for simple opensearch downloads. There is a ticket out that we will be picking up soon to add an opensearch stub server similar to the csw stub server, but for now, the best way to test is to spin up an instance and test that product retrieval still works for all the various sources WITHOUT the data becoming corrupted. Ingest some sort of plain text file so that it can be easily double checked for missing characters. It would also probably be best to re-download the file immediately to check that the returned cache copy is not corrupted either.

**Any background context you want to provide?**

Since I needed to change ReliableResourceDownloader, it was necessary to change the URLResourceReader to also support Partial Content responses, so that had the side-effect of (hopefully) making all the sources support range headers. Hooray!

**Commit Messages:**

Product Retrievals will now properly request range headers and handle
Partial Content responses so that subsequent retries will not have
to redownload the entire product. Previously the endpoints would send
requests with range headers properly, but would not handle the response
correctly, potentially resulting in corrupted data.

After modifying the ReliableResourceDownloader[RRD], changes were needed
in the URLResourceReader to adjust for the fact that the RRD no longer
skipped ahead bytes when re-retrieving a resource. These changes also cause
all of the other endpoints that use it to support range header retrieval.
Hooray!

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/resource/download/ReliableResourceDownloader.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/resource/download/ReliableResourceDownloader.java
@@ -17,7 +17,6 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.Serializable;
 import java.util.Timer;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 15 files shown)


coyotesqrl:
👍

PR #1109, opened 2016-08-08
DDF-2334 Null values displayed in metacard organization attributes

ani6gup:
#### What does this PR do?
Provides null checks within metacard attributes. For instance, the phone number area code would return "(null) 123-4353" instead of "123-4353". The phone number and address metacard attributes now have more null checks. 
@mcalcote @vinamartin @gordocanchola @clockard @brianfelix 
@figliold
@stustison
Build and check it passed 
Bug was found while working on [DDF-2251](https://github.com/codice/ddf/pull/1053). Since the summary report pulled attribute values from the metacard, parts of these values such as the phone area code or the country of an address would show null if not provided. 
DDF-2334 
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
+++ b/catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
@@ -75,7 +75,8 @@ public class RegistryPackageConverter {
 
     private static final Map<String, String> METACARD_XML_NAME_MAP;
 
-    private static final InternationalStringTypeHelper INTERNATIONAL_STRING_TYPE_HELPER = new InternationalStringTypeHelper();
+    private static final InternationalStringTypeHelper INTERNATIONAL_STRING_TYPE_HELPER =
+            new InternationalStringTypeHelper();
 
     private static final SlotTypeHelper SLOT_TYPE_HELPER = new SlotTypeHelper();
 
@@ -262,7 +263,7 @@ public class RegistryPackageConverter {
         if (person.isSetTelephoneNumber()) {
             List<TelephoneNumberType> phoneNumbers = person.getTelephoneNumber();
             if (CollectionUtils.isNotEmpty(phoneNumbers)) {
-                phone = getPhoneNumberString(phoneNumbers.get(0));
+                phone = getPhoneNumber(phoneNumbers.get(0));
             }
         }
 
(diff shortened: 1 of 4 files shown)


catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -490,24 +497,28 @@ private static void setMetacardEmailAttribute(List<EmailAddressType> emailAddres
...
-                        digits.getExtension()));
+                if (CollectionUtils.isNotEmpty(metcardPhoneNumbers)) {
+                    metacard.setAttribute(metacardAttribute, (Serializable) metcardPhoneNumbers);
+                }
             }
         }
-        if (CollectionUtils.isNotEmpty(metacardPhoneNumbers)) {
-            metacard.setAttribute(metacardAttribute, (Serializable) metacardPhoneNumbers);
+    }
+
```
figliold: Would you be able to use StringUtils.joinWith() instead of this? https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#joinWith(java.lang.String,%20java.lang.Object...)

ani6gup: Yup, thank you -- was trying to find a cleaner way!

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -490,24 +497,28 @@ private static void setMetacardEmailAttribute(List<EmailAddressType> emailAddres
...
+    private static StringBuilder buildNonNullString(StringBuilder stringBuilder, String format,
+            String attributeValue, String delimeter) {
+        if (StringUtils.isNotBlank(attributeValue)) {
+            if (stringBuilder.length() > 0) {
+                stringBuilder.append(String.format("%s", delimeter));
```
stustison: I think you can just leave out the String.format in both places in this method and simply do `stringBuilder.append(...)`. The format call is redundant in this case.

ani6gup: modified it to just append `delimeter`

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -437,36 +438,42 @@ private static boolean isAttributeMultiValued(String attribute, MetacardImpl met
...
             String metacardAttribute, MetacardImpl metacard) {
+
         if (CollectionUtils.isNotEmpty(addresses)) {
+            StringBuilder addressString = new StringBuilder();
             PostalAddressType address = addresses.get(0);
 
-            String metacardAddress = String.format("%s, %s, %s %s, %s",
-                    address.getStreet(),
-                    address.getCity(),
+            addressString = buildNonNullString(addressString, "%s", address.getStreet(), "");
```
clockard: I would recommend using a method that just returns a non-null string. 
So this line would look more like
`addressString.append(getStringOrDefault(address.getStreet(),"");`
The getStringOrDefault is a super simple method then 
`private String getStringOrDefault(String value, String default) { return value == null ? default : value;}`
This way you just build up your address with appends

ani6gup: Thank you, will definitely try implementing this, it is much cleaner.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -490,24 +497,28 @@ private static void setMetacardEmailAttribute(List<EmailAddressType> emailAddres
 
     private static void setMetacardPhoneNumberAttribute(List<TelephoneNumberType> phoneNumbers,
             String metacardAttribute, MetacardImpl metacard) {
-        List<String> metacardPhoneNumbers = new ArrayList<>();
 
-        for (TelephoneNumberType digits : phoneNumbers) {
+        List<String> metcardPhoneNumbers = new ArrayList<>();
```
brianfelix: Should this variable be "metacardPhoneNumbers"?

brianfelix:
LGTM :+1:

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
ani6gup: Yup, thanks for the catch

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
clockard: After our conversation, realized this is more complicated than I thought it was at first. Ignore my comment

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
ani6gup: Realized joinWith was in version 3.5 of Apache and not released yet.

brianfelix:
Successfully built. LGTM :+1:

stustison:
👍

PR #1114, opened 2016-08-09
DDF-2345 Update Registry Node Info screen to handle invalid dates

mcalcote:
#### What does this PR do?

The Registry Node Information screen will not display information if an invalid date is in any node's information. This would result in node modals not being accessable and not being displayed in the UI. 

This PR fixes the issue by replacing any invalid dates/times with a valid date/time representing now. These new valid values are not saved to the node unless the user chooses to do so.

@vinamartin @clockard @gordocanchola @brianfelix @ani6gup @andrewkfiedler 

@pklinef
@stustison

1. Verify Invalid Dates are not accepted from the User
Boot up the Registry-App an attempt to enter an invalid date/time (can only be performed on FireFox or IE) for any of the date fields, and more specifically the custom fields.
Invalid dates should not be saved to the node.

2. Verify Invalid Dates from another Node do not break your node
Boot up the Registry-App and add a Remote Node with invalid date/time information. Once the node is connected, verify that the remote node is present in the node information tab and that the node modal can be opened/navigated. 

**Should be tested in Chrome, Firefox and IE**
This PR also addresses a minor IE syntax issue.

Notes from the ticket:
"When not using a chrome browser, the date fields in the Node Information modal are not forced to be any particular format. This can cause errors if a date is entered in an invalid format.
Need to add a regular expression to enforce the format of dates so they will always be correct. 

An additional check should also be done within the javascript that normalizes the dates on the registry node information screen. (Segment.js, line 44) In the case where there is an invalid date, put a blank date or another string to show that the format is invalid. This will make sure that even if bad data is coming from remote nodes, it will fail more gracefully. "

DDF-2345 Registry Node Information screen will not display information if an invalid date is in any node's information 
ddf-2345
screen shot 2016-08-09 at 10 02 31 am

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
+++ b/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
@@ -17,11 +17,12 @@
 define([
     'backbone',
     'underscore',
+    'moment',
     'jquery',
     'backboneassociation'
 
 
-], function (Backbone, _) {
+], function (Backbone, _, moment) {
 
     function getSlot(slots, key) {
         for (var index = 0; index < slots.length; index++) {
@@ -30,7 +31,6 @@ define([
             }
         }
     }
-
 
     var Field = {};
 
@@ -58,7 +58,20 @@ define([
                 }
             });
         },
-        isEmpty: function() {
+        isBlankDate: function () {
+            var timeLength = 0;
+            if (this.get('valueDate') !== undefined) {
+                timeLength = this.get('valueDate').length;
+            }
+            if (this.get('valueTime') !== undefined) {
+                timeLength += this.get('valueTime').length;
+            }
+            if (timeLength === 0) {
+                return true;
+            }
+            return false;
+        },
+        isEmpty: function () {
             return !this.get('value') || (_.isArray(this.get('value')) && this.get('value').length === 0);
         },
         addValue: function (val) {
@@ -73,8 +86,8 @@ define([
         },
         removeValue: function (index) {
             var values = this.get('value');
-            for (var count = 0; count < values.length; count++){
-                values[count] = this.get('value'+count);
+            for (var count = 0; count < values.length; count++) {
+                values[count] = this.get('value' + count);
             }
 
             values.splice(index, 1);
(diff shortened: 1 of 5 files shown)


catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
```diff
@@ -451,13 +493,20 @@ define([
             }
             //update and add values
             _.each(fields, function (field) {
-                field.saveData(backingData,ebrimTypes);
+                if (field.get('type') === 'date') {
```
clockard: Think this check should be moved to the Field.js validate method so the user will get some kind of indication the date is bad.

clockard:
Once comment is addressed LGTM 👍

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
mcalcote: Good call, will update.

brianfelix:
LGTM :+1:

vinamartin:
Looks good. 👍

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
```diff
@@ -26,13 +26,47 @@ define([
...
+            (function () {
+
+                function pad(number) {
+                    if (number < 10) {
+                        return '0' + number;
+                    }
+                    return number;
+                }
+
+                Date.prototype.toISOString = function () {
```
stustison: Why do you even need to do this? The only browser that wouldn't have this would be IE8 and below, which we don't even support. IE10+ only

mcalcote: Good point.

vinamartin:
Built this and all things passed. 
Manually verified the bug no longer exists. :shipit: 

screen shot 2016-08-15 at 11 20 56 am

vinamartin:
:shipit:

pklinef:
It looks like you are using date and time input types.  It does not look like that is very well supported for Firefox and IE which is probably how invalid values were being entered.  Backbone validate will help but it would be better if we used a date and time picking tool that can enforce the correct format for all our supported browsers.  

I think that would be a separate ticket from this one which adds model validation.

http://caniuse.com/#feat=input-datetime

@codice/ui

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
```diff
@@ -134,6 +134,10 @@ define([
                     if(this.get('valueUpperLon') && (this.get('valueUpperLon') > 180 || this.get('valueUpperLon')< -180)){
                         error = 'Invalid decimal longitude value. Must be -180 < lat < 180';
                     }
+                } else if(type === 'date'){
+                    if (isNaN(new Date(this.get('valueDate') + 'T' + this.get('valueTime') + 'Z'))) {
```
pklinef: Might consider using underscore for this check.

http://underscorejs.org/#isDate

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
```diff
@@ -26,13 +26,21 @@ define([
...
+    var typeCounters = {
         Service: 1,
         Organization: 1,
         Content: 1,
         ServiceBinding: 1
     };
 
+    function checkDateValidity(values) {
+        var dateVal = new Date(values);
+        if (!isNaN(dateVal.getTime())) {
```
pklinef: Here too.

http://underscorejs.org/#isDate

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
```diff
@@ -41,7 +49,13 @@ define([
                 properties['value' + i] = values[i];
             }
         } else if (values && field.get('type') === 'date') {
-            var normalizedDateTime = new Date(values).toISOString();
+            var normalizedDateTime;
+
+            if (checkDateValidity(values)) {
+                normalizedDateTime = new Date(values).toISOString();
+            } else {
+                normalizedDateTime = new Date().toISOString();
```
pklinef: Could move this to initialization on line 52 to remove the else condition.

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
```diff
@@ -451,13 +466,18 @@ define([
             }
             //update and add values
             _.each(fields, function (field) {
-                field.saveData(backingData,ebrimTypes);
+                if (field.get('type') === 'date') {
+                    var dateToParse = (new Date(field.get('valueDate') + 'T' + field.get('valueTime') + 'Z')).toISOString().split('T');
```
pklinef: I would recommend looking at moment.js for this kind of parsing and splitting.

http://momentjs.com/docs/

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
mcalcote: So it seems that _.isDate returns true if the date is invalid. This means that _.isDate(new Date("anything")) always is true which made me think to rely on isNaN again. I found a thread where the underscore devs discussed and ultimately rejected adding an _.isValidDate method, but there is one present in an underscore-contrib library that performs two evaluations `_.isDate(x) && !_.isNaN(x.getTime()` I would still have to create the new Date object so if it doesn't have any issues, it might make some sense to leave this  with a single isNaN check.

mcalcote: actually, I think `_.isDate(new Date(this.get('valueDate') + 'T' + this.get('valueTime') + 'Z')).getTime()` will work.

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
mcalcote: I think this runs into a similar issue as above. If i use _.isDate, it will always be true, even if invalid.

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
mcalcote: Good call, fixed

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
mcalcote: Cleaning this up using moment.js

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
mcalcote: Don't need this function after using moment.js

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Segment.js
mcalcote: incorporated moment.js

mcalcote:
A new ticket has been created to replace the existing time/date fields with a time/date picking tool: https://codice.atlassian.net/browse/DDF-2410

stustison:
👍

clockard:
Successfully tested that you can't add invalid dates through the ui in firefox.

mcalcote:
Most recent commit incorporates a fix for an issue where a blank date (in FF/IE) would result in the error message displaying.  I'm not totally happy with the flow (particularly in Field.js) so please feel free to weigh in on how I can improve it further.

One note of interest that had to be accounted for, it seems that moment.js acts differently depending on its environment. Specifically if `moment("").isValid()` is called in Chrome it returns `false`, while in FF/IE, it returns `true`.

vinamartin:
Just found a small bug, should be a quick change.
In Firefox, since it doesn't have the date picker you'll see an error message like this:
screen shot 2016-08-24 at 8 27 36 am

However Chrome's date picker is not the same format and does not match the error text:
screen shot 2016-08-24 at 8 28 46 am

Consider changing the error text.

mcalcote:
@vinamartin I agree that is unfortunate that the error text does not match in Chrome vs FF/IE. The issue stems from using the date/time fields which Chrome enforces itself. According to the Google Developer forums "Web authors have no way to change the date format because there currently is no standards to specify the format."

On the backend, all of the browsers handle the date in the format YYYY-MM-DD.

One point of note, the error should be seen much less frequently in Chrome since it has a date picker and the user must have a blank date with a valid time present for the error to appear (using the (x) button to clear the date specifically instead of just typing in the change over the existing value).

The error is formatted for the most likely use case, FF/IE, where that error text format matches the expected input format and there is no input guidance or enforcement on these fields. This is also where we were seeing the issue in the first place. 

Additionally, it is believed that our users primarily access this UI in FF/IE and as such the error text will be accurate for their experience. Further improvements should be addressed with ticket DDF-2410 https://codice.atlassian.net/browse/DDF-2410

pklinef:
@mcalcote Please backport to 2.9.x.

PR #1116, opened 2016-08-09
DDF-2362 Cleans and refactors the CRUD operation delegate methods for CatalogFrameworkImpl to highlight flow

coyotesqrl:
#### What does this PR do?
Highlights the logic in the methods delegated from `CatalogFrameworkImpl` and breaks apart the "junk drawer" of `OperationsCrudSupport` to better silo support methods for the delegate classes.

Also adds a new JUnit test to ensure that in fanout mode, the source id of metacards in a response are correctly updated.

@rzwiefel @ryeats  

@lessarderic
@pklinef

Full build/test run

N/A

DDF-2362

N/A

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -349,11 +249,20 @@ public CreateResponse create(CreateStorageRequest streamCreateRequest)
...
-            opsCrudSupport.handleStorageException(createStorageRequest,
-                    streamCreateRequest.getId(),
-                    e);
+            if (createStorageRequest != null) {
+                try {
+                    sourceOperations.getStorage()
+                            .rollback(createStorageRequest);
+                } catch (StorageException e1) {
+                    LOGGER.error("Unable to remove temporary content for id: "
+                            + createStorageRequest.getId(), e1);
```
shaundmorris: looks like you missed slf4j parameterized logging here.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -117,96 +125,29 @@ public void setHistorian(Historian historian) {
...
+        createRequest = validateLocalSource(createRequest);
 
-            for (AccessPlugin plugin : frameworkProperties.getAccessPlugins()) {
-                createRequest = plugin.processPreCreate(createRequest);
-            }
+        try {
+            createRequest = injectAttributes(createRequest);
+            createRequest = setDefaultValues(createRequest);
+            createRequest = updateCreateRequestPolicyMap(createRequest);
+            createRequest = processPrecreateAccessPlugins(createRequest);
```
rzwiefel: man there's so much similiarity... i wish there was just someway to clean this up ;)

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
```diff
@@ -466,29 +513,31 @@ private QueryResponse injectAttributes(QueryResponse response) {
...
         if (sourceResp != null) {
             if (sourceResp.getResults() == null) {
                 throw new UnsupportedQueryException(
                         "CatalogProvider returned null list of results from query method.");
             }
             if (sourceResp.getRequest() == null) {
-                sourceResp = new SourceResponseImpl(queryRequest,
-                        sourceResp.getProperties(),
-                        sourceResp.getResults());
+                sourceResp = new QueryResponseImpl(queryRequest,
```
coyotesqrl: Changed to return a `QueryResponse` so callers don't throw it on the floor.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
```diff
@@ -174,53 +132,15 @@ QueryResponse query(QueryRequest queryRequest, FederationStrategy strategy,
...
-            for (PostQueryPlugin service : frameworkProperties.getPostQuery()) {
-                try {
-                    queryResponse = service.process(queryResponse);
-                } catch (PluginExecutionException see) {
-                    LOGGER.warn("Error executing PostQueryPlugin: {}", see.getMessage(), see);
-                } catch (StopProcessingException e) {
-                    throw new FederationException("Query could not be executed.", e);
-                }
-            }
+            queryResponse = validateFixQueryResponse(queryResponse,
```
coyotesqrl: Getting the returned, updated value from `validateFixQueryResponse` instead of ignoring it.

lessarderic:
I haven't done a very thorough review yet, but had a quick, first-glance comment. It feels like we've basically pulled out chunks of the big `CatalogFrameworkImpl` class and created a bunch of `Support` classes from it. `Support` classes really feel like `Helper` classes, which aren't very OO or extensible and, IMHO, will end up having the same problem over time as the original class.

What I was hoping for with this refactoring was that we could step back, look at the domain objects we're dealing with (e.g., requests, responses, operations, plug-ins, etc.) and build the design/code around those concepts using design patterns such as builders, decorators, chain or responsibility, commands, etc., along with lambdas and generics for common (e.g., plug-in) processing.

Thoughts?

coyotesqrl:
While the few `*Support` classes definitely have that helper flavor, they've been reduced in size/scope to the point where they provide very small bits of functionality. We could break them up more into single-responsibility classes, but I think the extra DI overhead makes that a wash.

As for the `*Operations` classes, they're just straight delegates. Their public APIs are what, perhaps in an ideal world, would have been smaller interfaces that CFI would have exposed by delegating to referenced services. Instead, it's delegating to injected beans. But they do, I think, represent reasonable APIs.

The code in here has unfortunately gotten very sclerotic over time and doing a major overhaul on it is, I think, of lower priority. I'm happy to pursue those avenues, but this breakup and cleanup needed to happen first.

pklinef:
👍

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -349,11 +249,21 @@ public CreateResponse create(CreateStorageRequest streamCreateRequest)
...
+                            createStorageRequest.getId(),
+                            e1);
+                }
+            }
+            throw new IngestException(
+                    "Unable to store products for request: " + streamCreateRequest.getId(), e);
+
         } finally {
-            opsCrudSupport.commitAndCleanup(createStorageRequest,
+            opsStorageSupport.commitAndCleanup(createStorageRequest,
```
ryeats: At first glance i'm not sure this should be in a finally this seems weird that we would throw an exception above and then do this commitAndCleanup.  That being said this has nothing to do with this PR it looks like its always been this way.

coyotesqrl: I can look at cleaning that up further in one of the next passes.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
lessarderic: I agree. Couldn't we use the chain of responsibilities pattern and have the request go through a series of handlers instead of hard-coding the steps here? Or use decorators? I don't like the decorator approach as much as there would be a _lot_  of them which isn't as easy to follow, but I think it'd still be better than hard-coded, linear method calls.

And if our handlers are small, generic and somewhat configurable, we may be able to go as far as sharing them between different types of operations, e.g., create and update?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -72,6 +73,8 @@
 import ddf.catalog.source.InternalIngestException;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.util.impl.Requests;
+import ddf.security.SecurityConstants;
+import ddf.security.Subject;
 
 public class CreateOperations {
```
lessarderic: Don't know if we're planning on keeping these classes moving forward or if they are just steps in the overall refactoring, but if we plan to keep them, we really should add good JavaDoc comments to make sure we're clear on their scope and responsibilities.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -117,96 +125,29 @@ public void setHistorian(Historian historian) {
...
-                    createResponse.getProperties()
-                            .putAll(remoteCreateResponse.getProperties());
-                    createResponse.getProcessingErrors()
-                            .addAll(remoteCreateResponse.getProcessingErrors());
-                }
-            }
+            createRequest = processPreIngestPlugins(createRequest);
+            createRequest = validateCreateRequest(createRequest);
+            createResponse = getCreateResponse(createRequest);
+            createResponse = performRemoteCreate(createRequest, createResponse);
```
lessarderic: Same comment as above for response processing.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -232,20 +173,11 @@ public CreateResponse create(CreateRequest createRequest)
...
-                } catch (PluginExecutionException e) {
-                    LOGGER.info(
-                            "Plugin processing failed. This is allowable. Skipping to next plugin.",
-                            e);
-                }
-            }
+            createResponse = processPostIngestPlugins(createResponse);
         } catch (RuntimeException re) {
             LOGGER.warn(
                     "Exception during runtime while performing doing post create operations (plugins and pubsub)",
```
lessarderic: Error message looks a little weird, i..e, lots of `ing` going on :-)

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
coyotesqrl: The similarity is structural and looks obvious to the human eye, but internally there is far more variation. 

I believe the small handlers for create/update/etc. would either end up as a proliferation of distinct concrete implementors or we'd be passing lambdas all over the place to define the work for them to do.

I am certainly open to suggestions and can attempt a few strategies in one of the next phases of this update.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -117,96 +125,29 @@ public void setHistorian(Historian historian) {
...
-                            .addAll(remoteCreateResponse.getProcessingErrors());
-                }
-            }
+            createRequest = processPreIngestPlugins(createRequest);
+            createRequest = validateCreateRequest(createRequest);
+            createResponse = getCreateResponse(createRequest);
+            createResponse = performRemoteCreate(createRequest, createResponse);
 
         } catch (IngestException iee) {
             INGEST_LOGGER.warn("Ingest error", iee);
```
lessarderic: Since the ingest actually failed, shouldn't this be logged as an error?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
lessarderic: Also, shouldn't we log `RuntimeExceptions` as errors instead of warnings?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
coyotesqrl: Maybe. I kept its level where it had been previously. It's possible that it was intentionally set at WARN in order to prevent overloading the ingest log. I'll see if I can figure that out.

coyotesqrl: It was added as a warn log back in '14 in this commit: https://github.com/codice/ddf/commit/26d08d4d5c77fb6a2df0d45dfa3ff05f12bc4e49

As it does still throw the exception, it's _probably_ okay like that.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
coyotesqrl: Same message/level as it's been since late 2013, but I can clean up the text. I'm not sure it's worth bumping to error, however.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
lessarderic: I guess my hope here would be that, by creating smaller, simpler classes/lambdas, we would be able to extract the similarities between the different operation classes, make them more generic and possibly re-use some of those smaller classes across operations, if that makes sense. I'm not sure that moving code to private methods will allow us to achieve the same thing.

My concern I guess is that if we don't apply some of those patterns anywhere we won't be able to extract as much common code and reduce the amount of repeated code and/or logic as much as we should.

I'm fine waiting for another iteration to try some of this out though. Doesn't have to be part of this PR.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -349,11 +249,21 @@ public CreateResponse create(CreateStorageRequest streamCreateRequest)
 
             createResponse = create(createRequest);
         } catch (Exception e) {
```
lessarderic: Should we `catch RuntimeException` instead to make sure that if a new type of checked exception that gets thrown in the future we know this code needs to be updated to handle it?

coyotesqrl: I can specify the exceptions that are now thrown - `StorageException | IOException | RuntimeException` if that would be better.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -559,4 +474,171 @@ private CreateResponse validateFixCreateResponse(CreateResponse createResponse,
...
+        LOGGER.debug("Calling catalog.create() with {} entries.",
+                createRequest.getMetacards()
+                        .size());
+        if (!Requests.isLocal(createRequest)) {
+            return null;
+        }
+
+        CreateResponse createResponse = sourceOperations.getCatalog()
+                .create(createRequest);
+        createResponse = historian.version(createResponse);
```
lessarderic: Not familiar with how the `historian` works, but would it be better to have that call in the main `create()` method instead of "hidden" here?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/CreateOperations.java
```diff
@@ -559,4 +474,171 @@ private CreateResponse validateFixCreateResponse(CreateResponse createResponse,
...
+                LOGGER.info("Plugin processing failed. This is allowable. Skipping to next plugin.",
+                        e);
+            }
+        }
+        return createRequest;
+    }
+
+    private CreateRequest processPrecreateAccessPlugins(CreateRequest createRequest)
+            throws StopProcessingException {
+        for (AccessPlugin plugin : frameworkProperties.getAccessPlugins()) {
```
lessarderic: Couldn't we use `frameworkProperties.getAccessPlugins().forEach()` here?

coyotesqrl: Not easily as we're updating the createResponse on each invocation. It would require knowing that the call to `processPreCreate` is purely a side effect method.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
```diff
@@ -296,6 +173,193 @@ public DeleteResponse delete(DeleteRequest deleteRequest)
...
+
+    private DeleteResponse processPostDeleteAccessPlugins(DeleteResponse deleteResponse)
+            throws StopProcessingException {
+        for (AccessPlugin plugin : frameworkProperties.getAccessPlugins()) {
+            deleteResponse = plugin.processPostDelete(deleteResponse);
+        }
+        return deleteResponse;
+    }
+
+    private DeleteRequest populateDeleteRequestPolicyMap(DeleteRequest deleteRequest,
```
lessarderic: Looks like `deleteResponse` isn't used here.

Also, these methods look so much like each other in all the different `*Operations` classes, we really need to try and reduce that duplication in the follow up PRs.

coyotesqrl: It's used on lines 200-201.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
```diff
@@ -13,17 +13,302 @@
...
+                metacardMap.put(metacard.getId(), metacard);
+
+                ContentItem generatedContentItem = new ContentItemImpl(metacard.getId(),
+                        com.google.common.io.Files.asByteSource(tmpPath.toFile()),
+                        mimeTypeRaw,
+                        fileName,
+                        size,
+                        metacard);
+                contentItems.add(generatedContentItem);
+            } catch (Exception e) {
```
lessarderic: `catch RuntimeException`?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
```diff
@@ -13,17 +13,302 @@
...
+                                    attributeName,
+                                    defaultValue)));
+                });
+    }
+
+    private boolean hasNoValue(Attribute attribute) {
+        return attribute == null || attribute.getValue() == null;
+    }
+
+    private Metacard generateMetacard(String mimeTypeRaw, String id, String fileName,
```
lessarderic: Feels like this could be extracted to its own `MetacardFactory` class.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
```diff
@@ -13,17 +13,302 @@
...
+                .filter(attributeName -> hasNoValue(metacard.getAttribute(attributeName)))
+                .forEach(attributeName -> {
+                    registry.getDefaultValue(metacardType.getName(), attributeName)
+                            .ifPresent(defaultValue -> metacard.setAttribute(new AttributeImpl(
+                                    attributeName,
+                                    defaultValue)));
+                });
+    }
+
+    private boolean hasNoValue(Attribute attribute) {
```
lessarderic: Can't we put this on the `AttributeImpl` class?

lessarderic: Actually, scratch that. Could we add it to the `Attribute` interface with a default implementation, or will that break code using that interface?

coyotesqrl: I've forgotten where we landed on adding default methods to interfaces of public apis. I'd love to move it there if it's safe. @pklinef?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
coyotesqrl: There are a few checked exceptions thrown in this block as well.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
```diff
@@ -13,17 +13,302 @@
...
+    }
+
+
+    /**
+     * Updates any empty metacard attributes with those defined in the
+     * {@link DefaultAttributeValueRegistry}.
+     *
+     * @param metacard the metacard to update with default attribute values
+     */
+    void setDefaultValues(Metacard metacard) {
```
lessarderic: Maybe we should add this method to the `DefaultAttributeValueRegistry` interface and implementation instead of having it here?

coyotesqrl: That might be worth doing.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
```diff
@@ -13,17 +13,302 @@
...
+     */
     Metacard applyInjectors(Metacard original, List<AttributeInjector> injectors) {
         Metacard metacard = original;
         for (AttributeInjector injector : injectors) {
             metacard = injector.injectAttributes(metacard);
         }
         return metacard;
     }
+
+    void generateMetacardAndContentItems(List<ContentItem> incomingContentItems, Subject subject,
```
lessarderic: _Three_ output parameters... This method makes me sad :-(

coyotesqrl: It hurts me deep in my soul as well.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsStorageSupport.java
```diff
@@ -0,0 +1,147 @@
...
+        return storageRequest;
+    }
+
+    void commitAndCleanup(StorageRequest storageRequest, Optional<String> historianTransactionKey,
+            HashMap<String, Path> tmpContentPaths) {
+        if (storageRequest != null) {
+            try {
+                sourceOperations.getStorage()
+                        .commit(storageRequest);
+                historianTransactionKey.ifPresent(historian::commit);
```
lessarderic: Instead of having this logic here, shouldn't we just always delegate to the `historian` and let it decide what to do?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsStorageSupport.java
```diff
@@ -0,0 +1,147 @@
...
+                try {
+                    sourceOperations.getStorage()
+                            .rollback(storageRequest);
+                } catch (StorageException e1) {
+                    LOGGER.error("Unable to remove temporary content for id: {}",
+                            storageRequest.getId(),
+                            e1);
+                } finally {
+                    try {
+                        historianTransactionKey.ifPresent(historian::rollback);
```
lessarderic: Same question as above.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsStorageSupport.java
coyotesqrl: Might be able to do that. I'm not familiar with the internals of the Historian well enough to say if that would be an improvement.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -347,11 +232,21 @@ public UpdateResponse update(UpdateStorageRequest streamUpdateRequest)
             updateRequest.setProperties(streamUpdateRequest.getProperties());
             updateResponse = update(updateRequest);
         } catch (Exception e) {
```
lessarderic: `catch RuntimeException`?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -362,23 +257,27 @@ public UpdateResponse update(UpdateStorageRequest streamUpdateRequest)
     //
```
lessarderic: Unnecessary comment?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsStorageSupport.java
coyotesqrl: I guess I'm not clear how to extract cleanly to the Historian. I can ask the Historian to commit or rollback and have it check the key's value, but that would only be a one-line change in each of these two locations. I can't have the Historian manage the other storage calls and the exception processing, so I'm not sure how much improvement we'd get.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
coyotesqrl: I used this comment (and ones similar to it) to distinguish between those methods that were the delegated operations from `CatalogFramework` and those that are support methods, private or shared. It's not necessary, but I find the delineation useful.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsStorageSupport.java
rzwiefel: We are following the same paradigm as the storage provider. We have to do different things based on whether it commits or rollbacks  which is why historian is the same way. without making stateful instances of historians (which we would have to completely rewrite it) then it would not be an easy change.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/DeleteOperations.java
lessarderic: Yeah... Not sure how I could have missed that...

ryeats:
hero'd successfully

PR #1121, opened 2016-08-09
DDF-2372 Allow metacards to track which validators have been run

emanns95:
#### What does this PR do?
Adds which validators a metacard has been run through and indicates whether there are validation errors or warnings for each validator run.

@harrison-tarr 
@AzGoalie 

@coyotesqrl
@shaundmorris

Full Build

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
@@ -337,6 +337,18 @@ public class BasicTypes {
                 false /* tokenized */,
                 true /* multivalued */,
                 STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(Validation.FAILED_VALIDATORS_WARNINGS,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                true /* multivalued */,
+                STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(Validation.FAILED_VALIDATORS_ERRORS,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                true /* multivalued */,
+                STRING_TYPE));
         DESCRIPTORS = Collections.unmodifiableSet(descriptors);
 
         BASIC_METACARD = new MetacardTypeImpl(MetacardType.DEFAULT_METACARD_TYPE_NAME,
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/data/AttributeType.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/data/AttributeType.java
@@ -20,8 +20,7 @@ import java.io.Serializable;
  * <p>
  * {@link Attribute} values are bound to and the format expected to be used by those values
  *
- * @param <T>
- *            - The class used by values of this {@link AttributeType}
+ * @param <T> - The class used by values of this {@link AttributeType}
  */
 public interface AttributeType<T extends Serializable> extends Serializable {
 
@@ -45,7 +44,6 @@ public interface AttributeType<T extends Serializable> extends Serializable {
      * switch on <code>instanceof</code>, an {@link java.util.Enumeration} is used instead.
      *
      * @author michael.menousek@lmco.com
-     *
      */
     public enum AttributeFormat {
 
@@ -122,7 +120,7 @@ public interface AttributeType<T extends Serializable> extends Serializable {
          * {@link AttributeDescriptor#isIndexed() indexing},
          * {@link AttributeDescriptor#isTokenized() tokenizing}) will not be performed even if
          * indicated by the {@link AttributeDescriptor}. <br>
-         *
+         * <p>
          * <p>
          * <b>NOTE:</b> In order for classes to be deserialized by a {@link ddf.catalog.source.Source}, that class must
          * exist on the classpath of that {@link ddf.catalog.source.Source}. This may require additional parameters to
--- a/catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
+++ b/catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
@@ -105,6 +105,8 @@ public class MetacardValidityMarkerPlugin implements PreIngestPlugin {
             Map<String, Integer> counter) {
         Set<String> errors = new HashSet<>();
         Set<String> warnings = new HashSet<>();
+        Set<String> errorValidators = new HashSet<>();
+        Set<String> warningValidators = new HashSet<>();
 
         Metacard metacard = itemToMetacard.apply(item);
         Set<String> tags = metacard.getTags();
@@ -131,7 +133,13 @@ public class MetacardValidityMarkerPlugin implements PreIngestPlugin {
                             validatorName);
                     return null;
                 } else {
-                    getValidationProblems(validatorName, e, errors, warnings, counter);
+                    getValidationProblems(validatorName,
+                            e,
+                            errors,
+                            warnings,
+                            errorValidators,
+                            warningValidators,
+                            counter);
                 }
             }
         }
@@ -143,21 +151,28 @@ public class MetacardValidityMarkerPlugin implements PreIngestPlugin {
                 (List<Serializable>) new ArrayList<Serializable>(errors)));
         metacard.setAttribute(new AttributeImpl(Validation.VALIDATION_WARNINGS,
                 (List<Serializable>) new ArrayList<Serializable>(warnings)));
+        metacard.setAttribute(new AttributeImpl(Validation.FAILED_VALIDATORS_WARNINGS,
+                (List<Serializable>) new ArrayList<Serializable>(warningValidators)));
+        metacard.setAttribute(new AttributeImpl(Validation.FAILED_VALIDATORS_ERRORS,
+                (List<Serializable>) new ArrayList<Serializable>(errorValidators)));
 
         return item;
     }
 
     private void getValidationProblems(String validatorName, ValidationException e,
-            Set<String> errors, Set<String> warnings, Map<String, Integer> counter) {
+            Set<String> errors, Set<String> warnings, Set<String> errorValidators,
+            Set<String> warningValidators, Map<String, Integer> counter) {
         boolean validationErrorsExist = CollectionUtils.isNotEmpty(e.getErrors());
         boolean validationWarningsExist = CollectionUtils.isNotEmpty(e.getWarnings());
         if (validationErrorsExist || validationWarningsExist) {
             if (validationErrorsExist) {
                 errors.addAll(e.getErrors());
+                errorValidators.add(validatorName);
                 counter.merge(Validation.VALIDATION_ERRORS, 1, Integer::sum);
             }
             if (validationWarningsExist) {
                 warnings.addAll(e.getWarnings());
+                warningValidators.add(validatorName);
                 counter.merge(Validation.VALIDATION_WARNINGS, 1, Integer::sum);
             }
         } else {
@@ -206,7 +221,7 @@ public class MetacardValidityMarkerPlugin implements PreIngestPlugin {
                 validatorName);
     }
 
-    private String getValidatorName(MetacardValidator metacardValidator) {
+    protected String getValidatorName(MetacardValidator metacardValidator) {
         if (metacardValidator instanceof Describable) {
             return ((Describable) metacardValidator).getId();
         } else {
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
@@ -1017,9 +1017,7 @@ public class TestCatalog extends AbstractIntegrationTest {
 
     @Test
     public void testEnforceValidityErrorsOnly() throws Exception {
-        getServiceManager().startFeature(true,
-                "catalog-plugin-metacard-validation",
-                "sample-validator");
+        getServiceManager().startFeature(true, "sample-validator");
 
         //Configure to enforce validator
         configureEnforcedMetacardValidators(Collections.singletonList("sample-validator"));
@@ -1054,19 +1052,15 @@ public class TestCatalog extends AbstractIntegrationTest {
             deleteMetacard(id1);
             deleteMetacard(id2);
             deleteMetacard(id3, false);
-            getServiceManager().stopFeature(true,
-                    "catalog-plugin-metacard-validation",
-                    "sample-validator");
+            getServiceManager().stopFeature(true, "sample-validator");
             configureEnforceValidityErrorsAndWarningsReset();
-            configureShowInvalidMetacards("false", "true");
+            configureShowInvalidMetacardsReset();
         }
     }
 
     @Test
     public void testEnforceValidityWarningsOnly() throws Exception {
-        getServiceManager().startFeature(true,
-                "catalog-plugin-metacard-validation",
-                "sample-validator");
+        getServiceManager().startFeature(true, "sample-validator");
 
         //Configure to enforce validator
         configureEnforcedMetacardValidators(Collections.singletonList("sample-validator"));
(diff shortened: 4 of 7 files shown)


catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
```diff
@@ -216,4 +227,32 @@ public void setEnforceWarnings(boolean enforceWarnings) {
     public boolean getEnforceWarnings() {
         return enforceWarnings;
     }
+
+    protected class ValidationResult {
+        boolean errors;
```
AzGoalie: If you keep these package private you don't need the getters/setters. Otherwise id say make these private.

catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
```diff
@@ -111,6 +112,15 @@ public UpdateRequest process(UpdateRequest input)
                 boolean validationErrorsExist = CollectionUtils.isNotEmpty(e.getErrors());
                 boolean validationWarningsExist = CollectionUtils.isNotEmpty(e.getWarnings());
 
+                ValidationResult validationResult = new ValidationResult(false, false);
```
harrison-tarr: i think you could just say `ValidationResult validationResult = new ValidationResult(validationErrorsExist, validationWarningsExist)` and save a few lines

catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
```diff
@@ -216,4 +227,32 @@ public void setEnforceWarnings(boolean enforceWarnings) {
     public boolean getEnforceWarnings() {
         return enforceWarnings;
     }
+
+    protected class ValidationResult {
```
harrison-tarr: i would just make errors and warnings publicly accessible and just set them directly. I could be super wrong there, I don't know best practices for tuple implementations

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
```diff
@@ -337,6 +344,12 @@
                 false /* tokenized */,
                 true /* multivalued */,
                 STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(Validation.VALIDATORS,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                true /* multivalued */,
+                STRING_TYPE));
```
harrison-tarr: Should this be a `HASHMAP_TYPE`?

catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
emanns95: Since it saves a few lines, I'll do that unless anyone objects

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
```diff
@@ -99,6 +99,8 @@
      */
     public static final AttributeType<Short> SHORT_TYPE;
 
+    public static final AttributeType<HashMap> HASHMAP_TYPE;
```
coyotesqrl: Is there a reason you're creating a new type here and not just falling back to `OBJECT_TYPE`?

Regardless, using a concrete class for this definition instead of `Map` is definitely a problem.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
```diff
@@ -337,6 +344,12 @@
                 false /* tokenized */,
                 true /* multivalued */,
                 STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(Validation.VALIDATORS,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                true /* multivalued */,
```
coyotesqrl: Even assuming we'd want to keep this new map type, does it actually need to be multivalued? Is it N-Maps or a single Map of N key-value pairs?

catalog/core/catalog-core-api/src/main/java/ddf/catalog/data/AttributeType.java
```diff
@@ -117,12 +115,18 @@
         XML,
 
         /**
+         * {@link Attribute}s of this {@link AttributeFormat} <b>must</b> have a value that is a
+         * Java {@link java.util.HashMap}.
+         */
+        HASHMAP,
```
coyotesqrl: See my previous questions/comments on the need for this new type and further discussion as to the appropriate type it then might take.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
```diff
@@ -337,6 +344,12 @@
                 false /* tokenized */,
                 true /* multivalued */,
                 STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(Validation.VALIDATORS,
+                true /* indexed */,
```
coyotesqrl: I can't see the value in indexing this field.

catalog/plugin/catalog-plugin-metacard-validation/src/main/java/ddf/catalog/metacard/validation/MetacardValidityMarkerPlugin.java
```diff
@@ -216,4 +222,15 @@ public void setEnforceWarnings(boolean enforceWarnings) {
     public boolean getEnforceWarnings() {
         return enforceWarnings;
     }
+
+    protected static class ValidationResult {
```
coyotesqrl: This needs to implement `Serializable`.

Conversely, you could adopt a simple tokenization pattern for the validation results where you join the String representations of the two boolean values with a comma, colon, pipe, etc. Then on retrieval you'd split and convert to boolean to get your values.

coyotesqrl:
The new type needs to go away. Having completed the review, you should be using a non-indexed, single-valued `OBJECT_TYPE` to store your map of results. And as I also indicated, you'll either want to make your static helper class `Serializable` or swap it out for a tokenized string representation of the boolean values. 

I believe I'd opt for the latter approach, as it provides greater future flexibility in case we decide to track a third class of validation results. Otherwise, you'd need to explicitly define your read/write objects in that future version to handle deserializing results from the older version. With joined Strings, you can just look at tok[0] and tok[1] for now and in the future the absence of tok[2] and beyond could easily be managed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
brendan-hofmann: I don't think you can just define a new type here anyways, since these attribute formats correspond to the supported types in Solr, or at least our dynamic schema.

harrison-tarr: I agree that this should just be `OBJECT_TYPE`, but for what it's worth, I think `HashMap` was used here because `HashMap` implements serializable while `Map` doesn't.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
harrison-tarr: This is for the attribute descriptor, not the new type, so I think we do want to keep this. I don't think it should be multivalued though, the single map of N key-value pairs is what we want.

brendan-hofmann:
I need to give this some more consideration, but I am inherently opposed to tokenization as a design pattern for our taxonomies. It quickly leads to poorly defined constraints and values that are only parsable by select code that is 'in the know'.

I would also still like these values to be queryable if at all possible, which I believe means they have to be indexed as well.

All that being said, my first inclination would be some kind of xml serialization, since that is already a supported type and well-understood by both Solr and our catalog provider implementation.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
emanns95: Since I shouldn't be using a `HashMap` and `Map` isn't serializable, would you guys think it a better idea to instead keep track of a list of a `Serializable` object that contains the validator name, boolean for errors, and boolean for warnings?

coyotesqrl:
I see your point about using tokenization, so I'd recommend a lightweight serialized object that includes a `serialVersionUID`. It's unlikely it will change in the future in some incompatible way; at most it would add (an) additional boolean field(s).

You can skip implementing the read/write logic for now and let it serialize itself.

I don't feel strongly about this vs. XML serialization, however. I'd be happy either way.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
coyotesqrl: You can put a `Serializable` map into the `OBJECT_TYPE` field.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
emanns95: Is this okay now considering @brendan-hofmann's comment about wanting it indexed so it'll be queryable?

coyotesqrl: I personally don't see the value in making this a queryable field, but I could be wrong.

emanns95: For the case where a user might want to search on validators that have passed/failed

harrison-tarr: If we're going to make this queryable, we should make sure it works. Can you add an itest that ingests a metacard that fails a validator with errors and warnings, warnings, and neither and then query for them back out based on if the validator had errors, warnings, and neither? I'd like to see that it works and that the `<boolean, boolean>` tuple works.

shaundmorris: So was the decision to keep indexing `validators` or is there more discussion needed on it?

pklinef: I agree with Richard.  I do not think the validators active on the system would often change after initial setup.  That field would always have the same list if that is the common case.  

A list for validators that gave warnings and another for ones that gave errors might be useful when doing federated queries but not sure yet.  Would have to consider if a version is included with the validator name.

emanns95: I'm implementing it with two separate lists as @pklinef suggests. If that's the case, would we still care for it being queryable, @harrison-tarr?

harrison-tarr: I say yes

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/BasicTypes.java
emanns95: Wanted to note that this is no longer relevant as implementation has changed

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -2651,13 +2737,13 @@ private ValidatableResponse executeAdminOpenSearch(String format, String... quer
                 .then();
     }
 
-    protected String ingestXmlFromResource(String resourceName) throws IOException {
+    public String ingestXmlFromResource(String resourceName) throws IOException {
```
emanns95: I should actually change these back to protected. Changed it for testing

coyotesqrl:
👍

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1204,6 +1204,92 @@ public void testNoEnforceValidityErrorsOrWarnings() throws Exception {
     }
 
     @Test
+    public void testQueryErrors() throws Exception {
```
harrison-tarr: can this name be more descriptive?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1204,6 +1204,92 @@ public void testNoEnforceValidityErrorsOrWarnings() throws Exception {
...
+        getServiceManager().startFeature(true,
+                "catalog-plugin-metacard-validation",
```
harrison-tarr: metacard validation is started by default now, you shouldn't need to start it explicitly anymore

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1204,6 +1204,92 @@ public void testNoEnforceValidityErrorsOrWarnings() throws Exception {
...
+            deleteMetacard(id3);
+            getServiceManager().stopFeature(true,
+                    "catalog-plugin-metacard-validation",
+                    "sample-validator");
+            configureShowInvalidMetacards("false", "true");
+        }
+    }
+
+    @Test
+    public void testQueryWarnings() throws Exception {
```
harrison-tarr: same notes here

harrison-tarr:
Other than my notes about the itest names, lgtm

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
emanns95: Yeah it totally says it's already started every time I run it, but I figured better safe than sorry in case another test stops the feature for whatever reason. If this isn't a real concern I'll remove it

harrison-tarr: I don't know how performant it is to try to start a feature when one is already started. I would feel most comfortable with taking it out - if another test stops the feature, it's that test's responsibility to restart it.

emanns95:
@harrison-tarr Let me know if those new names are good or if I should change them again

shaundmorris:
👍

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1204,6 +1204,98 @@ public void testNoEnforceValidityErrorsOrWarnings() throws Exception {
     }
 
     @Test
+    public void testQueryByErrorFailedValidators() throws Exception {
+        getServiceManager().startFeature(true,
+                "catalog-plugin-metacard-validation",
+                "sample-validator");
```
emanns95: @harrison-tarr removing this part caused these tests to fail. Looking over, the sample validator should be configured to be used at the start of the test. I saw other tests (that I admittedly wrote) stop it at the end of their run, so I reasoned that running one of these tests alone should pass. It didn't, however, so I added it back in.

harrison-tarr: Sorry, I may have been confusing; the "catalog-plugin-metacard-validation" is automatically started, you still need to start the "sample-validator"

emanns95:
This passed full build and itests

coyotesqrl:
Successfully heroed.

PR #1140, opened 2016-08-16
DDF-2334 Null values displayed in metacard organization attributes

vinamartin:
https://github.com/codice/ddf/pull/1109

Provides null checks within metacard attributes. For instance, the phone number area code would return "(null) 123-4353" instead of "123-4353". The phone number and address metacard attributes now have more null checks.

@coyotesqrl
@lessarderic

### DIFF
--- a/catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
+++ b/catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
@@ -76,7 +76,8 @@ public class RegistryPackageConverter {
 
     private static final Map<String, String> METACARD_XML_NAME_MAP;
 
-    private static final InternationalStringTypeHelper INTERNATIONAL_STRING_TYPE_HELPER = new InternationalStringTypeHelper();
+    private static final InternationalStringTypeHelper INTERNATIONAL_STRING_TYPE_HELPER =
+            new InternationalStringTypeHelper();
 
     private static final SlotTypeHelper SLOT_TYPE_HELPER = new SlotTypeHelper();
 
@@ -263,7 +264,7 @@ public class RegistryPackageConverter {
         if (person.isSetTelephoneNumber()) {
             List<TelephoneNumberType> phoneNumbers = person.getTelephoneNumber();
             if (CollectionUtils.isNotEmpty(phoneNumbers)) {
-                phone = getPhoneNumberString(phoneNumbers.get(0));
+                phone = getPhoneNumber(phoneNumbers.get(0));
             }
         }
 
@@ -449,36 +450,48 @@ public class RegistryPackageConverter {
         return multiValued;
     }
 
-    private static String getPhoneNumberString(TelephoneNumberType digits) {
-        String phoneNumber;
-        if (StringUtils.isBlank(digits.getExtension())) {
-            phoneNumber = String.format("(%s) %s", digits.getAreaCode(), digits.getNumber());
-        } else {
-            phoneNumber = String.format("(%s) %s extension %s",
-                    digits.getAreaCode(),
-                    digits.getNumber(),
-                    digits.getExtension());
+    private static String getPhoneNumber(TelephoneNumberType digits) {
+        if (digits.getNumber()
+                .isEmpty()) {
+            return "N/A";
         }
 
-        return phoneNumber;
+        StringBuilder phoneNumberBuilder = new StringBuilder();
+
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getCountryCode(),
+                "",
+                "+%s");
+
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getAreaCode(),
+                " ",
+                "(%s)");
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder, digits.getNumber(), " ");
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getExtension(),
+                " ",
+                "ext %s");
+        return phoneNumberBuilder.toString();
     }
 
     private static void setMetacardAddressAttribute(List<PostalAddressType> addresses,
             String metacardAttribute, MetacardImpl metacard) {
+
         if (CollectionUtils.isNotEmpty(addresses)) {
+            StringBuilder addressString = new StringBuilder();
             PostalAddressType address = addresses.get(0);
 
-            String metacardAddress = String.format("%s, %s, %s %s, %s",
-                    address.getStreet(),
-                    address.getCity(),
-                    address.getStateOrProvince(),
-                    address.getPostalCode(),
-                    address.getCountry());
-
-            if (StringUtils.isNotBlank(metacardAddress)) {
-                metacard.setAttribute(metacardAttribute, metacardAddress);
+            addressString = buildNonNullString(addressString, address.getStreet(), " ");
+            addressString = buildNonNullString(addressString, address.getCity(), ", ");
+            addressString = buildNonNullString(addressString, address.getStateOrProvince(), ", ");
+            addressString = buildNonNullString(addressString, address.getPostalCode(), " ");
+            addressString = buildNonNullString(addressString, address.getCountry(), ", ");
+            if (StringUtils.isNotBlank(addressString.toString())) {
+                metacard.setAttribute(metacardAttribute, addressString.toString());
             }
         }
+
     }
 
     private static void setMetacardStringAttribute(String value, String metacardAttribute,
@@ -502,24 +515,35 @@ public class RegistryPackageConverter {
 
     private static void setMetacardPhoneNumberAttribute(List<TelephoneNumberType> phoneNumbers,
             String metacardAttribute, MetacardImpl metacard) {
-        List<String> metacardPhoneNumbers = new ArrayList<>();
 
-        for (TelephoneNumberType digits : phoneNumbers) {
-
-            if (StringUtils.isBlank(digits.getExtension())) {
-                metacardPhoneNumbers.add(String.format("(%s) %s",
-                        digits.getAreaCode(),
-                        digits.getNumber()));
-            } else {
-                metacardPhoneNumbers.add(String.format("(%s) %s extension %s",
-                        digits.getAreaCode(),
-                        digits.getNumber(),
-                        digits.getExtension()));
-            }
-        }
-        if (CollectionUtils.isNotEmpty(metacardPhoneNumbers)) {
+        if (phoneNumbers != null) {
+            List<String> metacardPhoneNumbers = phoneNumbers.stream()
+                    .map(RegistryPackageConverter::getPhoneNumber)
+                    .collect(Collectors.toList());
             metacard.setAttribute(metacardAttribute, (Serializable) metacardPhoneNumbers);
         }
+    }
+
+    private static StringBuilder buildNonNullString(StringBuilder stringBuilder,
+            String attributeValue, String delimiterBefore, String format) {
+        if (StringUtils.isNotBlank(attributeValue)) {
+            if (stringBuilder.length() > 0) {
+                stringBuilder.append(delimiterBefore);
+            }
+            if (StringUtils.isNotEmpty(format)) {
+                stringBuilder.append(String.format(format, attributeValue));
+            } else {
+                stringBuilder.append(attributeValue);
+            }
+        }
+        return stringBuilder;
+    }
+
+    private static StringBuilder buildNonNullString(StringBuilder stringBuilder,
+            String attributeValue, String delimiterBefore) {
+
+        return (buildNonNullString(stringBuilder, attributeValue, delimiterBefore, null));
+
     }
 
     private static void setAttributeFromMap(String metacardAttributeName,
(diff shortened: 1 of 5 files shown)


coyotesqrl:
Please copy the description from the #1109 PR into your description.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -490,24 +493,39 @@ private static void setMetacardEmailAttribute(List<EmailAddressType> emailAddres
...
+            }
+        }
+    }
 
-            if (StringUtils.isBlank(digits.getExtension())) {
-                metacardPhoneNumbers.add(String.format("(%s) %s",
-                        digits.getAreaCode(),
-                        digits.getNumber()));
+    private static StringBuilder buildNonNullString(StringBuilder stringBuilder,
+            String attributeValue, String format, String delimiterBefore) {
```
coyotesqrl: Feels weird to have the potentially empty value be penultimate.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -490,24 +493,39 @@ private static void setMetacardEmailAttribute(List<EmailAddressType> emailAddres
...
-        if (CollectionUtils.isNotEmpty(metacardPhoneNumbers)) {
-            metacard.setAttribute(metacardAttribute, (Serializable) metacardPhoneNumbers);
-        }
+        return stringBuilder;
+    }
+
+    private static StringBuilder buildNonNullString(StringBuilder stringBuilder,
+            String attributeValue, String delimiterBefore) {
+
+        return (buildNonNullString(stringBuilder, attributeValue, "", delimiterBefore));
```
coyotesqrl: Seems weird to me in a private method not to use `null` as your "I'm not here" flag.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -437,36 +438,38 @@ private static boolean isAttributeMultiValued(String attribute, MetacardImpl met
...
-            phoneNumber = String.format("(%s) %s", digits.getAreaCode(), digits.getNumber());
-        } else {
-            phoneNumber = String.format("(%s) %s extension %s",
-                    digits.getAreaCode(),
-                    digits.getNumber(),
-                    digits.getExtension());
-        }
-
-        return phoneNumber;
+    private static String getPhoneNumber(TelephoneNumberType digits) {
```
coyotesqrl: So we only support US phone numbers? No country codes?

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -490,24 +493,39 @@ private static void setMetacardEmailAttribute(List<EmailAddressType> emailAddres
 
     private static void setMetacardPhoneNumberAttribute(List<TelephoneNumberType> phoneNumbers,
             String metacardAttribute, MetacardImpl metacard) {
+
```
coyotesqrl: This should work and be a smidge clearer.
```java
if (phoneNumbers != null) {
    List<String> metacardPhoneNumbers = phoneNumbers.stream()
        .map(this::getPhoneNumber)
    metacard.setAttribute(metacardAttribute, (Serializable) metacardPhoneNumbers);
}

**Comment by coyotesqrl** [comment]
Are there any negative tests? I don't see any where we have empty phone number/address.

File: catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java (hunk 1 above)
Comment (vinamartin) [inline catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java]: Could you expand on this comment some more?

Comment (coyotesqrl) [inline catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java]: The overloaded method invokes this by passing in an empty string to `format` (which I've commented should probably be `null` instead, but that's a separate issue). It's more common when you've got an overloaded private method that the optional/null values be passed in at the end.

**Comment by coyotesqrl** [comment]
Small nits aside (and as these changes are being cherry picked from master I can certainly ignore them on this branch), 👍

**Comment by vinamartin** [comment]
@coyotesqrl I'm going to go ahead and address these comments on master

File: catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java (hunk 3 above)
Comment (vinamartin) [inline catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java]: Weird, we are supporting country codes but as you have pointed out we are not building it here. 
I'm fixing that up now. 

What are your thoughts on how that should be displayed?
I was thinking the following:

+1 (555) 555-5555 extension 555

Comment (coyotesqrl) [inline catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java]: That works for me. 

I'm assuming the area code/prefix/number/extension parts are pretty universal.

File: catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java (hunk 1 above)
Comment (vinamartin) [inline catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java]: Oh gotcha, yeah, now that I look at it this is pretty strange. I'll fix this up. 
As for your other comment, could you expand on that some more as well?

File: catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java (hunk 2 above)
Comment (coyotesqrl) [inline catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java]: Expanding: using an empty string for an internal method as a flag that "I should be ignored" seems strange. A null would be the more common indicator.

**Comment by vinamartin** [comment]
@coyotesqrl Could I get a re-review on this? I have addressed most of your comments for now. I'll cherry-pick this back and make a PR into master once I get a 👍

File: catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java [hunk 5]
```diff
@@ -437,36 +438,43 @@ private static boolean isAttributeMultiValued(String attribute, MetacardImpl met
...
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getAreaCode(),
+                " ",
+                "(%s)");
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder, digits.getNumber(), " ");
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getExtension(),
+                " ",
+                "extension %s");
```
coyotesqrl: I just want to throw in a comment here - I don't expect it to be addressed - that this is the sort of thing that kills us with I18N/L10N.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -437,36 +438,43 @@ private static boolean isAttributeMultiValued(String attribute, MetacardImpl met
...
-            String metacardAddress = String.format("%s, %s, %s %s, %s",
-                    address.getStreet(),
-                    address.getCity(),
-                    address.getStateOrProvince(),
-                    address.getPostalCode(),
-                    address.getCountry());
-
-            if (StringUtils.isNotBlank(metacardAddress)) {
-                metacard.setAttribute(metacardAttribute, metacardAddress);
+            addressString = buildNonNullString(addressString, address.getStreet(), " ");
```
coyotesqrl: Do you want the space character prefixing the first line of the address here?

coyotesqrl: Ah...never mind. When the StringBuilder has length 0 it's not used. Never mind.

coyotesqrl:
👍

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
lessarderic: `extension` makes the phone number fairly long and take a lot of space. Should we consider using `ext` instead?

vinamartin: Works for me. That's how we've been displaying it in the report anyway.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
vinamartin: I don't disagree. :)

lessarderic: We may want to create a constant for the string we use for extension.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
```diff
@@ -437,36 +438,43 @@ private static boolean isAttributeMultiValued(String attribute, MetacardImpl met
...
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getCountryCode(),
+                "",
+                "+s");
+
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder,
+                digits.getAreaCode(),
+                " ",
+                "(%s)");
+        phoneNumberBuilder = buildNonNullString(phoneNumberBuilder, digits.getNumber(), " ");
```
lessarderic: Does it even make sense to create a phone number string if the number is missing? Should we consider returning a blank string in that case?

catalog/spatial/registry/registry-ebrim-transformer/src/test/java/org/codice/ddf/registry/transformer/RegistryTransformerTest.java
```diff
@@ -118,26 +118,52 @@ public void testBasicInfo() throws Exception {
...
-                .getValue(), is("emailaddress@something.com"));
+    @Test
+    public void testNullMetacardAttributeValues() throws Exception {
+        MetacardImpl metacard = convert("/csw-null-metacard-attributes.xml");
+        assertRegistryMetacard(metacard);
+
+        assertThat(RegistryUtility.getStringAttribute(metacard,
+                RegistryObjectMetacardType.ORGANIZATION_NAME,
+                null), is("Codice"));
+        assertThat(RegistryUtility.getListOfStringAttribute(metacard,
```
lessarderic: We're not testing any of the scenarios where a piece of the address is `null` or missing. We're also not testing many of the other scenarios such as what if the extension isn't there.

Maybe we should extract `StreetAddressStringBuilder` and `PhoneNumberStringBuilder` classes and unit test those independently. That way you wouldn't have to do that low-level testing here.

lessarderic:
Just realized those are coming from a 2.10.0 PR that's already been merged so we may need to port those to the 2.10 branch as well :-(

👍  when the comments have been addressed.

catalog/spatial/registry/registry-ebrim-transformer/src/main/java/org/codice/ddf/registry/converter/RegistryPackageConverter.java
vinamartin: The UI has validation to prevent someone from putting in a phone number like this.
That should prevent that from happening.

coyotesqrl: Never rely on UI validations.

vinamartin: Again, I don't disagree. :)

vinamartin:
Made a ticket for the testing tech debt: https://codice.atlassian.net/browse/DDF-2419 Will address this soon. 

I'll be cherry-picking this back over to master, as well.

adimka:
Hero'd succesfully 👍

PR #1141, opened 2016-08-16
DDF-2368: Fixed resource usage not counting user's first download

rzwiefel:
#### What does this PR do?
fixes resource usage not counting against first users downloads

@bdeining @lcrosenbu @jrnorth @codice/core-api (which doesn't exist but should :) )

@jlcsmith
@lessarderic 

#### Checklist:

 - This was caused by the sources not putting the resource request properties into the response.

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
@@ -411,6 +411,8 @@ public class ResourceOperations extends DescribableImpl {
                 LOGGER.info("Unable to download resource", e);
             }
 
+            resourceResponse = putPropertiesInResponse(resourceRequest, resourceResponse);
+
             resourceResponse = validateFixGetResourceResponse(resourceResponse, resourceReq);
 
             HashMap<String, Set<String>> responsePolicyMap = new HashMap<>();
@@ -450,6 +452,21 @@ public class ResourceOperations extends DescribableImpl {
             throw new ResourceNotSupportedException(FAILED_BY_GET_RESOURCE_PLUGIN + e.getMessage());
         }
 
+        return resourceResponse;
+    }
+
+    private ResourceResponse putPropertiesInResponse(ResourceRequest resourceRequest,
+            ResourceResponse resourceResponse) {
+        if (resourceResponse != null) {
+            // must add the request properties into response properties in case the source forgot to
+            Map<String, Serializable> properties =
+                    new HashMap<>(resourceResponse.getProperties());
+            resourceRequest.getProperties()
+                    .forEach(properties::putIfAbsent);
+            resourceResponse = new ResourceResponseImpl(resourceResponse.getRequest(),
+                    properties,
+                    resourceResponse.getResource());
+        }
         return resourceResponse;
     }
 


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -411,6 +411,18 @@ ResourceResponse getResource(ResourceRequest resourceRequest, boolean isEnterpri
                 LOGGER.info("Unable to download resource", e);
             }
 
+            // must add the request properties into response properties in case the source forgot to
+            Map<String, Serializable> properties = new HashMap<>(resourceResponse.getProperties());
+            resourceRequest.getProperties()
+                    .entrySet()
+                    .stream()
+                    .forEach(e -> {
+                        properties.merge(e.getKey(), e.getValue(), (oldval, newval) -> oldval);
```
jrnorth: `putIfAbsent()`?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -411,6 +411,18 @@ ResourceResponse getResource(ResourceRequest resourceRequest, boolean isEnterpri
                 LOGGER.info("Unable to download resource", e);
             }
 
+            // must add the request properties into response properties in case the source forgot to
+            Map<String, Serializable> properties = new HashMap<>(resourceResponse.getProperties());
+            resourceRequest.getProperties()
```
jrnorth: You can also do `resourceRequest.getProperties().forEach((k, v) -> ...);`

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
rzwiefel: Done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
rzwiefel: did put if absent since thats the most concise

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -411,6 +411,17 @@ ResourceResponse getResource(ResourceRequest resourceRequest, boolean isEnterpri
                 LOGGER.info("Unable to download resource", e);
             }
 
+            if (resourceResponse != null) {
+                // must add the request properties into response properties in case the source forgot to
+                Map<String, Serializable> properties =
+                        new HashMap<>(resourceResponse.getProperties());
+                resourceRequest.getProperties()
```
jrnorth: `.forEach(properties::putIfAbsent)` should work.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -411,6 +411,17 @@ ResourceResponse getResource(ResourceRequest resourceRequest, boolean isEnterpri
                 LOGGER.info("Unable to download resource", e);
             }
 
+            if (resourceResponse != null) {
+                // must add the request properties into response properties in case the source forgot to
```
lessarderic: This method is already _huge_. Any chance we could move this to a private method?

lessarderic: Are we sure this is the right way to fix this? If we know what sources are not copying the properties, shouldn't we be fixing those sources instead? Also, what if a source must remove one of the request properties and not send it out? This change is going to prevent that from working.

kcwire: @lessarderic good points, as a counter argument - the catalogframework should not have to know how the individual sources are implemented and we can't enforce the properties to be maintained within each source.

lessarderic: Works for me then. It'd still be nice to extract a private method for this though :-)

rzwiefel: refactored method :+1:

jlcsmith:
Built and tested successfully.  LGTM 👍

lcrosenbu:
LGTM

shaundmorris:
Shouldn't we have backported this to 2.9.x?

bdeining:
@shaundmorris I'm pretty sure this should be backported

shaundmorris:
@rzwiefel, @jlcsmith, @lessarderic can you help make sure this is backported if you agree?

PR #1146, opened 2016-08-17
DDF-2390 updated log levels and log messages for all logs written

stustison:
#### What does this PR do?
Updates all log messages

@kcwire @lessarderic @pklinef @jlcsmith @shaundmorris 

@coyotesqrl 
@stustison
#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/admin/admin-poller-service/src/main/java/org/codice/ddf/catalog/admin/poller/AdminPollerServiceBean.java
+++ b/catalog/admin/admin-poller-service/src/main/java/org/codice/ddf/catalog/admin/poller/AdminPollerServiceBean.java
@@ -107,7 +107,7 @@ public class AdminPollerServiceBean implements AdminPollerServiceBeanMBean {
         try {
             objName = new ObjectName(AdminPollerServiceBean.class.getName() + SERVICE_NAME);
         } catch (MalformedObjectNameException e) {
-            LOGGER.error("Unable to create Admin Source Poller Service MBean with name [{}].",
+            LOGGER.info("Unable to create Admin Source Poller Service MBean with name [{}].",
                     AdminPollerServiceBean.class.getName() + SERVICE_NAME,
                     e);
         }
@@ -118,17 +118,17 @@ public class AdminPollerServiceBean implements AdminPollerServiceBeanMBean {
         try {
             try {
                 mBeanServer.registerMBean(this, objectName);
-                LOGGER.info(
+                LOGGER.debug(
                         "Registered Admin Source Poller Service Service MBean under object name: {}",
                         objectName.toString());
             } catch (InstanceAlreadyExistsException e) {
                 // Try to remove and re-register
                 mBeanServer.unregisterMBean(objectName);
                 mBeanServer.registerMBean(this, objectName);
-                LOGGER.info("Re-registered Admin Source Poller Service Service MBean");
+                LOGGER.debug("Re-registered Admin Source Poller Service Service MBean");
             }
         } catch (Exception e) {
-            LOGGER.error("Could not register MBean [{}].", objectName.toString(), e);
+            LOGGER.info("Could not register MBean [{}].", objectName.toString(), e);
         }
     }
 
@@ -136,10 +136,10 @@ public class AdminPollerServiceBean implements AdminPollerServiceBeanMBean {
         try {
             if (objectName != null && mBeanServer != null) {
                 mBeanServer.unregisterMBean(objectName);
-                LOGGER.info("Unregistered Admin Source Poller Service Service MBean");
+                LOGGER.debug("Unregistered Admin Source Poller Service Service MBean");
             }
         } catch (Exception e) {
-            LOGGER.error("Exception unregistering MBean [{}].", objectName.toString(), e);
+            LOGGER.info("Exception unregistering MBean [{}].", objectName.toString(), e);
         }
     }
 
@@ -158,20 +158,20 @@ public class AdminPollerServiceBean implements AdminPollerServiceBeanMBean {
                             try {
                                 return source.isAvailable();
                             } catch (Exception e) {
-                                LOGGER.warn("Couldn't get availability on source {}: {}",
+                                LOGGER.info("Couldn't get availability on source {}: {}",
                                         servicePID,
                                         e);
                             }
                         }
                     } catch (IOException e) {
-                        LOGGER.warn("Couldn't find configuration for source '{}'", source.getId());
+                        LOGGER.info("Couldn't find configuration for source '{}'", source.getId());
                     }
                 } else {
-                    LOGGER.warn("Source '{}' not a configured service", source.getId());
+                    LOGGER.info("Source '{}' not a configured service", source.getId());
                 }
             }
         } catch (InvalidSyntaxException e) {
-            LOGGER.error("Could not get service reference list");
+            LOGGER.info("Could not get service reference list");
         }
 
         return false;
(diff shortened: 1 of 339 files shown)


catalog/admin/admin-poller-service/src/main/java/org/codice/ddf/catalog/admin/poller/AdminPollerServiceBean.java
```diff
@@ -107,7 +107,7 @@ public AdminPollerServiceBean(ConfigurationAdmin configurationAdmin) {
         try {
             objName = new ObjectName(AdminPollerServiceBean.class.getName() + SERVICE_NAME);
         } catch (MalformedObjectNameException e) {
-            LOGGER.error("Unable to create Admin Source Poller Service MBean with name [{}].",
+            LOGGER.info("Unable to create Admin Source Poller Service MBean with name [{}].",
```
pklinef: We need to add integration tests to make sure all of the mbeans come up and work.  I am afraid we could ship with a broken mbean since this is now at info.

pklinef:
Please update the Log Level wiki to reflect these changes.  We can then use that resource in future PRs.

https://codice.atlassian.net/wiki/display/DDF/Log+Levels

coyotesqrl:
- `SystemPropertiesAdmin` line 311 - would this be better as `warn`?
- `ReuterSolrImport` lines 100-101 - better as a single log message?
- same with lines 104-105
- `SolrHttpWrapper` line 75 - should use param interpolation
- `EventProcessorImpl` line 355 - I believe this method can be deleted
- `RestSecurity` line 76 - I believe this method can be deleted
- `Security` line 300 - line too long; reformat
- `AbstractMetricsInterceptor` - no longer using the LOGGER; it can be removed and imports cleaned up
- `ConfigurationAdminMigration` line 226 - should use param interpolation instead of `String.format`
- `CommandJob` lines 117 and 119 - I'm not sure these shouldn't be `trace` or just removed.
- `AuthSSLX509TrustManager` lines 87-92 - should use param interpolation
- same with lines 106-111
- `ActiveSearch` line 154 - the logic has been changed here. If the result of `toCQL` on `request.getQuery()` is null, you'll now return null instead of `request.toString()`. That may be acceptable.
- `MetadataConfigurationParser` line 105 - let's remove the method name and incorrect line number "91" from the log message
- `EncryptionServiceImpl` lines 64-65 - better as a single log message?
- same with lines 79-80
- `IdpEndpoint` line 486 - should probably have a message here
- `UPBSTValidator` line 334 - should probably have a message here
- `UsernameTokenValidator` line 245 - should probably have a message here
- `X509PathTokenValidator` lines 233 and 236 - should probably have a message here
- `SolrClientFactory` line 127 - not sure I see the value of the two `debug` messages in a row here. Made more sense when they were at different levels.
- `SendEvent` line 241 - should use param interpolation 
- `CswSubscriptionEndpoint` line 595 - should use param interpolation
- `AbstractCswSource` line 1836 - should use param interpolation
- `GeoCoderEndpoint` line 110 - should use param interpolation
- `GeoNamesLocalIndex` line 80 - should use param interpolation instead of `String.format()`
- `AbstractFeatureConverter` line 242 - looks odd to be calling `toString()` on the exception
- `FeatureCollectionMessageBodyReaderWfs10` line 114 - looks odd to be calling `toString()` on the exception
- `WfsSource` line 439-440 - should this wrap and log the result of `handleWebApplicationException` instead?

stustison:
@coyotesqrl 
I'd like to leave all of the mbean exceptions at info.
Done.
Done.
Done.
Deleted.
Going to leave that one for now.
Formatting will just break at the (.
Removed logger.
Funny, didn't even notice that one.
Set to trace.
That one uses the wrong logger and I didn't want to change the loggers that classes were using since that can have more impact.
The logic didn't change, because .toCQL() never returns null, which means the request.toString() line was redundant and the log message was never going to be returned.
Removed.
Done.
Done.
Done.
Done.
Removed one log.
Done.
Done.
Done.
Done.
Done.
Agreed, I found a lot of places doing this. Seems like I missed this one.
And this one!
Yep, missed that.

coyotesqrl:
@stustison's replies to my giant comment were at least as hard to read as my comment! That said, 

👍

catalog/admin/admin-poller-service/src/main/java/org/codice/ddf/catalog/admin/poller/AdminPollerServiceBean.java
stustison: I don't believe that we would have noticed the error message anyways with the flood of messages that we had previously. I'll write a ticket to do mbean tests.

stustison:
I updated the wiki with the new logging guidelines.

lessarderic:
Regarding the guidelines, I'm not sure I'm comfortable with logging stack traces at debug only. Seems to me that, even though stack traces aren't useful to administrators, they are _really_ helpful to us during testing, integration, and at site. And since we're only logging at INFO and above by default, it will be really difficult to troubleshoot one-off and hard to reproduce problems.

Logs are not just for admins, they are also for developers and integrators. I would actually argue that they are only for developers and integrators and something else should be used to report problems to admins, but that's a different story :-)

If we feel that stack traces aren't useful to the admins (which I agree with), then maybe we should consider keeping the stack traces for error logs (and maybe some warnings) and have two appenders, one that includes the stack traces and one that doesn't (see http://stackoverflow.com/questions/348582/how-can-i-configure-log4j-to-not-print-the-exception-stacktrace).

Also, should we have guidelines related to when to log something? For instance, don't log anything (or at most a debug log with some context information) when an exception is caught and re-thrown, and let the service layer decide what to log (error, warning, info, with or without stack trace)?

Thoughts?

stustison:
@lessarderic As of right now, I haven't removed all of the stack traces for the warn/error messages. We can either log to different appenders, as you are suggesting, or we can leave it as is and change the guidelines. I wasn't going to tackle that in this PR.

brendan-hofmann:
@lessarderic I largely agree. Trying to make the same log serve developers, integrators, and admins simultaneously feels very much like a square peg/round hole situation to me.

The biggest problem I have with sending most of the stacktraces to `debug` (as the updated guidelines suggest) is that `debug` was already too noisy to be useful. I wouldn't want to troubleshoot anything without a debugger attached at this rate.

I think we also need to have a more detailed discussion about what exceptions are actually exceptional. Security, for example, logs a ton of stacktraces, but those exceptions are *supposed* to occur because that's the API of those services.

stustison:
@brendan-hofmann most of the security stack traces have also been set to debug level. Only the exceptional cases, like none of your claims handlers working will be logged higher.

brendan-hofmann:
That's exactly my point though. The security stacktraces log _constantly_, burying many more useful/infrequent stacktraces at the debug level. If there's only a message at warn/error with no direct way to get to the stacktrace, that's not very helpful.

I'm not trying to single out security, by the way. It's just one example, we have a lot of APIs that use exceptions for control flow. Federated sources can be another very noisy category.

stustison:
The guidelines also don't say that having stack traces at WARN or ERROR is forbidden. It simply says to avoid them, which I think is still fine.

stustison:
@brendan-hofmann that is why you should turn debug logging on for specific areas of the system that you're concerned with and not turn it on for every package.

lessarderic:
Feels to me like it should be the other way around for errors. You should almost always log the stack trace if you have one.

brendan-hofmann:
That presumes a generalized knowledge of the system that I think is a relative minority of the contributing developers. I am wary of too high a presumption of historical/architectural knowledge on the part of contributors.

stustison:
@brendan-hofmann If we are talking about a developer, you should be attaching a debugger. The log was already far too noisy to be used for debugging purposes anyways.

brendan-hofmann:
You still have to know _where_ to attach the debugger though. Almost the same problem, especially since so much of our architecture is decoupled, with no direct callchain between "what I changed" and "why isn't this working".

stustison:
I don't see how this change affects any of that.

PR #1156, opened 2016-08-18
DDF-2411 Improve presentation of Validate Command

emanns95:
#### What does this PR do?
Updates ValidateCommand to not print an extra number. Fixed --help message to provide accurate usage examples for CQL queries. Updated unit tests

@spearskw 
@harrison-tarr 
@brendan-hofmann 

@coyotesqrl
@lessarderic 

Full build. Use validate command to verify --help message is accurate and presentation is clear. Attempt with valid and invalid data.

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ValidateCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ValidateCommand.java
@@ -59,44 +59,54 @@ public class ValidateCommand implements Action {
 
     @Option(name = "--path", aliases = "-p", description = "The path to the file to be validated")
     @Completion(FileCompleter.class)
-    String path;
+    private String path;
 
     @Option(name = "--cqlQuery", aliases = "-q", description =
             "Search using CQL Filter expressions.\n" + "CQL Examples:\n"
-                    + "\tTextual:   search --cql \"title like 'some text'\"\n"
-                    + "\tTemporal:  search --cql \"modified before 2012-09-01T12:30:00Z\"\n"
-                    + "\tSpatial:   search --cql \"DWITHIN(location, POINT (1 2) , 10, kilometers)\"\n"
-                    + "\tComplex:   search --cql \"title like 'some text' AND modified before 2012-09-01T12:30:00Z\"")
-    String cqlQuery;
+                    + "\tTextual:   catalog:validate --cqlQuery \"title like 'some text'\"\n"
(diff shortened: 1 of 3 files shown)


spearskw:
LGTM :+1:

emanns95:
@coyotesqrl @jaymcnallie could I get you guys to look this over so I can get it into hero mode?

coyotesqrl:
👍

harrison-tarr:
Can we see screenshots of the difference in the message? LGTM though

adimka:
Hero'd successfully 👍

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ValidateCommand.java
```diff
@@ -84,15 +84,17 @@
     @Reference
     CatalogFramework catalog;
 
+    protected int numMetacardsWithErrorsOrWarnings;
```
lessarderic: Did we make this a `protected` field just to be able to assert on it in the unit tests? If so, we should reconsider since it breaks the class' encapsulation and doesn't really allow the unit tests to verify what really needs to be verified, i.e., that a report has been printed.

What I would suggest is keep `numMetacardsWithErrorsOrWarnings` a local variable instead of a field like it used to, make `ValidatePrinter` a non-static class, inject it in this command, mock it in the unit test and verify that `printSummary` is called with the right parameters.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ValidateCommand.java
```diff
@@ -84,19 +84,21 @@
     @Reference
     CatalogFramework catalog;
 
+    ValidatePrinter printer = new ValidatePrinter();
```
lessarderic: Instead of making this package private, we should take that in the constructor or have a setter method to set it and inject this using blueprint.

emanns95: @lessarderic  I’ve been looking at it with @spearskw and it’s more complicated than it looks. ValidatePrinter does not implement an interface. It was originally a set of helper methods we separated to its own class to break up functionality and not have one overcluttered class. I don’t think we’re able to inject a non-interface. I don’t think it’s necessary though because the purpose of this ticket was just to remove and extra print and fix the help message of the validate command before it gets released

PR #1159, opened 2016-08-18
DDF-2414: Create an OWASP runner that runs only on changed poms

tbatie:
#### What does this PR do?
@ryeats @adimka
@coyotesqrl
@figliold 
Pull down latest master. Add a commit that adds new lines to several poms. Do a clean install in the owasp-diff-runner module using the `-Powasp-diff-runner` flag and ensure it runs owasp against the changes poms and builds successfully. Next add a known vulnerability with a CVS of 7 or higher to a pom and build the module again with the `-Powasp-diff-runner` flag. This build should fail on OWASP from OWASP findings.
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests


tbatie:
This PR is definitely open to opinions/alternatives/feedback as I am not sure if I took the best approach with this.

libs/owasp-diff-runner/README.md
```diff
@@ -0,0 +1,25 @@
...
+-->
+
+Description:
+    -OWASP diff runner executes OWASP on any changed poms that reside in the current branch commits. This is intended to be used for pull requests but can be used locally as well.
+
+How to use:
+    - Activate the OWASP diff runner by specifying the `-Powasp-diff-runner` flag from the owasp-diff-runner module or any parent module.
+
+Important info:
+    - Compares ONLY commits, if you have changes that aren't committed then OWASP will not be ran against those changes.
```
coyotesqrl: s/b "...will not run against..."

libs/owasp-diff-runner/README.md
```diff
@@ -0,0 +1,25 @@
...
+    - OWASP diff runner runs against the local master branch and assumes the name of the branch is `master`. This also means if the local master branch or your current branch is not up to date with the latest master, incorrect pom differences may be ran.
```
coyotesqrl: s/b "...may be run."

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,176 @@
...
+
+    public static void main(String[] args) throws OwaspDiffRunnerException {
+
+        System.out.println(DASHES + " BEGIN OWASP DIFF AUDIT: " + DASHES);
+        mavenHome = getMavenHome();
+        localRepo = getLocalRepo();
+        String modulesOfChangedPoms = getModulesOfChangedPoms();
+        if (modulesOfChangedPoms.isEmpty()) {
+            System.out.println("No changed poms.");
+            return;
```
coyotesqrl: With a return here, you won't print out the `END OWASP DIFF AUDIT` message. It's not critical, but does make the output different.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        InvocationRequest request = new DefaultInvocationRequest();
+        request.setPomFile(new File(PROJECT_ROOT + File.separator + "pom.xml"));
+        request.setBaseDirectory(new File(PROJECT_ROOT));
+        request.setLocalRepositoryDirectory(new File(localRepo));
+
+        request.setGoals(Arrays.asList("dependency-check:check",
```
coyotesqrl: You didn't want to/weren't able to change this to full maven coordinates?

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,176 @@
...
+
+        InvocationResult mavenBuildResult;
+
+        try {
+            mavenBuildResult = invoker.execute(request);
+        } catch (MavenInvocationException e) {
+            throw new OwaspDiffRunnerException(
+                    OwaspDiffRunnerException.UNABLE_TO_RUN_MAVEN + modulesOfChangedPoms, e);
+        }
+        System.out.println(DASHES + " END OWASP DIFF AUDIT: " + DASHES);
```
coyotesqrl: This should probably be in a finally block so it always prints.

coyotesqrl:
👍

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
tbatie: I abstained from doing it since there is no version property for the owasp plugin. I would have had to change several poms to do so. It may be a good idea to do this when upgrade the plugin

ryeats:
👍  hero'd

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,184 @@
...
+
+    public static final String GIT_SHORT_BRANCH_NAME_COMMAND = "git rev-parse --abbrev-ref HEAD";
+
+    private static final String DASHES = StringUtils.repeat("-", 25);
+
+    public static final String BEGIN_OWASP_AUDIT = DASHES + " BEGIN OWASP DIFF AUDIT " + DASHES;
+
+    public static final String END_OWASP_AUDIT = DASHES + " END OWASP DIFF AUDIT " + DASHES;
+
+    public static final String PROJECT_ROOT = System.getProperty("user.dir")
```
figliold: Will this work on Windows?

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,184 @@
...
+    private static String getMavenHome() throws OwaspDiffRunnerException {
+        String mavenHome;
+        String mavenVersionInfo;
+
+        try {
+            mavenVersionInfo = IOUtils.toString(runTime.exec(MAVEN_VERSION_COMMAND)
+                    .getInputStream());
+
+            //parsing console response, confirmed to work with at least maven version 3.3.9
+            mavenHome = Arrays.stream(mavenVersionInfo.split(System.getProperty("line.separator")))
```
figliold: Do you think this might break since the output of maven may change?

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,184 @@
...
+                                + localRepoProperty);
+            }
+            return localRepoProperty;
+        }
+
+        try {
+            mavenHelpInfo = IOUtils.toString(runTime.exec(MAVEN_SETTINGS_COMMAND)
+                    .getInputStream());
+
+            //parsing console response, confirmed to work with at least maven version 3.3.9
```
figliold: This seems fragile also since it parses command output.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
tbatie: Yes, the user.dir system property works on windows

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
tbatie: I wanted to keep it stray away from depending on environments/environment variables. For example, I don't have my M2_HOME set on my own machine. The `mvn -v` is guaranteed to be there though.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
tbatie: This was one of my concerns but I did a search for the last time this was altered and it has not in a very long time, especially this local repo section. I couldn't find a environment independent way of getting the local repo path as well. I was hoping for a maven property that I could use but none seem to exist.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.maven.shared.invoker.DefaultInvocationRequest;
+import org.apache.maven.shared.invoker.DefaultInvoker;
+import org.apache.maven.shared.invoker.InvocationRequest;
+import org.apache.maven.shared.invoker.InvocationResult;
+import org.apache.maven.shared.invoker.Invoker;
+import org.apache.maven.shared.invoker.MavenInvocationException;
+import org.codehaus.plexus.util.StringUtils;
+
+public class OwaspDiffRunner {
+
```
figliold: Is there a certain min version of git that needs to be installed for all of these git commands to work?

tbatie: That I am unsure of. I'm not very concerned about it though, seeing how this tool must be triggered by a profile flag and is really intended to be ran in a CI environment which is configurable. I can leave a comment saying the git version it has been tested on for historical purposes though.

PR #1160, opened 2016-08-19
DDF-2416 Fix ip validation of guest tokens

brendan-hofmann:
#### What does this PR do?
Fix scoped ipv6 addresses in guest tokens.
@clockard 
@glenhein 
@harrison-tarr 
@pklinef
@stustison
Verify that your computer asks for a guest token with a scoped ipv6 address. I was able to reliably produce this by using `?dev=true` to change my cert. Make sure (in a private session) that you can access guest pages like the login page.
This should be backported to 2.9.x and the same fix applied to videographer in alliance.
DDF-2416
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/security/sts/security-sts-guestvalidator/src/main/java/org/codice/ddf/security/validator/guest/GuestValidator.java
+++ b/platform/security/sts/security-sts-guestvalidator/src/main/java/org/codice/ddf/security/validator/guest/GuestValidator.java
@@ -13,8 +13,9 @@
  */
 package org.codice.ddf.security.validator.guest;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.util.List;
-import java.util.regex.Pattern;
 
 import org.apache.cxf.sts.request.ReceivedToken;
 import org.apache.cxf.sts.token.validator.TokenValidator;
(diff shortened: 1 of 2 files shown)


shaundmorris:
Hmm, was there not a bamboo build for this PR for some reason?

stustison:
👍

platform/security/sts/security-sts-guestvalidator/src/main/java/org/codice/ddf/security/validator/guest/GuestValidator.java
```diff
@@ -65,10 +57,11 @@ private GuestAuthenticationToken getGuestTokenFromTarget(ReceivedToken validateT
     }
 
     private boolean validIpAddress(String address) {
-        return IPV4_PATTERN.matcher(address)
-                .matches() || IPV6_STD_PATTERN.matcher(address)
-                .matches() || IPV6_HEX_COMPRESSED_PATTERN.matcher(address)
-                .matches();
+        try {
+            return InetAddress.getByName(address) != null;
+        } catch (UnknownHostException e) {
```
glenhein: Any reasonable chance that a security exception could be thrown? "SecurityException - if a security manager exists and its checkConnect method doesn't allow the operation"

glenhein:
👍

platform/security/sts/security-sts-guestvalidator/src/main/java/org/codice/ddf/security/validator/guest/GuestValidator.java
brendan-hofmann: @stustison may know better, but I can't think of any *reasonable* chance. I don't find any usages of `java.lang.SecurityManager` in our codebase. Since there's no real expectation of it being thrown, I think it is appropriate not to swallow an unchecked exception.

That's my reasoning, but I could be persuaded otherwise. :)

brendan-hofmann:
@shaundmorris very mysterious. I can push a rebase and see if that triggers one.

platform/security/sts/security-sts-guestvalidator/src/main/java/org/codice/ddf/security/validator/guest/GuestValidator.java
stustison: I think it is fine as is.

brendan-hofmann:
There we go, that seems to have done it. 😄

PR #1125, opened 2016-08-10
DDF-2378 Add 'internal' to registry packages that export service interfaces as well as the rest endpoint path

mcalcote:
#### What does this PR do?
Add 'internal' to registry packages that export service interfaces and to the rest endpoint path

Most all of the osgi services that were created with the experimental banner in the class header but should be updated to contain 'internal' in the package so that it is very clear it is not an external interface and can be changed. 
This should also be the changed for the registry rest endpoint. It is currently ../services/registries/...
and should be changed to ../services/internal/registries/...

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged; if a component team is listed, at least one of its members needs to approve)?
@clockard @gordocanchola @vinamartin 

#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@pklinef
@shaundmorris

#### How should this be tested?
Build, ensure registry endpoints still function as expected

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #1163, opened 2016-08-19
DDF-2378 -> 2.9.x Add 'internal' to registry packages that export service interfaces as well as the rest endpoint path

mcalcote:
#### What does this PR do?

Add 'internal' to registry packages that export service interfaces and to the rest endpoint path

Most all of the osgi services that were created with the experimental banner in the class header but should be updated to contain 'internal' in the package so that it is very clear it is not an external interface and can be changed. 
This should also be the changed for the registry rest endpoint. It is currently ../services/registries/...
and should be changed to ../services/internal/registries/...

@pklinef
@shaundmorris
Bamboo Should Pass
This PR was merged on master: https://github.com/codice/ddf/pull/1125


pklinef:
👍

PR #1164, opened 2016-08-19
DDF-2422: TestFederation iTest Intermittent Failures

garrettfreibott:
### What does this PR do?

Fixes the TestFederation intermittent itests failures that have been occurring recently with the bamboo nightly master build and bamboo PR builds.

### Who is reviewing it?

@emanns95 
@brendan-hofmann 
@figliold 

### Choose 2 committers to review/merge the PR.

@kcwire 
@coyotesqrl 

### How should this be tested?

Bamboo PR tests passed a few times in a row, but the master build and other PR builds should be watched after this gets merged in to make sure that the issue is completely gone.

### DIFF
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
@@ -253,6 +253,7 @@ public class TestFederation extends AbstractIntegrationTest {
             CswSourceProperties cswStubServerProperties =
                     new CswSourceProperties(CSW_STUB_SOURCE_ID);
             cswStubServerProperties.put("cswUrl", CSW_STUB_SERVER_PATH.getUrl());
+            cswStubServerProperties.put(POLL_INTERVAL, CSW_SOURCE_POLL_INTERVAL);
             getServiceManager().createManagedService(CswSourceProperties.FACTORY_PID,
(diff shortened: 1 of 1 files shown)


figliold:
LGTM 👍

emanns95:
LGTM :+1:

coyotesqrl:
👍

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -253,6 +253,7 @@ public void beforeExam() throws Exception {
             CswSourceProperties cswStubServerProperties =
                     new CswSourceProperties(CSW_STUB_SOURCE_ID);
             cswStubServerProperties.put("cswUrl", CSW_STUB_SERVER_PATH.getUrl());
+            cswStubServerProperties.put(POLL_INTERVAL, CSW_SOURCE_POLL_INTERVAL);
```
garrettfreibott: Yes, the poll interval was changed to 10 minutes to match the other test sources. @figliold noticed that if his itests took long enough that the sources were re-polled, they because marked as unavailable. There should be more digging into why exactly this occurs in itests, but for the time being this should cause the unavailability issues with the cswStubServer to stop. As far as we can tell, this issue does not happen in production, only in the itests.

kcwire:
👍

PR #1167, opened 2016-08-19
DDF-2341 update PdfInputTransformer for new taxonomy

glenhein:
#### What does this PR do?

Update PdfInputTransformer for new taxonomy. This is the second attempt, which includes the content metadata extractors.

@bdeining 
@brendan-hofmann 
@jrnorth 
@rzwiefel 
@coyotesqrl 
@jlcsmith
Ingest a PDF and check SOLR for non-new-taxonomy fields.
DDF-2341
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
+++ b/catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
@@ -13,10 +13,8 @@
  */
 package ddf.catalog.transformer.input.pdf;
 
-import java.awt.Graphics2D;
-import java.awt.RenderingHints;
-import java.awt.image.BufferedImage;
-import java.io.ByteArrayOutputStream;
+import static org.apache.commons.lang3.Validate.notNull;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Calendar;
@@ -30,19 +28,14 @@ import java.util.stream.Collectors;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.time.FastDateFormat;
 import org.apache.pdfbox.pdmodel.PDDocument;
 import org.apache.pdfbox.pdmodel.PDDocumentInformation;
-import org.apache.pdfbox.pdmodel.PDPage;
-import org.apache.pdfbox.pdmodel.common.PDMetadata;
 import org.apache.pdfbox.pdmodel.encryption.InvalidPasswordException;
-import org.apache.pdfbox.rendering.ImageType;
-import org.apache.pdfbox.rendering.PDFRenderer;
-import org.apache.pdfbox.tools.imageio.ImageIOUtil;
 import org.apache.tika.parser.AutoDetectParser;
 import org.apache.tika.parser.ParseContext;
 import org.apache.tika.parser.Parser;
 import org.apache.tika.sax.ToTextContentHandler;
+import org.codice.ddf.platform.util.TemporaryFileBackedOutputStream;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.ServiceReference;
@@ -50,39 +43,75 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.ContentHandler;
 
-import com.google.common.html.HtmlEscapers;
-import com.google.common.io.FileBackedOutputStream;
 import com.google.common.net.MediaType;
 
 import ddf.catalog.content.operation.ContentMetadataExtractor;
 import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.Metacard;
-import ddf.catalog.data.impl.BasicTypes;
+import ddf.catalog.data.MetacardType;
 import ddf.catalog.data.impl.MetacardImpl;
 import ddf.catalog.data.impl.MetacardTypeImpl;
+import ddf.catalog.data.types.Contact;
+import ddf.catalog.data.types.Core;
+import ddf.catalog.data.types.Media;
+import ddf.catalog.data.types.Topic;
 import ddf.catalog.transform.CatalogTransformerException;
 import ddf.catalog.transform.InputTransformer;
 import ddf.catalog.transformer.common.tika.TikaMetadataExtractor;
 import ddf.catalog.util.impl.ServiceComparator;
 
 public class PdfInputTransformer implements InputTransformer {
+
     private static final Logger LOGGER = LoggerFactory.getLogger(PdfInputTransformer.class);
 
-    private static final int RESOLUTION_DPI = 44;
+    private final PDDocumentGenerator pdDocumentGenerator;
 
-    private static final float IMAGE_QUALITY = 1.0f;
+    private final GeoPdfParser geoParser;
 
-    private static final float IMAGE_HEIGHTWIDTH = 128;
-
-    private static final String FORMAT_NAME = "jpg";
-
-    private static final GeoPdfParser GEO_PDF_PARSER = new GeoPdfParser();
-
-    private static final FastDateFormat DATE_FORMAT = FastDateFormat.getInstance(
-            "yyyy-MM-dd'T'HH:mm:ssZZ");
+    private final PdfThumbnailGenerator pdfThumbnailGenerator;
 
     private Map<ServiceReference, ContentMetadataExtractor> contentMetadataExtractors =
             Collections.synchronizedMap(new TreeMap<>(new ServiceComparator()));
+
+    private MetacardType metacardType;
+
+    private boolean usePdfTitleAsTitle;
+
+    /**
+     * @param metacardType          must be non-null
+     * @param usePdfTitleAsTitle    must be non-null
+     * @param pdDocumentGenerator   must be non-null
+     * @param geoParser             must be non-null
+     * @param pdfThumbnailGenerator must be non-null
+     */
+    public PdfInputTransformer(MetacardType metacardType, Boolean usePdfTitleAsTitle,
+            PDDocumentGenerator pdDocumentGenerator, GeoPdfParser geoParser,
+            PdfThumbnailGenerator pdfThumbnailGenerator) {
+        notNull(metacardType, "metacardType must be non-null");
+        notNull(usePdfTitleAsTitle, "usePdfTitleAsTitle must be non-null");
+        notNull(pdDocumentGenerator, "pdDocumentGenerator must be non-null");
+        notNull(geoParser, "geoParser must be non-null");
+        notNull(pdfThumbnailGenerator, "pdfThumbnailGenerator must be non-null");
+
+        this.metacardType = metacardType;
+        this.usePdfTitleAsTitle = usePdfTitleAsTitle;
+        this.pdDocumentGenerator = pdDocumentGenerator;
+        this.geoParser = geoParser;
+        this.pdfThumbnailGenerator = pdfThumbnailGenerator;
+    }
+
+    @SuppressWarnings("unused")
+    public boolean isUsePdfTitleAsTitle() {
+        return usePdfTitleAsTitle;
+    }
+
+    /**
+     * @param usePdfTitleAsTitle must be non-null
+     */
+    public void setUsePdfTitleAsTitle(Boolean usePdfTitleAsTitle) {
+        notNull(usePdfTitleAsTitle, "usePdfTitleAsTitle must be non-null");
+        this.usePdfTitleAsTitle = usePdfTitleAsTitle;
+    }
 
     public void addContentMetadataExtractors(
             ServiceReference<ContentMetadataExtractor> contentMetadataExtractorRef) {
@@ -119,7 +148,7 @@ public class PdfInputTransformer implements InputTransformer {
     }
 
     private Metacard transformWithoutExtractors(InputStream input, String id) throws IOException {
-        try (PDDocument pdfDocument = PDDocument.load(input)) {
+        try (PDDocument pdfDocument = pdDocumentGenerator.apply(input)) {
             return transformPdf(id, pdfDocument);
         } catch (InvalidPasswordException e) {
             LOGGER.debug("Cannot transform encrypted pdf", e);
@@ -129,7 +158,7 @@ public class PdfInputTransformer implements InputTransformer {
 
     private Metacard transformWithExtractors(InputStream input, String id)
             throws IOException, CatalogTransformerException {
-        try (FileBackedOutputStream fbos = new FileBackedOutputStream(1000000)) {
+        try (TemporaryFileBackedOutputStream fbos = new TemporaryFileBackedOutputStream()) {
             try {
                 IOUtils.copy(input, fbos);
             } catch (IOException e) {
@@ -151,11 +180,11 @@ public class PdfInputTransformer implements InputTransformer {
             }
 
             try (InputStream isCopy = fbos.asByteSource()
-                    .openStream(); PDDocument pdfDocument = PDDocument.load(isCopy)) {
+                    .openStream(); PDDocument pdfDocument = pdDocumentGenerator.apply(isCopy)) {
 
                 return transformPdf(id, pdfDocument, plainText);
             } catch (InvalidPasswordException e) {
-                LOGGER.warn("Cannot transform encrypted pdf", e);
+                LOGGER.debug("Cannot transform encrypted pdf", e);
                 return initializeMetacard(id);
             }
         }
@@ -175,18 +204,21 @@ public class PdfInputTransformer implements InputTransformer {
                     .flatMap(Collection::stream)
                     .collect(Collectors.toSet());
 
-            metacard = new MetacardImpl(new MetacardTypeImpl(BasicTypes.BASIC_METACARD.getName(),
-                    BasicTypes.BASIC_METACARD,
+            metacard = new MetacardImpl(new MetacardTypeImpl(metacardType.getName(),
+                    metacardType,
                     attributes));
+
             for (ContentMetadataExtractor contentMetadataExtractor : contentMetadataExtractors.values()) {
                 contentMetadataExtractor.process(contentInput, metacard);
             }
         } else {
-            metacard = new MetacardImpl();
+            metacard = new MetacardImpl(metacardType);
         }
 
         metacard.setId(id);
-        metacard.setContentTypeName(MediaType.PDF.subtype());
+        metacard.setContentTypeName(MediaType.PDF.toString());
+        metacard.setAttribute(Media.TYPE, MediaType.PDF.toString());
+        metacard.setAttribute(Core.DATATYPE, "Document");
 
         return metacard;
     }
@@ -206,9 +238,10 @@ public class PdfInputTransformer implements InputTransformer {
 
         extractPdfMetadata(pdfDocument, metacard);
 
-        metacard.setThumbnail(generatePdfThumbnail(pdfDocument));
+        pdfThumbnailGenerator.apply(pdfDocument)
+                .ifPresent(metacard::setThumbnail);
 
-        Optional.ofNullable(GEO_PDF_PARSER.getWktFromPDF(pdfDocument))
+        Optional.ofNullable(geoParser.apply(pdfDocument))
                 .ifPresent(metacard::setLocation);
 
         return metacard;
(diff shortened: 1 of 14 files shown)


catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -157,10 +256,20 @@ private Metacard transformWithExtractors(InputStream input, String id)
             } catch (InvalidPasswordException e) {
                 LOGGER.warn("Cannot transform encrypted pdf", e);
                 return initializeMetacard(id);
+
```
jrnorth: Extra line

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -219,101 +330,48 @@ private Metacard transformPdf(String id, PDDocument pdfDocument, String contentI
...
-            addXmlElement("modificationDate", DATE_FORMAT.format(modificationDate), metadataField);
             LOGGER.debug("PDF Modification date was: {}", DATE_FORMAT.format(modificationDate));
         }
 
-        String title = documentInformation.getTitle();
-        if (StringUtils.isNotBlank(title)) {
-            metacard.setTitle(title);
-            addXmlElement("title", title, metadataField);
+        if (usePdfTitleAsTitle) {
+            String title = documentInformation.getTitle();
```
jrnorth: It might be worth extracting the "get value -> check not blank -> set attribute" pattern into a private method since you use it in multiple places.

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
```diff
@@ -55,16 +74,9 @@ class PdfInputTransformerSpecTest extends Specification {
 
     void verifySamplePdfMetacard(Metacard metacard) {
         verifyThumbnailSize(metacard.thumbnail)
-        assert metacard.contentTypeName == 'pdf'
-        assert metacard.title == 'Microsoft Word - Document1'
```
jrnorth: Shouldn't we verify the metacard title here?

jlcsmith:
@glenhein please rebase this PR with master

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -81,9 +85,99 @@
...
+    }
+
+    @SuppressWarnings("StatementWithEmptyBody")
+    private static byte[] generatePdfThumbnail(PDDocument pdfDocument) throws IOException {
+
+        PDFRenderer pdfRenderer = new PDFRenderer(pdfDocument);
+
+        if (pdfDocument.getNumberOfPages() < 1) {
+            /*
+             * Can there be a PDF with zero pages??? Should we throw an error or what? The
```
jlcsmith: If the PDF has zero pages I don't think we need to worry about thumbnail generation.

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
```diff
@@ -43,9 +53,18 @@ class PdfInputTransformerSpecTest extends Specification {
     }
 
     void setup() {
+        metacardType = new MetacardTypeImpl("pdf", [
+            new AssociationsAttributes(),
+            new ContactAttributes(),
+            new MediaAttributes(),
+            new DateTimeAttributes(),
```
bdeining: nit : It looks like DateTime and Location aren't used on the actual `MetacardTypeImpl`, so they don't need to be added to the mock one

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -157,10 +256,20 @@ private Metacard transformWithExtractors(InputStream input, String id)
...
+
             }
         }
     }
 
+    private Metacard transformWithoutExtractors(InputStream input, String id) throws IOException {
+        try (PDDocument pdfDocument = inputStreamPDDocumentFunction.apply(input)) {
+            return transformPdf(id, pdfDocument);
+        } catch (InvalidPasswordException e) {
+            LOGGER.warn("Cannot transform encrypted pdf", e);
```
stustison: debug

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
coyotesqrl: Agreed. Should probably return a null or empty byte[].

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -118,18 +226,9 @@ public Metacard transform(InputStream input, String id)
...
-        } catch (InvalidPasswordException e) {
-            LOGGER.warn("Cannot transform encrypted pdf", e);
-            return initializeMetacard(id);
-        }
-    }
-
     private Metacard transformWithExtractors(InputStream input, String id)
             throws IOException, CatalogTransformerException {
-        try (FileBackedOutputStream fbos = new FileBackedOutputStream(1000000)) {
+        try (TemporaryFileBackedOutputStream fbos = new TemporaryFileBackedOutputStream()) {
```
coyotesqrl: This method isn't using your new `inputStreamPDDocumentFunction`. Is that an issue?

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
coyotesqrl: There are also two `warn` methods up in `transformeWithExtractors` that should be changed to `debug`.

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: fixed

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: fixed

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
glenhein: by default, the input transformer does not set the title. There are separate test method in PdfInputTransformerTest that tests the title functionality.

coyotesqrl:
Other than my question re: `transformWithExtractors` not using the provided func, this lgtm.

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: changed to Optional<byte[]>

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
glenhein: removed

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: changed

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: not an issue, inputStreamPDDocumentFunction is used to make the static PDDocument::load a settable function for testing

coyotesqrl: It won't be an issue for your unit tests where you pass in the function?

glenhein: through the ctor

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
```diff
@@ -15,8 +15,17 @@ package ddf.catalog.transformer.input.pdf
...
 import ddf.catalog.content.operation.ContentMetadataExtractor
 import ddf.catalog.data.Metacard
+import ddf.catalog.data.MetacardType
 import ddf.catalog.data.impl.AttributeDescriptorImpl
 import ddf.catalog.data.impl.BasicTypes
+
+import ddf.catalog.data.impl.types.AssociationsAttributes;
+import ddf.catalog.data.impl.types.ContactAttributes;
+import ddf.catalog.data.impl.types.MediaAttributes;
+import ddf.catalog.data.impl.types.DateTimeAttributes;
```
jlcsmith: are any of the datetime attributes being used?

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
```diff
@@ -15,8 +15,17 @@ package ddf.catalog.transformer.input.pdf
...
+import ddf.catalog.data.impl.types.LocationAttributes;
```
jlcsmith: same question here

catalog/transformer/catalog-transformer-pdf/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -12,9 +12,36 @@
...
 
-    <bean id="pdfTransformer" class="ddf.catalog.transformer.input.pdf.PdfInputTransformer"/>
+    <bean id="pdfMetacardType" class="ddf.catalog.data.impl.MetacardTypeImpl">
+        <argument value="pdf"/>
+        <argument>
+            <list>
+                <bean class="ddf.catalog.data.impl.types.AssociationsAttributes"/>
+                <bean class="ddf.catalog.data.impl.types.ContactAttributes"/>
+                <bean class="ddf.catalog.data.impl.types.MediaAttributes"/>
+                <bean class="ddf.catalog.data.impl.types.ValidationAttributes"/>
```
jlcsmith: @glenhein - looks like the Topic category is missing here and possible DateTime.

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
glenhein: removed extra imports

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
glenhein: removed extra imports

catalog/transformer/catalog-transformer-pdf/src/main/resources/OSGI-INF/blueprint/blueprint.xml
glenhein: Added Topic, but I don't see DateTime being used in the transformer.

jlcsmith: Ok.  I saw DateTime being imported somewhere so I was just double checking.

jlcsmith:
All comments addressed.  Built and tested, LGTM 👍

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
```diff
@@ -15,8 +15,11 @@ package ddf.catalog.transformer.input.pdf
 
 import ddf.catalog.content.operation.ContentMetadataExtractor
 import ddf.catalog.data.Metacard
+import ddf.catalog.data.MetacardType
 import ddf.catalog.data.impl.AttributeDescriptorImpl
 import ddf.catalog.data.impl.BasicTypes
+import ddf.catalog.data.impl.MetacardTypeImpl
+import ddf.catalog.data.impl.types.*
```
jrnorth: We shouldn't be using `.*` imports.

jrnorth:
Apart from my last comment, LGTM

catalog/transformer/catalog-transformer-pdf/src/test/groovy/ddf/catalog/transformer/input/pdf/PdfInputTransformerSpecTest.groovy
glenhein: ok, btw, that was done by the code formatter

coyotesqrl: It shouldn't have if you've installed our IJ code style. It defaults to 99 imports from a package before collapsing to `*`.

jlcsmith: perhaps because it is groovy? ¯\_(ツ)_/¯

coyotesqrl: AH. Yes. That would do it. We have to fix the groovy formatter. :/

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -81,9 +84,110 @@
...
+    /**
+     * @param metacardType       must be non-null
+     * @param usePdfTitleAsTitle must be non-null
+     */
+    public PdfInputTransformer(MetacardType metacardType, Boolean usePdfTitleAsTitle) {
+        this(metacardType,
+                usePdfTitleAsTitle,
+                PDDocument::load,
+                GEO_PDF_PARSER::getWktFromPDF,
+                PdfInputTransformer::generatePdfThumbnail);
```
rzwiefel: Just curious if you're using these method references for a particular reason besides unit testing? It definitely makes the code a little harder to follow when these are otherwise just normal methods calls. It is generally a code smell to have code changes purely for unit testing.

coyotesqrl: Thinking about it a bit more, I think I agree with @rzwiefel here. Especially considering that `transformWithExtractors` isn't being tested with a unit-test defined function but getting the default, I'm not sure what's being gained.

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -219,101 +328,48 @@ private Metacard transformPdf(String id, PDDocument pdfDocument, String contentI
...
-             * original implementation assumed that a PDF would always have at least one
-             * page. That's what I've implemented here, but I don't like make those
-             * kinds of assumptions :-( But I also don't want to change the original
-             * behavior without knowing how it will impact the system.
-             */
+    private void setDateIfNotNull(Calendar calendar, MetacardImpl metacard, String attributeName,
+            String description) {
+        if (calendar != null) {
+            metacard.setAttribute(attributeName, calendar.getTime());
+            LOGGER.debug("PDF {} date was: {}", description, DATE_FORMAT.format(calendar));
```
rzwiefel: Don't think this debug statement is needed

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: @coyotesqrl why are you calling out transformWithExtractors? The method ref inputStreamPDDocumentFunction is called from transformWithoutExtractors as well.

glenhein: @rzwiefel if I don't use method refs, then I would have to create methods that can be overridden by the unit test. Either way, i need some way to support unit tests.

coyotesqrl: @glenhein Ignore my comment. I see that it's been updated to use the function on line 258. I'm still not in love with the testing-specific code smell. It does make this class more complicated than it otherwise needs to be for no obvious reason without the context of the unit tests. It might be cleaner if you could extract the logic for the three replaceable operations to another class, inject a bean for that in your blueprint, and mock that class and its three methods in your unit test.

It removes the smell but does lead to some extra complexity. However, from the perspective of someone looking at your class (and this new bean) in a vacuum it won't look strange.

jlcsmith:
@glenhein a couple comments remaining including @coyotesqrl's about the test ctor and then this should be good to go.

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: I convert the method refs to proper injected beans

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
```diff
@@ -30,59 +28,92 @@
...
+    private MetacardType metacardType;
 
-    private static final FastDateFormat DATE_FORMAT = FastDateFormat.getInstance(
-            "yyyy-MM-dd'T'HH:mm:ssZZ");
+    private boolean usePdfTitleAsTitle;
 
-    private Map<ServiceReference, ContentMetadataExtractor> contentMetadataExtractors =
-            Collections.synchronizedMap(new TreeMap<>(new ServiceComparator()));
+    /**
+     * Note: direct use of this constructor is meant for unit testing only
```
coyotesqrl: This doc note is no longer true, right?

coyotesqrl:
While I personally would have skipped the interfaces and just created the beans directly that's purely a style issue. This looks _much_ better now.

Just fix the one javadoc note and this is 👍

catalog/transformer/catalog-transformer-pdf/src/main/java/ddf/catalog/transformer/input/pdf/PdfInputTransformer.java
glenhein: fixed

catalog/spatial/registry/registry-api-impl/pom.xml
```diff
@@ -180,6 +180,11 @@
             <artifactId>registry-federation-admin-service</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
jlcsmith: Glen - why was this added here?

PR #1169, opened 2016-08-20
DDF-2424: Updating registry entries fails when versioning is enabled

rzwiefel:
@clockard @djblue @tbatie @pklinef 

@coyotesqrl 
@kcwire

### DIFF
--- a/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
+++ b/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
@@ -40,6 +40,7 @@ import org.osgi.service.cm.ConfigurationAdmin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableSet;
 import com.thoughtworks.xstream.converters.Converter;
 
 import ddf.catalog.Constants;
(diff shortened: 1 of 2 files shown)


catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -197,8 +198,8 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
 
         //This is a registry store so only allow registry requests through
         if (!filterAdapter.adapt(request.getQuery(),
-                new TagsFilterDelegate(Collections.singleton(RegistryConstants.REGISTRY_TAG),
-                        true))) {
+                new TagsFilterDelegate(Sets.newHashSet(RegistryConstants.REGISTRY_TAG,
```
coyotesqrl: As this was immutable before - `Collections.singleton()` returns an immutable set - we should switch to `ImmutableSet` here.

coyotesqrl:
Other than my mutability comment, this LGTM.

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
rzwiefel: fixed

pklinef:
👍

kcwire:
👍

PR #1183, opened 2016-08-24
DDF-2282  Updated RefreshRegistryEntries remote registry queries to be multi-threaded

clockard:
#### What does this PR do?
Multithreads the registry subscription polling.
@tbatie  @mcalcote 
@figliold
@lessarderic
Setup a couple of ddfs and make sure the subscription between them still works
DDF-2282
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

 - Also switched to an internal scheduler for polling instead of a camel route

### DIFF
--- a/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
+++ b/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
@@ -19,6 +19,12 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -46,7 +52,6 @@ import ddf.catalog.operation.Query;
 import ddf.catalog.operation.SourceResponse;
 import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
-import ddf.catalog.source.UnsupportedQueryException;
 
 /**
  * Though refreshRegistrySubscriptions in this class is a public method, it should never be called by any
@@ -54,11 +59,12 @@ import ddf.catalog.source.UnsupportedQueryException;
  * route and should avoid elevating privileges for any other service or being exposed to any
  * other endpoint.
  */
-@SuppressWarnings("PackageAccessibility")
 public class RefreshRegistryEntries {
     private static final Logger LOGGER = LoggerFactory.getLogger(RefreshRegistryEntries.class);
 
     private static final int PAGE_SIZE = 1000;
+
+    private static final int SHUTDOWN_TIMEOUT_SECONDS = 60;
 
     private List<RegistryStore> registryStores;
 
@@ -68,8 +74,26 @@ public class RefreshRegistryEntries {
 
     private boolean enableDelete = true;
 
+    private ScheduledExecutorService executor;
+
+    private int taskWaitTimeSeconds = 30;
+
+    private int refreshIntervalSeconds = 30;
+
+    private Future scheduledTask;
+
     public RefreshRegistryEntries() {
 
+    }
+
+    public void init() {
+        scheduledTask = executor.scheduleWithFixedDelay(() -> {
+            try {
+                refreshRegistryEntries();
+            } catch (FederationAdminException e) {
+                LOGGER.error("Problem refreshing registry entries.", e);
+            }
+        }, 10, refreshIntervalSeconds, TimeUnit.SECONDS);
     }
 
     /**
@@ -80,6 +104,10 @@ public class RefreshRegistryEntries {
      * @throws FederationAdminException
      */
     public void refreshRegistryEntries() throws FederationAdminException {
+
+        if (!registriesAvailable()) {
+            return;
+        }
 
         RemoteRegistryResults remoteResults = getRemoteRegistryMetacardsMap();
         Map<String, Metacard> remoteRegistryMetacardsMap =
@@ -148,6 +176,13 @@ public class RefreshRegistryEntries {
                 .isPresent();
     }
 
+    private boolean registriesAvailable() {
+        return registryStores.stream()
+                .filter(RegistryStore::isAvailable)
+                .findFirst()
+                .isPresent();
+    }
+
     private Map<String, List<Metacard>> getMetacardRegistryIdMap(Collection<Metacard> metacards) {
         Map<String, List<Metacard>> map = new HashMap<>();
         for (Metacard mcard : metacards) {
@@ -179,38 +214,68 @@ public class RefreshRegistryEntries {
         Map<String, Metacard> remoteRegistryMetacards = new HashMap<>();
         List<String> failedQueries = new ArrayList<>();
         List<String> storesQueried = new ArrayList<>();
-        List<String> localMetacardRegIds;
+        List<String> localMetacardRegIds = getLocalRegistryIds();
+
+        //Create the remote query task to be run.
+        List<Callable<RemoteResult>> tasks = new ArrayList<>();
+        for (RegistryStore store : registryStores) {
+            if (!store.isPullAllowed() || !store.isAvailable()) {
+                continue;
+            }
+            storesQueried.add(store.getRegistryId());
+            tasks.add(() -> {
+                SourceResponse response =
+                        store.query(new QueryRequestImpl(getBasicRegistryQuery()));
+                Map<String, Metacard> results = response.getResults()
+                        .stream()
+                        .map(Result::getMetacard)
+                        .filter(e -> !localMetacardRegIds.contains(RegistryUtility.getRegistryId(e)))
+                        .collect(Collectors.toMap(Metacard::getId, Function.identity()));
+                return new RemoteResult(store.getRegistryId(), results);
+
+            });
+        }
+
+        failedQueries.addAll(storesQueried);
+        List<RemoteResult> results = executeTasks(tasks);
+        results.stream()
+                .forEach(result -> {
+                    failedQueries.remove(result.getRegistryId());
+                    remoteRegistryMetacards.putAll(result.getRemoteRegistryMetacards());
+                });
+
+        return new RemoteRegistryResults(remoteRegistryMetacards, failedQueries, storesQueried);
+    }
+
+    private List<RemoteResult> executeTasks(List<Callable<RemoteResult>> tasks) {
+        List<RemoteResult> results = new ArrayList<>();
+        try {
+            List<Future<RemoteResult>> futures = executor.invokeAll(tasks);
+            for (Future<RemoteResult> future : futures) {
+                try {
+                    results.add(future.get(taskWaitTimeSeconds, TimeUnit.SECONDS));
+                } catch (ExecutionException e) {
+                    LOGGER.debug("Error executing query on a remote registry.", e);
+                } catch (TimeoutException e) {
+                    LOGGER.debug("Timeout occurred when querying a remote registry");
+                }
+            }
+        } catch (InterruptedException e) {
+            LOGGER.debug("Remote registry queries interrupted", e);
+        }
+        return results;
+    }
+
+    private List<String> getLocalRegistryIds() throws FederationAdminException {
         try {
             List<Metacard> localMetacards =
                     Security.runAsAdminWithException(() -> federationAdminService.getLocalRegistryMetacards());
-            localMetacardRegIds = localMetacards.stream()
+            return localMetacards.stream()
                     .map(e -> RegistryUtility.getRegistryId(e))
                     .collect(Collectors.toList());
         } catch (Exception e) {
             throw new FederationAdminException("Error querying for local metacards ", e);
         }
-        SourceResponse response;
-        for (RegistryStore store : registryStores) {
-            if (!store.isPullAllowed() || !store.isAvailable()) {
-                continue;
-            }
-
-            try {
-                storesQueried.add(store.getRegistryId());
-                response = store.query(new QueryRequestImpl(getBasicRegistryQuery()));
-                remoteRegistryMetacards.putAll(response.getResults()
-                        .stream()
-                        .map(Result::getMetacard)
-                        .filter(e -> !localMetacardRegIds.contains(RegistryUtility.getRegistryId(e)))
-                        .collect(Collectors.toMap(Metacard::getId, Function.identity())));
-            } catch (UnsupportedQueryException e) {
-                LOGGER.debug("Unable to contact {} for registry subscription query",
-                        store.getId(),
-                        e);
-                failedQueries.add(store.getRegistryId());
-            }
-        }
-        return new RemoteRegistryResults(remoteRegistryMetacards, failedQueries, storesQueried);
     }
 
     private void writeRemoteUpdates(List<Metacard> remoteMetacardsToUpdate)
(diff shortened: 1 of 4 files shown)


catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -68,10 +77,31 @@
...
     public RefreshRegistryEntries() {
 
     }
 
+    public void init() {
+        scheduledTask = executor.scheduleWithFixedDelay(() -> {
+            try {
+                refreshRegistryEntries();
+            } catch (FederationAdminException e) {
+                LOGGER.debug("Problem refreshing registry entries.", e);
```
mcalcote: How frequently might we expect this? I'm not suggesting putting this any higher, unless you feel it "might be important, such as failing sources or an endpoint failing for an internal reason".

clockard: It shouldn't happen often. Really only when we have an internal query fail for some reason. I put it at a debug level since it gets called periodically (30 seconds by default)

lessarderic: If this keeps happening because of a network problem or something similar, how will the admin know that the registry isn't being updated anymore?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -189,27 +230,44 @@ private RemoteRegistryResults getRemoteRegistryMetacardsMap() throws FederationA
         } catch (Exception e) {
             throw new FederationAdminException("Error querying for local metacards ", e);
         }
-        SourceResponse response;
+        List<Callable<RemoteResult>> tasks = new ArrayList<>();
```
lessarderic: Could we extract the 2 try/catch blocks and the for loop to separate private methods, or maybe encapsulate some of this logic into its own separate class? This method is fairly long and not all that easy to understand at first glance.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -176,8 +217,8 @@ private boolean hasMatch(Metacard local, List<Metacard> remoteMetacards) {
     }
 
     private RemoteRegistryResults getRemoteRegistryMetacardsMap() throws FederationAdminException {
-        Map<String, Metacard> remoteRegistryMetacards = new HashMap<>();
-        List<String> failedQueries = new ArrayList<>();
+        Map<String, Metacard> remoteRegistryMetacards = new ConcurrentHashMap<>();
```
lessarderic: Not sure why this has to be a `ConcurrentHashMap`.

lessarderic: Did we make this and `failedQueries` synchronized because they are returned in the `RemoteRegistryResults` object? As far as I can tell, those are only updated _after_ all the tasks have been run, so they should only be updated by a single thread, and do not seem to be updated outside of this method. If we expect them to be only modified here and not after they are set in `RemoteRegistryResults`, should we make them immutable instead of synchronized?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -292,13 +350,67 @@ public void setEnableDelete(boolean enableDelete) {
...
+    public void setExecutor(ScheduledExecutorService executor) {
+        this.executor = executor;
+    }
+
+    public void setRefreshInterval(int refreshInterval) {
+        if (this.refreshInterval == refreshInterval) {
+            return;
+        }
+        this.refreshInterval = refreshInterval;
+        if(initialized) {
```
lessarderic: Formatting.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -68,10 +77,31 @@
 
     private boolean enableDelete = true;
 
+    private boolean initialized = false;
+
+    private ScheduledExecutorService executor;
+
+    private int taskWaitTimeSeconds = 30;
+
+    private int refreshInterval = 30;
```
lessarderic: Should we also put the unit in the name?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -68,10 +77,31 @@
 
     private boolean enableDelete = true;
 
+    private boolean initialized = false;
```
lessarderic: Instead of using an `initialized` flag, couldn't we just check to see whether `scheduledTask` is `null`?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -68,10 +77,31 @@
...
+
+    private ScheduledExecutorService executor;
+
+    private int taskWaitTimeSeconds = 30;
```
lessarderic: Since`taskWaitTimeSeconds`, `refreshInterval` (and `initialized` and/or `initialized`) are accessed by multiple threads, should we make them `volatile`?

catalog/spatial/registry/registry-federation-admin-service-impl/src/test/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntriesTest.java
```diff
@@ -263,14 +278,38 @@ public void testNoUpdatesOnOlder() throws Exception {
...
         refreshRegistryEntries.refreshRegistryEntries();
 
         verify(federationAdminService, never()).updateRegistryEntry(any(Metacard.class));
     }
 
+    @Test
+    public void testDestroy() throws Exception {
+        when(executorService.awaitTermination(anyLong(), any(TimeUnit.class))).thenReturn(true);
+        refreshRegistryEntries.destroy();
+        verify(executorService, times(1)).awaitTermination(anyLong(), any(TimeUnit.class));
```
lessarderic: We should verify that the termination timeout is set to 60 seconds in at least one test.

catalog/spatial/registry/registry-federation-admin-service-impl/src/test/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntriesTest.java
```diff
@@ -263,14 +278,38 @@ public void testNoUpdatesOnOlder() throws Exception {
...
+
+    @Test
+    public void testDestroyInterupt() throws Exception {
+        when(executorService.awaitTermination(anyLong(),
+                any(TimeUnit.class))).thenThrow(new InterruptedException("interrupt"));
+        refreshRegistryEntries.destroy();
+        verify(executorService, times(1)).awaitTermination(anyLong(), any(TimeUnit.class));
+        verify(executorService, times(1)).shutdownNow();
+    }
+
```
lessarderic: Do we have tests to cover the new paths added to `getRemoteRegistryMetacardsMap`? For instance, should we make sure that store query exceptions are properly handled and don't stop the registry update (`RefreshRegistryEntries`, L255-L265), or that we properly handle a mixture of adds/updates/deletes?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -68,10 +77,31 @@
...
     }
 
+    public void init() {
+        scheduledTask = executor.scheduleWithFixedDelay(() -> {
+            try {
+                refreshRegistryEntries();
+            } catch (FederationAdminException e) {
+                LOGGER.debug("Problem refreshing registry entries.", e);
+            }
+        }, 10, refreshInterval, TimeUnit.SECONDS);
```
lessarderic: What will happen if it takes more than the refresh interval to update the registry entries?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -25,16 +28,20 @@
...
         <cm:managed-properties persistent-id="org.codice.ddf.registry.federationadmin.service.impl.RefreshRegistryEntries"
-                               update-strategy="container-managed" />
+                               update-strategy="container-managed"/>
         <property name="federationAdminService" ref="federationAdminService"/>
         <property name="filterBuilder" ref="filterBuilder"/>
         <property name="registryStores" ref="registryStore"/>
+        <property name="executor" ref="threadPool"/>
     </bean>
 
     <bean id="executor" class="java.util.concurrent.Executors" factory-method="newSingleThreadScheduledExecutor"/>
```
lessarderic: Can we give better names to the `executor` and `threadPool` beans? It's difficult to tell which one is used for what, especially since `executor` is also used as a property name on L38.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: Not sure why I didn't do it that way to start. Changed

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: Those variables are only used/accessed in the same one thread. Is there somewhere specific that you see the use of these variables in another threads context?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: Added seconds to the name to be consistent.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: Went ahead and changed this to be an error log message

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: Nothing. According to the javadocs once the task has finished (however long it takes) it will then wait the refreshInterval amount of time before running the task again.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: The concurrent and synchronized collections were a leftover from how i was trying to do it initially.  Will make the RemoteRegistryResults collections immutable.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: I was able to extract a couple of parts into private methods. Let me know if you think it is clearer this way.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: Fixed

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
clockard: Renamed

catalog/spatial/registry/registry-federation-admin-service-impl/src/test/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntriesTest.java
clockard: The timeout is a constant but I will update this test to check for it.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
lessarderic: `taskWaitTimeSeconds` is used in `refreshRegistryEntries`, which will run in the Camel thread if I'm not mistaken, and also in `setTimeWaitSeconds`, which is called by the Configuration Admin thread.

I should have taken a closer look the first time around. Looks like `refreshInterval` and `scheduledTask` are only used by a single thread, so those are ok the way they are.

catalog/spatial/registry/registry-federation-admin-service-impl/src/test/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntriesTest.java
clockard: Added a multi store test where one fails. Also added a test that does a create, update and delete in one refresh call.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
```diff
@@ -292,19 +358,73 @@ public void setEnableDelete(boolean enableDelete) {
...
-        List<String> failureList;
+        private List<String> failureList;
 
         public RemoteRegistryResults(Map<String, Metacard> remoteRegistryMetacards,
                 List<String> failureList, List<String> storesQueried) {
-            this.remoteRegistryMetacards = remoteRegistryMetacards;
-            this.failureList = failureList;
-            this.registryStoresQueried = storesQueried;
+            this.remoteRegistryMetacards = Collections.unmodifiableMap(remoteRegistryMetacards);
+            this.failureList = Collections.unmodifiableList(failureList);
```
tbatie: Would it be better to return a new map and new list instead of an unmodifiableMap and unmodifiableList in the getter methods?

tbatie:
:+1:

lessarderic:
:shipit: !

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RefreshRegistryEntries.java
clockard: I think probably so. I will update it.

mcalcote:
Manually confirmed subscription between two DDFs still works as per the testing instructions.

PR #1192, opened 2016-08-28
DDF-2423: Updating new attribute descriptors

rzwiefel:
#### What does this PR do?
Updating attribute descriptors that were not what we wanted them to be.  There will be a separate PR to update `DynamicSchemaResolver` and the `MetacardType` to attempt to prevent against duplicates coming in externally. 
@bdeining @brendan-hofmann 

@jlcsmith
@pklinef

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
@@ -36,13 +36,13 @@ public class CoreAttributes implements Core, MetacardType {
         descriptors.add(new AttributeDescriptorImpl(CHECKSUM,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(CHECKSUM_ALGORITHM,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(CREATED,
@@ -55,7 +55,7 @@ public class CoreAttributes implements Core, MetacardType {
                 true /* indexed */,
                 true /* stored */,
                 true /* tokenized */,
-                true /* multivalued */,
+                false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(EXPIRATION,
                 true /* indexed */,
@@ -66,7 +66,7 @@ public class CoreAttributes implements Core, MetacardType {
         descriptors.add(new AttributeDescriptorImpl(ID,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(LANGUAGE,
@@ -107,14 +107,14 @@ public class CoreAttributes implements Core, MetacardType {
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_DOWNLOAD_URL,
                 true /* indexed */,
-                false /* stored */,
+                true /* stored */,
                 true /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_SIZE,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_URI,
@@ -130,7 +130,7 @@ public class CoreAttributes implements Core, MetacardType {
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(THUMBNAIL,
-                true /* indexed */,
+                false /* indexed */,
                 true /* stored */,
                 false /* tokenized */,
                 false /* multivalued */,
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/AssociationsAttributes.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/AssociationsAttributes.java
@@ -36,13 +36,13 @@ public class AssociationsAttributes implements Associations, MetacardType {
         descriptors.add(new AttributeDescriptorImpl(DERIVED,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 true /* multivalued */,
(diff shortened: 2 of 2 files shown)


jlcsmith:
LGTM 👍

bdeining:
:+1:

peterhuffer:
👍

brendan-hofmann:
It looks like we don't even check `AttributeDescriptor.isTokenized()` anywhere in the Solr `CatalogProvider` classes. As far as I can tell, this means that many of these changes have no effect now, but could have unintended consequences in the future. Do we really not want to tokenize things like `Description` or the validation attributes? I'm pretty sure that we do, or at least they are used that way right now.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -84,7 +84,7 @@
         descriptors.add(new AttributeDescriptorImpl(METADATA,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
```
pklinef: I don't agree with this change.  Metadata should and is currently tokenized.  Searches would have to match the entire XML blob to match if tokenization is false.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -96,31 +96,31 @@
...
         descriptors.add(new AttributeDescriptorImpl(DERIVED_RESOURCE_URI,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_DOWNLOAD_URL,
-                true /* indexed */,
+                false /* indexed */,
```
pklinef: Why is this attribute and the next have indexed set to false?  I think they should be true.  Otherwise no one will be able to search on these attributes.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -96,31 +96,31 @@
...
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_URI,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
```
pklinef: I could see tokenization being useful for this attribute.  This will limit searches to exact url searches.  It would not be possible to search by hostname or filename for example if tokenized is false.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/ValidationAttributes.java
```diff
@@ -37,13 +37,13 @@
         descriptors.add(new AttributeDescriptorImpl(VALIDATION_ERRORS,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
```
pklinef: This will reduce the usefulness of these attributes to facet value searches since users would not know the full text of the validation errors and warnings without searching for valid values first.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
jlcsmith: @pklinef  - this change (and the others you mention) were done to align with the attribute descriptor definitions in BasicTypes.  I suppose we could update these definitions in BasicTypes.

brendan-hofmann: Shouldn't `BasicTypes` be deprecated in favor of `Core`/`CoreAttributes` anyways?

brendan-hofmann:
Based on @pklinef's additional comments, I think I disagree with the premise of this PR. 👎

rzwiefel:
We are going to fix them to make sense, and ensure further down the line duplicate attributes will not cause errors as they do currently. @brendan-hofmann, So look through the attributes and make sure (besides the ones phil commented on) that they make sense.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -54,8 +54,8 @@
         descriptors.add(new AttributeDescriptorImpl(DESCRIPTION,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
```
brendan-hofmann: I think description should be tokenized, otherwise it will only hit on exact matches.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -96,31 +96,31 @@
         descriptors.add(new AttributeDescriptorImpl(DERIVED_RESOURCE_DOWNLOAD_URL,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
```
brendan-hofmann: These derived URLs/URIs should be tokenized or deprecated. Indexing only the full URI/URL is not very useful from a query standpoint.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -96,31 +96,31 @@
...
-                true /* indexed */,
+                false /* indexed */,
                 false /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_SIZE,
-                true /* indexed */,
+                false /* indexed */,
```
brendan-hofmann: Javadoc says this is the size in bytes. If that is true, it should be a numeric type and indexed so that users can perform range queries.

brendan-hofmann:
Ok, in that case I left some additional comments.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -102,19 +102,19 @@
         descriptors.add(new AttributeDescriptorImpl(DERIVED_RESOURCE_URI,
                 true /* indexed */,
                 true /* stored */,
-                true /* tokenized */,
+                false /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.STRING_TYPE));
         descriptors.add(new AttributeDescriptorImpl(RESOURCE_DOWNLOAD_URL,
                 true /* indexed */,
                 false /* stored */,
```
brendan-hofmann: I don't know why this one isn't stored, but I think it should be?

brendan-hofmann:
The URLs/URIs and `VALIDATION_WARNINGS` should still be tokenized, I think.

jlcsmith:
Full build successful - Looks like all comments addressed 👍

PR #1193, opened 2016-08-29
DDF-2438 Fixed registry entry duplication issue in IdentificationPlugin

clockard:
#### What does this PR do?
Fixes a duplication issue where loopback registry configurations cause local nodes to be duplicated.
There had been duplication logic in the IdentificationPlugin before but was removed when we
added the internal registry entries. Added the logic back in a separate class and only apply it to registry entries that aren't internal.
@mcalcote @vinamartin 
@stustison 
@figliold
Install ddf and ddf-registry. Create a loopback remote registry. After a minute or so go to solr and verify there is only one registry metacard entry (metacard-tags=registry)
DDF-2438
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
+++ b/catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
@@ -1,0 +1,215 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p/>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p/>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.registry.identification;
+
+import java.security.PrivilegedActionException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import org.codice.ddf.registry.common.RegistryConstants;
+import org.codice.ddf.registry.common.metacard.RegistryObjectMetacardType;
+import org.codice.ddf.registry.common.metacard.RegistryUtility;
+import org.codice.ddf.security.common.Security;
+import org.opengis.filter.Filter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
+import ddf.catalog.filter.FilterBuilder;
+import ddf.catalog.operation.CreateResponse;
+import ddf.catalog.operation.DeleteResponse;
+import ddf.catalog.operation.QueryRequest;
+import ddf.catalog.operation.QueryResponse;
+import ddf.catalog.operation.UpdateResponse;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.plugin.PluginExecutionException;
+import ddf.catalog.plugin.PostIngestPlugin;
+import ddf.catalog.util.impl.Requests;
+
+/**
+ * The RegistryIdPostIngestPlugin collects a set of registry-ids that are currently in the system
+ * and makes them available to the IdentificationPlugin for duplication checking.
+ * <p>
+ * Initially the catalog is queried to initialize the set of registry-ids. After that the list is
+ * maintained by the process methods for create and delete responses.
+ */
+public class RegistryIdPostIngestPlugin implements PostIngestPlugin {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RegistryIdPostIngestPlugin.class);
+
+    private static final int RETRY_INTERVAL = 30;
+
+    private static final int SHUTDOWN_TIMEOUT_SECONDS = 60;
+
+    private static final int PAGE_SIZE = 1000;
+
+    private ScheduledExecutorService executorService;
+
+    private CatalogFramework catalogFramework;
+
+    private FilterBuilder filterBuilder;
+
+    private Set<String> registryIds = ConcurrentHashMap.newKeySet();
+
+    private Set<String> localRegistryIds = ConcurrentHashMap.newKeySet();
+
+    private Set<String> remoteMetacardIds = ConcurrentHashMap.newKeySet();
+
+    private Security security;
+
+    public RegistryIdPostIngestPlugin() {
+        security = Security.getInstance();
+    }
+
+    public RegistryIdPostIngestPlugin(Security security){
+        this.security = security;
+    }
+
+    public Set<String> getRegistryIds() {
+        return Collections.unmodifiableSet(registryIds);
+    }
+
+    public Set<String> getRemoteMetacardIds() {
+        return remoteMetacardIds;
+    }
+
+    public Set<String> getLocalRegistryIds() {
+        return localRegistryIds;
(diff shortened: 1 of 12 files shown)


catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+            QueryImpl query = new QueryImpl(registryFilter);
+            query.setPageSize(1000);
+            QueryRequest request = new QueryRequestImpl(query);
+
+            QueryResponse response = Security.runAsAdminWithException(() -> catalogFramework.query(
+                    request));
+
+            if (response == null) {
+                throw new PluginExecutionException(
+                        "Plugin failed to initialize plugin, unable to query identity node.");
```
mcalcote: This message could probably be a little more informative/helpful.

clockard: Is this better `"Failed to initialize plugin. Query for registry metacards came back null"`

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * The RegistryIdPostIngestPlugin collects a set of registry-ids that are currently in the system
+ * and makes them available to the IdentificationPlugin for duplication checking.
+ *
+ * Initially the catalog is queried to initialize the set of registry-ids. After that the list is
+ * maintained by the process methods for create and delete responses.
+ */
+public class RegistryIdPostIngestPlugin implements PostIngestPlugin {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(IdentificationPlugin.class);
```
figliold: Do you really want this to be RegistryIdPostIngestPlugin.class?

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    public void init() {
+        try {
+            List<Metacard> registryMetacards;
+            Filter registryFilter = filterBuilder.attribute(Metacard.TAGS)
+                    .is()
+                    .like()
+                    .text(RegistryConstants.REGISTRY_TAG);
+            QueryImpl query = new QueryImpl(registryFilter);
```
figliold: How did you come up with 1000?

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+            query.setPageSize(1000);
+            QueryRequest request = new QueryRequestImpl(query);
+
+            QueryResponse response = Security.runAsAdminWithException(() -> catalogFramework.query(
+                    request));
+
+            if (response == null) {
+                throw new PluginExecutionException(
```
figliold: Should this message be more specific and state the name of the plugin that failed to initialize?

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+                    .filter(e -> !StringUtils.isEmpty(e))
+                    .collect(Collectors.toList()));
+        } catch (PrivilegedActionException | PluginExecutionException e) {
+            LOGGER.debug("Error getting registry metacards. Will try again later");
+            executorService.schedule(this::init, RETRY_INTERVAL, TimeUnit.SECONDS);
+        }
+    }
+
+    public void destroy() {
+        executorService.shutdownNow();
```
figliold: Do you think we should follow the example in the javadocs for shutting down an executor service?
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
clockard: Nope, fixed

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
clockard: Updated to include plugin name

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
clockard: We have used that example for our destroy methods in several places. Guess i just forgot about it here. Updated with appropriate destroy logic for executor.

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
clockard: Used the same value we had in other places. 1000 is the maximum some sources can handle but that shouldn't matter here since we are dealing with the local catalog and locally created nodes. Basically we want all of them and we don't want to page through the results. In practice the most we have ever seen in a registry is ~30 and of those only a couple are locally created.  Let me know if you think we need to change this.

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+
+    private static final int RETRY_INTERVAL = 30;
+
+    private ScheduledExecutorService executorService;
+
+    private CatalogFramework catalogFramework;
+
+    private FilterBuilder filterBuilder;
+
+    private Set<String> registryIds = ConcurrentHashMap.newKeySet();
```
stustison: This isn't going to be persistent. What happens when you restart? Do you get duplicates again? If so, we need to consider other ways to store this.

clockard: That is what the init() method takes care of. It loads all the registry-ids we currently have into this set.

stustison:
👍

stustison:
Does this need back ported?

catalog/spatial/registry/registry-identification-plugin/src/main/java/org/codice/ddf/registry/identification/RegistryIdPostIngestPlugin.java
```diff
@@ -0,0 +1,161 @@
...
+        if (input != null && input.getDeletedMetacards() != null) {
+            input.getDeletedMetacards()
+                    .stream()
+                    .filter(RegistryUtility::isRegistryMetacard)
+                    .forEach(e -> registryIds.remove(RegistryUtility.getRegistryId(e)));
+        }
+        return input;
+    }
+
+    public void init() {
```
coyotesqrl: Do we execute catalog queries in the `init()` methods of any other plugins? While this should be okay, it strikes me funny.
1. Start catalog framework
2. Start this plugin, but as part of starting this plugin, execute a query against the catalog
3. Complete starting this plugin so it can be registered to run in the catalog framework

clockard: I don't know of any catalog plugins that do this although there are other bundles that query the catalog in their init method. I agree that this is a little funny though. Just not sure how else to accomplish this. This is an ingest plugin and we are performing queries so they don't directly overlap.

mcalcote:
Hero build passed :+1:

mcalcote:
Manual steps worked on the Hero build :+1: LGTM

PR #1205, opened 2016-08-30
DDF-2444 Added anyTags filter to include results with empty tags

emanns95:
#### What does this PR do?
Fixes itest failures caused by filters on Metacard.TAGS removing metacards without tags from the query.

@clockard 
@mcalcote 
@spearskw 
@rzwiefel 

@figliold
@lessarderic

Full build and tests

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
@@ -68,7 +68,7 @@ import ddf.catalog.util.impl.DescribableImpl;
 
 /**
  * Support class for resource delegate operations for the {@code CatalogFrameworkImpl}.
- *
+ * <p>
  * This class contains six delegated resource methods and methods to support them. No
  * operations/support methods should be added to this class except in support of CFI
  * resource operations.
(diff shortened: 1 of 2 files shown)


emanns95:
Not ready to merge yet

coyotesqrl:
Does the fix, when it is vetted and complete, need to be backported to 2.9?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -665,11 +663,24 @@ protected ResourceInfo getResourceInfo(ResourceRequest resourceRequest, String s
...
+
+        //no metacard tag
+        Filter nullTag = frameworkProperties.getFilterBuilder()
+                .attribute(Metacard.TAGS)
+                .is()
+                .empty();
+
+        //any or no metacard tag
+        anyTag = frameworkProperties.getFilterBuilder()
+                .anyOf(anyTag, nullTag);
```
rzwiefel: Just a bit, don't overwrite the variable. Makes it harder to follow. Declare a new var for final filter.

rzwiefel:
Also this is a pr for catalog framework, touching a specific persons code. You should be adding the person who's code it was originally as a reviewer. EG, me.  Also since it's catalog framework you should add the respective component owner as a committer (EG, Phil. Possibly Richard but not sure he feels comfortable with that burden yet)

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -665,11 +663,24 @@ protected ResourceInfo getResourceInfo(ResourceRequest resourceRequest, String s
     }
 
     private Query anyTag(Query query) {
+        //Any metacard tag
```
clockard: Need to do some kind of check to to see if this is a local query. If it is not then the tag filters shouldn't be applied.

emanns95:
Regarding backporting to 2.9, the change this is modifying is not on that branch, so I'd say no need.

emanns95:
@rzwiefel is going to be coming up with a different fix for this

PR #1207, opened 2016-08-30
DDF-2444: Fixes anytag queries from being federated in Resource Requests

rzwiefel:
@clockard @emanns95 

@lessarderic 
@pklinef

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
@@ -564,7 +564,7 @@ public class ResourceOperations extends DescribableImpl {
                     // if isEnterprise, go out and obtain the actual source
                     // where the product's metacard is stored.
                     QueryRequest queryRequest =
-                            new QueryRequestImpl(anyTag(propertyEqualToUriQuery),
+                            new QueryRequestImpl(anyTag(propertyEqualToUriQuery, site, isEnterprise),
                                     isEnterprise,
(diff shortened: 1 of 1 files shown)


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -664,7 +664,11 @@ protected ResourceInfo getResourceInfo(ResourceRequest resourceRequest, String s
         return new ResourceInfo(metacard, resourceUri);
     }
 
-    private Query anyTag(Query query) {
+    private Query anyTag(Query query, String site, boolean isEnterprise) {
+        if (isEnterprise || !getId().equals(site)) {
```
clockard: doesn't an enterprise search include the local catalog? Just thinking that in the enterprise case it would work for all sources except local.

emanns95:
Makes sense, glad it didn't require a bigger change :+1:

clockard:
This is a nice and simple solution. My only concern is with the enterprize query option.

rzwiefel:
If you're trying to do enterprise it just won't get tags. If local-only queries are the only one with tags it seems like an acceptable behavior imo. At least until we find a more solid solution with downstream endpoints.

rzwiefel:
@clockard Also for what we're using it for and what we're fixing, i don't think doing query deconstruction for enterprise (and the amount of work + risk that entails) justifies the change. Also the only use of this feature right now is history and this will suffice in the meanwhile.

clockard:
@rzwiefel sounds good 👍

PR #1215, opened 2016-08-31
DDF-2297 Federated queries sometimes fail to return results

dcruver:
#### What does this PR do?
This change fixes 2 bugs in the processing of federated search results.
First, it fixes a threading bug in SearchController that was blocking results
from being returned to the UI. Code had been added to handle exceptions, but
that code wasn't running in its own thread. As a result, the error-checking
prevented the method from returning.

Second, it fixes a minor bug in the case when the local solr results are
empty and those results weren't being merged into the set of cached
results. This was preventing the progress bar from being updated correctly.

@roelens8 
@bdeining 

@pklinef
@stustison

1) install the bundle provided by the bug submitter (available here: https://groups.google.com/forum/#!topic/ddf-users/F_kKYMsiEzg)
2) Perform some queries with various numbers in the search box (10, 20, 30, etc)
   - Verify that the progress bar doesn't hang up
   - Verify that you get back the expected number of results, especially with inputs above 30.

There is some confusion about the purpose of the timeout setting that the bug reporter mentions. There is no connection between that parameter and the described behavior.

DDF-2297

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/SearchController.java
+++ b/catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/SearchController.java
@@ -173,21 +173,23 @@ public class SearchController {
         }
 
         for (Map.Entry<String, Future> entry : futures.entrySet()) {
-            String sourceId = entry.getKey();
-            Future future = entry.getValue();
-            try {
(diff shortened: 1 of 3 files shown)


pklinef:
👍

roelens8:
👍

bdeining:
👍

stustison:
👍

PR #1228, opened 2016-09-02
DDF-2453: LogViewer Pause/Resume Polling

garrettfreibott:
#### What does this PR do?
This PR adds a button to the top left of the logviewer to allow admins to pause and resume polling of the backend. Admins often want to be able to view the current state of the logs when there is some sort of configuration issue, so this button was primarily added so that the logs could be paused after an issue to allow them to be closely inspected without the tailing feature of the viewer causing the logs to be pushed out of the viewer's context. This should also addresses any concerns with update scrolling behavior.

Also cleaned up the redux code quite a bit by using redux-thunk middleware and combineReducers() to make future additions easier.

#### Who is reviewing it?
@tbatie 
@djblue 
@andrewkfiedler 
@ricklarsen 
@codice/ui

@kcwire 
@lessarderic
@figliold 

Unit tests were added, so a build should be adequate. If wanted, manual testing is as simple as opening the logviewer and making sure that there is no "retrieveLogEvents" network activity when polling is paused. Network activity can be viewed live in the developer tools of most browsers.

This ticket originally wanted to address scrolling issues in the logviewer. After discussing the use-case for this situation, we decided that the best solution would be to add a pause/play button to allow the admin to pause log polling in order to prevent both the window from scrolling, as well as ensure that the finite logviewer queue does not eventually push the logs under inspection out of the finite queue.

Polling the backend:
image

Polling is paused:
image

Error occurred while polling, banner appears and polling is automatically paused:
image


michaelmenousek:
thanks for the screenshot.  it doesn't make sense to me.  Polling is an implementation detail.  I recommend looking for examples for convention to use.

garrettfreibott:
@michaelmenousek The log viewer tickets were created because admins specifically wanted a way to stop the network traffic itself when not needed, and we wanted to make it as obvious as possible that pressing this button is one way to do that.

I've done research into convention and it most just have a pause/play button with no text. I'm concerned this will make the button even more ambiguous as to whether it's pausing just the logviewer or pausing DDF's logging (a VERY common concern when the viewer was first released, and I've been trying to make that more obvious ever since). I've also considered "TAILING" and "LIVE".

I'm open to suggestions if you have any.

michaelmenousek:
what about (pause) and (return to live)?

andrewkfiedler:
Perhaps `disconnect` and `connect`

garrettfreibott:
@michaelmenousek @andrewkfiedler 
Here are what the suggestions look like. I removed the play/stop icons - since the text now describes the action and not the state, it would be redundant.

- pause / return to live
image
image

- connect / disconnect
image
image

- a live toggle button was also suggested at one point
image
image

garrettfreibott:
Seeing them side-by-side, i kind of like the live button option, so I made the design a bit more flat to match the rest of the ui and changed it to say offline. Thoughts?
image
image

michaelmenousek:
i'm not a fan of "offline" or "disconnect" because they make it sound bad.   Prefer "paused" but I'll defer to @andrewkfiedler

andrewkfiedler:
Users should be aware that clicking this button could make them miss important logs, so I don't think it matters if the language makes it sound bad.  It is bad.  

I'll think about it a bit more, but it would be nice to hear from a few other voices on what they prefer. 

As long as the functionality works, I'm fine with iterating on the language later.

vinamartin:
I like `Live`/`Paused`. For some reason the `Disconnected` or `Offline` make me think of it as a network problem vs. something intentional, but that's just me.

andrewkfiedler:
I agree with the network problem worries.

Lambeaux:
+1 for `Live / Paused`

peterhuffer:
I'll throw my hat in for a Stop and Play symbols. Similar to the icons used for starting/stopping a feature. I think this would be consistent with behavior of other parts of the application, and we can skip the confusion of words. Make the play button have a green background and the stop button have a red background.
screen shot 2016-09-06 at 1 20 06 pm
screen shot 2016-09-06 at 1 20 02 pm

pklinef:
What do other log viewers do?

Felix web console requires you to manually push a button to refresh when you want to see new log messages.  Solr appears to refresh every minute but appends to the end of the page so the scroll does not move.  Hawt.io polls every 5 seconds and also appends to the end of the page to avoid messages moving.  Kibana can sort with newest at the top but allow the user to control the refresh rate and one of the options is to not refresh.

garrettfreibott:
Before starting this ticket, I talked with @kcwire, the writer of these log viewer tickets. We came the the conclusion that the best solution for the use-case is to have a button to pause/resume polling for admins wanting to view the current state of the logs without worrying about the logs under examination being bumped off of the finite queue. This also gives the admins a simple way to stop network traffic caused by the logviewer for any reason if needed.

As for the word on the button, there seem to be about 10 different opinions. I'm going to go with "LIVE/PAUSED" with the "RED/GREY" color scheme as it seems to be the only option with more than a single person backing it.

pklinef:
Examples from Kibana.  Auto-refresh is disabled by default.

Auto-refresh currently off.
auto-refresh off

Auto-refresh on and playing.
auto-refresh on

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
```diff
@@ -13,33 +13,73 @@
...
+export const togglePolling = () => ({
+  type: 'TOGGLE_POLLING'
+})
+
+// retrieves new logs from the backend and appends any new ones
+export const fetch = (getLogs) => (dispatch, getState) => {
+  if (getState().isPolling && !getState().isFetching) {
+    dispatch(beginFetching())
+    getLogs((err, logs) => {
+      if (err) {
```
pklinef: We should notify if there are errors instead of silently hiding them.

distribution/docs/src/main/resources/_contents/running-contents.adoc
```diff
@@ -1384,6 +1384,8 @@ Monitor system logs with the *LogViewer*, a convenient "viewing portal" for inco
 
 The LogViewer displays the most recent 500 log messages by default, but will grow to a maximum of 5000 messages.
 
+The top left button labeled POLLING/PAUSED will toggle live retrieval of incoming logs from the backend.
```
lessarderic: Remember to update when we all agree on the button labels.

Also, should we explain the side-effects of pausing the logs, i.e., logs may be lost.

distribution/docs/src/main/resources/_contents/running-contents.adoc
```diff
@@ -1384,6 +1384,8 @@ Monitor system logs with the *LogViewer*, a convenient "viewing portal" for inco
...
+
 Log events can be filtered by:
 
 * Log level (`ERROR`, `WARNING`, etc).
```
lessarderic: I know this isn't part of this change, but do you think we should clarify the log level filtering a little bit better, i.e., how the Karaf's logging level will affect the logs retrieved by the log viewer and how the log viewer filtering will only impact the logs being displayed?

distribution/docs/src/main/resources/_contents/running-contents.adoc
lessarderic: Ignore my comment about the side-effects, just saw it's already described below.

distribution/docs/src/main/resources/_contents/running-contents.adoc
lessarderic: Just saw the `Note` below that explains this. Do you think it'd be better to just put the explanation here and get rid of the note?

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
```diff
@@ -13,33 +13,78 @@
...
+})
+
+// toggle polling to the backend for new logs
+export const togglePolling = () => ({
+  type: 'TOGGLE_POLLING'
+})
+
+// retrieves new logs from the backend and appends any new ones
+export const fetch = (getLogs) => (dispatch, getState) => {
+  console.log("1. check fetch: " + getState().isFetching)
```
lessarderic: Are those logs for debugging purposes? If so, should we remove them?

garrettfreibott: oh whoops, ya. i was debugging the fetching logic this morning. I'll remove those.

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
```diff
@@ -13,33 +13,78 @@
...
+  if (getState().isPolling && !getState().isFetching) {
+    dispatch(beginFetching())
+    console.log("2. after setting fetch: " + getState().isFetching)
+    console.log("3. Making the call...")
+    getLogs((err, logs) => {
+      console.log("4. Finished the call...")
+      if (err) {
```
lessarderic: Not exactly sure what to do here but it feels weird to just ignore the error...

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
```diff
@@ -13,33 +13,78 @@
...
-  }
+export const getUnique = (oldLogs, newLogs) => {
+  return newLogs.map((entry) => ({hash: MD5(entry), ...entry}))
+    .filter((a) => oldLogs.find((b) => a.hash === b.hash) === undefined)
+}
+
+// timeout loop for polling the backend
+export const fetchLoop = (getLogs, interval) => (dispatch, getState) => {
+  dispatch(fetch(getLogs))
+  setTimeout(() => dispatch(fetchLoop(getLogs)), interval || 5000)
```
lessarderic: Should we have a `const` for the `5000`?

platform/admin/core/admin-core-logviewer/src/main/webapp/js/components/polling-button/polling-button.less
```diff
@@ -0,0 +1,35 @@
+.button-base {
+  width: 100%;
+  height: 30px;
+  color: #fff;
+  font-size: 16px;
+  font-family: Helvetica;
+  background-color: #18bc9c;
+  border-color: #18bc9c;
+  //float: right;
```
lessarderic: We should remove the commented out lines (here and L12).

platform/admin/core/admin-core-logviewer/src/main/webapp/js/filter.js
```diff
@@ -18,32 +18,41 @@ import includes from 'lodash/includes'
 import lvls from './levels'
 const levels = lvls()
 
-const arrayToObject = function (key, array) {
-  var o = {}
-  array.forEach(function (element) {
+const arrayToObject = (key, array) => {
```
lessarderic: Not clear what this function does. Can we either add a comment or give it a better name?

platform/admin/core/admin-core-logviewer/src/test/js/actions.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  const getState = () => ({
+    isPolling: true,
+    isFetching: false,
+    logs: oldLogs
+  })
+
+  fetch(getLogs)(dispatch, getState)
+})
```
lessarderic: Do we test that we truncate the list of logs when we reach the maximum?

lessarderic:
👍  when all comments have been addressed and Bamboo build is green.

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
garrettfreibott: Added a drop-down notification - will add screenshots soon.

distribution/docs/src/main/resources/_contents/running-contents.adoc
garrettfreibott: Updated the button label names.

distribution/docs/src/main/resources/_contents/running-contents.adoc
garrettfreibott: I moved the note above the warning so it immediately comes after the log level filtering blurb. I think it's best to keep it inside the note tag to draw extra attention to it since it's a point of concern that has been brought up a couple times before.

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
garrettfreibott: Fixed.

platform/admin/core/admin-core-logviewer/src/main/webapp/js/components/polling-button/polling-button.less
garrettfreibott: Good catch, fixed.

platform/admin/core/admin-core-logviewer/src/main/webapp/js/filter.js
garrettfreibott: I added a comment.

platform/admin/core/admin-core-logviewer/src/test/js/actions.js
garrettfreibott: Nope, but we definitely should - just created that test inside test/js/reducer.js

platform/admin/core/admin-core-logviewer/src/main/webapp/js/actions/index.js
garrettfreibott: Added a drop-down banner to display the error and auto-pauses the polling to prevent more errors. I'll add screenshots in a moment.

garrettfreibott:
Screenshots have been updated with the finished button and error banner design.

kcwire:
👍

PR #1253, opened 2016-09-10
DDF-2474 Fixes issue with ingesting a large zip file with more metacards than space in blocking queue.

coyotesqrl:
#### What does this PR do?
Fixes bug where some records might not be ingested with no error logged if the zip file to be ingested contained more than `--batchsize` * `--multithreaded` metacards.

This also cleans up the threading logic, relying on a Phaser used as a CountUp/DownLatch instead of flags and Thread.sleeps().

This needs to be backported to 2.9.x but might require a slightly more complex cherrypick as some other changes have been introduced to master and not backported.

@bdeining @rzwiefel @tbatie 

@kcwire
@pklinef

- Full build/test run.
- Threading tests
    - Get some metacards to be ingested by any transformer and put them into a directory. I would recommend at least 100.
    - Ingest through the karaf command line (include the `-p` flag to speed the tests, passing various values for the `-b` (`--batchsize`) and `-m` (`--multithreaded`) flags to test different possible edge cases. e.g.
        - `-b 1 -m 1` 
        - `-b 1 -m 20` 
        - `-b [number of metacards in test]`
        - `-b [number of metacards in test] -m 1`
        - `-b [number of metacards in test] -m 3`
        - `-b [factor of metacard count]` (For 100, using 10 or 20 would be good for this)
        - `-b [factor of metacard count - 1]`
        - `-b [factor of metacard count + 1]`
        - `-b [metacard count/2]`
        - `-b [metacard count/2 - 1]`
        - `-b [metacard count/2 + 1]`
        - `-b [metacard count - 1]`
        - `-b [metacard count + 1]`
        - `-b [prime number less than metacard count]`
    - Run each of the above tests several times to ensure there are no timing issues. There is no need to remove the records from the catalog between tests. If the threading is going to fail, it's going to do so whether the records are already present of not.
- Zip ingest test
    - Create zip file of metacards to ingest, either manually or through `catalog:dump`
    - Sign the zip file. This can be accomplished by running `jarsigner -keystore $DDF_HOME/etc/keystores/serverKeystore.jks [zipfile] localhost`, assuming your DDF instance is named `localhost`.
    - Ingest the zip file, first with default batch and thread size, then by adjusting those parameters so their product is smaller than the number of metacards in the zip file. e.g.
        - `ingest --include-content [zipfile]`
        - `ingest --include-content -b 2 -m 2 [zipfile]` (If you have more than four metacards in your zip, these values would trigger the error prior to this fix.)

N/A

DDF-2474

N/A

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
@@ -25,6 +25,7 @@ import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
@@ -340,16 +341,19 @@ public class IngestCommand extends CatalogCommands {
 
     private int totalFileCount(File inputFile) throws IOException {
         if (inputFile.isDirectory()) {
-            int cnt = 0;
+            int currentFileCount = 0;
             try (DirectoryStream<Path> stream = Files.newDirectoryStream(inputFile.toPath())) {
                 for (Path entry : stream) {
-                    cnt++;
+                    if (!entry.toFile()
+                            .isHidden()) {
+                        currentFileCount++;
+                    }
                 }
-                return cnt;
+                return currentFileCount;
             }
         }
 
-        return 1;
+        return inputFile.isHidden() ? 0 : 1;
     }
 
     /**
@@ -508,8 +512,8 @@ public class IngestCommand extends CatalogCommands {
     }
 
     private void moveToFailedIngestDirectory(File source) {
-        File destination = new File(
-                failedIngestDirectory.getAbsolutePath() + File.separator + source.getName());
+        File destination = Paths.get(failedIngestDirectory.getAbsolutePath(), source.getName())
+                .toFile();
 
         if (!source.renameTo(destination)) {
             printErrorMessage("Unable to move source file [" + source.getAbsolutePath() + "] to ["
@@ -565,16 +569,20 @@ public class IngestCommand extends CatalogCommands {
         }
 
         if (result != null) {
-            try {
-                phaser.register();
-                metacardQueue.put(result);
-            } catch (InterruptedException e) {
-                phaser.arriveAndDeregister();
+            putMetacardOnQueue(metacardQueue, result);
+        }
+    }
 
-                INGEST_LOGGER.error("Thread interrupted while waiting to 'put' metacard: {}",
-                        result.getId(),
-                        e);
-            }
+    private void putMetacardOnQueue(ArrayBlockingQueue<Metacard> metacardQueue, Metacard metacard) {
+        try {
+            phaser.register();
+            metacardQueue.put(metacard);
+        } catch (InterruptedException e) {
+            phaser.arriveAndDeregister();
+
+            INGEST_LOGGER.error("Thread interrupted while waiting to 'put' metacard: {}",
+                    metacard.getId(),
+                    e);
         }
     }
 
@@ -601,13 +609,10 @@ public class IngestCommand extends CatalogCommands {
                     fileCount.set(metacardList.size());
 
                     for (Metacard metacard : metacardList) {
-                        phaser.register();
-                        metacardQueue.put(metacard);
+                        putMetacardOnQueue(metacardQueue, metacard);
                     }
                 }
-            } catch (IOException | CatalogTransformerException | InterruptedException e) {
-                // Clear the queue, to make this an all-or-nothing transaction
-                metacardQueue.clear();
+            } catch (IOException | CatalogTransformerException e) {
                 LOGGER.info("Unable to transform zip file into metacard list.", e);
                 INGEST_LOGGER.warn("Unable to transform zip file into metacard list.", e);
             }
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
@@ -19,6 +19,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
+import java.io.UncheckedIOException;
+import java.nio.file.DirectoryStream;
 import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
@@ -33,12 +35,13 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Phaser;
 import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.apache.commons.io.FilenameUtils;
@@ -113,13 +116,11 @@ public class IngestCommand extends CatalogCommands {
             .appendSuffix(" second", " seconds")
             .toFormatter();
 
+    private final Phaser phaser = new Phaser();
+
     private final AtomicInteger ingestCount = new AtomicInteger();
 
     private final AtomicInteger ignoreCount = new AtomicInteger();
-
-    private final AtomicBoolean doneBuildingQueue = new AtomicBoolean();
-
-    private final AtomicInteger processingThreads = new AtomicInteger();
 
     private final AtomicInteger fileCount = new AtomicInteger(Integer.MAX_VALUE);
 
@@ -128,8 +129,6 @@ public class IngestCommand extends CatalogCommands {
     private File failedIngestDirectory = null;
 
     private InputTransformer transformer = null;
-
-    private InputCollectionTransformer zipDecompression;
 
     @Argument(name = "File path or Directory path", description =
             "File path to a record or a directory of files to be ingested. Paths are absolute and must be in quotes."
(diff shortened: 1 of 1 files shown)


kcwire:
👍

bdeining:
LGTM

rzwiefel:
👍

coyotesqrl:
@stustison asked for some metrics on any performance ramifications for the threading changes. I ran ingests using three different data sets against current master and this branch. The differences in performance were negligible:

Dataset Size | Import Transformer | Master Time to Ingest | Updated Threading
-------------| ------------------ | --------------------- | -----------------
21578        | xml                | 1:02                  | 1:16
34475        | xml                | 5:04                  | 4:53
1000000      | geojson            | 1:23:05               | 1:20:04

stustison:
👍  Thanks!

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -465,96 +491,108 @@ private boolean processBatch(CatalogFacade catalog, ArrayList<Metacard> metacard
...
+                moveToFailedIngestDirectory(file);
+            }
+            printErrorMessage(String.format("Failed to ingest file [%s].", file.getAbsolutePath()));
+            INGEST_LOGGER.warn("Failed to ingest file [{}].", file.getAbsolutePath());
+        }
+
+        if (result != null) {
+            try {
+                metacardQueue.put(result);
+                phaser.register();
```
pklinef: I think you need to move this register before adding to the queue.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -465,96 +491,108 @@ private boolean processBatch(CatalogFacade catalog, ArrayList<Metacard> metacard
...
+                    metacardFileMapping = generateFileMap(new File(inputFile.getParent(),
+                            CONTENT_PATH));
+                    fileCount.set(metacardList.size());
+                    for (Metacard metacard : metacardList) {
+                        metacardQueue.put(metacard);
                     }
-                } catch (IOException | CatalogTransformerException e) {
-                    LOGGER.info("Unable to transform zip file into metacard list.", e);
-                    INGEST_LOGGER.warn("Unable to transform zip file into metacard list.", e);
+                    phaser.bulkRegister(metacardList.size());
```
pklinef: Same comment that register should be before adding to the queue.

pklinef:
I think this command will fail if the system cannot keep up with the queue.  The phaser only supports up to 65535 parties without tiering the phasers.

>Implementation notes: This implementation restricts the maximum number of parties to 65535. Attempts to register additional parties result in IllegalStateException. However, you can and should create tiered phasers to accommodate arbitrarily large sets of participants.
http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Phaser.html#bulkRegister(int)

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -171,6 +182,12 @@
 
     @Override
     protected Object executeWithSubject() throws Exception {
+        if (batchSize * multithreaded > MAX_QUEUE_SIZE) {
```
coyotesqrl: Made this change to protect from too large a queue of metacards and to avoid the complexity of managing tiered Phasers.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -549,9 +566,11 @@ private void addFileToQueue(ArrayBlockingQueue<Metacard> metacardQueue, long sta
 
         if (result != null) {
             try {
-                metacardQueue.put(result);
                 phaser.register();
+                metacardQueue.put(result);
```
coyotesqrl: Registering (counting up) first and then counting it back down should an exception be thrown.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -571,15 +590,20 @@ private void processIncludeContent(ArrayBlockingQueue<Metacard> metacardQueue) {
...
+                        .filter(Objects::nonNull)
+                        .collect(Collectors.toList());
+
                 if (metacardList.size() != 0) {
                     metacardFileMapping = generateFileMap(new File(inputFile.getParent(),
                             CONTENT_PATH));
                     fileCount.set(metacardList.size());
+
                     for (Metacard metacard : metacardList) {
+                        phaser.register();
```
coyotesqrl: Counting up first instead of doing the bulk register after.

coyotesqrl:
Comments added on the newer commit.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -314,17 +278,89 @@ protected Object executeWithSubject() throws Exception {
         return null;
     }
 
+    private File getInputFile() {
```
pklinef: Instead of returning `null`, did you consider changing this to return `Optional<File>`?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -314,17 +278,89 @@ protected Object executeWithSubject() throws Exception {
...
+                console.println(transformerId + " is an invalid input transformer.");
+                return null;
+            }
+        }
+        return inputFile;
+    }
+
+    private int totalFileCount(File inputFile) throws IOException {
+        if (inputFile.isDirectory()) {
+            int cnt = 0;
```
pklinef: I would recommend renaming to something more descriptive like `currentDirectoryFileCount`.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -314,17 +278,89 @@ protected Object executeWithSubject() throws Exception {
...
+            transformer = getTransformer();
+            if (transformer == null) {
+                console.println(transformerId + " is an invalid input transformer.");
+                return null;
+            }
+        }
+        return inputFile;
+    }
+
+    private int totalFileCount(File inputFile) throws IOException {
```
pklinef: Should this method return 0 if `inputFile` is hidden?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -314,17 +278,89 @@ protected Object executeWithSubject() throws Exception {
...
+        if (inputFile.isDirectory()) {
+            int cnt = 0;
+            try (DirectoryStream<Path> stream = Files.newDirectoryStream(inputFile.toPath())) {
+                for (Path entry : stream) {
+                    cnt++;
```
pklinef: Should we also not count paths associated with files that are hidden?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -465,96 +508,115 @@ private boolean processBatch(CatalogFacade catalog, ArrayList<Metacard> metacard
     }
 
     private void moveToFailedIngestDirectory(File source) {
-        File destination = new File(failedIngestDirectory.getAbsolutePath() + File.separator + source.getName());
+        File destination = new File(
```
pklinef: I am a fan of `Paths.get()`.

`Paths.get(failedIngestDirectory.getAbsolutePath(), source.getName()).toFile()`

https://docs.oracle.com/javase/7/docs/api/java/nio/file/Paths.html#get(java.lang.String,%20java.lang.String...)

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -168,71 +178,23 @@
     @Option(name = "--include-content", required = false, aliases = {}, multiValued = false, description = "Ingest a zip file that contains metacards and content using the default transformer.  The specified zip must be signed externally using DDF certificates.")
     boolean includeContent = false;
```
pklinef: I wish this boolean and option were `ingestZip`.  I keep wanting to confuse it with content storage which is unrelated to this.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -465,96 +508,115 @@ private boolean processBatch(CatalogFacade catalog, ArrayList<Metacard> metacard
...
-                    INGEST_LOGGER.warn("Unable to transform zip file into metacard list.", e);
                 }
-            } else {
-                LOGGER.info(
-                        "No Zip Transformer found.  Unable to transform zip file into metacard list.");
-                INGEST_LOGGER.warn(
-                        "No Zip Transformer found.  Unable to transform zip file into metacard list.");
+            } catch (IOException | CatalogTransformerException | InterruptedException e) {
+                // Clear the queue, to make this an all-or-nothing transaction
+                metacardQueue.clear();
```
pklinef: Don't we need to arriveAndDeregister based on the size of metacardQueue?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
coyotesqrl: As it's invoked in only one place locally and immediately tested, I'm not sure there's a lot of difference either way. It just requires the creation of Optionals with nulls in them when the input file doesn't exist or should be skipped.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
coyotesqrl: Agreed.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
coyotesqrl: Yeah. I hadn't done a lot of cleanup within the methods that were already there. Should have taken the extra time.

coyotesqrl:
Changes pushed in a separate commit. It might be unnecessary with the updated github review to keep it separate but just in case it isn't, I wanted this to be quick to look at.

bdeining:
Full Hero build successful.  Ran through all suggested test procedures and observed expected ingest behavior throughout.  LGTM

PR #1254, opened 2016-09-12
DDF-2266 Fix registry store cleanup bugs

clockard:
#### What does this PR do?
Fixes a bug with the registry store cleanup handler where the store reference can be removed before the cleanup logic can be run. 
Also adds a simple retry to publication updates
@mcalcote @vinamartin @tbatie 
@coyotesqrl
@figliold 
Bring up two instances of ddf with the registry running. Have one of the registries subscribe to the other. Verify that you get the other nodes information. Now remove the remote registry you just created. Verify that the node you received from the other registry has been removed.
DDF-2266
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
+++ b/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
@@ -142,7 +142,15 @@ public class RegistryPublicationServiceImpl implements RegistryPublicationServic
                 .collect(Collectors.toCollection(HashSet::new));
 
         if (CollectionUtils.isNotEmpty(locations)) {
-            federationAdminService.updateRegistryEntry(metacard, locations);
+            try {
+                federationAdminService.updateRegistryEntry(metacard, locations);
+            } catch (FederationAdminException e) {
+                //This should not happen often but could occur if the remote registry removed the metacard
+                //that was to be updated. In that case performing an add will fix the problem. If the failure
+                //was for another reason like the site couldn't be contacted then the add will fail
+                //also and the end result will be the same.
+                federationAdminService.addRegistryEntry(metacard, locations);
+            }
             metacard.setAttribute(new AttributeImpl(RegistryObjectMetacardType.LAST_PUBLISHED,
                     Date.from(ZonedDateTime.now()
                             .toInstant())));
--- a/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreCleanupHandler.java
+++ b/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreCleanupHandler.java
@@ -67,12 +67,6 @@ public class RegistryStoreCleanupHandler implements EventHandler {
         }
     }
 
-    public void unbindRegistryStore(ServiceReference serviceReference) {
-        if (serviceReference != null) {
-            registryStorePidToServiceMap.remove(serviceReference.getProperty(Constants.SERVICE_PID));
-        }
-    }
-
     @Override
     public void handleEvent(Event event) {
         Object eventProperty = event.getProperty(EventConstants.EVENT);
@@ -95,6 +89,8 @@ public class RegistryStoreCleanupHandler implements EventHandler {
         if (service == null) {
             return;
         }
+        registryStorePidToServiceMap.remove(servicePid);
+
         executor.execute(() -> {
             String registryId = service.getRegistryId();
             try {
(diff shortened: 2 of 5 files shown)


catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
```diff
@@ -142,7 +142,11 @@ public void update(Metacard metacard) throws FederationAdminException {
                 .collect(Collectors.toCollection(HashSet::new));
 
         if (CollectionUtils.isNotEmpty(locations)) {
-            federationAdminService.updateRegistryEntry(metacard, locations);
+            try {
+                federationAdminService.updateRegistryEntry(metacard, locations);
+            } catch (FederationAdminException e){
```
coyotesqrl: Formatting is off here; missing a space.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
```diff
@@ -142,7 +142,11 @@ public void update(Metacard metacard) throws FederationAdminException {
...
+                federationAdminService.addRegistryEntry(metacard, locations);
```
coyotesqrl: The only reason an exception would occur above is if underlying records weren't present to be updated? No other cause where that exception might be thrown and a call to `addRegistryEntry` would be in error?

And, if it is the case that a failed updated should automatically lead to an add, is there a way to manage that process flow without using exception processing? Using exceptions for flow control is a code smell.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
clockard: Fixed

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
```diff
@@ -142,7 +142,11 @@ public void update(Metacard metacard) throws FederationAdminException {
                 .collect(Collectors.toCollection(HashSet::new));
 
         if (CollectionUtils.isNotEmpty(locations)) {
-            federationAdminService.updateRegistryEntry(metacard, locations);
+            try {
+                federationAdminService.updateRegistryEntry(metacard, locations);
+            } catch (FederationAdminException e) {
+                federationAdminService.addRegistryEntry(metacard, locations);
```
coyotesqrl: _I'm adding this comment again so it doesn't get lost by being against an older commit._

The only reason an exception would occur above is if underlying records weren't present to be updated? No other cause where that exception might be thrown and a call to addRegistryEntry would be in error?

And, if it is the case that a failed updated should automatically lead to an add, is there a way to manage that process flow without using exception processing? Using exceptions for flow control is a code smell.

clockard: It is not the only reason that an exception could be thrown but is the most likely reason. addRegistryEntry shouldn't be a problem in any case though. This was just the simplest way to do this for an exceptional case. If we wanted to avoid doing this we would have to perform a remote query prior to every update to check for an 'exceptional' case. I can change it if you feel that we should be checking first.

coyotesqrl: The update would only be throwing an exception after a remote call too, right? If the odds are vanishingly small then I'd be okay leaving it this way (with a comment explaining the flow and the low likelihood of a failure); otherwise, it would be better to switch to an explicit call to check.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/RegistryPublicationServiceImpl.java
```diff
@@ -142,7 +142,15 @@ public void update(Metacard metacard) throws FederationAdminException {
                 .collect(Collectors.toCollection(HashSet::new));
 
         if (CollectionUtils.isNotEmpty(locations)) {
-            federationAdminService.updateRegistryEntry(metacard, locations);
+            try {
+                federationAdminService.updateRegistryEntry(metacard, locations);
+            } catch (FederationAdminException e) {
+                //This should not happen often but could occur if the remote registry removed the metacard
```
coyotesqrl: This comment is sufficient for me. The fact it's a rare condition makes this less bothersome code.

coyotesqrl:
👍

tbatie:
:+1:

brianfelix:
Manually tested. LGTM :+1:

PR #1224, opened 2016-09-01
DDF-2398 added revised Endpoint docs to documentation.

ricklarsen:
#### What does this PR do?

Initial draft of Endpoint section. Subtask of DDF-2309
#### Who is reviewing it?

@brendan-hofmann @ahoffer @peterhuffer @mweser 
#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).

@pklinef
@shaundmorris
#### How should this be tested?

docs render correctly, build succeeds, review outline
#### Any background context you want to provide?

This section contains some new endpoints that were previously not documented. Some sections are incomplete, and some may need to be removed altogether. 
#### What are the relevant tickets?

https://codice.atlassian.net/browse/DDF-2309
https://codice.atlassian.net/browse/DDF-2398
#### Screenshots (if appropriate)

[draft-documentation.pdf](https://github.com/codice/ddf/files/450685/draft-documentation.pdf)
#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #1260, opened 2016-09-12
DDF-2468 Add support for more FTP commands

glenhein:
#### What does this PR do?

STOU - Full support is added. The FTP server will return a success message that
includes the metacard id, which the client can interpret as the unique name given
to the object.

SITE - Since the FTP server does not have any site-specific commands, the server
now returns the status code 202, which indicates that SITE is not implemented.

DELE, RETR, RMD, MKD, APPE, RNTO - The FTP server returns a response indicating
success, even though no action was taken. Some clients, for example, always
attempt to make a new directory and then CD to it. This isn't necessary from
the DDF point of view. Previously, the server would return an error that would
cause the client to fail.

Add support for clients that PUT a dot-file and then rename the dot-file to
the final filename. The rename command will result in the object being stored
in the catalog.

Add itest for clients that the create a directory, change to that directory, and
then PUT content.

The CWD always returns a success response to support clients that use a
MKDIR/CWD/PUT command sequence.

@peterhuffer 
@beyelerb 
@troymohl 
@ricklarsen 

@beyelerb
@kcwire
Build and install DDF. Enable the catalog-ftp feature. From the command line, start an ftp client:

$ ftp admin@localhost 8021
ftp> binary
ftp> sunique on
ftp> put some-file.pdf

From the catalog search, confirm that the file was uploaded.

For the dot-file/rename feature:

$ ftp admin@localhost 8021
ftp> binary
ftp> put some-file.pdf .foo
ftp> rename .foo some-file.pdf

From the catalog search, confirm that the file was uploaded.

DDF-2468
#### Checklist:
- [X] Documentation Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/ftp/pom.xml
+++ b/catalog/ftp/pom.xml
@@ -127,22 +127,22 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.70</minimum>
+                                            <minimum>0.67</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.55</minimum>
+                                            <minimum>0.52</minimum>
                                         </limit>
                                         <limit>
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.43</minimum>
+                                            <minimum>0.45</minimum>
                                         </limit>
                                         <limit>
                                             <counter>LINE</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.73</minimum>
+                                            <minimum>0.70</minimum>
                                         </limit>
                                     </limits>
                                 </rule>
@@ -164,7 +164,8 @@
                             catalog-core-api-impl,
                             shiro-core,
                             commons-beanutils,
-                            commons-lang3
+                            commons-lang3,
+                            mina-core
                         </Embed-Dependency>
                         <Export-Package/>
                     </instructions>
--- a/catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
+++ b/catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
@@ -17,9 +17,14 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetAddress;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.ftpserver.ftplet.DataConnection;
 import org.apache.ftpserver.ftplet.DataConnectionFactory;
 import org.apache.ftpserver.ftplet.DefaultFtpReply;
@@ -31,10 +36,9 @@ import org.apache.ftpserver.ftplet.FtpRequest;
 import org.apache.ftpserver.ftplet.FtpSession;
 import org.apache.ftpserver.ftplet.FtpletResult;
 import org.apache.ftpserver.impl.IODataConnectionFactory;
+import org.codice.ddf.platform.util.TemporaryFileBackedOutputStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.google.common.io.FileBackedOutputStream;
 
 import ddf.catalog.CatalogFramework;
 import ddf.catalog.content.data.ContentItem;
@@ -51,18 +55,30 @@ import ddf.mime.MimeTypeResolutionException;
 import ddf.security.Subject;
 
 /**
- * Handler for incoming FTP requests from the client. Supports the PUT and MPUT operations - all other operations respond with 550 request not taken response.
- * When PUTing a new file, it is not temporarily written to the file system. Instead, the new file's data stream is ingested directly into the catalog.
+ * Handler for incoming FTP requests from the client. Supports the PUT, MPUT,
+ * DELE, MKD, RETR, RMD, APPE, RNTO, STOU, and SITE operations. When PUTing a new
+ * file, it is not temporarily written to the file system. Instead, the new
+ * file's data stream is ingested directly into the catalog. If PUT submits
+ * a file that begins with a dot (eg. .foo), then the data is not immediately
+ * stored into the catalog. Instead, the system waits for a RNTO command for
+ * the dot-file and then stores the data in the catalog. Clients may PUT multiple
+ * dot-files and then execute multiple RNTO commands. Any dot-files that are not
+ * renamed at the end of the ftp session will be discarded and resources will
+ * be released. The CWD command always return a successful response. This is
+ * to support clients that use a MKDIR/CWD/PUT command sequence.
  */
 public class FtpRequestHandler extends DefaultFtplet {
     private static final Logger LOGGER = LoggerFactory.getLogger(FtpRequestHandler.class);
 
-    private static final String UNSUPPORTED_METHOD_RESPONSE =
-            "Unsupported method, only PUT and MPUT are supported.";
-
     private static final String SUBJECT = "subject";
 
     private static final String STOR_REQUEST = "STOR";
+
+    private static final String STOU_REQUEST = "STOU";
+
+    private static final String ATTR_TEMP_FILENAMES = "org.codice.ddf.temp-files";
+
+    private static final String CWD = "CWD";
 
     private CatalogFramework catalogFramework;
 
@@ -81,8 +97,12 @@ public class FtpRequestHandler extends DefaultFtplet {
         return FtpletResult.DEFAULT;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public FtpletResult onDisconnect(FtpSession session) throws FtpException, IOException {
+
+        removeAllTempFilesFromSession(session);
+
         return FtpletResult.DEFAULT;
     }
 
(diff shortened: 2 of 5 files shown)


glenhein:
Note: I still need to add new unit tests.

glenhein:
Note: I still need to add support for clients that upload to .filename and then rename

glenhein:
Still need to "Updating documentation of supported FTP commands"

glenhein:
FYI, I'm going to add another itest that covers another ticket, since it's all in the same code base anyways.

beyelerb:
The CD command returns an error - that will prohibit the sequence of MKDIR followed by CD to the newly made directory. Since we don't have the concept of a file system inside our server, this should probably return a successful response as well. The only mismatch then would be the PWD - which always returns the '/' path.

glenhein:
I'll look into it. The DDF code doesn't actually handle CD commands, it's handled in the FTPlet code. Also, Peter said it was working.

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
```diff
@@ -94,14 +114,73 @@ public FtpletResult onLogin(FtpSession session, FtpRequest request)
     }
 
     /**
-     * @param session The current {@link FtpSession}
-     * @param request The current {@link FtpRequest}
-     * @return {@link FtpletResult#SKIP} - signals successful ingest and to discontinue and further processing on the FTP request
-     * @throws FtpException general exception for Ftplets
-     * @throws IOException  thrown when there is an error fetching data from client
+     * Find the temp file map of TFBOS.
```
peterhuffer: Maybe we can make this an `@link` doc for the TFBOS.

distribution/docs/src/main/resources/_contents/_catalog-contents/integrating-catalog-contents.adoc
```diff
@@ -1135,7 +1135,11 @@ The configurable properties for the FTP Endpoint are accessed from the *FTP Endp
 
 ====== Using the endpoint
 
-The FTP endpoint supports the PUT and MPUT operations. Any other operation, such as DELETE, will return a 550 action not taken response.
+The FTP endpoint supports the PUT, MPUT DELE, RETR, RMD, APPE, RNTO, STOU, and SITE operations.
+
+The FTP endpoint supports files being upload as a dot-file (eg. .foo) and then being renamed to
```
ricklarsen: `e.g.,`

distribution/docs/src/main/resources/_contents/_catalog-contents/integrating-catalog-contents.adoc
```diff
@@ -1135,7 +1135,11 @@ The configurable properties for the FTP Endpoint are accessed from the *FTP Endp
...
+the final filename (eg. some-file.pdf). The endpoint will complete the ingestation process when
+the rename command is sent.
```
ricklarsen: I've added this content to https://github.com/codice/ddf/pull/1224 so it'll be in the new outline

distribution/docs/src/main/resources/_contents/_catalog-contents/integrating-catalog-contents.adoc
```diff
@@ -1135,7 +1135,11 @@ The configurable properties for the FTP Endpoint are accessed from the *FTP Endp
 
 ====== Using the endpoint
 
-The FTP endpoint supports the PUT and MPUT operations. Any other operation, such as DELETE, will return a 550 action not taken response.
+The FTP endpoint supports the PUT, MPUT DELE, RETR, RMD, APPE, RNTO, STOU, and SITE operations.
+
+The FTP endpoint supports files being upload as a dot-file (eg. .foo) and then being renamed to
+the final filename (eg. some-file.pdf). The endpoint will complete the ingestation process when
```
ricklarsen: ingest process?

catalog/ftp/pom.xml
```diff
@@ -164,7 +164,8 @@
                             catalog-core-api-impl,
```
peterhuffer: Are we able to increase the jacoco numbers in this pom now that we have tests covering all the overridden methods?

glenhein: i'll give it a try

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
```diff
@@ -362,4 +511,17 @@ private void notNull(Object object, String name) {
             throw new IllegalArgumentException(name + " cannot be null");
         }
     }
+
+    public FtpletResult beforeCommand(FtpSession session, FtpRequest request)
```
peterhuffer: `@Override` on this method.

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
```diff
@@ -151,65 +232,118 @@ public FtpletResult onUploadStart(FtpSession session, FtpRequest request)
...
 
-                if (createResponse != null) {
-                    List<Metacard> createdMetacards = createResponse.getCreatedMetacards();
+                session.write(new DefaultFtpReply(FtpReply.REPLY_226_CLOSING_DATA_CONNECTION,
+                        "Closing data connection"));
+                LOGGER.debug("Sending FTP status code 226 to client - closing data connection");
+
+            } catch (FtpException fe) {
+                LOGGER.debug("Failure to create metacard for file {}", fileName, fe);
+                throw new FtpException("Failure to create metacard for file " + fileName);
```
troymohl: Same comment as above about exceptions

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
```diff
@@ -151,65 +232,118 @@ public FtpletResult onUploadStart(FtpSession session, FtpRequest request)
...
-                final AutoCloseable ac = outputStream::reset) {
-            DataConnection dataConnection = connFactory.openConnection();
-            dataConnection.transferFromClient(session, outputStream);
+        if (isDotFile(request.getArgument())) {
+            DataConnection dataConnection;
+            try {
+                dataConnection = connFactory.openConnection();
+            } catch (Exception e) {
+                LOGGER.debug("Error getting the output data stream from the FTP session", e);
+                throw new IOException("Error getting the output stream from FTP session");
```
troymohl: Is there a reason we are changing the exception thrown instead of either re-throwing it, or just letting it pass through to the caller since it's already declared as a checked exception?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFtp.java
```diff
@@ -250,6 +249,67 @@ public void testFtpsPut() throws Exception {
...
+        String dotFilename = ".foo";
+        String finalFilename = "test.txt";
+
+        ftpPut(client, dotFilename, SAMPLE_DATA);
+        boolean ret = client.rename(dotFilename, finalFilename);
+        assertTrue(ret);
+
+        Response response = executeOpenSearch("xml", "q=*", "count=100");
+
+        response.then()
```
peterhuffer: We should wrap these in a `WaitCondition`s to avoid a race condition.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFtp.java
```diff
@@ -250,6 +249,67 @@ public void testFtpsPut() throws Exception {
...
+
+        boolean ret = client.makeDirectory(newDir);
+        assertTrue(ret);
+        ret = client.changeWorkingDirectory(newDir);
+        assertTrue(ret);
+        ftpPut(client, finalFilename, SAMPLE_DATA);
+
+        Response response = executeOpenSearch("xml", "q=*", "count=100");
+
+        response.then()
```
peterhuffer: Same here. We should wrap these in a `WaitCondition`s to avoid a race condition.

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
```diff
@@ -151,65 +232,118 @@ public FtpletResult onUploadStart(FtpSession session, FtpRequest request)
...
+            }
+
+            if (createResponse != null) {
+                if (LOGGER.isDebugEnabled()) {
+                    createResponse.getCreatedMetacards()
+                            .forEach(metacard -> LOGGER.debug("Content item created with id = {}",
+                                    metacard.getId()));
+                }
+            } else {
+                throw new FtpException();
```
troymohl: Seems like this would be more appropriate thrown above where the LOGGER.info is happening so it could wrap the `IngestException` or `SourceUnavailableException` with more details rather than just throwing an empty exception.

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: no idea, that was from the original code

https://github.com/codice/ddf/blob/master/catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java#L202

Maybe @peterhuffer knows

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: depends, can catalogFramework.create return null without throwing an exception?

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
peterhuffer: It looks like `openConnection` will throw an `FtpException` on an SSL error. I must've ignored this initially since we didn't support FTPS, and forgot about it during the FTPS ticket. It also throws the `IOException` when getting the data stream. In any case, this can be updated to catch each one separately now.

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: from the javadoc, catalogFramework.complete should only return non-null values, so I reworked this exception

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: Ultimately, the Exception types needs to be wrapped in an FtpException or IOException because of how the methods are declared by Ftplet. So, it's either wrap it here, or wrap it one stack frame higher. I will change it so the IOException contains the Exception so it doesn't get lost.

troymohl: @glenhein Sorry, I didn't notice the Exception vs IOException there. It probably makes more sense to catch IOException here and then rethrow that, and below catch generic "Exception" and then wrap that in an FtpException. 

In fact i think if you catch "Exception" before FtpException it will never hit the FtpException case as FtpException subclasses Exception.

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: I moved it up

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: changed

catalog/ftp/src/main/java/ddf/catalog/ftp/ftplets/FtpRequestHandler.java
glenhein: changed

beyelerb:
👍 lgtm - tested the mkdir/cd and that works now - should be good for all known requirements, if others come along later we can work them on a different ticket.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFtp.java
```diff
@@ -14,6 +14,7 @@
 package ddf.test.itests.catalog;
 
 import static org.codice.ddf.itests.common.WaitCondition.expect;
+import static org.hamcrest.Matchers.equalTo;
```
peterhuffer: Don't think this import is being used anywhere - failed check-style on this.

glenhein: fixed

peterhuffer:
Reminder to increase the `SET_CLIENT_AUTH_TIMEOUT_SEC` from 1 to 5 minutes.

glenhein:
@peterhuffer changed SET_CLIENT_AUTH_TIMEOUT_SEC to 5 minutes

PR #1261, opened 2016-09-12
DDF-2461 Syncs registry name with site name

vinamartin:
#### What does this PR do?
Previously, the default registry name would be `ddf.distribution` before being set by a user if the registry was selected at install. 
The name is now bound to the site name system property. 
The editable field is now disabled and I've grayed it out for clarity.

@djblue @clockard @brianfelix 

@figliold
@pklinef

Install a ddf and select the registry app to be installed during the installer process.
Change the site name to any other name using the installer. 
Verify that after restart, the name is the name chosen during the installer. Add additional local nodes and verify they can be renamed (the name is not grayed out). 

DDF-2461

screen shot 2016-09-12 at 12 08 26 pm

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Field.view.js
+++ b/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Field.view.js
@@ -82,9 +82,13 @@ define([
                     data = this.model.toJSON();
                     data.validationError = this.model.get('error');
                     data.errorIndices = this.model.errorIndices;
+                    if (this.model.get('identityNode') && this.model.get('key') === 'Name'){
+                        data.editable = false;
+                    }
                     if (this.readOnly) {
                         data.editable = false;
                     }
+
                 }
                 return data;
             },
--- a/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Segment.view.js
+++ b/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Segment.view.js
@@ -62,7 +62,8 @@ define([
 
                 var standardFields = [];
                 var advancedFields = [];
-                this.model.get('fields').forEach( function (field) {
+                this.model.get('fields').forEach(function (field) {
+                    field.set('identityNode', this.model.get('identityNode'));
                     if (!field.get('custom') && !field.get('advanced')) {
                         standardFields.push(field);
                     }
@@ -70,7 +71,7 @@ define([
                         advancedFields.push(field);
                         this.addvancedFields = true;
                     }
-                });
+                }.bind(this));
                 this.formFieldsView = new Field.FieldCollectionView({collection: new Backbone.Collection(standardFields), readOnly: this.readOnly});
                 this.formSegmentsView = new Segment.SegmentCollectionView({
                     model: this.model,
--- a/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
+++ b/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
@@ -14,7 +14,6 @@
 
 package org.codice.ddf.registry.federationadmin.service.impl;
 
-
 import static org.codice.ddf.registry.schemabindings.EbrimConstants.RIM_FACTORY;
 
 import java.io.IOException;
@@ -38,6 +37,7 @@ import org.codice.ddf.registry.federationadmin.service.internal.FederationAdminE
 import org.codice.ddf.registry.federationadmin.service.internal.FederationAdminService;
 import org.codice.ddf.registry.schemabindings.helper.InternationalStringTypeHelper;
 import org.codice.ddf.registry.schemabindings.helper.MetacardMarshaller;
+import org.codice.ddf.registry.schemabindings.helper.RegistryPackageTypeHelper;
 import org.codice.ddf.registry.schemabindings.helper.SlotTypeHelper;
 import org.codice.ddf.security.common.Security;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.CswConstants;
@@ -52,7 +52,6 @@ import oasis.names.tc.ebxml_regrep.xsd.rim._3.ExtrinsicObjectType;
 import oasis.names.tc.ebxml_regrep.xsd.rim._3.RegistryPackageType;
 import oasis.names.tc.ebxml_regrep.xsd.rim._3.SlotType1;
 import oasis.names.tc.ebxml_regrep.xsd.rim._3.VersionInfoType;
-
 
 /**
  * Creates a registry identity node when DDF Registry is first instantiated.
@@ -86,9 +85,17 @@ public class IdentityNodeInitialization {
     public void init() {
         try {
             Security.runAsAdminWithException(() -> {
-                Optional<Metacard> optional = federationAdminService.getLocalRegistryIdentityMetacard();
-                optional.ifPresent(e -> System.setProperty(RegistryConstants.REGISTRY_ID_PROPERTY,
-                        RegistryUtility.getRegistryId(e)));
+                Optional<Metacard> optional =
+                        federationAdminService.getLocalRegistryIdentityMetacard();
+                if (optional.isPresent()) {
+                    Metacard metacard = optional.get();
+                    System.setProperty(RegistryConstants.REGISTRY_ID_PROPERTY,
+                            RegistryUtility.getRegistryId(metacard));
+                    if (!metacard.getTitle()
+                            .equals(SystemInfo.getSiteName())) {
+                        updateIdentityNodeName(metacard);
+                    }
+                }
                 if (!optional.isPresent()) {
                     createIdentityNode();
                 }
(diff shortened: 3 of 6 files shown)


catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Field.view.js
```diff
@@ -82,9 +82,13 @@ define([
                     data = this.model.toJSON();
                     data.validationError = this.model.get('error');
                     data.errorIndices = this.model.errorIndices;
+                    if (this.model.get('identityNode') && this.model.attributes.key === 'Name'){
```
clockard: Should use `this.model.get('key')` instead of going to the attributes directly.

brianfelix:
After fixing Chris's comment, LGTM :+1:

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Segment.view.js
```diff
@@ -62,15 +62,16 @@ define([
...
+                    field.set('identityNode', this.model.get('identityNode'));
                     if (!field.get('custom') && !field.get('advanced')) {
                         standardFields.push(field);
                     }
                     if(field.get('advanced')){
                         advancedFields.push(field);
                         this.addvancedFields = true;
                     }
-                });
+                }.bind(this));
```
clockard: Why `bind(this)`?

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Field.view.js
vinamartin: Oops didn't know you could do that. :) Thanks, I'll fix this

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
```diff
@@ -86,9 +85,17 @@
     public void init() {
         try {
             Security.runAsAdminWithException(() -> {
-                Optional<Metacard> optional = federationAdminService.getLocalRegistryIdentityMetacard();
+                Optional<Metacard> optional =
+                        federationAdminService.getLocalRegistryIdentityMetacard();
                 optional.ifPresent(e -> System.setProperty(RegistryConstants.REGISTRY_ID_PROPERTY,
                         RegistryUtility.getRegistryId(e)));
+                if (optional.isPresent()) {
```
clockard: Seems like this section and the `optional.ifPresent` right above could be combined.

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Segment.view.js
vinamartin: Have to look inside this for the `identityNode` bool

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
vinamartin: Ah yeah I think so too. I'll fix that up

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
```diff
@@ -104,6 +111,30 @@ public void destroy() {
         executorService.shutdown();
     }
 
+    private void updateIdentityNodeName(Metacard metacard) throws FederationAdminException {
+        RegistryPackageType registryPackage = federationAdminService.getRegistryObjectByRegistryId(
```
clockard: Since we already have the metacard we can just just the MetacardMarshaller's getRegistryPackageFromMetacard(metacard) instead of making another catalog call.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
```diff
@@ -104,6 +111,30 @@ public void destroy() {
...
+                RegistryUtility.getRegistryId(metacard));
+
+        String siteName = SystemInfo.getSiteName();
+        RegistryPackageTypeHelper registryPackageTypeHelper = new RegistryPackageTypeHelper();
+
+        registryPackageTypeHelper.getExtrinsicObjects(registryPackage.getRegistryObjectList())
+                .stream()
+                .filter(extrinsicObjectType -> extrinsicObjectType.getObjectType()
+                        .equals(RegistryConstants.REGISTRY_NODE_OBJECT_TYPE))
+                .forEach(extrinsicObjectType -> extrinsicObjectType.setName(
```
clockard: You can swap out `forEach` with `first` since there should only be one matching extrinsic object.

vinamartin: True, will change

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
```diff
@@ -104,6 +111,30 @@ public void destroy() {
...
+                        internationalStringTypeHelper.create(siteName)));
+
+        Metacard identityMetacard = getRegistryMetacardFromRegistryPackage(registryPackage);
+        if (identityMetacard != null) {
+            identityMetacard.setAttribute(new AttributeImpl(RegistryObjectMetacardType.REGISTRY_IDENTITY_NODE,
```
clockard: Shouldn't need to set these attributes since they are already on the metacard stored in solr. The transient logic we have for updates should copy them over from the original.

vinamartin: Wasn't sure if I need them or not. Should I just get rid of this whole block then?

clockard:
Seems like we should add a unit test to verify the IdentityNodeInitialization is working correctly.

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
clockard: I think so, except for the update call :)

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/view/Segment.view.js
stustison: I might be mis-remembering, but I'm pretty sure you can also just do `forEach(function() {...}, this);`

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
vinamartin: Actually, could you provide an example, I'm not really sure I get how to make this conversion.

vinamartin:
@clockard Good call, just noticed I broke a unit test, so good reason to add one. :)

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitialization.java
clockard: ```.filter(.....).findFirst().ifPresent(extrinsicObjectType -> extrinsicObjectType.setName(internationalStringTypeHelper.create(siteName)))```

catalog/spatial/registry/registry-federation-admin-service-impl/src/test/java/org/codice/ddf/registry/federationadmin/service/impl/IdentityNodeInitializationTest.java
```diff
@@ -113,13 +122,38 @@ public void initWithNoPreviousEntry() throws Exception {
...
         when(federationAdminService.getLocalRegistryIdentityMetacard()).thenReturn(Optional.of(
                 testMetacard));
         identityNodeInitialization.init();
         verify(federationAdminService, never()).addRegistryEntry(any(Metacard.class));
     }
 
     @Test
+    public void initWithPreviousEntryWithNameChange() throws Exception {
+        System.setProperty(SystemInfo.SITE_NAME, CHANGED_TEST_SITE_NAME);
+
```
figliold: I think the path given to convert might not work on Windows.  See https://codice.atlassian.net/wiki/display/DDF/Windows+Compatibility#WindowsCompatibility-ResourceCodingPractices:HowtoAvoidFilePitfalls

vinamartin: That's a fair concern, I think there's actually a different way I can be doing this that doesn't actually require a path at all. I'm going to try and update it now.

vinamartin:
@figliold Took some advice from @clockard and got rid of the path entirely. Let me know if that fixes up your comment.

mcalcote:
Tested with manual steps, works as describe :+1: (Chrome & IE)

PR #1263, opened 2016-09-13
DDF-2477: Admin UI no longer requires setting the system port. That is derived from the HTTP or HTTPS port and the selected default protocol.

troymohl:
#### What does this PR do?
Updates the Admin UI so that the system.port is derived from the protocol selection and the port number entered for either HTTP or HTTPS port.

@lcrosenbu 
@peterhuffer 

@kcwire
@pklinef

Perform a build and install of DDF. Change the port and/or protocol in the Admin UI. Verify <install>/etc/system.properties has the correct value for property: org.codice.ddf.system.port

https://codice.atlassian.net/browse/DDF-2477

screen shot 2016-09-14 at 9 49 07 am

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/platform/admin/core/admin-core-api/src/main/java/org/codice/ddf/ui/admin/api/impl/SystemPropertiesAdmin.java
+++ b/platform/admin/core/admin-core-api/src/main/java/org/codice/ddf/ui/admin/api/impl/SystemPropertiesAdmin.java
@@ -42,6 +42,8 @@ import org.slf4j.LoggerFactory;
 
 public class SystemPropertiesAdmin implements SystemPropertiesAdminMBean {
 
+    public static final String HTTP_PROTOCOL = "http://";
+
     private static final String DEFAULT_LOCALHOST_DN = "localhost.local";
 
     private MBeanServer mbeanServer;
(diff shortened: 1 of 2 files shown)


platform/admin/core/admin-core-api/src/main/java/org/codice/ddf/ui/admin/api/impl/SystemPropertiesAdmin.java
```diff
@@ -288,6 +288,28 @@ private void updateProperty(String key, Map<String, String> updatedProperties,
         }
     }
 
+    private void updatePortProperty(Map<String, String> updatedProperties,
```
clockard: If you called this method after all the other system properties have been set you could simplify this method by retrieving the protocol and port numbers using the SystemBaseUrl helpers which will never return null.

troymohl: Updated the code to simplify it with your recommendation.

pklinef:
We should consider also removing the default protocol.  If a user sets it to HTTP, the default web context policy will cause the STS and Solr to fail.  Even if they fixed the web context policy, the STS would be wide open.

stustison:
Yeah, using HTTP against a SOAP service that requires an HTTPS token or one of our web apps would fail. So it comes down to whether or not we want to advertise HTTP in any scenario. I think the answer is no, but I've been hesitant throwing out the comment on this PR that @pklinef just made.

troymohl:
@pklinef @stustison : I have removed the protocol configuration option from the AdminUI. The default port and protocol still exist as system properties. The protocol is going to always be HTTPS (default) and the port will always use the HTTPS Port.

pklinef:
👍

lcrosenbu:
:+1:

peterhuffer:
👍

pklinef:
Bamboo failed due to OutOfMemory exception for build.

PR #1267, opened 2016-09-13
DDF-2455 SchematronValidationService incorrectly appears in sources

adimka:
#### What does this PR do?
Backport of DDF-2455

@tbatie @ryeats 

@coyotesqrl
@shaundmorris 

1. Make sure Schematron Validation Services doesn't appear as an source configuration option in the Catalog UI upon first starting the system.
2. Confirm the Admin Poller Service modal works on the System tab of the Admin UI
3. Confirm Schematron's the default settings appear in the modal (see screenshot).
4. Add a few patterns (*csw*, *Csw*, *Federated*, *federated*) and restart ddf to confirm that adding to the filter list removes sources.
5. Since the modal has been acting up, re-install ddf and confirm the modal appears on the first click without having to refresh the page.

DDF-2455

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/admin/admin-poller-service/pom.xml
+++ b/catalog/admin/admin-poller-service/pom.xml
@@ -57,6 +57,30 @@
             <groupId>ddf.action.core</groupId>
             <artifactId>action-core-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.spockframework</groupId>
+            <artifactId>spock-core</artifactId>
+            <version>1.0-groovy-2.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>cglib</groupId>
+            <artifactId>cglib-nodep</artifactId>
+            <version>3.2.0</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.objenesis</groupId>
+            <artifactId>objenesis</artifactId>
+            <version>2.1</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>2.4.7</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
     <build>
         <resources>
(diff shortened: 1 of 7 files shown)


coyotesqrl:
👍

catalog/admin/admin-poller-service/pom.xml
```diff
@@ -57,6 +57,11 @@
             <groupId>ddf.action.core</groupId>
             <artifactId>action-core-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>ddf.lib</groupId>
+            <artifactId>spock-shaded</artifactId>
+            <version>${ddf.version}</version>
```
shaundmorris: @adimka may need to look into the spock-shaded version here. I don't believe we had it on DDF 2.9.x. So you may have to actually put in the version here for the dependency to resolve correctly. I believe this is why the Bamboo build failed.

coyotesqrl: Ah, heck. You'd think I'd have noticed that. :(

Shaun is right. You'll need to add the full set of dependencies and the build plugin.

catalog/admin/admin-poller-service/pom.xml
```diff
@@ -59,6 +59,30 @@
             <groupId>ddf.action.core</groupId>
             <artifactId>action-core-api</artifactId>
         </dependency>
+        <dependency>
```
coyotesqrl: This didn't require the gmavenplus plugin? Is it coming out of a pom farther up in the hierarchy?

adimka: It appears to be in the root pom.

coyotesqrl: Well, yes and know. The 1.4 version of the gmavenplus plugin is in the root pom but it is only going to run if the `findbugs` profile is on. So during a full build it will be invoked and all will be good, but if someone were to want to run tests with findbugs turned off (which, honestly, they shouldn't but...) it would not be run.

Go ahead and add the 1.5 version to this module for now. We may consider doing a bit of Spock cleanup (maybe going so far as backporting the shaded jar) in 2.9.x after the next release but not now.

shaundmorris:
The bamboo build had an issue with the container coming up. I'm assuming it is un-related. Thoughts?

coyotesqrl:
I'm like 98% sure it is unrelated but to be sure we don't have a broken bundle have asked @adimka to run a full hero build on another server. He should have the results of that shortly and we'll be able to merge.

adimka:
The hero build was successful. :+1: Should be good to go.

shaundmorris:
👍  for some reason, I couldn't start a review on this one 😄

coyotesqrl:
Might have been because _someone_ was really quick to merge it. ;)

shaundmorris:
LOL, @coyotesqrl I think it was the case yesterday as well, I wanted to approve but didn't figure it out on this one or I just needed sleep.

PR #1268, opened 2016-09-14
DDF-2480 Fixed Updload dialog to reopen if closed during an upload

timothytierney:

@andrewkfiedler
@jrnorth
@azgoalie
@emanns95 
@ahoffer
@NGoss
@spearskw

@stustison 
@jlcsmith

Open the file upload dialog and select two large files to upload. While the uploads are in progress, close the dialog using the x. Once the dialog has been closed, it cannot be re-opened.

This was caused by a change made in DDF-1643

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
+++ b/catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
@@ -37,8 +37,6 @@ define([
             },
             initialize: function () {
                 // there is no automatic chaining of initialize.
-                Modal.prototype.initialize.apply(this, arguments);
-
                 this.model = new Backbone.Model({
                     state: 'start'
                 });


emanns95:
Confirmed works with IE10, Chrome, Firefox 👍

catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
```diff
@@ -37,8 +37,6 @@ define([
             },
             initialize: function () {
                 // there is no automatic chaining of initialize.
-                Modal.prototype.initialize.apply(this, arguments);
```
stustison: We can't do this here. Removing this line causes the view to not be cleaned up. But we create a new modal in the IngestMenu.js file, which then causes this view to become a zombie. I think we need to look at the `isUnfinished` function and figure out why that isn't keeping the original modal around.

timothytierney: The change introduced in DDF-1643 added in a view.destroy() in the modal.js initialize function  when the model is hidden, so it destroy's the view which caused this issue.

emanns95:
Another PR for this issue has been opened: https://github.com/codice/ddf/pull/1281

PR #1276, opened 2016-09-15
DDF-2361 - (2.9.x) Enable fanout and ddf registry to be used at the same time in the same ddf instance

gordocanchola:
#### What does this PR do?
This ticket adds the ability for fanout to write to its local store. A blacklist is added to give the admin control over what metacard tags should be rejected while in fanout mode.

@coyotesqrl 
@pklinef 
DDF-2361
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

…nout is enabled

Fanout is changing to allow access to the local store. The fanout requests will be rejected if fanout is enabled and the request contains a metacard with a tag on the blacklist.

CatalogFrameWorkImpl.java
  Adding a fanoutTagBlacklist that will be used to determine which catalog requests should be blocked.
  Changing the tests for fanout flag to now check the blacklist

blueprint.xml
  Updating catalogFrameworkImpl with metatype info and default values for fanoutProxyTagBlacklist and filterAdapter

metatype.xml
  Adding fanoutTagBlackList
  Adding fanoutProxyTagBlacklist

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CatalogFrameworkImpl.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CatalogFrameworkImpl.java
@@ -50,6 +50,7 @@ import javax.activation.MimeType;
 import javax.activation.MimeTypeParseException;
 import javax.xml.bind.DatatypeConverter;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
@@ -107,7 +108,9 @@ import ddf.catalog.data.impl.MetacardImpl;
 import ddf.catalog.data.impl.ResultImpl;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.federation.FederationStrategy;
+import ddf.catalog.filter.FilterAdapter;
 import ddf.catalog.filter.FilterDelegate;
+import ddf.catalog.filter.delegate.TagsFilterDelegate;
 import ddf.catalog.filter.impl.LiteralImpl;
 import ddf.catalog.filter.impl.PropertyIsEqualToLiteral;
 import ddf.catalog.filter.impl.PropertyNameImpl;
@@ -224,6 +227,12 @@ public class CatalogFrameworkImpl extends DescribableImpl implements CatalogFram
 
     private boolean fanoutEnabled = false;
 
+    private List<String> fanoutTagBlacklist = new ArrayList<>();
+
+    private List<String> fanoutProxyTagBlacklist;
+
+    private FilterAdapter filterAdapter;
+
     private StorageProvider storage;
 
     private FrameworkProperties frameworkProperties;
@@ -256,6 +265,18 @@ public class CatalogFrameworkImpl extends DescribableImpl implements CatalogFram
 
     public void setFanoutEnabled(boolean fanoutEnabled) {
         this.fanoutEnabled = fanoutEnabled;
+    }
+
+    public void setFanoutTagBlacklist(List<String> fanoutTagBlacklist) {
+        this.fanoutTagBlacklist = fanoutTagBlacklist;
+    }
+
+    public void setFanoutProxyTagBlacklist(List<String> fanoutProxyTagBlacklist) {
+        this.fanoutProxyTagBlacklist = fanoutProxyTagBlacklist;
+    }
+
+    public void setFilterAdapter(FilterAdapter filterAdapter) {
+        this.filterAdapter = filterAdapter;
     }
 
     public void setNotificationEnabled(boolean notificationEnabled) {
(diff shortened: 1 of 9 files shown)


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1008,13 +1009,15 @@ public void testGetMetacardResourceStatusCached()
...
+        String resourceCachedXpath = String.format("//*[@name='%s']/*/text()",
+                Metacard.RESOURCE_CACHE_STATUS);
 
         get(url + "?transform=resource").then();
 
         get(url).
-            then().
-                body(hasXPath(resourceCachedXpath, is("true")));
+                then()
+                .
```
vinamartin: Not sure if it's just how github is showing it, but this looks really weirdly formatted.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFanout.java
```diff
@@ -14,31 +14,48 @@
 package ddf.test.itests.catalog;
```
vinamartin: For all the tests you've updated here, the coverage numbers haven't changed at all?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CatalogFrameworkImpl.java
```diff
@@ -3447,4 +3451,61 @@ public URI getResourceUri() {
...
+    private boolean isMetacardBlacklisted(Metacard metacard) {
+        if (metacard == null) {
+            return true;
+        }
+
+        Set<String> tags = new HashSet<>(metacard.getTags());
+
+        // default to resource tag
+        if (tags.isEmpty()) {
+            tags.add("resource");
```
vinamartin: Should we be using a constant here?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFanout.java
```diff
@@ -14,31 +14,48 @@
...
 @RunWith(PaxExam.class)
 @ExamReactorStrategy(PerClass.class)
 public class TestFanout extends AbstractIntegrationTest {
 
     private static final String LOCAL_SOURCE_ID = "ddf.distribution";
 
+    // Using default resource tag as the one to blacklist
+    private static final List<String> TAG_BLACKLIST = Collections.singletonList("resource");
+
+    private static final TestCatalog TEST_CATALOG = new TestCatalog();
```
clockard: If we are going to be using this functionality across multiple test classes we should extract the common code into something catalog helper.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
gordocanchola: It was weirdly formatted. Fixed it.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFanout.java
gordocanchola: Numbers went up a tiny bit. Changed them

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/CatalogFrameworkImpl.java
gordocanchola: Change to use a constant

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFanout.java
gordocanchola: Pulled out the common code to a new CatalogIngest class in test-itests-common

coyotesqrl:
I can make small comments about code cleanup/refactoring in a few places, but I'm not clear what this is accomplishing or how/why it's doing it this way. We keep shoving things into `CatalogFramework`, polluting it with more and more complex conditional branches, and we really shouldn't be doing that unless absolutely no other solutions exist.

On first blush, it seems that you should be able to accomplish this all with plugins:
* `PreCreateStoragePlugin`
* `PreIngestPlugin`
* etc.

I'm slightly less concerned about this change than I am to its sibling which is going into master. This version of `CatalogFrameworkImpl` is already a huge mass of conditional code and customizations; however, the version in master is (while far from perfect) much less so. Additionally, if you were able to use plugins to solve this issue, then your change to master and to 2.9 would be nearly identical.

Perhaps I'm missing some salient point that makes it such that this logic must be directly implemented in the framework, but that would need to be clarified to me.

At worst, I would think a new plugin type could be created to isolate this logic and then we need merely add injection points to the framework for that plugin.

gordocanchola:
Hi @coyotesqrl 
These changes are being made so that a registry can be run in fanout mode. The existing fanout checks in CatalogFrameworkImpl would prevent any writes to the local store. The registry information is stored in the local store. The blacklist was added to give the framework a way to restrict writing specific metacard tags to the local store.
The plugin solution was discussed and ruled out. I don't remember the details as to why it was ruled out. @pklinef was involved in that decision and my have a better explanation.

coyotesqrl:
We really need to find a way to extract all the fanout logic out of here. However for 2.9.x, this is acceptable.

mcalcote:
Manually confirmed Hero Build/Test.

PR #1281, opened 2016-09-16
DDF-2480 Fixed upload dialog to reopen if closed during an upload

emanns95:
#### What does this PR do?
Fixes the Search UI file upload dialog so it can be reopened if its closed while resources are being uploaded.

@andrewkfiedler
@jrnorth
@azgoalie
@emanns95 
@ahoffer
@NGoss
@spearskw
@timothytierney

@shaundmorris 
@stustison

Open the file upload dialog and select two large files to upload. While the uploads are in progress, close the dialog using the x. Once the dialog has been closed, it cannot be re-opened.

This PR relates to: https://github.com/codice/ddf/pull/1268

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/search-ui/standard/src/main/webapp/js/view/Modal.js
+++ b/catalog/ui/search-ui/standard/src/main/webapp/js/view/Modal.js
@@ -30,10 +30,11 @@ define([
             initialize: function () {
                 // destroy view after animation completes.
                 // extending views must call: Modal.prototype.initialize.apply(this, arguments);
-                var view = this;
                 this.$el.one('hidden.bs.modal', function () {
-                    view.destroy();
-                });
+                    if (this.shouldDestroy()) {
+                        this.destroy();
+                    }
+                }.bind(this));
             },
             show: function () {
                 this.$el.modal({
@@ -43,6 +44,9 @@ define([
             },
             hide: function () {
                 this.$el.modal('hide');
+            },
+            shouldDestroy: function () {
+                return true;
             }
         });
         return BaseModal;
--- a/catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestMenu.js
+++ b/catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestMenu.js
@@ -44,10 +44,13 @@ define([
                 this.listenTo(wreqr.vent, 'upload:finish', this.onUploadFinish);
             },
             showModal: function() {
-                var keepCurrentModal = this.modal && this.modal.isUnfinished();
-                if (!keepCurrentModal) {
+                if (this.modal === undefined) {
+                    this.modal = new IngestModal();
+                } else if (this.modal.isFinished()) {
+                    this.modal.destroy();
                     this.modal = new IngestModal();
                 }
+
                 wreqr.vent.trigger('showModal', this.modal);
             },
             onUploadStart: function() {
--- a/catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
+++ b/catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
@@ -47,6 +47,15 @@ define([
                 this.listenTo(this.collection, "add", this.checkIfDialogComplete);
                 this.listenTo(this.collection, "remove", this.checkIfDialogComplete);
             },
+            shouldDestroy: function() {
+                if (this.model === undefined) {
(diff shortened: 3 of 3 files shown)


catalog/ui/search-ui/standard/src/main/webapp/js/view/Modal.js
```diff
@@ -43,6 +43,10 @@ define([
             },
             hide: function () {
                 this.$el.modal('hide');
+            },
+            remove: function () {
+                var view = this;
```
andrewkfiedler: `this.destroy()`

catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
```diff
@@ -202,6 +199,10 @@ define([
             isUnfinished: function () {
```
jrnorth: I think this function can be removed now.

catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestMenu.js
```diff
@@ -44,8 +44,10 @@ define([
                 this.listenTo(wreqr.vent, 'upload:finish', this.onUploadFinish);
             },
             showModal: function() {
-                var keepCurrentModal = this.modal && this.modal.isUnfinished();
-                if (!keepCurrentModal) {
+                if (typeof this.modal === 'undefined') {
```
andrewkfiedler: I'm guessing this could be: `this.model === undefined`

emanns95: Hmm it's actually not entering that condition the same way with that change

catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
```diff
@@ -36,9 +36,6 @@ define([
                 'click .finish-upload': 'finishUpload'
             },
             initialize: function () {
-                // there is no automatic chaining of initialize.
-                Modal.prototype.initialize.apply(this, arguments);
```
stustison: So this still just broke all of the other modals. I think there is a better way to do this and still keep the other ones working. The main issue is that hiding the modal causes the view to be destroyed because a listener is being attached. What you really need to do is update Modal.js so that a modal can pass in its own function for destroying the view and if the modal doesn't pass in its own function then it just uses the default one it has now. Then that function that you pass in can actually take into account whether or not the view is done doing its thing. If it isn't, you don't destroy it.

catalog/ui/search-ui/standard/src/main/webapp/js/view/ingest/IngestModal.view.js
```diff
@@ -199,9 +196,9 @@ define([
                     this.$('.upload-info').toggleClass('hidden', true);
                 }
             },
-            isUnfinished: function () {
```
stustison: This is good, isUnFinished is just weird.

emanns95:
Still needs to change the function calls in the other places that call Modal.js's initialize function to pass in a function that'll tell it to destroy when hidden

shaundmorris:
@emanns95 we should plan on backporting this one as well

catalog/ui/search-ui/standard/src/main/webapp/js/view/Modal.js
```diff
@@ -27,13 +27,15 @@ define([
...
-            initialize: function () {
+            initialize: function (view, shouldDestroy) {
                 // destroy view after animation completes.
                 // extending views must call: Modal.prototype.initialize.apply(this, arguments);
-                var view = this;
                 this.$el.one('hidden.bs.modal', function () {
-                    view.destroy();
-                });
+                    if (view !== undefined && shouldDestroy ||
+                        shouldDestroy === undefined) {
```
stustison: Not sure the conditional is quite right, you're never actually calling "shouldDestroy", should just be something like `if(shouldDestroy === undefined || shouldDestroy())` We shouldn't need to update the preferences modal to make this change.

emanns95: Yeah it's not right. I'm going to take a look at it now

timothytierney:
LGTM :+1:

PR #1290, opened 2016-09-20
DDF-2382 Fix registry-id key used in configurations.

clockard:
#### What does this PR do?
Changes the key used to store registry id information in configuration files from the metacard metatype of registry.registry-id to just a constant of registry-id. The `.` in the name causes the property to be lost when passed to backbone UIs on the front end.
@mcalcote @gordocanchola 
@figliold
@shaundmorris

Install ddf with registry. 
Create a secondary local node with a service binding
Go to the sources tab and make a change to the source that was generated for the secondary local node
Update something on the secondary node (like description) and save it. 
Reload the sources tab and verify that there is still only the one entry without an id in the name
DDF-2382
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests


mcalcote:
Manually confirmed fix :shipit:

PR #1292, opened 2016-09-20
DDF-2478 Add service to convert country codes from FIPS 10-4 into ISO 3166-1 alpha 3 and vice versa

roelens8:
#### What does this PR do?

-Adds an OSGi service to convert country codes from FIPS 10-4 into ISO 3166-1 alpha and vice versa.
-LocalCountryCodeConverter.java: loads in a static file (fipsToIso.properties) containing the country code mappings, builds MultiMaps from the mappings to perform the conversions.

-Removes the ISOCountryFormatConverter.java.

This DDF API is designed to be expanded to support not only converting country codes using a local static file, but also converting country codes via external services when available in the future.

@dcruver 
@ryeats 
@jlcsmith 
@harrison-tarr 
@peterhuffer 
@lessarderic 
@jaymcnallie
@pklinef

Run a full build. Make sure integration and unit tests pass. Verify the service install correctly.

Verify the country mappings from this link are the same as the mappings from the country codes mapping file from this PR. I did not include the countries that did not have a FIPS 10-4 to ISO 3166-1 alpha 3 mapping. http://www.opengeocode.org/download/fips2iso.txt

Also, run the `migration:export` admin command and make sure the `fipsToIso3.properties` file gets exported to the `$DDF_HOME/etc/exported/etc` directory.

There needs to be a way to normalize country codes from inputs eg. NITFs, that contain a 2 letter country code (FIPS 10-4), into a common format DDF uses (ISO 3166-1 alpha 3) so that when a user wants to query a metacard by country, he would search by the normalized country code and still retrieve metacards from inputs that have a country code in FIPS 10-4. This service will provide the ability to normalize those country codes.

DDF-2478
#### Checklist:
- [X] Documentation Updated
- [X] Update / Add Unit Tests


libs/geospatial/src/main/java/org/codice/ddf/libs/location/FIPSFormatConverter.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * @param iso3alphaCountryCode a FIPS 10-4 country code
+     * @return an ISO 3166-1 country code
+     */
+    public static String convertIso3ToFips(String iso3alphaCountryCode) {
+        Map<String, String> countryCodesMap = getCountryCodesMap();
+
+        String fipsCountryCode = countryCodesMap.entrySet()
```
ryeats: Could use the guava BiMap so that you can do be 0(1) both ways

peterhuffer:
Do we want to combine this into the same utility class with https://github.com/codice/ddf/blob/master/libs/geospatial/src/main/java/org/codice/ddf/libs/location/ISOFormatConverter.java and call it something like CountryCodeConverter?

roelens8:
Sure, I don't see a problem with combing both into the same utility class.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/FIPSFormatConverter.java
roelens8: I originally tried that, but found out there are two FIPS countries that map to the same ISO alpha 3 country.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,105 @@
...
+     * Converts ISO 3166-1 alpha 3 into FIPS 10-4.
+     * <p>
+     * Special case:
+     * ISO 3166-1 alpha 3 country: "PSE" corresponds to FIPS 10-4 countries: "WE" and "GZ",
+     * in this case, the following string will be returned: "WE,GZ"
+     *
+     * @param iso3alphaCountryCode a FIPS 10-4 country code
+     * @return an ISO 3166-1 country code
+     */
+    public static String convertIso3ToFips(String iso3alphaCountryCode) {
```
peterhuffer: I think returning a list of Strings on this method will be more friendly to the caller rather than returning a tokenized response. And it will make it more clear that this method can return multiple country codes.

roelens8: That's true otherwise the caller would have to parse the results.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,105 @@
...
+            return Arrays.asList("WE", "GZ");
+        }
+        String fipsCountryCode = getCountryCodesMap().entrySet()
+                .stream()
+                .filter(entry -> Objects.equals(entry.getValue(), iso3alphaCountryCode))
+                .map(Map.Entry::getKey)
+                .findFirst()
+                .orElse(null);
+        if (StringUtils.isBlank(fipsCountryCode)) {
+            throw new MissingResourceException(
```
pklinef: Would it be better to return an empty list instead of throwing a runtime exception?

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,105 @@
...
+     * @return an ISO 3166-1 alpha-3 country code
+     */
+    public static String convertIso2ToIso3(String language, String alpha2CountryCode) {
+        Locale locale = new Locale(language, alpha2CountryCode);
+        return locale.getISO3Country();
+    }
+
+    private static Map<String, String> getCountryCodesMap() {
+        if (countryCodesMap == null || countryCodesMap.isEmpty()) {
+            countryCodesMap = PropertiesLoader.toMap(PropertiesLoader.loadProperties(FIPS_TO_ISO3));
```
pklinef: I would suggest putting this into a static block so it only loads once.  This properties file will never change during runtime and does not need to be reloaded every time a conversion is made.

roelens8: My current solution will load the countryCodesMap once, because once the static variable has been initialized it will never be null again. However, this lazy init is not thread safe without it being synchronized, so I will put it in a static block.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
roelens8: Fixed to return an empty list instead on an exception. Also removed the thrown exception in the convertFipsToIso3 method, it will now return null if there is no valid conversion.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
roelens8: Fixed.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,106 @@
...
+     * ISO 3166-1 alpha 3 country: "PSE" corresponds to FIPS 10-4 countries: "WE" and "GZ",
+     * in this case, the returned list will contain those two countries. In all other cases,
+     * the returned list will contain one country.
+     *
+     * @param iso3alphaCountryCode a FIPS 10-4 country code
+     * @return an ISO 3166-1 country code in a List<String> or an empty list if there isn't a valid
+     * conversion.
+     */
+    public static List<String> convertIso3ToFips(String iso3alphaCountryCode) {
+        if ("PSE".equals(iso3alphaCountryCode)) {
```
lessarderic: Since it is possible to have a single alpha 3 code map to multiple FIPS 10-4 codes, maybe we should build another map (See Multimap: http://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimap.html) from the properties file and not have a special case here. Hard-coded special cases aren't great as it is really easy to not know about them when updating the country code properties file and forget to add and/or remove the hard-coded cases.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,106 @@
...
+                .stream()
+                .filter(entry -> Objects.equals(entry.getValue(), iso3alphaCountryCode))
+                .map(Map.Entry::getKey)
+                .findFirst()
+                .orElse(null);
+        if (StringUtils.isBlank(fipsCountryCode)) {
+            LOGGER.debug(
+                    "Could not find a corresponding FIPS 10-4 country code given the ISO 3166-1"
+                            + " alpha 3 country code: {}" + iso3alphaCountryCode);
+            return new ArrayList<>();
```
lessarderic: We should use `Collections.emptyList()` instead.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        return Arrays.asList(fipsCountryCode);
```
lessarderic: `Collections.singletonList(fipsCountryCode)`.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
lessarderic: Also, we may want to have a guard clause for the case where `iso3alphaCountryCode` is `null` or empty, i.e.:

```
if (StringUtils.isEmpty(iso3alphaCountryCode)) {
    return Collections.emptyList();
}
```

libs/geospatial/src/test/java/org/codice/ddf/libs/location/CountryCodeConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    @Test
+    public void testFipsToIso3ConversionInvalidCountryCode() {
+        String iso3 = CountryCodeConverter.convertFipsToIso3("");
+        assertThat(iso3, nullValue());
+    }
+
+    @Test
+    public void testIso3ToFipsConversion() {
+        List<String> fipsCountryCodes = CountryCodeConverter.convertIso3ToFips("RUS");
+        assertThat(fipsCountryCodes.get(0), is("RS"));
```
lessarderic: We should use `containsInAnyOrder("RS")` here to make sure that only one country code was returned.

libs/geospatial/src/test/java/org/codice/ddf/libs/location/CountryCodeConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    public void testIso3ToFipsSpecialCaseConversion() {
+        List<String> fipsCountryCodes = CountryCodeConverter.convertIso3ToFips("PSE");
+        assertThat(fipsCountryCodes, hasItems("WE", "GZ"));
```
lessarderic: We should use `containsInAnyOrder("WE", "GZ")` here to make sure that only one country code was returned.

libs/geospatial/src/main/java/org/codice/ddf/libs/location/CountryCodeConverter.java
```diff
@@ -0,0 +1,106 @@
...
+     * Special case:
+     * ISO 3166-1 alpha 3 country: "PSE" corresponds to FIPS 10-4 countries: "WE" and "GZ",
+     * in this case, the returned list will contain those two countries. In all other cases,
+     * the returned list will contain one country.
+     *
+     * @param iso3alphaCountryCode a FIPS 10-4 country code
+     * @return an ISO 3166-1 country code in a List<String> or an empty list if there isn't a valid
+     * conversion.
+     */
+    public static List<String> convertIso3ToFips(String iso3alphaCountryCode) {
```
lessarderic: Add `@Nullable` on the parameter since we accept `null`.

lessarderic:
@pklinef: should we consider moving the country code properties file somewhere under the `/etc` directory? Since those country codes can and do change, seems to me like we should allow admins to change this file without us having to rebuild and release a new version of the software.

If we do, we should add a note in the documentation and release notes to that effect @roelens8.

pklinef:
@lessarderic I was not aware that they kept updating those lists.  Based on that I agree that we should not use a bundled resource but instead read the file from `/etc`.  

I would suggest reading that file statically once into a data structure.  The tradeoff is that we would need to restart if the file changed but I think that is acceptable given that the file will rarely change and the conversion could be called frequently (per metacard).

bradh:
After a NITF TEM yesterday morning on country codes, I think this might be the wrong approach. The issue is that the ISO-3166 is the standard, but not the code lists. The code lists change (four times per year), and there is already correlation tables available at https://nsgreg.nga.mil/genc/registers.jsp?register=GPC

Perhaps there should be a way to use the existing API; or to provide the same API as a service on DDF.

coyotesqrl:
That's a reasonable solution in production environments with access to the Internet.

lessarderic:
I like @bradh's suggestion of making this an OSGi service. This gives us the most flexibility to support these mappings (external service, static file, etc.).

roelens8:
I agree that making this a service provides more flexibility whether it's using a static file or calling out to a remote web service. I'm going to create a DDF service from my current solution and allow the country codes mapping file to be configurable by a user to cover the case the country codes mapping would need to be changed. And if there is an existing API we can use down the line to map FIPS 10-4 to ISO 3166-1 Alpha 3 country codes, it shouldn't be difficult to modify the DDF service to use that API.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.codice.ddf.spatial.country.converter.api.CountryCodeConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Multimap;
+
+import ddf.security.PropertiesLoader;
+
+public class LocalCountryCodeConverter implements CountryCodeConverter {
```
lessarderic: Short class description would be nice, including that this implementation is backed by a configuration file. May also want to provide what the format of the file needs to be, i.e., properties file with list of comma-separated values.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
```diff
@@ -0,0 +1,94 @@
...
+    public List<String> convertFipsToIso3(@Nullable String fipsCountryCode) {
+        return convert(fipsCountryCode, true);
+    }
+
+    public List<String> convertIso3ToFips(@Nullable String iso3alphaCountryCode) {
+        return convert(iso3alphaCountryCode, false);
+    }
+
+    public String convertIso2ToIso3(String language, String alpha2CountryCode) {
+        Locale locale = new Locale(language, alpha2CountryCode);
```
lessarderic: Since all we're doing here is delegate to Java's `Locale`, do we really need to make this part of the service? Especially given the fact that this code has nothing to do with the country mappings.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
```diff
@@ -0,0 +1,94 @@
...
+        return locale.getISO3Country();
+    }
+
+    private List<String> convert(String countryCode, boolean fipsToIso) {
```
lessarderic: Instead of passing in a `boolean` that changes this method's behavior, should we instead have one common private method (e.g., `boolean isValidCountry(countryCode)`) called from `convertIso3Fips()` and `convertIso2ToIso3()`, have the code in the if/else moved back to `convertIso3Fips()` and `convertIso2ToIso3()` methods, and remove `convert(String, boolean)`?

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
```diff
@@ -0,0 +1,94 @@
...
+                LOGGER.debug(
+                        "Could not find a corresponding ISO 3166-1 alpha 3 country code given the"
+                                + "FIPS 10-4 country code: {}",
+                        countryCode);
+                return Collections.emptyList();
+            }
+        } else {
+            convertedCountryCodes = countryCodesMap.entries()
+                    .stream()
+                    .filter(entry -> Objects.equals(entry.getValue(), countryCode))
```
lessarderic: Since this is a `MultiMap`, `entry.getValue()` will return a `List<String>`, so we can't just compare with `countryCode`. Not sure why this isn't being caught be the unit tests.

Also, any performance concerns here? Do we expect this method to be called a lot, e.g., while ingesting metacards, by plugins, when performing security checks, etc.? If so, we may want to consider using two separate maps.

catalog/spatial/country/spatial-country-local-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,34 @@
...
+           http://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd
+           http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0
+           http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.1.0.xsd">
+
+    <bean id="localCountryCodeConverter"
+          class="org.codice.ddf.spatial.country.converter.impl.LocalCountryCodeConverter">
+        <cm:managed-properties
+                persistent-id="org.codice.ddf.spatial.country.converter.impl.LocalCountryCodeConverter"
+                update-strategy="container-managed"/>
+        <property name="fipsToIso3MappingFile" value="${ddf.home}/etc/fipsToIso.properties"/>
```
lessarderic: Since this is a properties file, we need to update the migration code to copy the file over, https://github.com/codice/ddf/blob/master/platform/platform-migratable/src/main/java/org/codice/ddf/platform/migratable/impl/PlatformMigratable.java if we move this service to platform. If we keep it under catalog, let me know and we can see where this can be added.

distribution/ddf-common/src/main/resources/etc/fipsToIso.properties
```diff
@@ -0,0 +1,249 @@
+# FIPS 10-4 to ISO 3166-1 alpha 3 country codes from http://www.opengeocode.org/
```
lessarderic: Instead of putting this file here, we should keep it inside the bundle and copy it when it gets installed (see https://github.com/codice/ddf/blob/master/platform/security/security-services-app/src/main/resources/features.xml#L96 for an example).

catalog/spatial/country/spatial-country-local-impl/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.99</minimum>
```
roelens8: This has been fixed.

catalog/spatial/country/spatial-country-api/src/main/java/org/codice/ddf/spatial/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,55 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.spatial.country.converter.api;
+
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+/**
+ * Utility to convert country codes into various formats.
```
roelens8: Yes, I'll change it.

catalog/spatial/country/spatial-country-api/src/main/java/org/codice/ddf/spatial/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public interface CountryCodeConverter {
+
+    String ENGLISH_LANG = "en";
```
roelens8: I agree, will fix that.

catalog/spatial/country/spatial-country-api/src/main/java/org/codice/ddf/spatial/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Converts FIPS 10-4 into ISO 3166-1 alpha 3. If a FIPS 10-4 code maps to more than one
+     * ISO 3166-1 alpha 3 code, all of them will be returned in a list of strings.
+     *
+     * @param fipsCountryCode a FIPS 10-4 country code
+     * @return an ISO 3166-1 alpha-3 country code in a List<String> or null if there isn't a
```
roelens8: You are correct, I will fix this error.

catalog/spatial/country/spatial-country-api/src/main/java/org/codice/ddf/spatial/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param iso3alphaCountryCode an ISO 3166 alpha 3 country code
+     * @return an ISO 3166-1 country code in a List<String> or an empty list if there isn't a valid
+     * conversion.
+     */
+    List<String> convertIso3ToFips(@Nullable String iso3alphaCountryCode);
+
+    /**
+     * Converts ISO 3166-1 alpha 2 into ISO 3166-1 alpha 3.
+     *
+     * @param language          the language to return the converted country code in
```
roelens8: Will fix.

catalog/spatial/country/spatial-country-api/src/main/java/org/codice/ddf/spatial/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param alpha2CountryCode an ISO 3166-1 alpha-2 country code
+     * @return an ISO 3166-1 alpha-3 country code
+     */
+    String convertIso2ToIso3(String language, String alpha2CountryCode);
```
roelens8: Sure, I will add @Nullable to each method parameter and state that if any are null, a NullPointerException will be thrown.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
roelens8: Before this PR we had a class `ISOFormatConverter` that contained the one method to convert Iso2 to Iso3. I figured it would be better to combine that class into one service, so that we don't have to worry about using multiple classes to do specific country code conversions. As far as this method having anything to do with the country code mappings file, it doesn't and
this it doesn't depend on that file either, meaning the method can be called whether the feature is configured with the county codes mapping file or not.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
roelens8: I'll remove the convert() method, add a couple private methods to check whether the country code is valid and that the country code map is not null, and then add the if/else code blocks back into their corresponding methods.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
roelens8: Included a description stating what this class does.

distribution/ddf-common/src/main/resources/etc/fipsToIso.properties
roelens8: Fixed.

catalog/spatial/country/spatial-country-api/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.codice.ddf.spatial</groupId>
+        <artifactId>country</artifactId>
+        <version>2.10.0-SNAPSHOT</version>
+    </parent>
+    <name>DDF :: Spatial :: Country :: API</name>
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>bundle</packaging>
+    <groupId>org.codice.ddf.spatial.country</groupId>
```
roelens8: Moved the service under platform to have access to `PlatformMigratable` in order to copy files from the bundled resource into the install directory and to allow other modules including security, catalog..., etc. access to this service.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
roelens8: Removed this method form the interface. It is not part of the service anymore.

catalog/spatial/country/spatial-country-local-impl/src/main/java/org/codice/ddf/spatial/country/converter/impl/LocalCountryCodeConverter.java
roelens8: Stated in the documentation, that if there is a single FIPS 10-4 country that maps to multiple ISO 3166-1 country, the format of the mapping is required to be for example: AB=ABC,BCD.

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
```diff
@@ -58,3 +58,16 @@ The Platform application is installed by default on a standard installation.
 |
 
 |===
+
```
lessarderic: @ricklarsen, can you take a quick look to make sure the description and formatting fit with the rest of the docs?

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
```diff
@@ -58,3 +58,16 @@ The Platform application is installed by default on a standard installation.
...
+==== Country Code Converter
+
+In the Platform Application, you have the option to install a feature called `platform-country-local`.
+This feature provides an OSGi service to convert country codes into various formats by performing a local lookup.
+
+In order to utilize the FIPS 10-4 to ISO 3166-1 alpha 3 and vice versa country code conversions, the `FIPS 10-4 to ISO 3166-1 alpha 3 File Location` field in the feature's configuration must be entered with the file path of a properties file containing the FIPS 10-4 to ISO 3166-3 alpha 3 country code mappings. However, the feature is already configured by default upon installation, with the file location field pointing to a properties file containing country code mappings from http://www.opengeocode.org/. This file can be found at: ${branding-lowercase}.home/etc/fipsToIso.properties, and should only be modified in the case the country code mappings would ever change.
```
lessarderic: Don't think we typically use `${branding-lowercase}.home` to refer to the home directory. @ricklarsen?

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
```diff
@@ -58,3 +58,16 @@ The Platform application is installed by default on a standard installation.
...
+
+The mapping file contains each mapping in the following format, where the country on the left side of the equals sign is FIPS and the country or countries on the right side of the equals sign is ISO, and a new line separating each mapping:
+
+If there is a single FIPS 10-4 country that maps to a single ISO 3166-1 alpha 3 country: AB=ABC
```
lessarderic: May want to make this a bulleted list.

platform/country-converter/platform-country-converter-local/src/main/java/org/codice/ddf/country/converter/local/LocalCountryCodeConverter.java
```diff
@@ -0,0 +1,95 @@
...
+                        .splitToList(countryMappingEntry.getValue());
+                for (String isoCountry : mappedCountries) {
+                    String fipsCountry = countryMappingEntry.getKey();
+                    fipsToIsoMap.put(fipsCountry, isoCountry);
+                    isoToFipsMap.put(isoCountry, fipsCountry);
+                }
+            }
+        }
+        if (fipsToIsoMap.isEmpty() || isoToFipsMap.isEmpty()) {
+            LOGGER.warn("Country code maps are empty. Verify the file path field in the feature's "
```
lessarderic: Might be useful to add `fipsToIso3MappingFile` to the log message.

platform/country-converter/platform-country-converter-api/src/main/java/org/codice/ddf/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Converts FIPS 10-4 into ISO 3166-1 alpha 3. If a FIPS 10-4 code maps to more than one
+     * ISO 3166-1 alpha 3 code, all of them will be returned in a list of strings.
+     *
+     * @param fipsCountryCode a FIPS 10-4 country code
+     * @return an ISO 3166-1 alpha-3 country code in a List<String> or an empty list if there isn't
+     * a valid conversion.
+     */
+    List<String> convertFipsToIso3(@Nullable String fipsCountryCode);
```
pklinef: This interface is too specific to FIPS and ISO currently.  There should just be one method that takes a country code, its format, and the desired target format.  As it currently is, you will have n^2 methods, where n is the number of support country code types, as we add support for other country codes in the future.

This also means you need a normalized format for the configuration files.  We could select ISO to be the normalized form.  If we did, I would expect the file to be just fips.properties since all mappings would be converting to ISO.

roelens8: Agreed, the interface is to specific. I will make it generic as requested.

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
ricklarsen: yes.

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
roelens8: Fixed.

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
roelens8: Made the requested changes.

distribution/docs/src/main/resources/_contents/_platform-contents/managing-platform-contents.adoc
```diff
@@ -58,3 +58,18 @@ The Platform application is installed by default on a standard installation.
...
+
+A mappings file is always required to convert between some country code format X and the ISO 3166-1 alpha 3 format.
+
+The mappings file contains each mapping in the following format, where the country on the left side of the equals sign is the source format e.g. FIPS 10-4 and the country or countries on the right side of the equals sign are required to be ISO 3166-1 alpha 3 (as that's the normalized country code format), and a new line separating each mapping:
+
+If there is a single country code from the format X that maps to a single ISO 3166-1 alpha 3 country: AB=ABC
+
+If there is a single country code from the format X that maps to multiple ISO 3166-1 alpha 3 countries: ZZ=ZZZ,YYY
```
roelens8: Fixed.

platform/country-converter/platform-country-converter-api/src/main/java/org/codice/ddf/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,63 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.country.converter.api;
+
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+/**
+ * Service to convert country codes into various formats. In order to convert country codes from
```
roelens8: Should this description stay in the interface or be moved to the implementation class, LocalCountryCodeConverter? The reason why it's currently in the interface, is so that when an integration wants to use the API, he'll know exactly how to use it without knowing what the implementation is, otherwise, it may be difficult for the integrator to know what the values of the sourceFormat and desiredFormat should be as they're just strings.

platform/country-converter/platform-country-converter-api/src/main/java/org/codice/ddf/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,63 @@
...
+     * The string value of the country code formats used for "sourceFormat" and "desiredFormat" must
+     * be equal to the name of their corresponding country code mappings file, excluding the file
+     * extension. EX. file name: "fips_10_4.properties", format used for conversion: "fips_10_4"
+     *
+     * @param countryCode   a country code.
+     * @param sourceFormat  source format of the country code.
+     * @param desiredFormat desired format of the country code to convert into.
+     * @return country code(s) in a List of Strings.
+     */
+    List<String> convert(@Nullable String countryCode, String sourceFormat, String desiredFormat);
```
lessarderic: I like the more generic method signature. However, it will cause problems if we want to provide two different ways of converting a country code from format A to format B, e.g., one implementation using a web service and one using a properties file (which was the goal of @bradh's comment on the initial version of this PR).

Since every conversion from one format to another may or may not be able to use the same conversion mechanism (e.g., A -> B and B -> A can use a web service, C -> D needs to use a properties files, etc.), the current interface would force the implementation class to support all conversion strategies and contain pretty ugly conditions. In other words, if the source format is A and the desired format is B, call this web service. Otherwise, if the source is C and the desired format is D, use this properties file, etc.

One option to keep things simple and flexible would be to use service properties. This would allow us to provide fine-grained implementations of this service interface, one for each conversion type. For instance, when a client needs to convert a country code from format A to format B, they would look up the `CountryCodeConverter` service using a filter like `&(sourceFormat=A)(destinationFormat=B)`. The method would only need to take in a single parameter, i.e., the country code to convert:

`List<String> convert(@Nullable String countryCode)`

This way, every X -> Y country code conversion can have its own implementation. Hope this makes sense.

lessarderic: Sorry, looks like my comment wasn't clear :-(

My suggestion wasn't really the opposite of Phil's solution. I was basically proposing to use the source and destination formats for looking up the service instead of passing those arguments to the method itself.

platform/country-converter/platform-country-converter-api/src/main/java/org/codice/ddf/country/converter/api/CountryCodeConverter.java
```diff
@@ -0,0 +1,63 @@
...
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface CountryCodeConverter {
+
+    /**
+     * Variable representing the normalized country code format.
+     */
+    String ISO_3166_1_ALPHA_3 = "ISO_3166_1_ALPHA3";
```
roelens8: Moved to the implementation.

roelens8:
After further discussion, it was concluded the best approach for now is to revert the PR back to the state when the interface was specific to ISO and FIPS.

distribution/docs/src/main/resources/_contents/_draft-platform-contents/country-code-converter-contents.adoc
```diff
@@ -0,0 +1,32 @@
+==== Country Code Converter
+
+In the Platform Application, you have the option to install a feature called `platform-country-local`.
+This feature provides an OSGi service to convert country codes into various formats by performing a local lookup.
+
+In order to utilize the FIPS 10-4 to ISO 3166-1 alpha 3 and vice versa country code conversions, the `FIPS 10-4 to ISO 3166-1 alpha 3 File Location` field in the feature's configuration must be entered with the file path of a properties file containing the FIPS 10-4 to ISO 3166-3 alpha 3 country code mappings. The feature is already configured by default upon installation, with the file location field pointing to a properties file containing country code mappings from http://www.opengeocode.org/. This file can be found at: ${branding-lowercase}.home/etc/fipsToIso3.properties, and should only be modified in the case the country code mappings would ever change.
```
lessarderic: Just thought of something. The configuration file is located in the `/etc` directory and can be updated by an admin if needed. However, if the file is changed, the in-memory map will not get updated unless the bundle is restarted. We probably should add some instructions explaining how to update the mappings file and restart the bundle. @ricklarsen, is this the best place to put these instructions?

distribution/docs/src/main/resources/_contents/_draft-platform-contents/country-code-converter-contents.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+The Country Code Converter is not installed by default with a standard installation.
+To install:
+
+. Navigate to the *${admin-console}*.
+. Select the *${ddf-platform}* application.
+. Select the *Features* tab.
+. Install the `platform-country-local` feature.
+
+===== Configuring Country Code Converter
```
lessarderic: I wonder if we need to allow admins to change the path to the mapping file. It adds yet another configuration option that I'm not sure is really that useful. On top of that, if we keep it configurable, we will need to change the `PlatformMigratable` code since the file could be in a different directory or have a different name.

distribution/docs/src/main/resources/_contents/_draft-platform-contents/country-code-converter-contents.adoc
roelens8: Good point! I will update the documentation to explain that if the mappings file is updated the bundle will need to be restarted and then provide instructions on how to do that.

distribution/docs/src/main/resources/_contents/_draft-platform-contents/country-code-converter-contents.adoc
roelens8: Yeah, it's probably better that admins do not have access to change the mappings file path. I will remove the metatype and remove the configuring section from the documentation.

platform/country-converter/platform-country-converter-api/src/main/java/org/codice/ddf/country/converter/api/CountryCodeConverter.java
roelens8: Reverted PR back to being specific as that was the decision made to move forward with.

harrison-tarr:
Successfully heroed.

harrison-tarr:
Manually verified that the bundles start when I install the feature. The file in the PR is the same as the file linked. I used the diff tool at https://text-compare.com/ to compare. All the countries that have a FIPS 10-4 and a ISO 3166-1 alpha 3 code are in the file introduced in this PR. Also verified that the `migration:export` worked.

distribution/docs/src/main/resources/_contents/_draft-platform-contents/country-code-converter-contents.adoc
```diff
@@ -0,0 +1,21 @@
+==== Country Code Converter
```
pklinef: Since this is an internal API, should we move this documentation to a README.adoc somewhere in the  platform/country-converter area?

jaymcnallie: Yes.

PR #1306, opened 2016-09-26
DDF-2498 Added ability to specify source and target for Migrate Command

emmberk:
#### What does this PR do?
This PR adds `--from` and `--to` options to the Migrate Command to specify the providers, and it adds a `--list` option to list the available providers.
@bdeining @glenhein @peterhuffer 
@beyelerb
@kcwire
Add an external provider, and test that the `catalog:migrate` and the options work as expected when migrating from this new external provider to the `SolrHttpCatalogProvider`.
https://codice.atlassian.net/browse/DDF-2498
#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/MigrateCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/MigrateCommand.java
@@ -13,12 +13,16 @@
  */
 package org.codice.ddf.commands.catalog;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.felix.gogo.commands.Command;
+import org.apache.felix.gogo.commands.Option;
 import org.codice.ddf.commands.catalog.facade.CatalogFacade;
 import org.codice.ddf.commands.catalog.facade.Provider;
 import org.geotools.filter.text.cql2.CQL;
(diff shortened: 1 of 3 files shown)


catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/MigrateCommand.java
```diff
@@ -43,44 +47,66 @@
...
+        if (listProviders) {
+            if (providers.size() == 0) {
+                console.println("There are no available providers.");
+                return null;
+            }
+            console.println("Available providers:");
+            providers.stream()
+                    .map(p -> p.getClass()
+                            .getSimpleName())
+                    .forEach(id -> console.println("\t" + id));
```
peterhuffer: Should we return on this and end the operation when the `--list` option is used? I can't imagine something like `catalog:migrate --list -to x -from y` being done.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/MigrateCommand.java
```diff
@@ -43,44 +47,66 @@
...
-        + "from a Configured External Provider into the Catalog.  The migrate command currently "
-        + "picks the first Catalog provider as the FROM and the second Catalog provider as the TO. "
-        + "Additional support will be added to specify which provider will be migrated TO and FROM "
-        + "in the future.")
+        + "from a Configured External Provider to another Provider.")
 public class MigrateCommand extends DuplicateCommands {
     private static final Logger LOGGER = LoggerFactory.getLogger(MigrateCommand.class);
 
+    @Option(name = "--list", required = false, aliases = {
+            "-list"}, multiValued = false, description = "Print a list of external providers before migrating.")
```
peterhuffer: Is this necessarily only printing external providers? Won't this `--list` option also print the default local provider? Similar comment for the `--from` and `--to` options.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/MigrateCommand.java
emmberk: Updated to complete command after listing providers.

glenhein:
Emily suggested two different approaches to hero the code. I used the approach where I commented out the code that enforces the requirement that there are at least two providers. That way I could migrate to/from the same instance.

`ddf@local>catalog:migrate -from SolrHttpCatalogProvider -to SolrHttpCatalogProvider
FROM provider ID: SolrHttpCatalogProvider
TO provider ID: SolrHttpCatalogProvider
Starting migration.
 100% [===================================================>]     6 records/sec	
 2 record(s) migrated; 0 record(s) failed; completed in 0.305 seconds.
`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/MigrateCommand.java
```diff
@@ -43,44 +47,68 @@
...
+            providers.stream()
+                    .map(p -> p.getClass()
+                            .getSimpleName())
+                    .forEach(id -> console.println("\t" + id));
+
+            return null;
+        }
+
         if (batchSize > MAX_BATCH_SIZE || batchSize < 1) {
             console.println("Batch Size must be between 1 and 1000.");
```
beyelerb: use MAX_BATCH_SIZE here?

PR #1322, opened 2016-09-30
DDF-2521: Refactored implementations of SaxEventHandler

AzGoalie:
#### What does this PR do?
Cleans up redundant code with the SaxEventHandler class and its implementations.
@harrison-tarr @spearskw 
@coyotesqrl
@lessarderic
A full build should pass
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/transformer/catalog-transformer-streaming-api/src/main/java/org/codice/ddf/transformer/xml/streaming/AbstractSaxEventHandler.java
+++ b/catalog/transformer/catalog-transformer-streaming-api/src/main/java/org/codice/ddf/transformer/xml/streaming/AbstractSaxEventHandler.java
@@ -1,0 +1,71 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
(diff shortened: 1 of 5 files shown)


coyotesqrl:
You should provide some heroing instructions; otherwise, LGTM.

AzGoalie:
The heroing process should just be a full build and pass. Updated original post.

catalog/transformer/catalog-transformer-streaming-api/src/main/java/org/codice/ddf/transformer/xml/streaming/AbstractSaxEventHandler.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public void endPrefixMapping(String prefix) throws SAXException {
+
+    }
+
+    @Override
+    public void endElement(String uri, String localName, String qName) throws SAXException {
+
```
harrison-tarr: In my opinion `endElement` and `characters` should not have implementations in this class. I think _usually_ they're used, and when someone is writing a SaxEventHandlerImpl, I think it should be a conscious decision to not put an implementation. I think it would lead to more legible, understandable classes. Maybe even `startPrefixMapping`. 

That being said, I can also understand and appreciate only having to implement the methods necessary for your implementation, so I could be swayed either way. I think having fewer default implementations is more outsider friendly, but having more is more aesthetic and elegant.

AzGoalie: I'll remove the `endElement` and `characters` implementations, but `startPrefixMapping` doesn't look like it's needed as much.

PR #1327, opened 2016-10-04
DDF-2507 Added MetacardType attribute registration to XmlInputTransformer

emanns95:
#### What does this PR do?
The attributes for metacards created using the XmlInputTransformer do not register the MetacardType Attributes. These attributes are now registered.

@brendan-hofmann 
@harrison-tarr 
@jrnorth 

@coyotesqrl
@kcwire

Full build and ingest xml data with a numerical field. Check that the field can be searched on.

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/transformer/catalog-transformer-streaming-impl/pom.xml
+++ b/catalog/transformer/catalog-transformer-streaming-impl/pom.xml
@@ -80,11 +80,6 @@
             <groupId>ddf.platform.util</groupId>
             <artifactId>platform-util</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.felix</groupId>
-            <artifactId>org.apache.felix.utils</artifactId>
-            <version>1.2.0</version>
-        </dependency>
     </dependencies>
     <build>
         <plugins>
@@ -102,14 +97,9 @@
                             gml-v_3_1_1-schema,
                             jaxb2-basics-runtime,
                             catalog-transformer-streaming-lib,
-                            platform-util,
-                            org.apache.felix.utils
+                            platform-util
                         </Embed-Dependency>
-                        <Export-Package/>
-                        <Import-Package>
-                            !org.apache.felix.utils.collections,
-                            *
-                        </Import-Package>
+                        <Export-Package />
                     </instructions>
                 </configuration>
             </plugin>
--- a/catalog/transformer/catalog-transformer-streaming-impl/pom.xml
+++ b/catalog/transformer/catalog-transformer-streaming-impl/pom.xml
@@ -80,6 +80,11 @@
             <groupId>ddf.platform.util</groupId>
             <artifactId>platform-util</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.felix</groupId>
+            <artifactId>org.apache.felix.utils</artifactId>
+            <version>1.2.0</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
@@ -97,9 +102,14 @@
                             gml-v_3_1_1-schema,
                             jaxb2-basics-runtime,
                             catalog-transformer-streaming-lib,
-                            platform-util
+                            platform-util,
+                            org.apache.felix.utils
                         </Embed-Dependency>
-                        <Export-Package />
+                        <Export-Package/>
+                        <Import-Package>
+                            !org.apache.felix.utils.collections,
+                            *
+                        </Import-Package>
                     </instructions>
                 </configuration>
             </plugin>
--- a/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -39,11 +39,6 @@
     <bean id="dynamicMetacardTypeRegister"
           class="org.codice.ddf.transformer.xml.streaming.lib.MetacardTypeRegister">
         <property name="id" value=""/>
-        <property name="saxEventHandlerFactories" ref="saxEventHandlerFactoriesList"/>
-        <property name="saxEventHandlerConfiguration">
-            <list>
-            </list>
-        </property>
     </bean>
 
     <cm:managed-service-factory id="XmlInputTransformerFactory"
--- a/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -64,7 +64,6 @@
                 <list>
                 </list>
             </property>
-            <property name="context" ref="blueprintBundleContext"/>
 
             <cm:managed-properties persistent-id=""
                                    update-strategy="container-managed"/>
--- a/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -64,7 +64,7 @@
                 <list>
                 </list>
             </property>
-
+            <property name="context" ref="blueprintBundleContext"/>
 
             <cm:managed-properties persistent-id=""
                                    update-strategy="container-managed"/>
--- a/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -32,8 +32,19 @@
                     availability="optional">
         <reference-listener ref="saxEventHandlerFactoriesList" bind-method="bindPlugin"
                             unbind-method="unbindPlugin"/>
+        <reference-listener ref="dynamicMetacardTypeRegister" bind-method="bind"
+                            unbind-method="unbind"/>
     </reference-list>
 
+    <bean id="dynamicMetacardTypeRegister"
+          class="org.codice.ddf.transformer.xml.streaming.lib.MetacardTypeRegister">
+        <property name="id" value=""/>
+        <property name="saxEventHandlerFactories" ref="saxEventHandlerFactoriesList"/>
+        <property name="saxEventHandlerConfiguration">
+            <list>
+            </list>
+        </property>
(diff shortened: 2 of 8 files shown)


catalog/transformer/catalog-transformer-streaming-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -64,7 +64,7 @@
                 <list>
                 </list>
             </property>
-
+            <property name="context" ref="blueprintBundleContext"/>
```
kcwire: do not pass in the `bundleContext` as it can become stale.  It is best to use FrameworkUtils to get a fresh bundle context each time you need it.
https://github.com/codice/ddf/blob/master/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java#L1401

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/XmlInputTransformer.java
```diff
@@ -260,9 +269,37 @@ public MetacardType getMetacardType() {
...
+        DynamicMetacardType dynamicMetacardType = new DynamicMetacardType(attributeDescriptors, id);
+        registerMetacardType(dynamicMetacardType);
+        return dynamicMetacardType;
+    }
+
+    private void registerMetacardType(MetacardType metacardType) {
+        unregisterMetacardType(metacardType);
+
+        ServiceRegistration serviceRegistration =
+                context.registerService(MetacardType.class.getName(), metacardType, null);
```
kcwire: You should include service properties that has the name of the metacard.
https://github.com/codice/ddf/blob/master/catalog/transformer/catalog-transformer-pdf/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L33

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/XmlInputTransformer.java
```diff
@@ -260,9 +269,37 @@ public MetacardType getMetacardType() {
...
         }
         attributeDescriptors.addAll(BasicTypes.BASIC_METACARD.getAttributeDescriptors());
-        return new DynamicMetacardType(attributeDescriptors, id);
+
+        DynamicMetacardType dynamicMetacardType = new DynamicMetacardType(attributeDescriptors, id);
+        registerMetacardType(dynamicMetacardType);
+        return dynamicMetacardType;
+    }
+
+    private void registerMetacardType(MetacardType metacardType) {
```
coyotesqrl: Can multiple threads be invoking this transformer at once? If so, all access to the `metacardTypeServiceRegistrations` map and the un/registration logic needs to be synchronized/guarded.

Also...you're reregistering on every call to transform in inputStream. That seems a bit expensive and problematic.

harrison-tarr: I recommend moving the dynamic metacard type generation from the transform method to the factory configuration update and then register the metacard type there.

emanns95:
@kcwire @coyotesqrl I think I've addressed the concerns

catalog/transformer/catalog-transformer-streaming-impl/src/test/java/org/codice/ddf/transformer/xml/streaming/impl/TestXmlInputTransformer.java
```diff
@@ -219,8 +225,12 @@ public void testGml3Conversion() throws FileNotFoundException, CatalogTransforme
...
             throws FileNotFoundException, CatalogTransformerException, ValidationException {
         inputStream = new FileInputStream("src/test/resources/metacard1.xml");
-        xmlInputTransformer = new XmlInputTransformer();
-        xmlInputTransformer.setSaxEventHandlerConfiguration(Collections.singletonList("gml-handler"));
+        xmlInputTransformer = new XmlInputTransformer() {
+            @Override
+            public BundleContext getContext() {
+                return mock(BundleContext.class);
+            }
+        };        xmlInputTransformer.setSaxEventHandlerConfiguration(Collections.singletonList("gml-handler"));
```
jrnorth: I think this formatting is wrong.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/XmlInputTransformer.java
```diff
@@ -47,6 +53,9 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(XmlInputTransformer.class);
 
+    private Map<String, ServiceRegistration> metacardTypeServiceRegistrations =
+            new HashMap<String, ServiceRegistration>();
```
jrnorth: Don't need the type parameters the second time.

catalog/transformer/catalog-transformer-streaming-impl/pom.xml
```diff
@@ -97,9 +102,14 @@
...
                             jaxb2-basics-runtime,
                             catalog-transformer-streaming-lib,
-                            platform-util
+                            platform-util,
+                            org.apache.felix.utils
                         </Embed-Dependency>
-                        <Export-Package />
+                        <Export-Package/>
+                        <Import-Package>
+                            !org.apache.felix.utils.collections,
```
brendan-hofmann: This is a smell, we should be managing exclusions through the maven dependencies instead. Otherwise you wind up with different behavior at build and at run.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/MetacardTypeRegister.java
```diff
@@ -0,0 +1,163 @@
...
+    public Set<AttributeDescriptor> getAttributeDescriptors() {
+        return attributeDescriptors;
+    }
+
+    @Override
+    public AttributeDescriptor getAttributeDescriptor(String attributeName) {
+        return attributeDescriptors.stream()
+                .filter(p -> p.getName()
+                        .equals(attributeName))
+                .collect(Collectors.toList())
```
coyotesqrl: If you just want the first, don't waste the time processing the whole collection and building a new List. Just do

```java
return attributeDescriptors.stream()
    .filter(p -> p.getName().equals(attributeName))
    .findFirst()
    .orElse(null);
```
That should do it.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/MetacardTypeRegister.java
```diff
@@ -34,19 +34,19 @@
...
-    private List<SaxEventHandlerFactory> saxEventHandlerFactories;
-
-    private List<String> saxEventHandlerConfiguration;
+    private List<SaxEventHandlerFactory> saxEventHandlerFactories = new ArrayList<>();
 
     private DynamicMetacardType metacardType;
 
     private String id = "DEFAULT_ID";
 
     public void bind(SaxEventHandlerFactory saxEventHandlerFactory) {
```
coyotesqrl: You'll now need to synchronize your reads/writes to the `saxEventHandlerFactories` list (and calls to `setMetacardType`. Simplest to just sync the `bind`, `unbind`, and `setMetacardType` methods. You could use a lock instead but that's probably overkill.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/MetacardTypeRegister.java
```diff
@@ -59,7 +59,7 @@ public void unbind(SaxEventHandlerFactory saxEventHandlerFactory) {
     public void setMetacardType() {
```
coyotesqrl: This doesn't need to be public that I can see. It'd be better as a private method if that's possible.

catalog/transformer/catalog-transformer-streaming-lib/src/test/java/org/codice/ddf/transformer/xml/streaming/lib/TestMetacardTypeRegister.java
```diff
@@ -0,0 +1,77 @@
...
+        metacardTypeRegister = new MetacardTypeRegister() {
+            @Override
+            public BundleContext getContext() {
+                return mock(BundleContext.class);
+            }
+        };
+    }
+
+    @Test
+    public void testBind() throws Exception {
```
jrnorth: So this test is really just checking that the registered metacard type has the same `AttributeDescriptors` as `BASIC_METACARD`? If that's the case you can just `assertThat(results, is(BasicTypes.BASIC_METACARD.getAttributeDescriptors()))`.

catalog/transformer/catalog-transformer-streaming-lib/src/test/java/org/codice/ddf/transformer/xml/streaming/lib/TestMetacardTypeRegister.java
```diff
@@ -0,0 +1,77 @@
...
+
+    SaxEventHandlerFactory mockFactory;
+
+    @Before
+    public void setup() {
+        //mock factory
+        mockFactory = mock(SaxEventHandlerFactory.class);
+        doReturn(BasicTypes.BASIC_METACARD.getAttributeDescriptors()).when(mockFactory)
+                .getSupportedAttributeDescriptors();
+        List mockFactories = ImmutableList.of(mockFactory);
```
jrnorth: Is this used anywhere?

catalog/transformer/catalog-transformer-streaming-lib/src/test/java/org/codice/ddf/transformer/xml/streaming/lib/TestMetacardTypeRegister.java
emanns95: Agreed. Fixed

PR #1171, opened 2016-08-22
DDF-2325: Populates Location.COUNTRY_CODE field for geolocated metacards

peterhuffer:
#### What does this PR do?
- Adds ability for local and websearch gazetteers to get country code for a wktLocation
- Adds a PreIngestPlugin for populate Location.COUNTRY_CODE field for metacard if Metacard.GEOGRAPHY is present
- Adds a utility in `ddf/libs` for converting ISO 3166-1 alpha2 to alpha3 format.

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged; if a component team is listed, at least one of its members needs to approve)?
@glenhein @emmberk @jrnorth @ricklarsen 
#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@kcwire
@shaundmorris
#### How should this be tested?
Unit tests. There is no visible affect from this addition due to many components not supporting the taxonomy yet.
#### What are the relevant tickets?
https://codice.atlassian.net/browse/DDF-2325
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #1332, opened 2016-10-05
DDF-2531 Updated GeoCoderPlugin to handle UpdateRequests

emmberk:
#### What does this PR do?
This PR updates the GeoCoderPlugin to handle UpdateRequests.
@peterhuffer @glenhein @mweser @brjeter
@beyelerb
@kcwire
Confirm unit tests pass.
Relates to https://github.com/codice/ddf/pull/1171.

The spatial-geocoding-plugin in the DDF Spatial app needs to be installed for the plugin to work.
[DDF-2531](https://codice.atlassian.net/browse/DDF-2531)
#### Checklist:
- [ ] Documentation Updated
      - [ ] Change log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
+++ b/catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
@@ -13,7 +13,6 @@
  */
 package org.codice.ddf.spatial.geocoding.plugin;
 
-import java.io.Serializable;
 import java.util.Map;
 import java.util.Optional;
 
@@ -46,6 +45,8 @@ public class GeoCoderPlugin implements PreIngestPlugin {
 
     private int radiusInKm = 10;
 
+    private GeoCoder geoCoder;
+
     public GeoCoderPlugin(ServiceSelector<GeoCoder> geoCoderFactory) {
         if (geoCoderFactory == null) {
             throw new IllegalArgumentException(
@@ -62,35 +63,26 @@ public class GeoCoderPlugin implements PreIngestPlugin {
             return input;
         }
 
-        for (Metacard metacard : input.getMetacards()) {
-            Optional<String> wktLocation = Optional.ofNullable(metacard.getLocation());
+        geoCoder = geoCoderFactory.getService();
+        input.getMetacards()
+                .forEach(this::setCountryCode);
 
-            if (wktLocation.isPresent()) {
-                if (countryCodeNeeded(metacard)) {
-                    GeoCoder geocoder = geoCoderFactory.getService();
-
-                    if (geocoder != null) {
-                        Optional<String> alpha3CountryCode =
-                                geocoder.getCountryCode(wktLocation.get(), radiusInKm);
-
-                        if (alpha3CountryCode.isPresent()) {
-                            setCountryCodeAttribute(metacard, alpha3CountryCode.get());
-                            LOGGER.trace(
-                                    "Setting metacard country code to {} for metacard with id {}",
-                                    alpha3CountryCode.get(),
-                                    metacard.getId());
-                        }
-                    }
-                }
-            }
-        }
         return input;
     }
 
     @Override
     public UpdateRequest process(UpdateRequest input)
             throws PluginExecutionException, StopProcessingException {
-        // TODO: Implement when AdaptedMetacard updated to use taxonomy
+        if (input == null || input.getUpdates() == null) {
+            return input;
+        }
+
+        geoCoder = geoCoderFactory.getService();
+        input.getUpdates()
+                .stream()
+                .map(Map.Entry::getValue)
+                .forEach(this::setCountryCode);
+
         return input;
     }
 
@@ -118,11 +110,33 @@ public class GeoCoderPlugin implements PreIngestPlugin {
         return radiusInKm;
     }
 
-    private boolean countryCodeNeeded(Metacard metacard) {
-        return metacard.getAttribute(Location.COUNTRY_CODE) == null;
+    /**
+     * Sets the country code attribute of the {@param metacard} when the {@param metacard} has a
+     * location and there is not already a country code attribute. Does not update the country code
+     * if the country code attribute is already set.
+     *
+     * @param metacard
+     */
+    private void setCountryCode(Metacard metacard) {
+        Optional<String> wktLocation = Optional.ofNullable(metacard.getLocation());
+
+        if (geoCoder != null && wktLocation.isPresent() && !hasCountryCode(metacard)) {
+            Optional<String> alpha3CountryCode = geoCoder.getCountryCode(wktLocation.get(),
+                    radiusInKm);
+
+            geoCoder.getCountryCode(wktLocation.get(), radiusInKm)
+                    .ifPresent(countryCode -> {
+                        LOGGER.trace("Setting metacard country code to {} for metacard with id {}",
+                                alpha3CountryCode.get(),
+                                metacard.getId());
+
+                        metacard.setAttribute(new AttributeImpl(Location.COUNTRY_CODE,
+                                alpha3CountryCode.get()));
+                    });
+        }
     }
 
-    private void setCountryCodeAttribute(Metacard metacard, Serializable countryCode) {
-        metacard.setAttribute(new AttributeImpl(Location.COUNTRY_CODE, countryCode));
+    private boolean hasCountryCode(Metacard metacard) {
+        return metacard.getAttribute(Location.COUNTRY_CODE) != null;
     }
 }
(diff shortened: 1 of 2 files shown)


catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -62,35 +61,20 @@ public CreateRequest process(CreateRequest input)
...
-                            LOGGER.trace(
-                                    "Setting metacard country code to {} for metacard with id {}",
-                                    alpha3CountryCode.get(),
-                                    metacard.getId());
-                        }
-                    }
-                }
-            }
-        }
+        input.getMetacards().stream().forEach(metacard -> setCountryCodeAttributeIfNeeded(metacard));
```
glenhein: I don't think you need ".stream()", since java.util.List has "forEach".

catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -118,11 +102,27 @@ public int getRadius() {
...
-    private boolean countryCodeNeeded(Metacard metacard) {
-        return metacard.getAttribute(Location.COUNTRY_CODE) == null;
+    private void setCountryCodeAttributeIfNeeded(Metacard metacard) {
+        Optional<String> wktLocation = Optional.ofNullable(metacard.getLocation());
+
+        if (wktLocation.isPresent() && countryCodeNeeded(metacard)) {
+            GeoCoder geocoder = geoCoderFactory.getService();
+
+            if (geocoder != null) {
+                Optional<String> alpha3CountryCode = geocoder.getCountryCode(wktLocation.get(), radiusInKm);
```
glenhein: I think this would work too:

`geocoder.getCountryCode(wktLocation.get(), radiusInKm).ifPresent(countryCode -> {

  metacard.setAttribute(new AttributeImpl(Location.COUNTRY_CODE, alpha3CountryCode.get()));

                    LOGGER.trace("Setting metacard country code to {} for metacard with id {}",
                            alpha3CountryCode.get(),
                            metacard.getId());
})`

catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -62,35 +61,20 @@ public CreateRequest process(CreateRequest input)
...
 
     @Override
     public UpdateRequest process(UpdateRequest input)
             throws PluginExecutionException, StopProcessingException {
-        // TODO: Implement when AdaptedMetacard updated to use taxonomy
+        if (input == null || input.getUpdates() == null) {
+            return input;
+        }
+
+        input.getUpdates().forEach(updateEntry -> setCountryCodeAttributeIfNeeded(updateEntry.getValue()));
```
coyotesqrl: Could also do...
```java
        input.getUpdates()
                .stream()
                .map(Map.Entry::getValue)
                .forEach(this::setCountryCodeAttributeIfNeeded);
```

catalog/spatial/geocoding/spatial-geocoding-plugin/src/test/groovy/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPluginTest.groovy
```diff
@@ -161,6 +128,90 @@ class GeoCoderPluginTest extends Specification {
...
+
+        geoCoderPlugin = initGeoCoderPlugin(countryCode, false);
+
+        geoCoderPlugin.updateConfiguration(map)
+
+        expect:
+        geoCoderPlugin.getRadius() == 15
+    }
+
     def initGeoCoderPlugin(Optional<String> countryCode, boolean overridDefaultGeocoder) {
```
emmberk: Typo: `overridDefaultGeocoder` should be `overrideDefaultGeocoder`

peterhuffer:
I'll bring it up here, but @emmberk and I talked about this. Currently we won't update any metacard that has a country code, even if its Geography is updated. Curious what other people think of this. Ie, should we update the country code if the metacard geography is updated? I was unsure about it, and thought we would want to keep the country code.

emmberk:
@peterhuffer, I asked @millerw8, and he said that we probably wouldn't want to overwrite the country code if it already exists because there could potentially be an issue, for example, where a location is on the border if 2 countries, and the country code is already set to one of the countries instead of the other

catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -118,11 +106,31 @@ public int getRadius() {
         return radiusInKm;
     }
 
-    private boolean countryCodeNeeded(Metacard metacard) {
-        return metacard.getAttribute(Location.COUNTRY_CODE) == null;
+    private void setCountryCodeAttributeIfNeeded(Metacard metacard) {
```
millerw8: Not to be nit-picky... but I do not love the names of these functions.

setCountryCode or updateCountryCode instead of setCountryCodeIfNeeded reads a lot better, a simple explanation in a method documentation would make it more readable.

Also, I think hasCountryCode would be a lot better than countryCodeNeeded

peterhuffer:
Successfully did full build against an empty m2 repo and passed all tests.

catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -118,11 +106,38 @@ public int getRadius() {
...
+     * Sets the country code attribute of the {@param metacard} when the {@param metacard} has a
+     * location and there is not already a country code attribute. Does not update the country code
+     * if the country code attribute is already set.
+     *
+     * @param metacard
+     */
+    private void setCountryCode(Metacard metacard) {
+        Optional<String> wktLocation = Optional.ofNullable(metacard.getLocation());
+
+        if (wktLocation.isPresent() && hasCountryCode(metacard)) {
```
beyelerb: I think I disagree with naming the method hasCountryCode - now it's confusing - hasCountryCode returns true when it doesn't have a country code attribute. Plus the description says it does not update the cc if the cc attribute is set, but this if statement says if the wkt is present and the metacard has a country code. It works but doesn't make sense trying to read it.

catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -118,11 +106,38 @@ public int getRadius() {
...
+     * location and there is not already a country code attribute. Does not update the country code
+     * if the country code attribute is already set.
+     *
+     * @param metacard
+     */
+    private void setCountryCode(Metacard metacard) {
+        Optional<String> wktLocation = Optional.ofNullable(metacard.getLocation());
+
+        if (wktLocation.isPresent() && !hasCountryCode(metacard)) {
+            GeoCoder geocoder = geoCoderFactory.getService();
```
kcwire: Is there any reason we need to get the GeoCoder for each metacard? I would think obtaining to once for the entire list would be more efficient.

catalog/spatial/geocoding/spatial-geocoding-plugin/src/main/java/org/codice/ddf/spatial/geocoding/plugin/GeoCoderPlugin.java
```diff
@@ -118,11 +110,33 @@ public int getRadius() {
...
+        if (geoCoder != null && wktLocation.isPresent() && !hasCountryCode(metacard)) {
+            Optional<String> alpha3CountryCode = geoCoder.getCountryCode(wktLocation.get(),
+                    radiusInKm);
+
+            geoCoder.getCountryCode(wktLocation.get(), radiusInKm)
+                    .ifPresent(countryCode -> {
+                        metacard.setAttribute(new AttributeImpl(Location.COUNTRY_CODE,
+                                alpha3CountryCode.get()));
+
+                        LOGGER.trace("Setting metacard country code to {} for metacard with id {}",
```
beyelerb: Could you move the trace statement before the operation it is describing? That way if it fails, the log will show what is trying to be done when it fails. Otherwise, it will only print out if it was successful.

PR #1342, opened 2016-10-11
DDF-2528: Adding AttributeValidator for Location.COUNTRY_CODE

brjeter:
#### What does this PR do?
This PR adds an attribute validator to verify that the Location.COUNTRY_CODE attribute is a valid ISO 3166-1 alpha-3 country code.
@glenhein @jrnorth @peterhuffer 
@kcwire 
@millerw8
- Run a full build.
- Verify all tests pass (specifically the ValidationParserSpecTest and CountryCodeValidatorTest)
- Test ingestion of a record with an invalid location.country-code:
  - Run local DDF instance.
  - Configure `Metacard Validation Filter Plugin` to not filter metacards with errors/warnings.
  - Configure `Catalog Federation Strategy` to show validation errors.
  - Ingest a record with an invalid `location.country-code` attribute value.
  - Search for the record and verify there is an error message under `Validation Errors` under the `Details` tab.
- Test ingestion of a record with a valid location.country-code:
  - See above, but verify there are no error messages under `Validation Errors` on the metacard.

The Location.COUNTRY_CODE attribute is now populated by the GeoCoderPlugin. This validation has been added to verify the attribute is populated correctly. See https://github.com/codice/ddf/pull/1171

https://codice.atlassian.net/browse/DDF-2528
#### Checklist:
- [x] Documentation Updated
	- [ ] Change log Updated
- [X] Update / Add Unit Tests


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
```diff
@@ -0,0 +1,103 @@
...
+     * {@inheritDoc}
+     * <p>
+     * Validates only the values of {@code attribute} that are {@link String}s.
+     */
+    @Override
+    public Optional<AttributeValidationReport> validate(Attribute attribute) {
+        Preconditions.checkArgument(attribute != null, "The attribute cannot be null.");
+
+        final String name = attribute.getName();
+        for (final Serializable value : attribute.getValues()) {
```
brjeter: I like it. Fixed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
```diff
@@ -0,0 +1,103 @@
...
+                countryCodesMap.forEach(report::addSuggestedValue);
+                return Optional.of(report);
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private void setupISO3CountryCodes() {
+        countryCodesMap = new HashSet<String>();
```
brjeter: Fixed it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
coyotesqrl: One additional thing you could do, to clean up that lambda in the middle, is extract a new private method and then invoke it with `this::newMethod`. You'd have to add an additional map call:
```java
return attribute.getValues().stream()
        .filter(String.class::isInstance)
        .map(String.class::cast)
        .filter(s -> !countryCodesMap.contains(s))
        .map(s -> name)
        .map(this::newMethod)
        .map(AttributeValidationReport.class::cast)
        .findFirst();
```

It has two advantages: firstly it makes the streaming code cleaner; secondly, it makes it more explicit that the value is being dropped on the floor after it's been determined it is not in the `countryCodesMap`.

brjeter: The trick is that the code in there requires the `name` variable for the error message, and I won't have access to that local variable in the private function. I'm not sure how to get around that other than setting the `name` as a field of the class which I'm not a huge fan of doing in this situation.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
```diff
@@ -0,0 +1,104 @@
...
+     * <p>
+     * Validates only the values of {@code attribute} that are {@link String}s.
+     */
+    @Override
+    public Optional<AttributeValidationReport> validate(Attribute attribute) {
+        Preconditions.checkArgument(attribute != null, "The attribute cannot be null.");
+
+        final String name = attribute.getName();
+
+        return attribute.getValues().stream()
```
rzwiefel: Don't think your formatter is right here. Might wanna check you have it installed and make sure this file is formatted.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
brjeter: Oops, I didn't catch that. But I was just realizing as I was changing the code that I actually need access to the `name` and the `String` value in order to print a more meaningful message in the report. Something like
```
report.addViolation(new ValidationViolationImpl(Collections.singleton(name),
                s + "is not a valid ISO_3166-1 Alpha3 country code.",
                ValidationViolation.Severity.ERROR));
```

coyotesqrl: Cool. Then you don't want that map operation, definitely.

You still might consider extracting a method like `AttributeValidationReport newMethod(String name, String countryCode)` to make the code cleaner, but otherwise it's good with me.

brjeter: Good feedback, thanks. I extracted the method out.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
brjeter: The files have been formatted

catalog/core/catalog-core-validationparser/src/main/java/ddf/catalog/validation/impl/ValidationParser.java
```diff
@@ -329,6 +330,9 @@ private AttributeValidator getValidator(Outer.Validator validator) {
             }
             return new RangeValidator(min, max);
         }
+        case "country": {
```
jrnorth: How about calling this `iso3_country` or something similar? We might want to add other types of country codes in the future.

catalog/core/catalog-core-validationparser/src/test/groovy/ddf/catalog/validation/impl/ValidationParserSpecTest.groovy
```diff
@@ -131,7 +131,7 @@ class ValidationParserSpecTest extends Specification {
         attributeRegistry.lookup(attribute2Name).isPresent()
 
         and: "the two attribute validators are registered in the attribute validator registry"
```
jrnorth: The messages for the assertions you changed need to be updated.

jrnorth:
You'll need to update the JSON validator documentation to include the new country validator.

Also it looks like there's an empty file called "list" in this PR that should be removed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void setupISO3CountryCodes() {
+        countryCodes = Arrays.stream(Locale.getISOCountries())
+                .map(iso2CC -> new Locale("", iso2CC))
+                .map(Locale::getISO3Country)
+                .collect(Collectors.toSet());
+    }
+
+    @Override
```
millerw8: any reason we need to override equals/hashcode?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
```diff
@@ -0,0 +1,107 @@
...
+
+        final String attributeName = attribute.getName();
+
+        return attribute.getValues()
+                .stream()
+                .filter(String.class::isInstance)
+                .map(String.class::cast)
+                .filter(s -> !countryCodes.contains(s))
+                .map(s -> this.buildReport(attributeName, s))
+                .findFirst();
```
millerw8: Does this mean if we are validating location.country-code which I believe is multi-valued, and we enter CSWD and WFDA as two values, we will only get warned on one of them?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
brjeter: I implemented them because other AttributeValidator implementations overrode them. I assumed the decision was made to do this for all validators for a particular reason. Perhaps to programmatically tell the difference between AttributeValidators without knowing the implementations.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
brjeter: Yes that's correct. Should it be behaving differently?

millerw8: Yeah, I think aggregating all the invalid value warnings would be ideal.

emmberk:
Successfully completed mvn clean install :+1:

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/CountryCodeValidator.java
```diff
@@ -0,0 +1,107 @@
...
+                countryCodeValue + " is not a valid ISO_3166-1 Alpha3 country code.",
+                ValidationViolation.Severity.ERROR));
+
+        //TODO: Should the country codes be included in the error report?
+        countryCodes.forEach(report::addSuggestedValue);
+        return report;
+    }
+
+    private void setupISO3CountryCodes() {
+        countryCodes = Arrays.stream(Locale.getISOCountries())
```
brjeter: Nope, that's good feedback. Were you thinking a static block to initialize the `countryCodes`?

rzwiefel: You don't even need a static block.
Could just change line 41 to be
```java
private static final Set<String> countryCodes = Arrays.stream(Locale.getISOCountries())
    .map(iso2CC -> new Locale("", iso2CC))
    .map(Locale::getISO3Country)
    .collect(Collectors.toSet());
```

rzwiefel: Unless you think a static block will look/read cleaner. Up to you

distribution/docs/src/main/resources/_contents/_catalog-contents/integrating-catalog-contents.adoc
```diff
@@ -1730,7 +1730,8 @@ The `validator` key must have a value of one of the following:
  * `arguments`: (unlimited) [list of strings: each argument is one case-sensitive, valid enumeration value]
   - `enumerationignorecase`
   * `arguments`: (unlimited) [list of strings: each argument is one case-insensitive, valid enumeration value]
-
+ - `iso3_country` (validates that a String is a valid ISO 3166-1 alpha-3 country code)
```
kcwire: It appears we have an ignoreCase validator but it is not installed by default or documented.

ricklarsen:
Should this get a mention in the change log?

catalog/plugin/catalog-plugin-metacard-validation/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -61,7 +61,7 @@
 
 
     <!-- Register in the OSGi Service Registry -->
-    <service ref="pre-ingest-plugin" interface="ddf.catalog.plugin.PreIngestPlugin"/>
+    <service ref="pre-ingest-plugin" interface="ddf.catalog.plugin.PreIngestPlugin" ranking="-1"/>
```
jrnorth: Why was this changed to -1?

catalog/core/catalog-core-validationparser/src/test/groovy/ddf/catalog/validation/impl/ValidationParserSpecTest.groovy
```diff
@@ -130,8 +130,8 @@ class ValidationParserSpecTest extends Specification {
         attributeRegistry.lookup(attribute1Name).isPresent()
         attributeRegistry.lookup(attribute2Name).isPresent()
 
-        and: "the two attribute validators are registered in the attribute validator registry"
-        attributeValidatorRegistry.getValidators(attribute1Name).size() == 2
+        and: "the three attribute validators are registered in the attribute validator registry"
```
jrnorth: Need to update these descriptions.

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/validation/impl/ISO3CountryCodeValidatorTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @Test
+    public void testInvalidISO3CountryCodesIgnoreCase() {
+        validateWithErrors(new AttributeImpl("location.country_code",
+                Arrays.asList(new String[] {"zzz", "abc"})), 2, false);
+    }
+
+    private void validateNoErrors(final Attribute attribute, boolean ignoreCase) {
+        ISO3CountryCodeValidator validator = new ISO3CountryCodeValidator(ignoreCase);
+        final Optional<AttributeValidationReport> reportOptional = validator.validate(attribute);
+        assertThat(reportOptional.isPresent(), CoreMatchers.is(false));
```
jrnorth: You can just use the `is()` method you statically imported.

catalog/plugin/catalog-plugin-metacard-validation/src/main/resources/OSGI-INF/blueprint/blueprint.xml
brjeter: Talked with Keith about this. We wanted the `MetacardValidityMarkerPlugin` to be the last pre-ingest plugin to process just in case any pre-ingest plugins modify/add attributes to the metacard. In fact, part of the reason for this PR was because the GeoCoder plugin was added as a pre-ingest plugin that adds the `location.country-code` attribute to the metacard but this was happening after validation which defeated the purpose.

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/validation/impl/ISO3CountryCodeValidatorTest.java
brjeter: Fixed.

PR #1371, opened 2016-10-24
[2.9.x] DDF-2520: Protects FileSystemStorageProvider from null resource URIs

rzwiefel:
#### What does this PR do?

see title.

@jrnorth 
@peterhuffer 

@lessarderic
@stustison

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/operation/impl/ReadStorageResponseImpl.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/operation/impl/ReadStorageResponseImpl.java
@@ -16,20 +16,31 @@ package ddf.catalog.content.operation.impl;
 import java.io.Serializable;
 import java.util.Map;
 
+import javax.annotation.Nullable;
+
 import ddf.catalog.content.data.ContentItem;
 import ddf.catalog.content.operation.ReadStorageRequest;
 import ddf.catalog.content.operation.ReadStorageResponse;
 import ddf.catalog.content.operation.StorageRequest;
 import ddf.catalog.operation.impl.ResponseImpl;
 
-public class ReadStorageResponseImpl extends ResponseImpl<ReadStorageRequest> implements
-        ReadStorageResponse {
+public class ReadStorageResponseImpl extends ResponseImpl<ReadStorageRequest>
+        implements ReadStorageResponse {
     private ContentItem contentItem;
+
+    /**
+     * Instantiates an empty, new ResponseImpl
+     *
+     * @param request the original request
+     */
+    public ReadStorageResponseImpl(ReadStorageRequest request) {
+        super(request, null);
+    }
 
     /**
      * Instantiates a new ResponseImpl
      *
-     * @param request    - the original request
+     * @param request the original request
      * @param properties
      */
     public ReadStorageResponseImpl(ReadStorageRequest request,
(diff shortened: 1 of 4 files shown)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/operation/impl/ReadStorageResponseImpl.java
```diff
@@ -27,6 +27,15 @@
     private ContentItem contentItem;
 
     /**
+     * Instantiates an empty, new ResponseImpl
+     *
+     * @param request    - the original request
```
lessarderic: Nit: You don't need to put a `-` between the parameter name and the description. The javadoc tool takes care of that for you.

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -142,6 +142,10 @@ public CreateStorageResponse create(CreateStorageRequest createRequest)
     public ReadStorageResponse read(ReadStorageRequest readRequest) throws StorageException {
         LOGGER.trace("ENTERING: read");
 
+        if (readRequest.getResourceUri() == null) {
+            return new ReadStorageResponseImpl(readRequest);
```
lessarderic: This `ReadStorageResponseImpl` object will contain a `null` item, which means that its `getContentItem()` will return `null` and could break existing code.

Fortunately, the interface that returns this is marked experimental so this isn't an API change and as long as we know any of our code isn't going to break because of this change, it should be ok to simply add `@Nullable` on the `getContentItem()` return type.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/operation/impl/ReadStorageResponseImpl.java
rzwiefel: I just copied it from below, but also good to know 👍

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
rzwiefel: Whats funny is in my catalog UI code (in the other fork) i checked and somehow I'm already checking if this is null... probably because I don't trust our API's. But hoping everyone else did the same 😄 
aka, I will investigate to make sure none of that nonsense is occuring in here or "in the wild".

rzwiefel: Added nullable. 👌

rzwiefel: Checked DDF and a little bit of downstream code and didn't find anybody really using it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/operation/impl/ReadStorageResponseImpl.java
rzwiefel: fixed!

rzwiefel:
all comments addressed

rzwiefel:
Full build succeeded

PR #1379, opened 2016-10-26
DDF-2556: Corrected caching logic to clone metacards on cache ingest

jlcsmith:
#### What does this PR do?
- Added defensive copy of metacards on CacheBulkProcessor.add()
- Added .equals() to MetacardImpl and fixed .hashCode()
- Updated MetacardImpl to use ConcurrentHashMap

@bdeining @rzwiefel @lcrosenbu @kcwire 

@pklinef
@coyotesqrl 

Full clean build - verify unit and integration tests pass.

The issue was originally triggered with the Catalog UI.  
- Add the Catalog UI app
- add a federated source (CSW loopback for example)
- ingest some data
- perform a federated search from the Catalog UI.  
- Verify no caching exceptions in the log. 
- Verify the cache is populated (search --cache *)

https://codice.atlassian.net/browse/DDF-2556
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
@@ -26,7 +26,10 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.slf4j.Logger;
@@ -110,7 +113,7 @@ public class MetacardImpl implements Metacard {
          * serialized object is maintained. For instance, if a null check is added in the
          * constructor, the same check should be added in the readObject() method.
          */
-        map = new HashMap<String, Attribute>();
+        map = new HashMap<>();
         if (type != null) {
             this.type = type;
         } else {
@@ -158,13 +161,20 @@ public class MetacardImpl implements Metacard {
                     MetacardType.class.getName() + " instance should not be null.");
         }
         map = new HashMap<>();
-        for (AttributeDescriptor descriptor : metacard.getMetacardType().getAttributeDescriptors()) {
-            Attribute metacardAttribute = metacard.getAttribute(descriptor.getName());
-            if (metacardAttribute == null || metacardAttribute.getValue() == null) {
-                continue;
-            }
-            map.put(descriptor.getName(), metacardAttribute);
+
+        if (metacard.getSourceId() != null) {
+            setSourceId(metacard.getSourceId());
         }
+
+        map = metacard.getMetacardType()
+                .getAttributeDescriptors()
+                .stream()
+                .filter(Objects::nonNull)
+                .filter(attributeDescriptor -> attributeDescriptor.getName() != null)
+                .filter(attributeDescriptor -> metacard.getAttribute(attributeDescriptor.getName())
+                        != null)
+                .map(AttributeDescriptor::getName)
+                .collect(Collectors.toMap(Function.identity(), metacard::getAttribute));
     }
 
     @Override
@@ -555,9 +565,9 @@ public class MetacardImpl implements Metacard {
 
         Attribute attribute = getAttribute(attributeName);
 
-        if (attribute == null) {
+        if (attribute == null || attribute.getValue() == null) {
             if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug("Attribute {} was not found, returning null", attributeName);
+                LOGGER.debug("Attribute {} was not found or value was missing.  Returning null.", attributeName);
             }
             return null;
         }
@@ -816,7 +826,7 @@ public class MetacardImpl implements Metacard {
          */
         stream.defaultReadObject();
 
-        map = new HashMap<String, Attribute>();
+        map = new HashMap<>();
 
         wrappedMetacard = null;
 
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
@@ -26,10 +26,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Collectors;
 
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.slf4j.Logger;
@@ -166,15 +163,13 @@ public class MetacardImpl implements Metacard {
             setSourceId(metacard.getSourceId());
         }
 
-        map = metacard.getMetacardType()
-                .getAttributeDescriptors()
-                .stream()
-                .filter(Objects::nonNull)
-                .filter(attributeDescriptor -> attributeDescriptor.getName() != null)
-                .filter(attributeDescriptor -> metacard.getAttribute(attributeDescriptor.getName())
-                        != null)
-                .map(AttributeDescriptor::getName)
-                .collect(Collectors.toConcurrentMap(Function.identity(), metacard::getAttribute));
+        map = new HashMap<>();
+        for (AttributeDescriptor attribute : metacard.getMetacardType()
+                .getAttributeDescriptors()) {
+            if(attribute != null && attribute.getName() != null) {
+                map.put(attribute.getName(), metacard.getAttribute(attribute.getName()));
+            }
+        }
     }
 
     @Override
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
@@ -174,7 +174,7 @@ public class MetacardImpl implements Metacard {
                 .filter(attributeDescriptor -> metacard.getAttribute(attributeDescriptor.getName())
                         != null)
                 .map(AttributeDescriptor::getName)
-                .collect(Collectors.toMap(Function.identity(), metacard::getAttribute));
+                .collect(Collectors.toConcurrentMap(Function.identity(), metacard::getAttribute));
     }
 
     @Override
--- a/catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/adapter/AdaptedMetacard.java
+++ b/catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/adapter/AdaptedMetacard.java
@@ -15,6 +15,7 @@ package ddf.catalog.transformer.xml.adapter;
 
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 
@@ -188,11 +189,11 @@ public class AdaptedMetacard implements Metacard {
         if (Metacard.ID.equals(name)) {
             return this.id;
         }
-        for (Attribute attribute : attributes) {
-            if (attribute == null || StringUtils.isEmpty(attribute.getName())) {
-                continue;
-            } else if (name.equals(attribute.getName())) {
-                return attribute;
+        for (Attribute attribute : Collections.unmodifiableList(attributes)) {
+            if (attribute != null && !StringUtils.isEmpty(attribute.getName())) {
+                if (name.equals(attribute.getName())) {
+                    return attribute;
+                }
             }
         }
         return null;
--- a/catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/adapter/AdaptedMetacard.java
+++ b/catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/adapter/AdaptedMetacard.java
@@ -15,7 +15,7 @@ package ddf.catalog.transformer.xml.adapter;
 
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 
@@ -189,7 +189,7 @@ public class AdaptedMetacard implements Metacard {
         if (Metacard.ID.equals(name)) {
             return this.id;
         }
-        for (Attribute attribute : Collections.unmodifiableList(attributes)) {
+        for (Attribute attribute : Arrays.asList((attributes.toArray(new Attribute[attributes.size()])))) {
             if (attribute != null && !StringUtils.isEmpty(attribute.getName())) {
                 if (name.equals(attribute.getName())) {
                     return attribute;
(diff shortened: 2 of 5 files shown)


pklinef:
I am concerned that this fix only works if `MetacardImpl` is used.  It is possible for a source to return its own implementation of the `Metacard` interface.  Though it might impact performance, I think it would be safer to do a defensive copy in the federation strategy before submitting the caching job.

https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java#L552-L561

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -843,15 +858,60 @@ private void readObject(ObjectInputStream stream) throws IOException, ClassNotFo
 
     }
 
+    @Override
+    public boolean equals(Object o) {
```
rzwiefel: Did we need this for something specifically? I think before we almost updated it but didnt' because we cant' guarantee that all metacard implementations will be using our `MetacardImpl` as its base, so using equals to compare two metacards is not deterministic.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -156,10 +159,22 @@ public MetacardImpl(Metacard metacard, MetacardType type) {
...
+        if (metacard.getSourceId() != null) {
+            setSourceId(metacard.getSourceId());
         }
+
+        map = new ConcurrentHashMap<>();
+
+        metacard.getMetacardType()
+                .getAttributeDescriptors()
+                .stream()
+                .filter(attributeDescriptor -> attributeDescriptor != null)
```
rzwiefel: `.filter(Objects::nonNull)`

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -110,7 +113,7 @@ public MetacardImpl(MetacardType type) {
          * serialized object is maintained. For instance, if a null check is added in the
          * constructor, the same check should be added in the readObject() method.
          */
-        map = new HashMap<String, Attribute>();
+        map = new ConcurrentHashMap<>();
```
rzwiefel: Just to be sure, Can we not guard the place that is concurrently accessing this over using a concurrent hashmap? These internals are *really* heavily accessed, so this could bring performance implications.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -156,10 +159,22 @@ public MetacardImpl(Metacard metacard, MetacardType type) {
...
+
+        metacard.getMetacardType()
+                .getAttributeDescriptors()
+                .stream()
+                .filter(attributeDescriptor -> attributeDescriptor != null)
+                .filter(attributeDescriptor -> attributeDescriptor.getName() != null)
+                .filter(attributeDescriptor -> metacard.getAttribute(attributeDescriptor.getName())
+                        != null)
+                .forEach(attributeDescriptor -> map.put(attributeDescriptor.getName(),
+                        metacard.getAttribute(attributeDescriptor.getName())));
```
rzwiefel: instead of a `.forEach`, why not do a collect and set whole container to the result?
eg
`map = ...
  .map(AttributeDescriptor::getName)
  .collect(Collectors.toMap(Function.identity(), metacard::getAttribute));

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
rzwiefel: Oh and looks like phil beat me to this comment above.
https://github.com/codice/ddf/pull/1379#issuecomment-256426288

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -550,9 +565,9 @@ public void setSecurity(HashMap<String, List<String>> security) {
 
         Attribute attribute = getAttribute(attributeName);
 
-        if (attribute == null) {
+        if (attribute == null || attribute.getValue() == null) {
             if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug("Attribute {} was not found, returning null", attributeName);
+                LOGGER.debug("Attribute " + attributeName + " was not found, returning null");
```
jlcsmith: I have no idea why I did that :)  Yeah I'll fix that

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
coyotesqrl: I'm not clear how this is helping to solve the underlying problem. All this concurrent map is going to do is prevent multiple threads from colliding with each other as they update the internals of this metacard, but it won't help with the case where the metacard is being modified after being added to the `CBP`.

jlcsmith: Correct.  I noticed concurrent modification exceptions after I made the change to the CacheBulkProcessor which led me to this change.

coyotesqrl: Those will probably be resolved if you follow @rzwiefel's advice to do a single collect operation as opposed to the `forEach` processing.

rzwiefel: Except that is in the constructor, so the underlying map that is being used to build this metacard is (maybe?) the one being changed? If the calling code (like phil pointed out) is explicitly running two things concurrently on the metacard, it would just be safer to have the copy happen in the cachingfederation strategy. That way you're purposely creating a safe copy for this concurrent process.

<sidenote> 
_**Sigh**_, this is fueling my desire to do a weekend project and make an `ImmutableMetacardImpl`.. that is actually immutable but still has our wrapping semantics.
</sidenote>

coyotesqrl: Considering our wrapping semantics have some fundamental issues to be addressed, you should think about relaxing on your weekend instead. :)

rzwiefel: Fundamental issues, yes. But I would think we'd need wrapping semantics to allow people to start using something like them instead of regular MetacardImpl... I guess we could leave it out and say its a tradeoff ¯\_(ツ)_/¯. sorry for turning your PR into a random idea ground @jlcsmith  XD

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
jlcsmith: Yeah we might not need this after all.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
jlcsmith: went back to Hashmap

jlcsmith: Once I fixed the streaming logic in the constructor, I was able to do away with ConcurrentHashMap

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -156,10 +159,20 @@ public MetacardImpl(Metacard metacard, MetacardType type) {
...
         }
+
+        map = metacard.getMetacardType()
+                .getAttributeDescriptors()
+                .stream()
+                .filter(Objects::nonNull)
+                .filter(attributeDescriptor -> attributeDescriptor.getName() != null)
+                .filter(attributeDescriptor -> metacard.getAttribute(attributeDescriptor.getName())
+                        != null)
+                .map(AttributeDescriptor::getName)
```
coyotesqrl: This is cleaner, but could be even a bit better by moving this `map` up two lines:

```java
//...
.filter(Objects::nonNull)
.map(AttributeDescriptor::getName)
.filter(Objects::nonNull)
.filter(attrName -> metacard.getAttribute(attrName) != null)
.collect(//...
```

jlcsmith:
Still seeing ConcurrentModificationExceptions in the MetacardImpl constructor.  I am switching back to using a ConcurrentHashMap to protect the internal map.

coyotesqrl:
I still don't see how that could be occurring in the ctor, as nothing else should be accessing that map and the same time and there is nothing unusual about the way the map is being modified.

Switching to a concurrent map is going to have a performance impact across the entire application and we really should avoid doing so unless there is an clearly understood and compelling reason to do so.

jlcsmith:
@coyotesqrl - it appears that are cases where multiple threads are reading from that internal map while it is being modified - for instances here https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java#L677.

coyotesqrl:
We have a bigger problem if there are threads attempting to read the contents of an object prior to the object being completely constructed.

catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/adapter/AdaptedMetacard.java
```diff
@@ -188,11 +189,11 @@ public final Attribute getAttribute(String name) {
         if (Metacard.ID.equals(name)) {
             return this.id;
         }
-        for (Attribute attribute : attributes) {
-            if (attribute == null || StringUtils.isEmpty(attribute.getName())) {
-                continue;
-            } else if (name.equals(attribute.getName())) {
-                return attribute;
+        for (Attribute attribute : Collections.unmodifiableList(attributes)) {
```
coyotesqrl: To avoid `ConcurrentModificationExceptions` caused by other threads modifying the attributes on this metacard while it is being iterated.

jrnorth: But `unmodifiableList()` only provides an unmodifiable view of the list. Couldn't another thread modify the backing list while it's being iterated here?

coyotesqrl: Ugh. Right you are. This is still not a sufficient guard. This mutable list is incredibly painful.

rzwiefel: I think we need to be doing this copy/protection higher up in the call chain. Before we actually have to worry about concurrent modifications.

coyotesqrl: Joe is, of course, correct. The change from yesterday is still not sufficient.

Wrapping with `Collections.unmodifiableList` isn't going to be enough. Making a defensive copy in the loop, while a more expensive option, is a safer solution.

```java
for (Attribute attribute : new ArrayList<>(attributes)) {
```

coyotesqrl: One more time. Sorry Jason.

After a deeper dive, there is lingering concern that the defensive copy above itself could still produce a `ConcurrentModificationException`. That appears not to be a problem as of Java 6, but I'm still leery of it. A guaranteed defensive copy can be gotten by changing the line to

```java
for (Attribute attribute: Arrays.asList(attributes.toArray())) {
```

However, there are issues with the defensive copy regardless of which way it is achieved. Most notably, if the collection *is* changed from beneath this thread, the attributes being iterated will not represent the actual state of the metacard. A comment should be added indicating this small but non-zero concern. This issue, notably, would still exist with a Concurrent collection; the use of that would only protect against CMEs.

The only way, short of an immutable implementation, of resolving this completely would be to synchronize all access to the attribute list and impacting performance; however, as it leaks out for jaxb marshalling, that isn't really feasible.

jlcsmith: Understood Richard.  I'd love to see us get away from jaxb anyways.. making an immutable metacard implementation more realistic.

catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/adapter/AdaptedMetacard.java
```diff
@@ -188,11 +189,11 @@ public final Attribute getAttribute(String name) {
         if (Metacard.ID.equals(name)) {
             return this.id;
         }
-        for (Attribute attribute : attributes) {
-            if (attribute == null || StringUtils.isEmpty(attribute.getName())) {
-                continue;
-            } else if (name.equals(attribute.getName())) {
-                return attribute;
+        for (Attribute attribute : Arrays.asList((attributes.toArray(new Attribute[attributes.size()])))) {
```
coyotesqrl: It's not necessary to provide the size to the type-hinting array you pass in, so it's just as correct (and a little cheaper) to implement this as 
`for (Attribute attribute : Arrays.asList((attributes.toArray(new Attribute[0])))) {`

Otherwise, this looks great.

jrnorth: Also the `asList()` isn't necessary since you can just iterate over the array directly.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -156,9 +156,17 @@ public MetacardImpl(Metacard metacard, MetacardType type) {
             throw new IllegalArgumentException(
                     MetacardType.class.getName() + " instance should not be null.");
         }
+
+        if (metacard.getSourceId() != null) {
```
pklinef: It looks like this should be added to `MetacardImpl(Metacard metacard)` as well so both constructors are consistent.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -156,9 +156,17 @@ public MetacardImpl(Metacard metacard, MetacardType type) {
...
+            setSourceId(metacard.getSourceId());
+        }
+
         map = new HashMap<>();
-        for (AttributeDescriptor attribute : metacard.getMetacardType().getAttributeDescriptors()) {
-            map.put(attribute.getName(), metacard.getAttribute(attribute.getName()));
+        for (AttributeDescriptor attribute : metacard.getMetacardType()
```
pklinef: Why does this constructor try to iterate over the attributes instead of using `wrappedMetacard` like the other constructor?  Seems like that was inconsistent before this PR.

coyotesqrl:
@jlcsmith You've merge conflicts on this PR.

PR #1405, opened 2016-11-08
DDF-2583: Added user bind methods to LDAP.

djblue:
#### What does this PR do?

Adds user bind methods to LDAP.

#### Who is reviewing it?

@adimka 
@garrettfreibott 
@tbatie 

@coyotesqrl 
@stustison

[DDF-2583](https://codice.atlassian.net/browse/DDF-2583)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/platform/security/sts/security-sts-ldapclaimshandler/src/main/java/ddf/security/sts/claimsHandler/LdapClaimsHandler.java
+++ b/platform/security/sts/security-sts-ldapclaimshandler/src/main/java/ddf/security/sts/claimsHandler/LdapClaimsHandler.java
@@ -32,6 +32,8 @@ import org.forgerock.opendj.ldap.LDAPConnectionFactory;
 import org.forgerock.opendj.ldap.LdapException;
 import org.forgerock.opendj.ldap.SearchResultReferenceIOException;
 import org.forgerock.opendj.ldap.SearchScope;
+import org.forgerock.opendj.ldap.requests.BindRequest;
+import org.forgerock.opendj.ldap.requests.Requests;
 import org.forgerock.opendj.ldap.responses.BindResult;
 import org.forgerock.opendj.ldap.responses.SearchResultEntry;
 import org.forgerock.opendj.ldif.ConnectionEntryReader;
@@ -50,6 +52,8 @@ public class LdapClaimsHandler extends org.apache.cxf.sts.claims.LdapClaimsHandl
     private String bindUserCredentials;
 
     private String bindUserDN;
+
+    private String bindMethod;
 
     private boolean overrideCertDn = false;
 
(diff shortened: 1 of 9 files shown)


platform/security/sts/security-sts-ldapclaimshandler/src/main/java/ddf/security/sts/claimsHandler/RoleClaimsHandler.java
```diff
@@ -207,7 +215,34 @@ public ProcessedClaimCollection retrieveClaimValues(ClaimCollection claims,
 
             connection = connectionFactory.getConnection();
             if (connection != null) {
-                BindResult bindResult = connection.bind(bindUserDN, bindUserCredentials.toCharArray());
+                BindRequest request;
+                switch (bindMethod) {
+                case "Simple":
```
rzwiefel: @stustison Case sensitive admin config options make me shudder. Any reason these need to be case sensitive options (regardless if this is a human typing the value or a wizard). [I'm thinking like the setup of idp.. remembering if you need to do upper or lower IDP is a pain]

platform/security/sts/security-sts-ldapclaimshandler/src/main/java/ddf/security/sts/claimsHandler/LdapClaimsHandler.java
```diff
@@ -112,8 +116,34 @@ public ProcessedClaimCollection retrieveClaimValues(ClaimCollection claims,
 
             connection = connectionFactory.getConnection();
             if (connection != null) {
-                BindResult bindResult = connection.bind(bindUserDN,
-                        bindUserCredentials.toCharArray());
+                BindRequest request;
```
coyotesqrl: An optional way to shorten this a little would be to create a static map in class initialization like this...
```java
            Map<String, Optional<BiFunction<String, char[], BindRequest>>> lookup = ImmutableMap.of(
                    "Simple",
                    Optional.of(Requests::newSimpleBindRequest),
                    "SASL",
                    Optional.of(Requests::newPlainSASLBindRequest),
//...
            );
```
Then replace the whole switch statement with 
```java
                BindRequest request = lookup.get(bindMethod)
                        .orElse(Requests::newSimpleBindRequest)
                        .apply(bindUserDN, bindUserCredentials.toCharArray());
```

coyotesqrl: Regardless, it would be best if the duplicate code blocks here and in `RoleClaimsHandler` (and if possible) `SslLdapLoginModule` could be extracted out to a common utility method.

rzwiefel: regardless of what happens i vote for getting rid of case sensitivity since it makes no impact in this case. SIMPLE, Simple, or SiMpLe, it aint gonna matter and is confusing to configurators.

platform/security/sts/security-sts-ldapclaimshandler/src/main/java/ddf/security/sts/claimsHandler/RoleClaimsHandler.java
rzwiefel: Or to developers writing wizards for said options.

tbatie:
docs need to be updated

platform/security/sts/security-sts-ldapclaimshandler/src/main/java/ddf/security/sts/claimsHandler/RoleClaimsHandler.java
stustison: No reason that I can think of, that is just how I did it in this patch. If people want to change it, that is totally fine with me.

stustison:
The default configurations that point to the embedded LDAP need to be updated to have the new field.

PR #1404, opened 2016-11-08
DDF-2581: TestFederation testAsyncDownloadActionPresentUsingCometDClient fails intermittently

lessarderic:
#### What does this PR do?
Added the ability to mark specific unit tests as unstable and be able to enable or disable them using a system property.

Updated the README.md file to include this new flag.

Also marked `TestFederation.testAsyncDownloadActionPresentUsingCometDClient()` as unstable until the root cause of the intermittent failures is addressed in a follow on PR.

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@brendan-hofmann 
@tbatie 

#### Select at least one member from relevant component team(s) from below (at least one component team member needs to approve the PR).
[Build](https://github.com/orgs/codice/teams/build)

#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@coyotesqrl
@pklinef 

#### How should this be tested? (List steps with links to updated documentation)
Run integration tests with and without the `-DincludeUnstableTests=true` flag and make sure the unstable test is ran or skipped.

#### What are the relevant tickets?
[DDF-2581](https://codice.atlassian.net/browse/DDF-2581)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests
- [x] Update / Add Integration Tests

PR #1409, opened 2016-11-09
DDF-2581: TestFederation testAsyncDownloadActionPresentUsingCometDClient fails intermittently

lessarderic:
#### What does this PR do?
Replaced `unstableTest()` method with jUnit Rule and annotation.

@brendan-hofmann 
@tbatie 

@coyotesqrl
@pklinef

Run the integration tests with and without the `-DincludeUnstableTests` and make sure `testAsyncDownloadActionPresentUsingCometDClient` is run or skipped based on the system property.

This is a follow up of https://github.com/codice/ddf/pull/1404

[DDF-2581](https://codice.atlassian.net/browse/DDF-2581)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/ConditionalIgnoreTest.java
+++ b/distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/ConditionalIgnoreTest.java
@@ -1,0 +1,115 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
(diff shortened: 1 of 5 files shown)


lessarderic:
The change was based on a comment from @pklinef on my previous PR and something I had myself found that looked like a better approach. We could always have a specific annotation (e.g., @Unstable) instead of the generic conditional one if people feel that it would be a better option.

And yes, I did think about adding unit tests for the annotation but wanted to get this in sooner so people could start using it. I'll try to add something real quick since you've mentioned it :-)

distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/ConditionalIgnoreTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.codice.ddf.itests.common.annotations.SkipUnstableTest;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.JUnitCore;
+import org.junit.runner.Result;
+import org.junit.runner.notification.Failure;
+import org.junit.runner.notification.RunListener;
+
+public class ConditionalIgnoreTest {
```
tbatie: This may be overkill/nit but it is possible to test the annotation in a pax exam environment in a unit test similar to the way we test our other annotations https://github.com/codice/ddf/blob/master/distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/PaxExamRuleTest.java#L42.

lessarderic:
Will add a note to the ticket to consider adding a PaxExam test for this.

PR #1415, opened 2016-11-10
DDF-2520: Fixes windows path issue in storage provider

rzwiefel:
#### What does this PR do?
Windows does not allow ":" in a file path. We assumed that we could translate the URI in storage provider to a native file path. This is not always the case. We now validate each URI to make sure it conforms to our storage provider spec.

anyone

@jlcsmith 
@kcwire

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java
@@ -13,8 +13,6 @@
  **/
 package ddf.catalog.content.data.impl;
 
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringUtils;
@@ -26,35 +24,52 @@ import ddf.catalog.content.data.ContentItem;
  * required by the {@link ddf.catalog.content.StorageProvider}.
  */
 public class ContentItemValidator {
-
     private static final Pattern QUALIFIER_PATTERN = Pattern.compile("\\w[a-zA-Z0-9_\\-]+");
+
+    // https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29
+    // xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
+    //    8      4     3    3       12
+    private static final String X = "[0-9a-fA-F]"; // a hex value
+
+    private static final String Y = "[89abAB]"; // second reserved bits
+
+    private static final String UUID_V4 = "x{8}-?x{4}-?4x{3}-?yx{3}-?x{12}".replace("x", X)
+            .replace("y", Y); // Note that the dashes in the UUID are optional.
+
+    private static final String CONTENT_REGEX = "content:uuid(#qualifier)?".replace("uuid", UUID_V4)
+            .replace("qualifier", QUALIFIER_PATTERN.pattern()); // Note the qualifier is optional
+
+    private static final Pattern CONTENT_PATTERN = Pattern.compile(CONTENT_REGEX);
 
     private ContentItemValidator() {
     }
 
-    public static void validate(ContentItem item) throws IllegalArgumentException {
+    public static boolean validate(ContentItem item) {
+        if (item == null || StringUtils.isBlank(item.getUri())) {
+            return false;
+        }
+
         if (StringUtils.isNotBlank(item.getQualifier())) {
-            validateInput(item.getQualifier(), QUALIFIER_PATTERN);
-        }
-
-        if (StringUtils.isNotBlank(item.getUri())) {
-            try {
-                new URI(item.getUri());
-            } catch (URISyntaxException e) {
-                throw new IllegalArgumentException("Unable to create content URI.", e);
+            boolean qualifierValid = validateInput(item.getQualifier(), QUALIFIER_PATTERN);
+            if (!qualifierValid) {
+                return false;
             }
-        } else {
-            throw new IllegalArgumentException("Cannot have an empty content URI.");
         }
 
+        if (CONTENT_PATTERN.matcher(item.getUri())
+                .matches()) {
+            return true;
+        }
+
+        return false;
     }
 
-    private static void validateInput(final String input, Pattern pattern) {
-        if (!pattern.matcher(input)
+    private static boolean validateInput(final String input, Pattern pattern) {
+        if (pattern.matcher(input)
                 .matches()) {
-            throw new IllegalArgumentException(
-                    "Illegal characters found while validating [" + input +
-                            "]. Allowable values must match: " + pattern.pattern());
+            return true;
         }
+
+        return false;
     }
 }
--- a/catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
+++ b/catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
@@ -114,7 +114,10 @@ public class FileSystemStorageProvider implements StorageProvider {
 
         for (ContentItem contentItem : contentItems) {
             try {
-                ContentItemValidator.validate(contentItem);
+                if (!ContentItemValidator.validate(contentItem)) {
+                    LOGGER.warn("Item is not valid: {}", contentItem);
+                    continue;
+                }
                 Path contentIdDir = getTempContentItemDir(createRequest.getId(),
                         new URI(contentItem.getUri()));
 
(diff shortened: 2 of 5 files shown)


catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -233,6 +233,12 @@ public DeleteStorageResponse delete(DeleteStorageRequest deleteRequest)
                 // For deletion we can ignore the qualifier and assume everything under a given ID is
                 // to be removed.
                 Path contentIdDir = getContentItemDir(new URI(deletedContentItem.getUri()));
+
+                if (contentIdDir == null) {
+                    throw new StorageException(
```
jlcsmith: We shouldn't throw an exception here as it will skip all the other metacards in the delete request.

jlcsmith: In fact, I would also say that the catch block below (line 257) should not re-throw the exception.

jlcsmith:
@rzwiefel - we need to consider cases where the metacard has no associated content in the storage provider (eg, where metacard resource-uri is not content:<id>).  It appears that the StorageProvider code you modified assumes that the metacards in the delete request always have an associated item in the content store which is not correct (I could be mis-reading the code though).

rzwiefel:
@lessarderic unfortunately I can't make a positive test for it because it is a windows specific issue. Unless we have a strategy for doing that now.

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -455,9 +466,17 @@ private Path getTempContentItemDir(String requestId, URI contentUri) {
     private Path getContentItemDir(URI contentUri) {
         List<String> pathParts = getContentFilePathParts(contentUri.getSchemeSpecificPart(),
                 contentUri.getFragment());
-
-        return Paths.get(baseContentDirectory.toAbsolutePath()
-                .toString(), pathParts.toArray(new String[pathParts.size()]));
+        try {
+            return Paths.get(baseContentDirectory.toAbsolutePath()
```
brendan-hofmann: I think we want canonical, rather than absolute, paths here. Absolute paths can still contain symlinks, redirects, and OS expansion characters that may or may not be resolved correctly. [This](http://stackoverflow.com/a/1099370) is a good summary.

rzwiefel: Made it so FileSystemStorageProvider Automatically canonicalizes the path when the `baseContentDirectory` is set so everything in here can just use that path as is and it should be :100: (as opposed to everyone repeating the same work every time)

brendan-hofmann: 👍

rzwiefel:
@brendan-hofmann @lessarderic @jlcsmith 

Updated PR.
 * Updated ContentItemValidator to now only allow valid Content items based on the URI
 * Behavior is now to skip any content item that we "don't own"

There will still be the one issue of handling characters in the URI that are not valid filesystem path characters. https://codice.atlassian.net/browse/DDF-2650

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -484,6 +502,9 @@ private Path getContentItemDir(URI contentUri) {
 
     private Path getContentFilePath(URI uri) throws StorageException {
         Path contentIdDir = getContentItemDir(uri);
+        if (contentIdDir == null) {
+            throw new StorageException("Could find file path for URI :" + uri.toString());
```
jlcsmith: nit: typo "Could find..." should be "Could not find..."

jlcsmith:
@rzwiefel - are there tests to cover bulk deletes where one of the content items is invalid?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java
```diff
@@ -26,35 +24,52 @@
  * required by the {@link ddf.catalog.content.StorageProvider}.
  */
 public class ContentItemValidator {
-
     private static final Pattern QUALIFIER_PATTERN = Pattern.compile("\\w[a-zA-Z0-9_\\-]+");
```
brendan-hofmann: Why validate with regexes instead of letting the URI and UUID constructors figure it out for you? Just get and assert on the URI scheme as well and you're done.

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
rzwiefel: This was not supposed to be in this PR, removing.

rzwiefel:
@jlcsmith We don't actually run validation on the delete content items. Not sure if that was intentional or not. I can ask @kcwire  since he was in here last. Any preferences?

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -459,9 +464,17 @@ private Path getTempContentItemDir(String requestId, URI contentUri) {
...
                 contentUri.getFragment());
-
-        return Paths.get(baseContentDirectory.toAbsolutePath()
-                .toString(), pathParts.toArray(new String[pathParts.size()]));
+        try {
+            return Paths.get(baseContentDirectory.toString(),
+                    pathParts.toArray(new String[pathParts.size()]));
+        } catch (InvalidPathException e) {
+            LOGGER.debug("Invalid path: [{}/{}]",
+                    baseContentDirectory.toAbsolutePath()
```
brendan-hofmann: I know it's just debug logging but  could still use canonical path here to be consistent. 😄

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java
rzwiefel: Because flow control with Exceptions is gross. Also Thats what they were doing before and we might as well actually explicitly define (like a regex) what we are expecting. It is far clearer than creating a URI and doing nothing with it purely for the side effect of the exception for validation.

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
rzwiefel: It should already be canonical path. I can remove the toabsolute path tho 👍

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java
rzwiefel: Also we allow the UUID's to not contain the dashes (and most cases don't) which is how the actual UUID expects the UUID to come in formatted

rzwiefel: They are not the validation we support, so they will not work in this case.

rzwiefel: One of the purposes for this change is because a URI is **too** broad which caused bugs. So we can't use URI. Also as previously stated the UUID will not work since we support no dashes. We really only support content:uuid#qualifier, so hence the regex to validate.

brendan-hofmann: I had forgotten that we strip dashes from the UUID. (as an aside we only do that due to a solr limitation) With that in mind, I see the need for regexes instead.

I'm not sure that we really note this anywhere, or what the correct place to do so would even be. This validator seems like it could use an explanatory comment. As you noted, more naive interpretations have already caused us problems in the past.

rzwiefel: I can most certainly add a note about it here. Also, I do have explicit tests with comments that show we allow dash and dash-less id's which hopefully will help.

rzwiefel:
@jlcsmith Added validation for delete and test for delete

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java
pklinef: @brendan-hofmann Stripping dashes from the UUID predates the use of Solr.  I don't know why they were stripped originally.  Maybe @shaundmorris remembers.

PR #1425, opened 2016-11-14
DDF-2598 - Improves and increases logging during exam setup for itests.

coyotesqrl:
#### What does this PR do?
- Raises the default level for printing inactive bundles to ERROR
- Logs any failed config files from the etc directory that do not start in time
- Adds printing of inactive bundles to many more of the failure condition states, as failed bundles are a likely underlying cause

@mweser @vinamartin @emanns95 

N/A

@lessarderic
@shaundmorris

Intentionally breaking a bundle so it cannot resolve and then running an itest that relies on that bundle would trigger the logging of inactive bundles.
Creating a configuration file that does not deploy would trigger the logging of unstarted/remaining config files, followed by the logging of inactive bundles.

For my testing, I ran an itest through the debugger in order to force some of the failure conditions - it is trivial to run through the `waitForAllConfigurations` method and then set the `files` array and `timeoutLimit` from within the debugger to trigger a failure state; however, it would be better if in heroing the failure conditions were created more naturally.

We too often have `exam setup failures` in our CI environments without sufficient information logged to help analyze the problems. As they are intermittent issues that rarely present on our development boxes, they are difficult to track down and resolve.

DDF-2598

N/A

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AdminConfig.java
+++ b/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AdminConfig.java
@@ -15,16 +15,13 @@ package org.codice.ddf.itests.common;
 
 import java.io.IOException;
 import java.util.Dictionary;
-import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.lang.StringUtils;
-import org.codice.ddf.itests.common.config.ConfigurationPredicate;
 import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.service.cm.Configuration;
 import org.osgi.service.cm.ConfigurationAdmin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 
 public class AdminConfig {
     public static final String LOG_CONFIG_PID = "org.ops4j.pax.logging";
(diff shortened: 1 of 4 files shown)


distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/ServiceManagerImpl.java
```diff
@@ -66,10 +68,22 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import com.jayway.restassured.response.Response;
 
 public class ServiceManagerImpl implements ServiceManager {
+    private static final Map<Integer, String> BUNDLE_STATES = ImmutableMap.of(Bundle.INSTALLED,
```
shaundmorris: Would uninstalled add any value here?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/ServiceManagerImpl.java
```diff
@@ -66,10 +68,20 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import com.jayway.restassured.response.Response;
 
 public class ServiceManagerImpl implements ServiceManager {
+    private static final Map<Integer, String> BUNDLE_STATES =
+            new ImmutableMap.Builder<Integer, String>().put(Bundle.UNINSTALLED, "UNINSTALLED")
```
coyotesqrl: @shaundmorris You were correct. I'm not sure how I dropped off the UNINSTALLED state when I copied those over. 

I did have to change to using a `Builder` to make the map because the overloaded `of` methods on `ImmutableMap` only support up to five entries.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AdminConfig.java
```diff
@@ -74,51 +71,6 @@ public Configuration getConfiguration(String s) throws IOException {
...
-     * @param pid       ID of the {@link Configuration} object that the {@link ConfigurationPredicate}
-     *                  will use to validate the condition
-     * @param predicate predicate to wait on
-     * @param timeoutMs wait timeout
-     * @return {@link Configuration} object
-     * @throws IOException          thrown if the {@link Configuration} object cannot be accessed
-     * @throws InterruptedException thrown if the wait times out
-     * @deprecated Use {@link WaitCondition} instead.
-     */
-    public Configuration waitForConfiguration(String pid, ConfigurationPredicate predicate,
```
lessarderic: Thanks for removing this!

PR #1433, opened 2016-11-15
DDF-2596 Reduce number of log entries by ApplicationImpl

ahoffer:
#### What does this PR do?
Stops printing the log entry "Error occured while trying to parse information for application. Application created but may have missing information." Problem in string parsing routines that expected the format "xml-specs-api-2.5.0". Names like "test-dependencies" caused the error.

New behavior tolerates both formats.
New behavior exposes exception if repository cannot be loaded. Repository methods getName(), getFeatures(), both attempt to load the repository.
Updates two features.xml files. Names conform to expectations.
New behavior throws exception if name cannot be parsed unambiguously.
 
@clockard @brendan-hofmann @vinamartin @pklinef 
@garrettfreibott 
@pklinef
@shaundmorris
The hero build should NOT contains the log entry "Error occured while trying to parse information for application. Application created but may have missing information." 
Part of the effort to remove unnecessary log entries.
[DDF-2596](https://codice.atlassian.net/browse/DDF-2596/)
#### Checklist:
- [N/A] Documentation Updated
- [N/A] Change Log Updated
- [Yes ] Update / Add Unit Tests

### DIFF
--- a/distribution/test/itests/test-itests-dependencies-app/src/main/filtered-resources/features.xml
+++ b/distribution/test/itests/test-itests-dependencies-app/src/main/filtered-resources/features.xml
@@ -12,7 +12,7 @@
  *
  **/
 -->
-<features name="ddf-itest-dependencies">
+<features name="ddf-itest-dependencies-${project.version}">
 
     <feature name="ddf-itest-dependencies">
         <feature>third-party-dependencies</feature>
@@ -26,7 +26,9 @@
         <bundle>mvn:org.apache.httpcomponents/httpclient-osgi/${httpclient.version}</bundle>
         <bundle>mvn:ddf.test.thirdparty/rest-assured/${project.version}</bundle>
 
-        <bundle>wrap:mvn:ddf.security/ddf-security-common/${project.version}$Bundle-SymbolicName=itest-test-security-common</bundle>
+        <bundle>
+            wrap:mvn:ddf.security/ddf-security-common/${project.version}$Bundle-SymbolicName=itest-test-security-common
+        </bundle>
         <bundle>mvn:ddf.test.itests/test-itests-common/${project.version}</bundle>
 
         <bundle>mvn:ddf.test.thirdparty/restito/${project.version}</bundle>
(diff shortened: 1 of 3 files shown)


distribution/test/itests/test-itests-dependencies-app/src/main/filtered-resources/features.xml
```diff
@@ -12,7 +12,7 @@
  *
  **/
 -->
-<features name="ddf-itest-dependencies">
+<features name="ddf-itest-dependencies-0.0.0">
```
pklinef: Any reason not to use ${project.version} like other feature files?

platform/admin/core/admin-core-appservice/src/test/java/org/codice/ddf/admin/application/service/impl/ApplicationImplTest.java
```diff
@@ -77,15 +81,33 @@ public void testAppGetters() throws Exception {
                         .size());
     }
 
+    @Ignore
```
pklinef: Can you add a reason for why this is ignored or a ticket when it can be reenabled?

ahoffer: Added this:
    // There is no known reason for creating an instance of ApplicationImpl if the repository
    // has no name. Getting a repositories name or its features triggers the repository to be 
    // loaded from storage. If that fails, it throws an exception. An Application without features
    // or a name is not useful.

distribution/test/itests/test-itests-dependencies-app/src/main/filtered-resources/features.xml
ahoffer: The original code was defaulting to 0.0.0 if the version was absent. But it makes sense to use  ${project.version}

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
```diff
@@ -62,25 +62,15 @@
...
-            } else {
-                name = repo.getName();
-                version = "0.0.0";
-            }
-            features = new HashSet<>(Arrays.asList(repo.getFeatures()));
+            repoName = repo.getName();
+            features.addAll(Arrays.asList(repo.getFeatures()));
         } catch (Exception e) {
-            LOGGER.warn(
-                    "Error occured while trying to parse information for application. Application created but may have missing information.");
```
shaundmorris: Okay so by `reduce` we are completely removing logging here? Just wanted to make sure that was the case.

ahoffer: We're throwing an exception instead of logging in every case. No one is catching the exception and it gets logged. I felt it deserved more than a warning because an exception means an application cannot be installed.

ahoffer:
Passed hero build at: 2016-11-21T12:24:27-07:00

PR #1458, opened 2016-11-28
DDF-2625: Fixes case insensitive enumeration validator

rzwiefel:
#### What does this PR do?
Fixes Enumeration Validator to not mutate the values passed in. 
@troymohl @bdeining @andrewkfiedler 

@coyotesqrl
@jlcsmith

https://codice.atlassian.net/browse/DDF-2625

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
@@ -18,13 +18,14 @@ import java.util.Collections;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.TreeSet;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
 
 import ddf.catalog.data.Attribute;
 import ddf.catalog.validation.AttributeValidator;
@@ -54,10 +55,8 @@ public class EnumerationValidator implements AttributeValidator {
 
         this.ignoreCase = ignoreCase;
 
-        this.values = Collections.unmodifiableSet(values.stream()
-                .filter(Objects::nonNull)
-                .map(ignoreCase ? String::toLowerCase : String::toString)
-                .collect(Collectors.toSet()));
+        this.values = new TreeSet<>(ignoreCase ? String.CASE_INSENSITIVE_ORDER : null);
+        this.values.addAll(Sets.filter(values, Objects::nonNull));
     }
 
     /**
@@ -74,16 +73,11 @@ public class EnumerationValidator implements AttributeValidator {
         String name = attribute.getName();
 
         for (final Serializable value : attribute.getValues()) {
-            final String stringValue = ignoreCase ?
-                    String.valueOf(value)
-                            .toLowerCase() :
-                    String.valueOf(value);
-
-            if (!values.contains(stringValue)) {
+            if (!values.contains(value)) {
                 final AttributeValidationReportImpl report = new AttributeValidationReportImpl();
                 // TODO (jrnorth) - escape the value.
                 report.addViolation(new ValidationViolationImpl(Collections.singleton(name),
-                        name + " has an invalid value: [" + stringValue + "]",
+                        name + " has an invalid value: [" + value + "]",
                         Severity.ERROR));
                 values.forEach(report::addSuggestedValue);
                 return Optional.of(report);


troymohl:
My only input is that this doesn't obey the Set interface as this implementation is inconsistent with .equals. Since we never return the Set directly it should be ok. The ticket seems to indicate that  `AttributeValidationReport.suggestedValues` needs to match something (original json file?) for the UI to work. We should probably add that information to the javadoc for `AttributeValidationReport` in case other validators are added that don't meet that expectation.

rzwiefel:
@troymohl It does obey the set interface, based upon your comparator. https://docs.oracle.com/javase/8/docs/api/java/util/TreeSet.html If your comparator is not, then it will be inconsistent. 

And also I would argue that you should **never** modify input data unless you explicitly are told it is allowed. You should always assume its not safe to mutate the input data and store it as is.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
```diff
@@ -54,10 +53,8 @@ public EnumerationValidator(final Set<String> values, final boolean ignoreCase)
 
         this.ignoreCase = ignoreCase;
 
-        this.values = Collections.unmodifiableSet(values.stream()
-                .filter(Objects::nonNull)
-                .map(ignoreCase ? String::toLowerCase : String::toString)
-                .collect(Collectors.toSet()));
+        this.values = new TreeSet<>(ignoreCase? String.CASE_INSENSITIVE_ORDER : null);
```
rzwiefel: Probably

rzwiefel: Though clearly checkstyle doesn't care. And in this case it makes it slightly more terse, so ima leave it in less someone has really strong opinions.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
```diff
@@ -54,10 +53,8 @@ public EnumerationValidator(final Set<String> values, final boolean ignoreCase)
 
         this.ignoreCase = ignoreCase;
 
-        this.values = Collections.unmodifiableSet(values.stream()
-                .filter(Objects::nonNull)
```
coyotesqrl: Are you no longer worried about there being a null value in the incoming set? You'll throw an NPE on the call to `addAll` below.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
```diff
@@ -54,10 +53,8 @@ public EnumerationValidator(final Set<String> values, final boolean ignoreCase)
...
-                .map(ignoreCase ? String::toLowerCase : String::toString)
-                .collect(Collectors.toSet()));
+        this.values = new TreeSet<>(ignoreCase? String.CASE_INSENSITIVE_ORDER : null);
+        this.values.addAll(values);
```
coyotesqrl: Could also change to something like...
```java
Supplier<Set<String>> supl =
        () -> ignoreCase ? new TreeSet<>(String.CASE_INSENSITIVE_ORDER) : new TreeSet();
this.values = values.stream()
        .filter(Objects::nonNull)
        .collect(Collectors.toCollection(supl));
```
...or take that and wrap in an unmodifiable collection is previously.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
rzwiefel: They cannot send a null value in the set.

coyotesqrl: There's nothing preventing it, other than your knowledge of the callers.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
rzwiefel: this is argubly is more complex.

and this set of values is never exposed anywhere, so there is no need to make it unmodifiable.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
rzwiefel: Filtering out nulls now.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/validation/impl/validator/EnumerationValidator.java
rzwiefel: Looks like the formatter did it anyways for you.

PR #1464, opened 2016-12-01
DDF-2620 Allow updates to metacards that are pending NRT indexing

pklinef:
#### What does this PR do?
It takes up to a second for a created, updated, or deleted metacard to become visible in the Solr near real time (NRT) index.  During this time, the catalog framework will not allow updates and deletes due to security checks if the active user has access to the current version of the metacard.  This adds a pending NRT index that is indexed by ID only.  This will allow all catalog operations to pass security checks before the metacards are fully indexed by Solr.  Pending metacards will still not be returned for non-ID queries.

@brendan-hofmann 
@rzwiefel 

@coyotesqrl
@pklinef

I tested using something similar to the following command.  Make sure to replace the two `<path to ddf clone>` places holders.  Inspect the created metacard and verify it has the details for metacard2.xml instead of metacard1.xml.
`curl -X POST --header "Content-Type: text/xml" -d @<path to ddf clone>/distribution/test/itests/test-itests-common/src/main/resources/metacard1.xml https://localhost:8993/services/catalog -u admin:admin --insecure -i | grep "id:" | grep -E -o "[0-9a-f]{32}" | xargs -I{} curl -X PUT --header "Content-Type: text/xml" -d @<path to ddf clone>/distribution/test/itests/test-itests-common/src/main/resources/metacard2.xml https://localhost:8993/services/catalog/{} -u admin:admin --insecure -v -i`

[DDF-2620](https://codice.atlassian.net/browse/ddf-2620)

#### Checklist:
- [ n/a ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/MetacardIdEqualityFilterDelegate.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/MetacardIdEqualityFilterDelegate.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.source.solr;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import ddf.catalog.data.Metacard;
+import ddf.catalog.filter.impl.SimpleFilterDelegate;
+
+/**
+ * Filter delegate that returns all IDs used in metacard ID equality filters.
+ */
+public class MetacardIdEqualityFilterDelegate extends SimpleFilterDelegate<Set<String>> {
+
+    @Override
+    public <S> Set<String> defaultOperation(Object property, S literal, Class<S> literalClass,
+            Enum operation) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public <S> Set<String> propertyIsEqualTo(String propertyName, S literal, Class<S> literalClass,
+            ComparisonPropertyOperation operation) {
+        if (Metacard.ID.equals(propertyName)) {
+            return Collections.singleton(literal.toString());
+        } else {
+            return defaultOperation(propertyName, literal, literalClass, operation);
+        }
+    }
+
+    @Override
+    public Set<String> and(List<Set<String>> operands) {
+        return joinIds(operands);
+    }
+
+    @Override
+    public Set<String> or(List<Set<String>> operands) {
+        return joinIds(operands);
+    }
+
+    private Set<String> joinIds(List<Set<String>> operands) {
+        return operands.stream()
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+    }
+}
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/MetacardIdEqualityFilterDelegate.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/MetacardIdEqualityFilterDelegate.java
@@ -44,6 +44,12 @@ public class MetacardIdEqualityFilterDelegate extends SimpleFilterDelegate<Set<S
     }
 
     @Override
+    public <S> Set<String> propertyIsNotEqualTo(String propertyName, S literal,
+            Class<S> literalClass, ComparisonPropertyOperation operation) {
+        return Collections.emptySet();
+    }
+
+    @Override
     public Set<String> and(List<Set<String>> operands) {
         return joinIds(operands);
     }
@@ -53,9 +59,15 @@ public class MetacardIdEqualityFilterDelegate extends SimpleFilterDelegate<Set<S
         return joinIds(operands);
     }
 
+    @Override
+    public Set<String> not(Set<String> operand) {
+        return Collections.emptySet();
+    }
+
     private Set<String> joinIds(List<Set<String>> operands) {
         return operands.stream()
                 .flatMap(Collection::stream)
                 .collect(Collectors.toSet());
     }
+
 }
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
@@ -24,6 +24,9 @@ import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.solr.client.solrj.SolrClient;
@@ -39,11 +42,16 @@ import org.codice.solr.factory.impl.ConfigurationStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
 import ddf.catalog.data.ContentType;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.MetacardCreationException;
+import ddf.catalog.data.Result;
 import ddf.catalog.data.impl.AttributeImpl;
 import ddf.catalog.data.impl.MetacardImpl;
+import ddf.catalog.data.impl.ResultImpl;
 import ddf.catalog.filter.FilterAdapter;
 import ddf.catalog.operation.CreateRequest;
 import ddf.catalog.operation.CreateResponse;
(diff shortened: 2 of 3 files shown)


catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -195,7 +212,36 @@ public void maskId(String id) {
...
+
+    private void queryPendingNrtIndex(QueryRequest request, SourceResponse response)
+            throws UnsupportedQueryException {
+        if (request == null || request.getQuery() == null) {
+            return;
+        }
+
+        Set<String> ids = filterAdapter.adapt(request.getQuery(),
+                new MetacardIdEqualityFilterDelegate());
+        if (ids.size() == 0 || ids.size() <= response.getResults()
```
rzwiefel: Are you sure that ids.size() being less than the number of results will always be a valid short circuit? 
i'm thinking something like
`id = t1 AND metacard.version.bla = somethingelse` Where the second condition matches 5 some odd metacards, so the results size is 5, but the ids size is 1.. since its less it short circuits. Though wouldn't we want to still fill in the one metacard id that was requested?

Or am I simply mislead and does this somehow not get executed if there is anything other than an ID query?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -195,7 +212,36 @@ public void maskId(String id) {
...
+                .size()) {
+            return;
+        }
+
+        for (Result result : response.getResults()) {
+            ids.remove(result.getMetacard()
+                    .getId());
```
rzwiefel: is this for speed? as opposed to doing a set difference or the like

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -304,6 +358,11 @@ public UpdateResponse update(UpdateRequest updateRequest) throws IngestException
                         "Found more metacards than updated metacards provided. Please ensure your attribute values match unique records.");
             }
 
+        } else if (pendingNrtIndex.getAllPresent(identifiers).size() > 0) {
+            Map<String, Metacard> matches = pendingNrtIndex.getAllPresent(identifiers);
```
rzwiefel: Not want to deal with having to store this in a variable instead of fetching it twice?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -422,6 +469,12 @@ public DeleteResponse delete(DeleteRequest deleteRequest) throws IngestException
                     identifiers.size());
             deleteListOfMetacards(deletedMetacards, identifierPaged, attributeName);
         }
+        addPendingNrtDeletedMetacards(deletedMetacards, identifiers);
```
rzwiefel: So the flow i've walked thru and have in my head for this method is:
```
If 
    the number of actually deleted metacards is less than the number they asked to be deleted AND any of the deleteRequested ids are currently in the pending cache,
then 
    get any deleteRequested metacards from the pending cache, 
        then remove any of the metacards from that that are in the deleted metacards response (ones that were actually deleted), 
whatever is remaining is things that were requested for deletion, but were only in the pending cache but not the actual delete response.```

So my question is, why would something be in the cache but not come back as in the list of deleted metacards?

**Comment by brendan-hofmann** [comment]
Successfully manually verified using that big long curl command. 👍

File: catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java [hunk 5]
```diff
@@ -457,7 +510,18 @@ private void createListOfDeletedMetacards(List<Metacard> deletedMetacards,
                 LOGGER.info("Metacard creation exception creating metacards during delete", e);
                 throw new IngestException("Could not create metacard(s).");
             }
+        }
+    }
+
+    private void addPendingNrtDeletedMetacards(List<Metacard> deletedMetacards,
+            List<? extends Serializable> identifiers) {
+        if (deletedMetacards.size() < identifiers.size() && pendingNrtIndex.getAllPresent(
```
coyotesqrl: As I'm understanding this, it is for the case where create/delete is called in rapid succession and the delete request needs to find the metacard to delete from the pending cache. I assume the request to delete will still be queued up by Solr and the NRT index will either
a) have the id added to it then quickly removed
b) never have the id added to it
?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
rzwiefel: Ignore. Loop is probably simplest :+1:

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/MetacardIdEqualityFilterDelegate.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import ddf.catalog.data.Metacard;
+import ddf.catalog.filter.impl.SimpleFilterDelegate;
+
+/**
+ * Filter delegate that returns all IDs used in metacard ID equality filters.
+ */
+public class MetacardIdEqualityFilterDelegate extends SimpleFilterDelegate<Set<String>> {
```
pklinef: This was designed to really only address the security check when an update/delete is done immediately after a create/update.  But I think you are right that this might cause issues for custom queries for IDs that include `not`s that are done immediately after a create/update.  I will add in some guards to be safe.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -304,6 +358,11 @@ public UpdateResponse update(UpdateRequest updateRequest) throws IngestException
                         "Found more metacards than updated metacards provided. Please ensure your attribute values match unique records.");
             }
 
+        } else if (pendingNrtIndex.getAllPresent(identifiers).size() > 0) {
+            Map<String, Metacard> matches = pendingNrtIndex.getAllPresent(identifiers);
+            for (Entry<String, Metacard> match : matches.entrySet()) {
```
pklinef: I tried this but it caused other tests to fail but it looks like they are failing since they are mutating the reference stored in the pending NRT index.  This worked after I added a defensive copy which now makes the pending NRT index safer.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -114,7 +115,8 @@
      * near real time index yet.  Cache is checked when ID queries fail from Solr.
      */
     private Cache<String, Metacard> pendingNrtIndex = CacheBuilder.newBuilder()
-            .expireAfterWrite(5, TimeUnit.SECONDS)
+            .expireAfterWrite(NumberUtils.toInt(System.getProperty("solr.provider.pending-nrt-ttl"),
```
brendan-hofmann: How hidden should this hidden attribute be? Will we document it anywhere? Should it be a constant?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -358,40 +359,39 @@ public UpdateResponse update(UpdateRequest updateRequest) throws IngestException
...
-        } else if (pendingNrtIndex.getAllPresent(identifiers).size() > 0) {
-            Map<String, Metacard> matches = pendingNrtIndex.getAllPresent(identifiers);
-            for (Entry<String, Metacard> match : matches.entrySet()) {
-                idToMetacardMap.put(match.getKey(), match.getValue());
+            for (SolrDocument doc : idResults.getResults()) {
+                Metacard old;
+                try {
+                    old = client.createMetacard(doc);
+                } catch (MetacardCreationException e) {
+                    throw new IngestException("Could not create metacard(s).");
```
brendan-hofmann: This message seems technically correct, but ambiguous/confusing in the context of an update request. Shouldn't we at least rethrow the `MetacardCreationException`, or log the stacktrace and rethrow the message?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
```diff
@@ -286,7 +288,7 @@ public CreateResponse create(CreateRequest request) throws IngestException {
         }
 
         pendingNrtIndex.putAll(output.stream()
-                .collect(Collectors.toMap(Metacard::getId, Function.identity())));
+                .collect(Collectors.toMap(Metacard::getId, m -> copyMetacard(m))));
```
brendan-hofmann: Is defensive logic something that we should unit test? Seems like the kind of thing that would be easy to break by accident.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
pklinef: By hidden I just mean that it is undocumented.  I don't think a constant is needed since the string value is not used anywhere else.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
pklinef: This defensive logic took a few tries before all tests were green.  Based on that I think there is already enough coverage that tests will fail if the defensive logic is broken.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrCatalogProvider.java
pklinef: This message was pre-existing and just moved but I will review the messages and logging.  It looks like there are other cases like this one in this case that could be improved.

pklinef:
@brendan-hofmann I don't have a unit test for multiple sequential updates but I think the create with immediate update is sufficient and should not require a rehero.

rzwiefel:
:shipit:

PR #1477, opened 2016-12-07
DDF-2638: Added support for solr cloud to persistence store

oconnormi:
#### What does this PR do?

Updates the Persistence Store to support solr cloud

@lessarderic 
@pklinef 

@figliold
@shaundmorris 

Update etc/system.properties to point at solr cloud and verify persistence works correctly

[DDF-2638](https://codice.atlassian.net/browse/DDF-2638)

#### Checklist:
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests


platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -169,12 +144,12 @@ private void doRollback(SolrClient coreSolrClient, String type) {
         List<Map<String, Object>> results = new ArrayList<>();
 
         // Set Solr Core name to type and create/connect to Solr Core
```
shaundmorris: should we update this comment as well?

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -289,26 +264,25 @@ public int delete(String type, String cql) throws PersistenceException {
         return idsToDelete.size();
     }
 
-    private SolrClient getSolrCore(String storeName) {
-        if (coreSolrClients.containsKey(storeName)) {
-            LOGGER.debug("Returning core {} from map of coreSolrClients", storeName);
-            return coreSolrClients.get(storeName);
+    private SolrClient getSolrClient(String storeName) {
+        if (solrClients.containsKey(storeName)) {
+            LOGGER.debug("Returning core {} from map of solrClients", storeName);
```
shaundmorris: should we update the log message to say "clients" instead of "core" ?

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -289,26 +264,25 @@ public int delete(String type, String cql) throws PersistenceException {
...
+            return solrClients.get(storeName);
         }
 
         // Must specify shard in URL so proper core is used
```
shaundmorris: should we update this comment as well?

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -289,26 +264,25 @@ public int delete(String type, String cql) throws PersistenceException {
         return idsToDelete.size();
     }
 
-    private SolrClient getSolrCore(String storeName) {
-        if (coreSolrClients.containsKey(storeName)) {
-            LOGGER.debug("Returning core {} from map of coreSolrClients", storeName);
-            return coreSolrClients.get(storeName);
+    private SolrClient getSolrClient(String storeName) {
```
lessarderic: ⭐️  We need to change the logic and add synchronization in this method to make sure we don't run into concurrency issues with the map of clients.

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -99,8 +74,8 @@ public void add(String type, Map<String, Object> properties) throws PersistenceE
         LOGGER.debug("Adding entry of type {}", type);
 
         // Set Solr Core name to type and create/connect to Solr Core
-        SolrClient coreSolrClient = getSolrCore(type);
-        if (coreSolrClient == null) {
+        SolrClient solrClient = getSolrClient(type);
+        if (solrClient == null) {
```
lessarderic: Should we log something everywhere `getSolrClient()` fails to return a client? Right now, the only that gets logged is a generic debug message, which won't help figure out what's going on if an update to the persistence store fails.

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
pklinef: The solrClients map is a ConcurrentHashMap.  The race was not a concern since the worst case was the racing threads would both get a client and set the map.  I am not sure if the cost of building the extra client is worth the extra complexity to synchronize.  Is there a different concurrency issue I am missing?

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -289,26 +264,25 @@ public int delete(String type, String cql) throws PersistenceException {
...
         // Must specify shard in URL so proper core is used
-        SolrClient coreSolrClient = null;
+        SolrClient solrClient = null;
         try {
-            Future<SolrClient> coreSolrClientFuture =
-                    HttpSolrClientFactory.getHttpSolrClient(solrUrl.getResolvedString(), storeName);
-            coreSolrClient = coreSolrClientFuture.get(30, TimeUnit.SECONDS);
-            coreSolrClients.put(storeName, coreSolrClient);
+            Future<SolrClient> clientFuture = clientFactory.newClient(storeName);
+            solrClient = clientFuture.get(30, TimeUnit.SECONDS);
```
pklinef: I think we should change the `solrClients` map to be a map of futures instead of clients so we can remove this blocking 30 second get and allow the retry logic to work.  That would make it work more like the remote Solr catalog provider.

Though I think we should do this eventually, this does not need to be done as part of this PR if you think it is too big of a change since it is out of scope.

platform/persistence/platform-persistence-core-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -18,11 +18,13 @@
 
     <reference id="encryptionService" interface="ddf.security.encryption.EncryptionService"/>
 
+    <bean id="solrFactory" class="org.codice.solr.factory.impl.SolrClientFactoryImpl" />
+
     <bean id="persistentStore" class="org.codice.ddf.persistence.internal.PersistentStoreImpl">
         <cm:managed-properties
```
pklinef: Since you removed the metatype file, I think you should remove this manage-properties section.

oconnormi: I knew I forgot something, done!

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
lessarderic: Thanks @pklinef. Thought I commented on that already but I guess I forgot.

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
lessarderic: If we change the code to use a `ConcurrentHashMap<String, Future<SolrClient>>` and a separate `ConcurrentHashMap<SolrClient>`, then the potential concurrency issues might disappear. Let's see what the code looks like after that change is in.

pklinef: @lessarderic Why would we need another `ConcurrentHashMap<SolrClient>` map?  Instead of having to manage two maps,  I think it would be ok to just have `ConcurrentHashMap<String, Future<SolrClient>>` and get the client each time from the future.

lessarderic: True. And I'm ok with not adding synchronization if we know that having two `SolrClient` instances for the same storeName for short periods of time isn't going to cause any issues now or down the road.

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
lessarderic: @oconnormi, don't forget the logs.

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -289,26 +264,28 @@ public int delete(String type, String cql) throws PersistenceException {
...
-    private SolrClient getSolrCore(String storeName) {
-        if (coreSolrClients.containsKey(storeName)) {
-            LOGGER.debug("Returning core {} from map of coreSolrClients", storeName);
-            return coreSolrClients.get(storeName);
+    private SolrClient getSolrClient(String storeName) {
+
+        Future<SolrClient> clientFuture = null;
+        SolrClient solrClient = null;
+
+        if (solrFutures.containsKey(storeName)) {
```
lessarderic: You should be able to replace this if statement with:

`Future<SolrClient> clientFuture = solrFutures.computeIfAbsent(storeName, storeName -> clientFactory.newClient(storeName));`

platform/persistence/platform-persistence-core-impl/src/main/java/org/codice/ddf/persistence/internal/PersistentStoreImpl.java
```diff
@@ -289,26 +264,28 @@ public int delete(String type, String cql) throws PersistenceException {
...
-        SolrClient coreSolrClient = null;
         try {
-            Future<SolrClient> coreSolrClientFuture =
-                    HttpSolrClientFactory.getHttpSolrClient(solrUrl.getResolvedString(), storeName);
-            coreSolrClient = coreSolrClientFuture.get(30, TimeUnit.SECONDS);
-            coreSolrClients.put(storeName, coreSolrClient);
+            solrClient = clientFuture.get(5, TimeUnit.SECONDS);
 
-            LOGGER.trace("EXITING: getSolrCore");
+            LOGGER.trace("EXITING: getSolrClient");
```
lessarderic: Since we don't have a log when entering this method and we don't log anything before L288, should we remove this trace?

LinkMJB:
test this please

figliold:
Clean build passed! 👍

PR #1479, opened 2016-12-08
DDF-2637 Remove tight dependencies on BrandingPlugin

harrison-tarr:
#### What does this PR do?

When using implementations of the BrandingPlugin other than the DdfBrandingPlugin, these can be installed at a late enough stage that the LandingPage times out waiting for the dependency.
Instead, the LandingPage should always start the DdfBrandingPlugin and then when another implementation becomes available, should use that instead.

In addition, this PR takes the fix from the LandingPage java and blueprint and applies it to any other classes that use a BrandingPlugin in DDF.

@roelens8 
@brendan-hofmann 
@jrnorth 
@kcwire 
@jaymcnallie 
Build and run DDF. Hot deploy a different branding plugin implementation with a ranking higher than 0. Verify that in branded places, the branding is replaced. (Eg, landing page, admin ui, search ui, etc)

The landing page was timing out on startup during itests in sufficiently slow systems. This protects it from happening in the future.
[DDF-2637](https://codice.atlassian.net/browse/DDF-2637)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistry.java
+++ b/platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistry.java
@@ -1,0 +1,69 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+
+package org.codice.ddf.branding;
+
+import java.io.IOException;
+import java.util.List;
+
+public interface BrandingRegistry {
+
+    /**
+     * This method returns the name portion of the name & version of the product, as defined
+     * by the highest priority {@link BrandingPlugin} in the system..
+     * E.g, if the full name is "DDF v1.0.0", this will return "DDF"
+     *
+     * @return a String representing the name of the product or in the case where the name cannot be found, the String "DDF"
+     */
+    String getProductName();
+
+    /**
+     * This method takes in a {@link BrandingPlugin} method reference (a {@link BrandingRegistryImpl.BrandingMethod}
+     * and calls it on the highest priority {@link BrandingPlugin} in the system.
+     *
+     * @return a String with the value of the {@link BrandingPlugin}'s method call or if the method call cannot be evaluated, an empty String ""
+     */
+    String getAttributeFromBranding(BrandingMethod supplier);
+
+    /**
+     * This method returns the version number portion of the name & version of the product, as defined
+     * by the highest priority {@link BrandingPlugin} in the system..
+     * E.g, if the full name is "DDF v1.0.0", this will return "v1.0.0"
+     *
+     * @return a String representing the version number of the product or in the case where the version number cannot be found, an empty String ""
+     */
+    String getProductVersion();
+
+    /**
+     * Sets the {@link List} of {@link BrandingPlugin}s on the {@link BrandingRegistry}
+     *
+     * @param brandingPlugins list of branding plugins associated with the system
+     */
+    void setBrandingPlugins(List<BrandingPlugin> brandingPlugins);
+
+    /**
+     * Gets the {@link List} of {@link BrandingPlugin}s from the {@link BrandingRegistry}
+     *
+     * @return the list of branding plugins associated with the system
+     */
+    List<BrandingPlugin> getBrandingPlugins();
+
+    /**
+     * A functional interface for {@link BrandingPlugin} methods (cannot use {@link java.util.function.Function} since some of them throw Exceptions)
+     */
+    @FunctionalInterface
+    interface BrandingMethod {
+        String apply(BrandingPlugin b) throws IOException;
+    }
+}
--- a/platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
+++ b/platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+
+package org.codice.ddf.branding;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BrandingRegistryImpl implements BrandingRegistry {
+    private static final Logger LOGGER = LoggerFactory.getLogger(BrandingRegistryImpl.class);
+
+    private List<BrandingPlugin> brandingPlugins = Collections.emptyList();
+
+    @Override
+    public String getProductName() {
+        return getBrandingPlugins().stream()
+                .map(plugin -> StringUtils.substringBeforeLast(plugin.getProductName(), " "))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElse("DDF");
+    }
+    
+    @Override
+    public String getAttributeFromBranding(BrandingMethod supplier) {
+        return getBrandingPlugins().stream()
+                .map(plugin -> {
+                    try {
+                        return supplier.apply(plugin);
+                    } catch (IOException e) {
+                        LOGGER.warn("Could not get the requested attribute from the Branding Plugin", e);
+                        return null;
+                    }
+                })
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElse("");
+    }
+
+    @Override
+    public String getProductVersion() {
+        return getBrandingPlugins().stream()
+                .map(plugin -> StringUtils.substringAfterLast(plugin.getProductName(), " "))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .orElse("");
+    }
+
+    @Override
+    public List<BrandingPlugin> getBrandingPlugins() {
+        return brandingPlugins;
+    }
+
+    @Override
+    public void setBrandingPlugins(List<BrandingPlugin> brandingPlugins) {
+        this.brandingPlugins = brandingPlugins;
+    }
+
+}
--- a/platform/branding-api/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/platform/branding-api/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -1,0 +1,31 @@
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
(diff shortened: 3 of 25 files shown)


harrison-tarr:
I got the bamboo build failure locally when I didn't build the whole DDF after a rebase. Is this bamboo build only trying to build part of the system?

brendan-hofmann:
Yes, bamboo builds incrementally for PRs.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
```diff
@@ -269,8 +245,10 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
...
     // package-private for unit testing
-    String compileTemplateWithProperties() {
-        logoToUse = StringUtils.isNotEmpty(logo) ? logo : productImage;
+    String compileTemplateWithProperties() throws IOException {
+        title = getTitle();
+        version = branding.getAttributeFromBranding(BrandingPlugin::getProductName);
+        logoToUse = StringUtils.isNotEmpty(logo) ? logo : getProductImage();
         // FieldValueResolver so this class' fields can be accessed in the template.
         // MapValueResolver so we can access {{@index}} in the #each helper in the template.
         final Context context = Context.newBuilder(this)
```
jrnorth: I think we can simplify this now that we're getting some of the template values from the branding registry. For example, we don't really need the instance variables `title`, `version`, or `logoToUse` anymore since we're calculating them on the fly. We could add a `JavaBeanValueResolver` to the context so we can access values in the template via `get*()` methods in this class.

Then, instead of having to set `title`, `version`, and `logoToUse` at the beginning of the method, `getTitle()` and `getVersion()` would be called to get the values of `title` and `version` in the handlebars template. We'd have to add a getter for `logoToUse`. And we could even add getters for the rest of the fields and we wouldn't need the `FieldValueResolver` anymore.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
```diff
@@ -269,8 +245,10 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp)
     }
 
     // package-private for unit testing
-    String compileTemplateWithProperties() {
-        logoToUse = StringUtils.isNotEmpty(logo) ? logo : productImage;
+    String compileTemplateWithProperties() throws IOException {
+        title = getTitle();
+        version = branding.getAttributeFromBranding(BrandingPlugin::getProductName);
```
jrnorth: Is this supposed to be `version = getVersion()`?

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * This method returns the name portion of the name & version of the product, as defined
+     * by the highest priority {@link BrandingPlugin} in the system..
+     * E.g, if the full name is "DDF v1.0.0", this will return "DDF"
+     *
+     * @return a String representing the name of the product
+     */
+    @Override
+    public String getProductName() {
+        return !getBrandingPlugins().isEmpty() ?
```
brendan-hofmann: I'd actually prefer the consistency of always using a getter if one exists. Makes debugging (find usages, etc.) much easier down the line.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
harrison-tarr: No, the `getVersion` returns the number (eg "v1.0.0") and the `getName` returns the name, eg "DDF" but the `BrandingPlugin::getProductName` returns the "whole thing" eg `DDF v1.0.0` which I think was the intent on this page.

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
harrison-tarr: I used the getter mainly for unit testing. I couldn't figure out how to mock a method that takes a method reference (see the `LandingPageTest`) so instead I had to instead mock out the `BrandingPlugins` instead. I think it does lead to cleaner code though

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+    String getAttributeFromBranding(BrandingMethod supplier) throws IOException;
+
+    String getProductVersion();
+
+    void setBrandingPlugins(List<BrandingPlugin> brandingPlugins);
+
+    List<BrandingPlugin> getBrandingPlugins();
+
+    /**
+     * A functional interface for {@link BrandingPlugin} methods (cannot use {@link java.util.function.Function} since some of them throw Exceptions)
```
harrison-tarr: fixed

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+    
+    @Override
+    public String getAttributeFromBranding(BrandingMethod supplier) {
+        return getBrandingPlugins().stream()
+                .map(plugin -> {
+                    try {
+                        return supplier.apply(plugin);
+                    } catch (IOException e) {
+                        LOGGER.warn("Could not get attributeFromBranding", e);
```
jrnorth: This log message is weird. Maybe `Could not get attribute from branding plugin`?

platform/platform-configuration/src/test/java/org/codice/ddf/configuration/PlatformUiConfigurationTest.java
```diff
@@ -64,7 +67,10 @@ public void testConfig() throws IOException {
         assertEquals("footer", jsonObject.get(PlatformUiConfiguration.FOOTER));
         assertEquals("background", jsonObject.get(PlatformUiConfiguration.BACKGROUND));
         assertEquals("color", jsonObject.get(PlatformUiConfiguration.COLOR));
-        assertEquals("product", jsonObject.get(PlatformUiConfiguration.VERSION));
+        assertEquals("product", jsonObject.get(PlatformUiConfiguration.TITLE));
+        //TODO Tarr Identify if this is a valid check (should "version" be considered the number or the product name.
+        //TODO (cont'd) Most other places refer to the version number as "version" and the product name, eg "DDF" as "product name" or "title"
+        //assertEquals("product", jsonObject.get(PlatformUiConfiguration.VERSION));
```
harrison-tarr: @stustison I believe you are the person who introduced this assertion. Should the `PlatformUiConfiguration.VERSION` be something like "product" or "1.0.0" or "product 1.0.0"?

platform/landing-page/src/test/java/org/codice/ddf/landing/TestLandingPage.java
```diff
@@ -89,10 +104,22 @@ public void testCompileTemplate() {
...
+    @Test
+    public void testDoGet() throws IOException, ServletException {
+        String compiledTemplate = landingPage.compileTemplateWithProperties();
+        HttpServletRequest request = mock(HttpServletRequest.class);
+        when(request.getRequestURI()).thenReturn("");
+        PrintWriter writer = mock(PrintWriter.class);
+        HttpServletResponse response = mock(HttpServletResponse.class);
+        when(response.getWriter()).thenReturn(writer);
+        landingPage.doGet(request, response);
+        verify(writer, Mockito.times(1)).write(compiledTemplate);
```
jrnorth: Static import for `times()`

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BrandingRegistryImpl implements BrandingRegistry {
```
jrnorth: Is returning an empty string in the case of no branding plugins/not being able to find a suitable value from any of the branding plugins part of this interface's contract? If so, it should be documented in the interface javadoc.

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
harrison-tarr: I agree. I think I was trying to use the method name in the log message, but it looks weird.

platform/landing-page/src/test/java/org/codice/ddf/landing/TestLandingPage.java
harrison-tarr: fixed

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
harrison-tarr: fixed

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
harrison-tarr: fixed, let me know if you like the wording

platform/branding-api/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,33 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+
+<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
+           xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0">
+
+    <bean id="brandingRegistry" class="org.codice.ddf.branding.BrandingRegistryImpl">
+        <cm:managed-properties persistent-id="org.codice.ddf.branding.registry.properties"
```
kcwire: Do you actually have any managed properties?

harrison-tarr: I don't know, I copied this from elsewhere. I left it in because I'm injecting these plugins the same way the other blueprints were (which is now abstracted away). If I don't need this `managed-properties` element, I'll remove it

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+     * @return a String with the value of the {@link BrandingPlugin}'s method call or, if the method call cannot be evaluated, an empty string ""
+     */
+    String getAttributeFromBranding(BrandingMethod supplier);
+
+    /**
+     * This method returns the version number portion of the name & version of the product, as defined
+     * by the highest priority {@link BrandingPlugin} in the system..
+     * E.g, if the full name is "DDF v1.0.0", this will return "v1.0.0"
+     *
+     * @return a String representing the version number of the product or, in the case where the name cannot be found, an empty string ""
```
jrnorth: "or, in the case where the version number cannot be found"

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+    
+    @Override
+    public String getAttributeFromBranding(BrandingMethod supplier) {
+        return getBrandingPlugins().stream()
+                .map(plugin -> {
+                    try {
+                        return supplier.apply(plugin);
+                    } catch (IOException e) {
+                        LOGGER.warn("Could not get the requested attribute from any Branding Plugin", e);
```
jrnorth: Not _any_ branding plugin, since an exception here only indicates that the current plugin didn't work.

platform/landing-page/src/test/java/org/codice/ddf/landing/TestLandingPage.java
```diff
@@ -80,7 +95,7 @@ public void testSetAnnouncements() {
     }
 
     @Test
-    public void testCompileTemplate() {
+    public void testCompileTemplate() throws IOException {
```
jrnorth: Does this method actually throw an IOException?

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistry.java
harrison-tarr: right, `in the case where the version number cannot be found` is required, so i can't surround it in commas

platform/branding-api/src/main/java/org/codice/ddf/branding/BrandingRegistryImpl.java
harrison-tarr: Good call, I guess I wasn't looking closely enough when I wrote the message

roelens8:
Successfully hero'd and followed testing procedures using Alliance's branding plugin implementation.

PR #1482, opened 2016-12-09
DDF-2642: Remove accordion view from sources tab.

djblue:
#### What does this PR do?

Remove accordion view from sources tab.

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?

@garrettfreibott 
@tbatie 
@vinamartin 
@andrewkfiedler 
@rzwiefel 
@mcalcote 
@clockard 

#### Choose 2 committers to review/merge the PR.

@coyotesqrl
@pklinef

#### How should this be tested? (List steps with links to updated documentation)

- Ensure that you can still create sources.
- Ensure that you can edit different source configurations.
- Ensure the closing a configuration doesn't add duplicate entries in the active bindings menu.

#### Any background context you want to provide?

The accordion view is the sources tab can cause confusion so we decided to revert the view back to how it use to work.

#### What are the relevant tickets?

[DDF-2642](https://codice.atlassian.net/browse/DDF-2642)

#### Checklist:

- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #1483, opened 2016-12-09
2.9.x-DDF-2642: Remove accordion view from sources tab.

djblue:
See #1482.


vinamartin:
Built and tested.
I'm assuming the rest of the bug fixes to this UI were backported. If so, :shipit:

PR #1486, opened 2016-12-10
DDF-2646 registry-schema-bindings won't build when build machine locale is not US

singhdan:


bradh:
That list of commits does not look like what should be reviewed merged here.

catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/helper/InternationalStringTypeHelper.java
```diff
@@ -41,6 +42,11 @@ public InternationalStringTypeHelper() {
     InternationalStringTypeHelper(Locale locale) {
         setLocale(locale);
     }
+    
+    InternationalStringTypeHelper(Locale locale, boolean findNearestMatch) {
```
bradh: This looks like a slightly different problem to what the original JIRA ticket describes, in that this adds new functionality where the original problem is more a bad set of test assumptions.

catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/helper/InternationalStringTypeHelper.java
```diff
@@ -85,13 +91,37 @@ public InternationalStringType create(String internationalizeThis) {
...
                 .filter(localizedString -> locale.toLanguageTag()
                         .equals(localizedString.getLang()))
                 .findFirst()
                 .map(LocalizedStringType::getValue);
+        
+        //If an exact match has not been found then look at the base language e.g. if en-GB 
+        //has not been found then try to find the first one in the list of localizedStrings 
+        //that has the same language e.g. en-US. The behaviour of this will vary depending
+        //on the order that the localized strings are loaded
+        if (!optionalLocalString.isPresent() && findNearestMatch)
```
bradh: This is probably going to produce unexpected behaviour, where en-AU falls back to en-GB in some cases and en-US in other cases. Perhaps specifying a fallback locale would be better than "I'll take anything in that language", which may cause unreadable results (e.g. pt-PT and pt-BR aren't mutually intelligible).

singhdan: Agreed, a fallback would be useful and more deterministic but that would mean that something like Locale.China (zh-cn) would fallback to en-US for example. Should both nearest match and default be implemented for flexibility? Or just put in a default as people should load up the languages they want and that will mean its clear when a localization is missing.

singhdan:
I'm going to close this pull request and try to tidy it up and try again

catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/helper/InternationalStringTypeHelper.java
bradh: I was more thinking along the lines of an API alternative like 

     InternationalStringTypeHelper(Locale locale, Locale fallbackLocale) {...}

So that you can do en-AU if I have it, otherwise en-US will do.

PR #1487, opened 2016-12-11
DDF-2646 Modify tests in registry-schema-bindings to allow them to build when machine is not US

singhdan:
Modified InternationalStringTypeHelperTest and RegistryPackageTypeHelperTest to specifically set Locales and also check for the local of the build machine


coyotesqrl:
I can't be sure without a little investigation, but I believe this is a much bigger change than needed. I'm not sure any change to production code is required at all; rather, I think only the test assumptions need to be updated.

coyotesqrl:
@singhdan if you could please try the attached patch (I had to tack on the `.txt` extension so Github would accept the upload) I believe you'll find it resolves your issue much more simply and safely. If so, please update your PR with those changes.

If I'm misunderstanding the problem you're encountering, be sure to let me know.

[InternationalStringTypeHelper.patch.txt](https://github.com/codice/ddf/files/646299/InternationalStringTypeHelper.patch.txt)

singhdan:
That worked, thanks. But had to modify RegistryPackageTypeHelperTest too

catalog/spatial/registry/registry-schema-bindings/src/test/java/org/codice/ddf/registry/schemabindings/helper/RegistryPackageTypeHelperTest.java
```diff
@@ -24,6 +24,7 @@
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Locale;
```
coyotesqrl: This import is in the wrong order. It should be moved up one line so it is between `List` and `Map` (alphabetically sorted).

Once that is resolved, this PR can be merged.

catalog/spatial/registry/registry-schema-bindings/src/test/java/org/codice/ddf/registry/schemabindings/helper/InternationalStringTypeHelperTest.java
```diff
@@ -51,6 +64,13 @@ public void testGetString() throws Exception {
         InternationalStringType ist = getTestInternationalStringType();
 
         String istString = istHelper.getString(ist);
+
+        if (!isDefaultLocaleInTestSet()) {
```
lessarderic: To avoid this conditional, could we use the `getTestInternationalStringType()` only in this test? We could have a separate test for the default Locale that populates the `InternationalStringType` with an entry for the default Locale and some other undefined Locale and makes sure the string for the default Locale is returned.

I believe this would allow us to remove `isDefaultLocaleInTestSet()` and all the logic related to that check.

catalog/spatial/registry/registry-schema-bindings/src/test/java/org/codice/ddf/registry/schemabindings/helper/RegistryPackageTypeHelperTest.java
singhdan: Moved the import up so it's now alphabetically sorted

catalog/spatial/registry/registry-schema-bindings/src/test/java/org/codice/ddf/registry/schemabindings/helper/InternationalStringTypeHelperTest.java
singhdan: Updated tests to split default Locale into a separate test and remove the conditional

catalog/spatial/registry/registry-schema-bindings/src/test/java/org/codice/ddf/registry/schemabindings/helper/InternationalStringTypeHelperTest.java
```diff
@@ -63,6 +81,14 @@ public void testGetString() throws Exception {
     }
 
     @Test
+    public void testGetStringWithForDefaultLocale() throws Exception {
```
lessarderic: Nit: Would you mind removing the `For` from the method name, i.e., `testGetStringWithDefaultLocale()`?

singhdan:
No problem. Done


________________________________

PR #1490, opened 2016-12-12
DDF-2641 TransactionRequestConverter uses AttributeRegistry to determ…

kcwire:
#### What does this PR do?
The TransactionRequestConverter was previously only using the CswMetacardType to evaluate attribute values. It now uses the AttributeRegistry which contains all attributes available.  There is also a fix for multi-valued attributes (before it would only use the first value in a list)
@jlcsmith 
@glenhein 
[OGC](https://github.com/orgs/codice/teams/OGC)
@coyotesqrl
@kcwire
Use the new test requests in the unit tests and verify additional attributes can be updated on a metacard via CSW. (Postman should make it easy).
[DDF-2641](https://codice.atlassian.net/browse/DDF-2641)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

…ine the attribtue type.

### DIFF
--- a/catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/CswRecordConverter.java
+++ b/catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/CswRecordConverter.java
@@ -80,7 +80,7 @@ public class CswRecordConverter implements Converter, MetacardTransformer, Input
 
     private static XMLInputFactory factory;
 
-    private static MetacardType metacardType;
+    private MetacardType metacardType;
 
     static {
         factory = XMLInputFactory.newInstance();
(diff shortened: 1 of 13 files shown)


catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/CswRecordConverter.java
```diff
@@ -306,16 +306,17 @@ public static Serializable convertStringValueToMetacardValue(
...
-            HierarchicalStreamReader reader, CswAxisOrder cswAxisOrder) {
-        AttributeDescriptor attributeDescriptor =
-                metacardType.getAttributeDescriptor(attributeName);
-        if (attributeDescriptor != null) {
-            return CswUnmarshallHelper.convertRecordPropertyToMetacardAttribute(attributeDescriptor.getType()
-                    .getAttributeFormat(), reader, cswAxisOrder);
-        }
-        return null;
-    }
+    // TODO - remove?
```
coyotesqrl: Did you want to leave these two TODO removes in here for future attention or delete the code now? If leaving them in, should probably generate a Jira spike to determine whether they can safely be removed.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/TransactionRequestConverter.java
```diff
@@ -332,14 +350,10 @@ private Serializable getRecordPropertyValue(HierarchicalStreamReader reader, Str
             Serializable newValue;
             if (reader.hasMoreChildren()) {
```
coyotesqrl: Shouldn't this also be a `while`?

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/TransactionRequestConverter.java
```diff
@@ -350,6 +364,21 @@ private Serializable getRecordPropertyValue(HierarchicalStreamReader reader, Str
         }
     }
 
+    private Serializable readPropertyValue(HierarchicalStreamReader reader, String cswField) {
+        Optional<AttributeDescriptor> descriptor = registry.lookup(cswField);
```
coyotesqrl: Should guard here against a null `registry`.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/TransactionRequestConverter.java
kcwire: no. this element is an `anyType` but it must be constrained within a `<Value>`.  At least how I'm reading the schema the `<Value>` element can repeat but the content of that element cannot.

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/TransactionRequestConverter.java
kcwire: done.

PR #1499, opened 2016-12-14
DDF-2639 Restrict access to updating or creating resource URIs on metacards

ahoffer:
#### What does this PR do?
Allow administrators to allow privileged users to update metacard resource URI or add a resource URI when creating a metacard. 
@brendan-hofmann @mweser @emanns95 @spearskw
@stustison 
@figliold
@stustison
It has been manually tested. 
No.
#### Checklist:
- [x] Documentation Updated
- [x] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
+++ b/catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
@@ -1,0 +1,179 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.security;
+
+import static ddf.catalog.Constants.LOCAL_DESTINATION_KEY;
+import static ddf.catalog.Constants.OPERATION_TRANSACTION_KEY;
+
+import java.io.Serializable;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
+import ddf.catalog.operation.OperationTransaction;
+import ddf.catalog.operation.Query;
+import ddf.catalog.operation.ResourceRequest;
+import ddf.catalog.operation.ResourceResponse;
+import ddf.catalog.plugin.PolicyPlugin;
+import ddf.catalog.plugin.PolicyResponse;
+import ddf.catalog.plugin.StopProcessingException;
+import ddf.catalog.plugin.impl.PolicyResponseImpl;
+import ddf.security.permission.Permissions;
+
+/**
+ * Restricts how resource URIs are updated and created.
+ * There are security risk to allowing users to update URIs as well as allowing users to
+ * pass in a URI when the metacard is created.
+ */
+public class ResourceUriPolicy implements PolicyPlugin {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ResourceUriPolicy.class);
+
+    private String[] createPermissions;
+
+    private String[] updatePermissions;
+
+    public ResourceUriPolicy() {
+    }
+
+    public ResourceUriPolicy(String[] createPermissions, String[] updatePermissions) {
+        setCreatePermissions(createPermissions);
+        setUpdatePermissions(updatePermissions);
+    }
+
+    @Override
+    public PolicyResponse processPreCreate(Metacard input, Map<String, Serializable> properties)
+            throws StopProcessingException {
+
+        if (!isLocalRequest(properties)) {
+            return new PolicyResponseImpl();
+        }
+
+        if (input.getResourceURI() != null && StringUtils.isNotEmpty(input.getResourceURI()
+                .toString())) {
+            return new PolicyResponseImpl(null,
+                    Permissions.parsePermissionsFromString(getCreatePermissions()));
+        }
+
+        return new PolicyResponseImpl();
+    }
+
+    @Override
+    public PolicyResponse processPreUpdate(Metacard input, Map<String, Serializable> properties)
+            throws StopProcessingException {
+
+        if (!isLocalRequest(properties)) {
+            return new PolicyResponseImpl();
+        }
+
+        PolicyResponseImpl policyResponse = new PolicyResponseImpl(null,
+                Permissions.parsePermissionsFromString(getUpdatePermissions()));
+
+        List<Metacard> previousStateMetacards = ((OperationTransaction) properties.get(
+                OPERATION_TRANSACTION_KEY)).getPreviousStateMetacards();
+
+        Metacard previous;
+        try {
+            previous = previousStateMetacards.stream()
+                    .filter((x) -> x.getId()
+                            .equals(input.getId()))
+                    .findFirst()
+                    .get();
+        } catch (NoSuchElementException e) {
+            LOGGER.debug("Cannot locate metacard {} for update. Applying permissions to the item",
+                    input.getId());
+            return policyResponse;
+        }
+
+        return requiresPermission(input.getResourceURI(), previous.getResourceURI()) ?
+                policyResponse :
+                new PolicyResponseImpl();
+    }
+
+    private boolean isLocalRequest(Map<String, Serializable> properties) {
+        return (boolean) properties.get(LOCAL_DESTINATION_KEY);
+    }
+
+    private boolean requiresPermission(URI input, URI catalog) {
+
+        return !uriToString(input).equals(uriToString(catalog));
+    }
+
+    private String uriToString(URI uri) {
+        return uri == null ? "" : uri.toString();
+    }
+
+    @Override
+    public PolicyResponse processPreDelete(List<Metacard> metacards,
+            Map<String, Serializable> properties) throws StopProcessingException {
+
+        return new PolicyResponseImpl();
+    }
+
+    @Override
+    public PolicyResponse processPostDelete(Metacard input, Map<String, Serializable> properties)
+            throws StopProcessingException {
+        return new PolicyResponseImpl();
+    }
+
+    @Override
+    public PolicyResponse processPreQuery(Query query, Map<String, Serializable> properties)
+            throws StopProcessingException {
+
+        return new PolicyResponseImpl();
+    }
+
+    @Override
+    public PolicyResponse processPostQuery(Result input, Map<String, Serializable> properties)
+            throws StopProcessingException {
+        return new PolicyResponseImpl();
+    }
+
+    @Override
+    public PolicyResponse processPreResource(ResourceRequest resourceRequest)
+            throws StopProcessingException {
+
+        return new PolicyResponseImpl();
+    }
+
+    @Override
+    public PolicyResponse processPostResource(ResourceResponse resourceResponse, Metacard metacard)
+            throws StopProcessingException {
+        return new PolicyResponseImpl();
+    }
+
+    public String[] getCreatePermissions() {
+        return createPermissions == null ? null : createPermissions.clone();
+    }
+
+    public void setCreatePermissions(String[] createPermissions) {
+
+        this.createPermissions = createPermissions == null ? null : createPermissions.clone();
+    }
+
+    public String[] getUpdatePermissions() {
+        return updatePermissions == null ? null : updatePermissions.clone();
+    }
+
+    public void setUpdatePermissions(String[] updatePermissions) {
+        this.updatePermissions = updatePermissions == null ? null : updatePermissions.clone();
+    }
+}
--- a/catalog/security/catalog-security-resourceplugin/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/catalog/security/catalog-security-resourceplugin/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -1,0 +1,33 @@
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<blueprint xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0"
+           xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0">
+    <bean id="resourceUriPlugin"
+          class="org.codice.ddf.catalog.security.ResourceUriPolicy">
+        <cm:managed-properties
+                persistent-id="org.codice.ddf.catalog.security.ResourceUriPolicy"
+                update-strategy="container-managed"/>
+        <property name="createPermissions">
+            <array>
+                <value>http://schemas.xmlsoap.org/ws/2005/05/identity/claims/role=guest</value>
+            </array>
+        </property>
+        <property name="updatePermissions">
+            <array>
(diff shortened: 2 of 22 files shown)


catalog/ui/search-ui/standard/pom.xml
```diff
@@ -274,6 +274,18 @@
             </activation>
```
stustison: I don't think you meant to have these doc changes in here. It looks like you're rolling back a patch that was applied yesterday.

distribution/docs/pom.xml
```diff
@@ -57,25 +57,42 @@
                 <activeByDefault>true</activeByDefault>
             </activation>
                 <properties>
+                    <asciidoctor.maven.plugin.version>1.5.3</asciidoctor.maven.plugin.version>
                     <packaging>pom</packaging>
                     <module>docs</module>
                     <skipTests>false</skipTests>
                 </properties>
+                <repositories>
```
stustison: Same with these doc changes.

pom.xml
```diff
@@ -157,11 +158,10 @@
         <ddf-spatial>DDF Spatial</ddf-spatial>
         <ddf-ui>DDF Search UI</ddf-ui>
 
-        <asciidoctor.maven.plugin.version>1.5.3</asciidoctor.maven.plugin.version>
-        <asciidoctorj.version>1.5.4.1</asciidoctorj.version>
-        <asciidoctorj.diagram.version>1.5.0</asciidoctorj.diagram.version>
-        <asciidoctorj.pdf.version>1.5.0-alpha.11</asciidoctorj.pdf.version>
-        <jruby.version>1.7.25</jruby.version>
+        <asciidoctor.maven.plugin.version>1.5.2</asciidoctor.maven.plugin.version>
```
stustison: This should not be changed. You're rolling back a patch that fixed builds in Windows.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -0,0 +1,187 @@
...
+        QueryResponse queryResponse;
+        try {
+            queryResponse = catalogFramework.query(queryRequest);
+        } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            throw new RuntimeException(e);
+        }
+
+        List<Result> results = queryResponse.getResults();
+        if (results.isEmpty()) {
+            throw new StopProcessingException("Cannot locate metacard to be updated");
```
figliold: Would it be helpful to have the metacard id in the exception message?

catalog/ui/search-ui/standard/pom.xml
ahoffer: That's weird. I'll check into it.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -0,0 +1,187 @@
...
+                .is()
+                .equalTo()
+                .text(id);
+        Query query = new QueryImpl(filter, 1, 1, null, true, 10000);
+        QueryRequest queryRequest = new QueryRequestImpl(query);
+        QueryResponse queryResponse;
+        try {
+            queryResponse = catalogFramework.query(queryRequest);
+        } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            throw new RuntimeException(e);
```
ahoffer: will do

catalog/security/catalog-security-resourceplugin/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<blueprint xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.1.0"
+           xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0">
+
```
figliold: Is it okay to have these two service references with availability="optional"?

catalog/security/catalog-security-resourceplugin/src/test/java/org/codice/ddf/catalog/security/ResourceUriPolicyTest.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+        };
+
+        policyPlugin.setCreatePermissions(createPermissionsArray);
+        policyPlugin.setUpdatePermissions(updatePermissionsArray);
+        return policyPlugin;
+    }
+
+    private Metacard getMockMetacard(String inputResourceUri) throws URISyntaxException {
+        Metacard inputMetacard = mock(MetacardImpl.class);
```
figliold: Shouldn't you just be able to mock(Metacard.class) instead?

catalog/security/catalog-security-resourceplugin/src/test/java/org/codice/ddf/catalog/security/ResourceUriPolicyTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        assertThat(itemPolicy.get(key), containsInAnyOrder(value));
+
+    }
+
+    private ResourceUriPolicy getPolicyPlugin(String catalogResourceUri,
+            String[] createPermissionsArray, String[] updatePermissionsArray)
+            throws URISyntaxException, StopProcessingException {
+
+        Metacard catalogMetacard = mock(MetacardImpl.class);
```
figliold: Shouldn't you just be able to mock(Metacard.class) instead?

distribution/docs/src/main/resources/_contents/_plugins/resource-uri-policy-plugin-contents.adoc
```diff
@@ -0,0 +1,18 @@
+==== Resource URI Policy Plugin
+
+The Resource URI Policy Plugin configures the attributes required for users to perform Create, Read, Update, and Delete operations on the catalog.
+
```
figliold: Is this true? Or, does it configure the attributes required to update a metacard's resource uri or create a metacard with a resource uri?

catalog/security/catalog-security-resourceplugin/src/test/java/org/codice/ddf/catalog/security/ResourceUriPolicyTest.java
```diff
@@ -0,0 +1,178 @@
...
+import ddf.catalog.plugin.PolicyResponse;
+import ddf.catalog.plugin.StopProcessingException;
+
+public class ResourceUriPolicyTest {
+
+    private String key;
+
+    @Before
+    public void setUp() throws Exception {
+    }
```
emanns95: Unused, can remove

distribution/docs/src/main/resources/_contents/_plugins/resource-uri-policy-plugin-contents.adoc
brendan-hofmann: It also only applies to Create/Update.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    private boolean requiresPermissionsToChangeFrom(URI input, URI catalog) {
+
+        return !uriToString(input).equals(uriToString(catalog));
+    }
+
+    private String uriToString(URI uri) {
+        return uri == null ? "" : uri.toString();
```
brendan-hofmann: Are `null` and `""` really equivalent? If the `resource-uri` is null on the update metacard, won't that have no effect on the existing `resource-uri`?

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
ahoffer: Done.

catalog/security/catalog-security-resourceplugin/src/test/java/org/codice/ddf/catalog/security/ResourceUriPolicyTest.java
ahoffer: That worked. I didn't realize you could mock out interfaces.

distribution/docs/src/main/resources/_contents/_plugins/resource-uri-policy-plugin-contents.adoc
ahoffer: Fixed.

catalog/security/catalog-security-resourceplugin/src/test/java/org/codice/ddf/catalog/security/ResourceUriPolicyTest.java
ahoffer: Done

catalog/security/catalog-security-resourceplugin/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -0,0 +1,34 @@
...
+ *
+ **/
+-->
+<metatype:MetaData xmlns:metatype="http://www.osgi.org/xmlns/metatype/v1.0.0">
+
+    <OCD description="Resource URI Policy Plugin"
+         name="Resource URI Policy Plugin"
+         id="org.codice.ddf.catalog.security.ResourceUriPolicy">
+
+        <AD description="Allow users to provide a URI when creating a metacard"
```
ahoffer: Done

catalog/security/catalog-security-resourceplugin/src/test/java/org/codice/ddf/catalog/security/ResourceUriPolicyTest.java
ahoffer: Converted to Metacard

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
ahoffer: I did some testing. Here is what I found:

Setting the `resource-uri` attribute to `null` causes an exception:

```
Caused by: java.lang.NullPointerException
	at ddf.catalog.data.impl.AttributeImpl.getValue(AttributeImpl.java:112)[357:catalog-rest-endpoint:2.10.0.SNAPSHOT]
	at ddf.catalog.data.impl.MetacardImpl.<init>(MetacardImpl.java:163)[357:catalog-rest-endpoint:2.10.0.SNAPSHOT]
	at ddf.catalog.source.solr.SolrCatalogProvider.copyMetacard(SolrCatalogProvider.java:533)
	at ddf.catalog.source.solr.SolrCatalogProvider.lambda$create$0(SolrCatalogProvider.java:292)
```

Removing the `resource-uri` from the map of attributes to update removes the `resource-uri` from the Solr index. To test this I put a breakpoint on line 416 of SolrCatalogProvider and removed the `resource-uri` from the metacard's attributes. 

>  {
>         "metacard-tags_txt":["resource",
>           "VALID"],
>         "resource-download-url_txt":["https://localhost:8993/services/catalog/sources/dib.distribution/443903e2e0c94a7d9ee7fa4cb8e959ed?transform=resource"],
>         "metadata_xml":["</intelProduct>"],
>         "id_txt":"443903e2e0c94a7d9ee7fa4cb8e959ed",
>         "checksum-algorithm_txt":["Adler32"],
>         "resource-size_txt":["7423"],
>         "created_tdt":["2016-12-15T18:24:57.367Z"],
>         "metacard.modified_tdt":["2016-12-15T18:34:20.498Z"],
>         "metacard.created_tdt":["2016-12-15T18:24:57.356Z"],
>         "checksum_txt":["6e2cb1b3"],
>         "title_txt":["PemFile.java.bava"],
>         "modified_tdt":["2016-12-15T18:34:20.677Z"],
>         "lux_xml":"A==",
>         "metacard_type_name_txt":["common"],
>         "metacard_type_obj": [],
>         "_version_":1553808126885494784}

The results are that if the `resource-uri` attribute is missing from the update request, or the value of the `resource-uri` attribute is null, the `resource-uri` of the metacard will be altered. Therefore, the user should have the correct privileges to update the `resource-uri` if that attribute is missing or if the attribute's value is null.

The conclusion is that an update missing the `resource-uri` attribute should be treated as an update and require permission. The same is true if the attribute is present, but it's value is null, however, an exception will be generated that prevents the update from being written to the Solr index.

brendan-hofmann: Good science. I agree with your conclusions. 👍

ahoffer:
This PR causes a race condition to happen in several itests. Prior to this PR, the update process did not retrieve the existing metacard from the catalog and compare its values to the values sent with the update. 

Different itests now fail, or sometimes do not fail with the message:

ddf.catalog.plugin.StopProcessingException: Cannot locate metacard to update, ID=2014ca7f59ac46f495e32b4a67a51290

The most likely explanation is the itest creates a metacard and then immediately tries to update it, before the metacard is fully in the catalog. I have seen several tests fail. For example, two tests failed in FanoutTest, but passed on the next run. In the Bamboo build, the test TestRegistryMethod.testCswRegistryStoreUpdate failed, something that never happened in my local tests.

Trying to patch every itests that creates and then updates metacard is difficult and it is easy to miss places where this happens. I think the best solution it to put a retry in the method that, as part of the update, retrieves the metacard from the catalog.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -0,0 +1,200 @@
...
+    public String[] getUpdatePermissions() {
+        return updatePermissions == null ? null : updatePermissions.clone();
+    }
+
+    public void setUpdatePermissions(String[] updatePermissions) {
+        this.updatePermissions = updatePermissions == null ? null : updatePermissions.clone();
+    }
+
+    protected Metacard getMetacardFromCatalog(String id) throws StopProcessingException {
+        Filter filter = filterBuilder.attribute(Metacard.ID)
```
clockard: You shouldn't need to query the catalog on an update. The metacard(s) getting updated have all ready been pulled from the catalog and stored in passed in properties under the key Constants.OPERATION_TRANSACTION_KEY. The OperationTransaction object holds the current metacards.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private CatalogFramework catalogFramework;
+
+    private FilterBuilder filterBuilder;
+
+    @Override
+    public PolicyResponse processPreCreate(Metacard input, Map<String, Serializable> properties)
+            throws StopProcessingException {
+
+        if (input.getResourceURI() != null && StringUtils.isNotEmpty(input.getResourceURI()
```
clockard: For all each of the PolicyPlugin methods you are going to want to check if the request is local. If the change is not local then I don't think this plugin needs to do anything. `Requests.isLocal(properties) ` should return true if it is a local request.

catalog/security/catalog-security-resourceplugin/src/main/resources/OSGI-INF/blueprint/blueprint.xml
clockard: should be able to remove both from the blueprint once the plugin is updated to use the OperationTransaction instead of querying for the metacard.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
```diff
@@ -13,6 +13,7 @@
  */
```
ahoffer: My code formatter was to "reorganize", so the file diff looks chopped-up. Something about my PR changes the runtime environment enough so that these tests fail, unless you set a breakpoint somewhere. In debug mode, the tests pass. A ticket was created for this issue by @vinamartin (DDF-2670).

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -0,0 +1,179 @@
...
+                    input.getId());
+            return policyResponse;
+        }
+
+        return requiresPermission(input.getResourceURI(), previous.getResourceURI()) ?
+                policyResponse :
+                new PolicyResponseImpl();
+    }
+
+    private boolean isLocalRequest(Map<String, Serializable> properties) {
```
clockard: Is there a reason you created this method instead of directly using `Requests.isLocal(Map<String, Serializable> props)`

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
clockard: This is kind of concerning that we are removing all of these registry itests.

PR #1495, opened 2016-12-13
DDF-2569: Add default implementations for aysnchronous plugin API

peterhuffer:
#### What does this PR do?
~_Note:  https://github.com/codice/ddf/pull/1503, which should be reviewed first._~ _Closed_
~_https://github.com/codice/ddf/pull/1602 is required before merge._~ _Closed_

- Provides default implementations for the API. The `InMemoryProcessingFramework` will process the request and submit its results back to the Catalog.
- Adds a PostIngestPlugin that will kick off the asynchronous processing by submitting jobs to the `ProcessingFramework`.

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@jrnorth @troymohl @emmberk @brendan-hofmann 

#### Select at least one member from relevant component team(s) from below (at least one component team member needs to approve the PR).
[Security](https://github.com/orgs/codice/teams/security)
[Core APIs](https://github.com/orgs/codice/teams/core-apis)

#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@lessarderic
@pklinef 

#### How should this be tested? (List steps with links to updated documentation)
Full build. Ensure there are no errors when manually installing the `catalog-plugin-async` feature in the `DDF Catalog` app.
There are currently no `PostProcessPlugin`s available for testing. There will be a future ticket to migrate long running plugins to be `PostProcessPlugin`s, namely the NITF thumbnail generation plugin.

#### Any background context you want to provide?
There was a need for long running processes to be done asynchronously in order to decrease the amount of time it takes to process large amounts of data.

This ticket relies on [DDF-2647](https://github.com/codice/ddf/pull/1491) and [DDF-2571](https://github.com/codice/ddf/pull/1503).

#### What are the relevant tickets?
This PR is many tickets in one. [DDF-2576](https://codice.atlassian.net/browse/DDF-2576) is the Epic.
#### Checklist:
- [x] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #1503, opened 2016-12-15
DDF-2571: Add API to support asynchronous plugins

peterhuffer:
#### What does this PR do?
- Adds an internal experimental API for asynchronous processing plugins.

@troymohl @emmberk @brendan-hofmann

@lessarderic
@pklinef

Full build.

There was a need for long running processes to be done asynchronously in order to decrease the amount of time it takes to process large amounts of data.

This PR is a prerequisite to [DDF-2569](https://github.com/codice/ddf/pull/1495) and was split up to make smaller PRs for easier reviews.

This PR is many tickets in one. [DDF-2576](https://codice.atlassian.net/browse/DDF-2576) is the Epic.
[DDF-2569](https://codice.atlassian.net/browse/DDF-2569)
[DDF-2571](https://codice.atlassian.net/browse/DDF-2571)
[DDF-2573](https://codice.atlassian.net/browse/DDF-2573)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/async/catalog-async-data-api/pom.xml
+++ b/catalog/async/catalog-async-data-api/pom.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>async</artifactId>
+        <groupId>ddf.catalog.async</groupId>
+        <version>2.10.0-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>catalog-async-data-api</artifactId>
+    <name>Catalog :: Async :: Data :: API</name>
+    <packaging>bundle</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>ddf.catalog.core</groupId>
+            <artifactId>catalog-core-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Name>${project.name}</Bundle-Name>
+                        <Export-Package>
+                            org.codice.ddf.catalog.async.data.api*
+                        </Export-Package>
+                    </instructions>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.0</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.0</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.0</minimum>
+                                        </limit>
+                                    </limits>
+                                </rule>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/catalog/async/catalog-async-data-api/pom.xml
+++ b/catalog/async/catalog-async-data-api/pom.xml
@@ -45,7 +45,7 @@
                         <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
                         <Bundle-Name>${project.name}</Bundle-Name>
                         <Export-Package>
-                            org.codice.ddf.catalog.async.data.api*
+                            org.codice.ddf.catalog.async.data.api.internal.*
                         </Export-Package>
                     </instructions>
                 </configuration>
--- a/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessRequest.java
+++ b/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessRequest.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.async.data.api.internal;
+
+import java.util.List;
+
+import ddf.catalog.operation.Operation;
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ */
+
+/**
+ * The {@code ProcessRequest} is a request that is capable of being processed by the {@link ProcessingFramework}.
+ *
+ * @param <T> A {@link ProcessItem} that contains the data to be processed.
+ */
+public interface ProcessRequest<T extends ProcessItem> extends Operation {
+
+    /**
+     * Gets the {@link ProcessItem} that will contain the data required for processing by the {@link ProcessingFramework}.
+     * The available {@link ProcessItem}s are:
+     * <ul>
+     *   <li>{@link ProcessCreateItem} which contains a {@link ddf.catalog.data.Metacard} and {@link ProcessResource}</li>
+     *   <li>{@link ProcessUpdateItem} which contains an updated {@link ddf.catalog.data.Metacard}, the
+     *   original {@link ddf.catalog.data.Metacard}, and the {@link ProcessResource}</li>
+     *   <li>{@link ProcessDeleteItem} which contains the {@link ddf.catalog.data.Metacard} being deleted</li>
+     * </ul>
+     *
+     * @return a {@link List} of {@link ProcessItem}s
+     */
+    List<T> getProcessItems();
+}
--- a/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessRequest.java
+++ b/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessRequest.java
@@ -25,17 +25,18 @@ import ddf.catalog.operation.Operation;
  */
 
 /**
- * The {@code ProcessRequest} is a request that is capable of being processed by the {@link ProcessingFramework}.
+ * The {@code ProcessRequest} is a request that is capable of being processed by {@link PostProcessPlugin}s
+ * in a {@link ProcessingFramework}.
  *
  * @param <T> A {@link ProcessItem} that contains the data to be processed.
  */
 public interface ProcessRequest<T extends ProcessItem> extends Operation {
 
     /**
-     * Gets the {@link ProcessItem} that will contain the data required for processing by the {@link ProcessingFramework}.
+     * Gets the {@link ProcessItem}s that contain the data required for processing by the {@link ProcessingFramework}.
      * The available {@link ProcessItem}s are:
      * <ul>
-     *   <li>{@link ProcessCreateItem} which contains a {@link ddf.catalog.data.Metacard} and {@link ProcessResource}</li>
+     *   <li>{@link ProcessCreateItem} which contains a {@link ddf.catalog.data.Metacard} and the {@link ProcessResource}</li>
      *   <li>{@link ProcessUpdateItem} which contains an updated {@link ddf.catalog.data.Metacard}, the
      *   original {@link ddf.catalog.data.Metacard}, and the {@link ProcessResource}</li>
      *   <li>{@link ProcessDeleteItem} which contains the {@link ddf.catalog.data.Metacard} being deleted</li>
--- a/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
+++ b/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
@@ -15,6 +15,7 @@ package org.codice.ddf.catalog.async.data.api.internal;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 
 /**
  * <p>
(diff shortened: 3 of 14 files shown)


peterhuffer:
@ricklarsen Do I nee

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessUpdateItem.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return the original resource
+     */
+    ProcessResource getProcessResource();
+
+    /**
+     * Determines if the new {@code Metacard} was modified during processing.
+     *
+     * @return {@code true} if modified, {@code false} otherwise
+     */
+    boolean isNewMetacardModified();
```
pklinef: I would suggest making this consistent with `ProcessResource` and rename these functions to `isModifed` and `markAsModified` for consistency.  The JavaDoc makes it clear that the state of the new metacard is the state being tracked for this object.

peterhuffer: Updated.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessingFramework.java
```diff
@@ -0,0 +1,58 @@
...
+ *  <li>{@link ProcessUpdateRequest}</li>
+ * </ul>
+ */
+public interface ProcessingFramework {
+
+    /**
+     * A property name to indicate that post processing has already been completed on the request
+     * in question.
+     */
+    String POST_PROCESS_COMPLETE = "post-process-complete";
```
lessarderic: This feels like an implementation detail.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessingFramework.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Submits a {@code ProcessCreateRequest} to be processed by the {@code ProcessingFramework}.
+     *
+     * @param input the {@code ProcessCreateRequest} to be processed
+     */
+    void submit(ProcessCreateRequest input);
```
lessarderic: If we use more generic and common interfaces, we may be able to simplify this interface down to something like this:

```
public interface ProcessingFramework {
    void submit(ProcessRequest<ProcessDeleteItem> input);
    void submit(ProcessRequest<ProcessResourceItem> input);
}
```

The question here is, will the implementations ever care if the `ProcessResourceItem` is a `ProcessCreateItem` or a `ProcessUpdateItem`. The current implementation doesn't and pretty much repeats the same code for both. That's not to say that the interface should be changed because of that. Just something to consider.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessingFramework.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.codice.ddf.catalog.plugin.async.api.internal;
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ *
+ * The {@code ProcessingFramework} processes all requests submitted to it. Available requests for processing
```
lessarderic: Given the number of interfaces, we may want to provide more information on how this one relates to the others and who's going to be calling it.

Also, I wonder if this interface should be part of the API bundle as it is not directly related to the other plug-in interface implementations. We could also consider putting it in its own API bundle, register the In-Memory implementation as an OSGi service and have the post-ingest plug-in look it up and calling it.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,101 @@
...
+     * Determines if the {@code ProcessResource} has been modified during processing.
+     *
+     * @return {@code true} if modified, {@code false} otherwise
+     */
+    boolean isModified();
+
+    /**
+     * Marks the {@code ProcessResource} as modified.
+     */
+    void markAsModified();
```
peterhuffer: The purpose of this and other like methods was that, since the `PostProcessPlugin`s may not touch or modify anything during their processing, we needed a way to show that. This allowed us not to make any calls back to the `CatalogFramework`, such as new creates or updates. Our specific example was, even though we had one `PostProcessPlugin` that touched only NITFs products, we were sending erroneous requests back to the `Catalog` for products like a JPG.

lessarderic: I understand that this might be useful in some circumstances. I just want to make sure it needs to be part of the API's interfaces in this case. In general, it's better to let the implementors decide whether they want their classes to be immutable or not and not force them a certain way. It is not always possible, but preferable.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessCreateItem.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ *
+ * The {@code ProcessCreateItem} represents the data that will be processed by the {@link ProcessingFramework}
+ * when handling a {@link ProcessCreateRequest}.
+ */
+public interface ProcessCreateItem {
```
pklinef: It crossed my mind while reviewing and seeing some of the duplication but I did not take the time to see if it would make sense like @lessarderic did.  That looks like an nice improvement to me.  Might be able to remove some duplicated code once you do this.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessingFramework.java
brendan-hofmann: I don't think I agree. It's a good practice to separate API and implementation from the get-go, even for experimental code.

It's fine if experimental APIs live in separate bundles, but please don't bundle the API and impl together. That quickly leads to a dependency nightmare as clients become unable to distinguish what they actually depend on, and as a result it becomes too easy to couple to implementation details instead of interfaces.

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessRequest.java
```diff
@@ -0,0 +1,47 @@
...
+ * @param <T> A {@link ProcessItem} that contains the data to be processed.
+ */
+public interface ProcessRequest<T extends ProcessItem> extends Operation {
+
+    /**
+     * Gets the {@link ProcessItem} that will contain the data required for processing by the {@link ProcessingFramework}.
+     * The available {@link ProcessItem}s are:
+     * <ul>
+     *   <li>{@link ProcessCreateItem} which contains a {@link ddf.catalog.data.Metacard} and {@link ProcessResource}</li>
+     *   <li>{@link ProcessCreateItem} which contains an updated {@link ddf.catalog.data.Metacard}, the
```
troymohl: ProcessUpdateItem?

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessRequest.java
```diff
@@ -0,0 +1,47 @@
...
+     *   original {@link ddf.catalog.data.Metacard}, and the {@link ProcessResource}</li>
+     *   <li>{@link ProcessCreateItem} which contains the {@link ddf.catalog.data.Metacard} being deleted</li>
```
troymohl: ProcessDeleteItem?

peterhuffer:
Updated to reflect comments.
Refactored the api and put into separate bundles.
Expanded on some of the javadoc.

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessUpdateItem.java
```diff
@@ -0,0 +1,36 @@
...
+import ddf.catalog.data.Metacard;
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ * <p>
+ *
+ * The {@code ProcessCreateItem} represents the data that will be processed by the {@link PostProcessPlugin}s
```
troymohl: ProcessUpdateItem

catalog/async/catalog-async-processingframework-api/src/main/java/org/codice/ddf/catalog/async/processingframework/api/internal/ProcessingFramework.java
```diff
@@ -0,0 +1,63 @@
...
+import org.codice.ddf.catalog.async.data.api.internal.ProcessRequest;
+import org.codice.ddf.catalog.async.data.api.internal.ProcessUpdateItem;
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ * <p>
+ * The {@code ProcessingFramework} processes all requests submitted to it. In the default implementation,
```
troymohl: Probably would leave default implementation information out of the interface javadoc.

catalog/plugin/catalog-plugin-async/src/main/java/org/codice/ddf/catalog/plugin/async/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ *
+ * A {@code ProcessResource} represents a local or remote resource that will be used during processing
+ * by the {@link ProcessingFramework}.
+ */
+public interface ProcessResource {
```
peterhuffer: We needed this to be able to be easily translatable back to a `ContentItem`, so it resembles it in a lot of ways. The reason we didn't use `ContentItem` is because a `ProcessResource` may be remote, while a `ContentItem` should always be local. Made some modifications to this interface, but I think we are going to leave it largely as is in the interest of time and since it is experimental.

catalog/async/catalog-async-data-api/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Name>${project.name}</Bundle-Name>
+                        <Export-Package>
+                            org.codice.ddf.catalog.async.data.api*
```
lessarderic: Should we add a `.` between `api` and `*`?

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,88 @@
...
+ * removed in a future version of the library. </b>
+ * </p>
+ *
+ * <p>
+ * A {@code ProcessResource} represents a local or remote resource that will be used during processing
+ * by the {@link ProcessingFramework}.
+ */
+public interface ProcessResource {
+
+    String DEFAULT_MIME_TYPE = "application/octet-stream";
```
lessarderic: Is there any value/reason we're defining this in the interface?

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+    String DEFAULT_FILE_NAME = "content_store_file.bin";
```
lessarderic: ⭐️ We may want to move this constant to the class since this is an implementation detail.

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Gets a URI that represents the resource in the catalog.
+     *
+     * @return the URI of the {@link ProcessResource}
+     */
+    String getUri();
```
lessarderic: ⭐️ Can't we return an `URI` instead of a `String` here?

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * An optional field to provide additional information about a {@link ProcessResource}
```
lessarderic: ⭐️ We may want to be more specific here about what "additional information" means here since different implementers would need to know what to return.

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
```diff
@@ -0,0 +1,88 @@
...
+     * @return the qualifier of the {@link ProcessResource}
+     */
+    String getQualifier();
+
+    /**
+     * The filename of the {@link ProcessResource}. If not set it will default to {@link #DEFAULT_FILE_NAME}
+     *
+     * @return the filename of the {@link ProcessResource}
+     */
+    String getFilename();
```
lessarderic: ⭐️ If the interface _must_ return the file information (which I wish it didn't have to 😉), could we have a `File getFile()` method instead of a `String getFilename()`?

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResourceItem.java
```diff
@@ -0,0 +1,39 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.async.data.api.internal;
+
+/**
+ * The {@code ProcessResourceItem} represents the the data that will be processed by the {@link ProcessingFramework}.
```
lessarderic: Typo: _the_ repeated twice.

catalog/async/catalog-async-plugin-api/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Name>${project.name}</Bundle-Name>
+                        <Export-Package>
+                            org.codice.ddf.catalog.async.plugin.api*
```
lessarderic: `.` between `api` and `*`?

catalog/async/catalog-async-processingframework-api/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Name>${project.name}</Bundle-Name>
+                        <Export-Package>
+                            org.codice.ddf.catalog.async.processingframework.api.internal*
```
lessarderic: `.` between `internal` and `*`?

catalog/async/catalog-async-plugin-api/src/main/java/org/codice/ddf/catalog/async/plugin/api/internal/PostProcessPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+ * A {@code PostProcessPlugin} represents a process that will be run asynchronously by the {@link ProcessingFramework}.
+ * The {@code PostProcessPlugin}s may only be run after the {@link ddf.catalog.data.Metacard}s have been
+ * ingested into the catalog. This will allow the {@code PostProcessPlugin}s to be able to use a copy
+ * of the {@link ddf.catalog.data.Metacard}, and optionally a copy of the associated {@link ddf.catalog.resource.Resource}
+ * of the catalog-stored record, in order to perform processing asynchronously.
+ */
+public interface PostProcessPlugin {
+
+    /**
+     * Submits a {@link ProcessRequest<ProcessCreateItem>} to be processed by the {@code PostProcessPlugin}.
```
lessarderic: We could use `{@link PostProcessPlugin}` instead of `{@code ...}`. It usually helps IDEs update the reference if the name is every changed.

catalog/async/catalog-async-data-api/pom.xml
peterhuffer: I followed the example in `catalog-core-api`. Not sure if it makes a difference in this case.

catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/ProcessResource.java
lessarderic: After sleeping on it, I don't think this comment makes a whole lot of sense. However, since the file name is currently only used when creating the `ContentItemImpl` (https://github.com/codice/ddf/pull/1495/files#diff-c5491fd87e13469fa7535a1005fb7c9fR226), and given that `ContentItemImpl` has a constructor that doesn't take a file name, it looks like we could do without this method in the interface.

peterhuffer: Ended up changing the `getFilename()` method to `getName()` since plugins can create and name content that does not yet exist in the catalog.

lessarderic:
Successfully built and ran locally. Bamboo and CI Pipeline failing because or known and unrelated build issues. Merging.

PR #1533, opened 2017-01-06
DDF-2423 Updated TestCatalog cache itest to reduce flakiness

mcalcote:
#### What does this PR do?
Changes the testCachedContentLengthHeader itest in TestCatalog to clear the catalog's cache before the test is run and also ensure the cache is queried. 

@rzwiefel @clockard @gordocanchola @tbatie 

@coyotesqrl
@stustison

Run TestCatalog#testCachedContentLengthHeader

This test has been reported as flaky a few times in the past. With these changes hopefully the test will fail more accurately. 

[DDF-2423](https://codice.atlassian.net/browse/DDF-2423)

screen shot 2017-01-06 at 10 14 41 am

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
@@ -83,6 +83,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpStatus;
 import org.codice.ddf.itests.common.AbstractIntegrationTest;
+import org.codice.ddf.itests.common.KarafConsole;
 import org.codice.ddf.itests.common.annotations.BeforeExam;
 import org.codice.ddf.itests.common.catalog.CatalogTestCommons;
 import org.codice.ddf.itests.common.config.UrlResourceReaderConfigurator;
@@ -135,6 +136,10 @@ public class TestCatalog extends AbstractIntegrationTest {
 
     private static final String DEFAULT_URL_RESOURCE_READER_ROOT_RESOURCE_DIRS = "data/products";
 
+    private static KarafConsole console;
+
+    private static final String CLEAR_CACHE = "catalog:removeall -f -p --cache";
+
     @Rule
     public TestName testName = new TestName();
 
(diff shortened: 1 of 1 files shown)


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -135,6 +136,10 @@
 
     private static final String DEFAULT_URL_RESOURCE_READER_ROOT_RESOURCE_DIRS = "data/products";
 
+    private static KarafConsole console;
+
+    private static final String CLEAR_CACHE = "catalog:removeall -f --cache";
```
rzwiefel: I wonder if we should use `-f -p` here  (using provider should wipe metacards of any tag instead of just resource tags)

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -887,7 +896,7 @@ public void testCachedContentLengthHeader() throws IOException {
                 .headers();
 
         //Get again to hit the cache
-        get(url).then()
+        get(url + "&mode=cache").then()
```
clockard: Shouldn't it automatically hit the cache the second time without the need for the `mode=cache` parameter?

mcalcote: Perhaps the itest doesn't have the same flow as a fully running DDF? I would guess the queries are formed differently. 

When I traced through CachingFederationStrategy during this test, both queries did not have a `QUERY_MODE` property set, so it looked like the cache wasn't being considered either time.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
mcalcote: Updated

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -863,6 +869,9 @@ public void testUpdateContentResourceUri() throws IOException {
 
     @Test
     public void testCachedContentLengthHeader() throws IOException {
+
+        console.runCommand(CLEAR_CACHE);
```
tbatie: Is there any type of wait we can do to assure the cache has been cleared before continuing the test?

rzwiefel: I think the karaf commands are synchronous, may be worth double checking

mcalcote: Karaf commands should be executed synchronously, however I know solr sometimes has a very brief delay on operations not involving a specific metacard id. Clearing the cache here may even be somewhat excessive considering we are adding the content and then operating with its metacard id specifically. 

Since we don't have a definitive reason for the test instability, I added this more as a precaution.

clockard:
Verified Build :shipit:

PR #1539, opened 2017-01-06
DDF-2692 Update revert action to refresh result data differently

andrewkfiedler:
- Previously the revert action was using a response from the endpoint.  However, this response could not be used directly as special parsing is done for results in the result collection.  Specifically, the thumbnail gets converted to a URI.
 - Now the metacard is immediately tagged as a revision, pending refresh from the cql endpoint.  If the refresh takes too long, the user is pushed back to the default view with a warning instructing them that they'll receive write control again once the data returns (if it makes sense based on the returned data).
 - Updated single metacard inspector view to throttle rather than debounce refreshes to current metacard data.  This makes for a more consistent experience if the refresh happens to come back a ms or so too late and the warning pops up.  If we didn't do this, they'd see the warning but wouldn't be affected by anything (they'd stay on the history view instead of being kicked back to the default view).  This should minimize confusion in those edge scenarios.
 - Updated a few views to no longer rerender on all model updates, as this can cause issues if the view happens to be destroyed.

@djblue 
@rzwiefel 
@pklinef
@andrewkfiedler 

Upload an item with a thumbnail.  Then search for the item, or expand it.  Give it some history by editing an attribute.  Now go to the history view and revert to the previous version. 

https://codice.atlassian.net/browse/DDF-2692

Before: https://gfycat.com/DazzlingEuphoricBandicoot
After: https://gfycat.com/CompetentInfamousCockerspaniel
After (high latency): https://gfycat.com/CanineTameBluefish

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/metacard-actions/metacard-actions.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/metacard-actions/metacard-actions.view.js
@@ -30,9 +30,6 @@ define([
         className: 'is-list',
         template: template,
         tagName: CustomElements.register('metacard-actions'),
-        modelEvents: {
-            'all': 'render'
-        },
         regions: {
             mapActions: '.map-actions'
         },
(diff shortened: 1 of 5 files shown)


rzwiefel:
Hero build successful.

rzwiefel:
Test successful.

PR #1541, opened 2017-01-07
DDF-2596 releases contain dashes in the version not deploying

ryeats:
#### What does this PR do?
A recent merge caused issues when creating a release for reflex because we used a dash after the version number since it was an out of band release.
@clockard @brendan-hofmann
@ahoffer 
@figliold
@shaundmorris
unit and itests
when trying to do an out of band release with a dash in the version this issue caused application deployment to fail.
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
+++ b/platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
@@ -113,7 +113,7 @@ public class ApplicationImpl implements Application, Comparable<Application> {
         int numberOfParts;
 
         // DDF-2596
-        repoNameParts = repoName.split("-(?=[0-9])");
+        repoNameParts = repoName.split("-(?=[0-9])", 2);
         numberOfParts = repoNameParts.length;
         switch (numberOfParts) {
         case 1:
@@ -124,9 +124,8 @@ public class ApplicationImpl implements Application, Comparable<Application> {
             name = repoNameParts[0];
             version = repoNameParts[1];
             break;
-        case 3:
-            throw new RuntimeException(String.format("Could not tokenize repository name of '%s'",
-                    repoName));
+        default:
+            LOGGER.error("Could not tokenize repository name of '%s'", repoName);
         }
     }
 
--- a/platform/admin/core/admin-core-appservice/src/test/java/org/codice/ddf/admin/application/service/impl/ApplicationImplTest.java
+++ b/platform/admin/core/admin-core-appservice/src/test/java/org/codice/ddf/admin/application/service/impl/ApplicationImplTest.java
@@ -23,7 +23,6 @@ import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import java.io.IOException;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.Set;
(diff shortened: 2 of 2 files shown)


platform/admin/core/admin-core-appservice/src/test/java/org/codice/ddf/admin/application/service/impl/ApplicationImplTest.java
```diff
@@ -108,10 +108,21 @@ public void testMissingVersionInName() throws Exception {
     @Test(expected = Exception.class)
     public void testUnparsableApplicationName() throws IOException {
         Repository repo = mock(Repository.class);
-        when(repo.getName()).thenReturn("MyName-0.0.0-1.1");
```
brendan-hofmann: Not sure why you changed this test case? It looks like it should work with the above change, if I'm reading the [String.split()](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#split(java.lang.String,%20int)) documentation correctly. The result should be `["MyName", "0.0.0-1.1"]`.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
```diff
@@ -124,7 +124,7 @@ private void setRepoNameAndVersion(String repoName) {
             name = repoNameParts[0];
             version = repoNameParts[1];
             break;
-        case 3:
+        default:
             throw new RuntimeException(String.format("Could not tokenize repository name of '%s'",
```
brendan-hofmann: I feel that this is a little too rigid now. A malformed (according to us) repo name shouldn't throw a `RuntimeException`. I would rather log a warning and use some sensible defaults instead.

platform/admin/core/admin-core-appservice/src/test/java/org/codice/ddf/admin/application/service/impl/ApplicationImplTest.java
ryeats: the test is UnparsableApplicationName and it started failing because it is parseable after my change so I fixed it to fail again.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationImpl.java
ryeats: default is equivalent to case: 0 and I could rename it to that, it would probably would fail before this point because the feature name would be empty and that should never happen. I can change it to log an error since it was really annoying that this causes the app to fail to deploy.

platform/admin/core/admin-core-appservice/src/test/java/org/codice/ddf/admin/application/service/impl/ApplicationImplTest.java
brendan-hofmann: Sorry, poorly phrased. Since the implementation was updated to handle `MyName-0.0.0-1.1`, I would like to still see a test for that, except it would pass now of course. :)

ryeats: Agreed, see testComplexApplicationName

brendan-hofmann: Ah, I see it now.

shaundmorris:
@ryeats was there a specific app name driving this change?

ryeats:
All apps were failing to deploy when I made an out of band release for one of our upstream projects and used a dash in the version number to denote that it wasn't a real release.

PR #1545, opened 2017-01-09
DDF-2695 Added "Rights" taxonomy attributes to better support neutralizing data

emanns95:
#### What does this PR do?
Adds:
`rights.privacy` to `ddf.rights`
`rights.intellectual-property` to `ddf.rights`
`rights.copyright` to `ddf.rights`

to better support a neutralized taxnonomy compatible with more data outputs.

@brendan-hofmann 
@ahoffer 

@lessarderic
@shaundmorris

Full build

[#2695](https://codice.atlassian.net/browse/DDF-2695)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
@@ -171,12 +171,6 @@ public class CoreAttributes implements Core, MetacardType {
                 true /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.STRING_TYPE));
-        descriptors.add(new AttributeDescriptorImpl(DATA_STORE,
-                true /* indexed */,
-                true /* stored */,
-                false /* tokenized */,
-                true /* multivalued */,
-                BasicTypes.STRING_TYPE));
         DESCRIPTORS = Collections.unmodifiableSet(descriptors);
     }
 
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
@@ -63,6 +63,12 @@ public class CoreAttributes implements Core, MetacardType {
                 false /* tokenized */,
                 false /* multivalued */,
                 BasicTypes.DATE_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(INFO_CUT_OFF,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                false /* multivalued */,
+                BasicTypes.DATE_TYPE));
         descriptors.add(new AttributeDescriptorImpl(ID,
                 true /* indexed */,
                 true /* stored */,
@@ -169,6 +175,12 @@ public class CoreAttributes implements Core, MetacardType {
                 true /* indexed */,
                 true /* stored */,
                 true /* tokenized */,
+                true /* multivalued */,
+                BasicTypes.STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(DATA_STORE,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.STRING_TYPE));
         DESCRIPTORS = Collections.unmodifiableSet(descriptors);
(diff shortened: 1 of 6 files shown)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java
```diff
@@ -171,6 +177,12 @@
                 true /* tokenized */,
                 true /* multivalued */,
                 BasicTypes.STRING_TYPE));
+        descriptors.add(new AttributeDescriptorImpl(DATA_STORE,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                false /* multivalued */,
```
shaundmorris: I assume a metacard cannot be in multiple data stores?

emanns95: Perhaps it could be, actually, in some use cases I can imagine. It may be better to make it multi-valued just in case.

emanns95:
@shaundmorris I don't think this is happening for any of the other attributes. Since the values would really be determined by how DDF is being integrated, I'm not sure how much value would be added with example values. I hope it'll be more clear when I update the confluence, which will happen after this PR.

catalog/core/catalog-core-api/src/main/java/ddf/catalog/data/types/Security.java
```diff
@@ -31,4 +31,22 @@
...
+
+    /**
+     * {@link ddf.catalog.data.Attribute} name for identifying whether the resource contains personally
+     * identifying data.
+     */
+    String PRIVACY = "security.privacy";
+
+    /**
+     * {@link ddf.catalog.data.Attribute} name for identifying whether the resource contains data under
+     * protection from reproduction and/or distribution without permission of the rights owner.
```
AzGoalie: Is this supposed to be the same as copyright?

emanns95:
@ahoffer @AzGoalie Just wanted to note the descriptions for intellectual property and copyright make a key distinction at the end between being the "rights" owner and the "copyright" owner. The rights owner can apply to other aspects such as trademark or patents.

We wanted to keep the terms somewhat vague so that it would be more useful for more applications (concerning dataStore and privacy). Make it too specific and we'll find ourselves quickly needing to add more attributes to properly map other formats. That being said, if we want to make them more specific, then they can be implemented instead in other projects that might need a larger taxonomy set in order to accommodate their needs. If that's the consensus then we can close this PR.

emanns95:
@shaundmorris trying to get this into the DDF-2.10.0 release

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/RightsAttributes.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+public class RightsAttributes implements Rights, MetacardType {
+
+    private static final Set<AttributeDescriptor> DESCRIPTORS;
+
+    private static final String NAME = "security";
+
+    static {
+        Set<AttributeDescriptor> descriptors = new HashSet<>();
+        descriptors.add(new AttributeDescriptorImpl(PRIVACY,
```
ahoffer: I don't like the privacy attribute. It's too vague. There will be little or no consistency in when people set the flag or how the flag is interpreted. It will be junk data from day 1.

I'm confused why we need or want information about privacy, but if we have to have it, it makes more sense to do something like this (thanks Wikipedia):

Atrribute 1 PROTECTED - Protected information is not classified. Protected information pertains to any sensitive information that does not relate to national security and cannot be disclosed under the access and privacy legislation because of the possible injury to particular public or private interests.

And then use enumerations/CVE:

Protected C (Extremely Sensitive protected information): is used to protect extremely sensitive information, which if compromised, could reasonably be expected to cause extremely grave injury outside the national interest. Examples could include bankruptcy, identities of informants in criminal investigations, etc.

Protected B (Particularly Sensitive protected information): is used to protect information that could cause severe injury or damage to the people or group involved if it was released. Examples include medical records, annual personnel performance reviews, income tax returns, etc.

Protected A (Low-Sensitive protected information): is applied to low sensitivity information that should not be disclosed to the public without authorization and could reasonably be expected to cause injury or embarrassment outside the national interest. Example of Protected A information could include employee number, pay deposit banking information, etc.

emanns95:
The conversation on this PR raised some good discussion points. We're going to hold off on adding these to the taxonomy until a more in depth conversation can resolve the issues.

PR #1552, opened 2017-01-10
DDF-2702 Add system usage modal, enable limited cors

andrewkfiedler:
#### What does this PR do?
 - Adds system usage modal which uses the system message property setting.
 - The system usage message can sometimes contain html, so I've quarantined it to an iframe with the sandbox attribute set.  This should keep us secure while allowing admins a degree of freedom when constructing messages.
 - Enables limited cors.  At this point it's only allowed for resources in the /css/ directoy, and only for GET and HEAD requests.  I went ahead and updated the grunt server proxy to add the same cors support, so development is easier.

@djblue 
@clockard 

@kcwire
@shaundmorris 

Go to the platform ui configuration in the Admin UI and set a system message.

https://codice.atlassian.net/browse/DDF-2702

Example

### DIFF
--- a/catalog/ui/catalog-ui-search/package.json
+++ b/catalog/ui/catalog-ui-search/package.json
@@ -28,6 +28,7 @@
     "connect-livereload": "0.3.2",
     "console-polyfill": "0.2.2",
     "copy-webpack-plugin": "3.0.1",
+    "cors": "2.8.1",
     "css-loader": "0.23.1",
     "exports-loader": "0.6.3",
     "express": "3.21.0",
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/SystemUsage.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/SystemUsage.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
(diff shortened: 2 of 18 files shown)


clockard:
Are we ok with not having an 'ok' or 'accept' button?

catalog/ui/catalog-ui-search/src/main/webapp/js/SystemUsage.js
```diff
@@ -0,0 +1,46 @@
...
+    if (hasMessage()) {
+        if (shownOncePerSession()) {
+            return true;
+        } else {
+            return hasNotSeenMessage();
+        }
+    } else {
+        return false;
+    }
+}
```
rzwiefel: I'm just putting this so you know what you're missing out on.

```javascript
if (!hasMessage()) {
    return false;
}
if (shownOncePerSession()) {
    return true;
}
return hasNotSeenMessage();
```

or if you're digging the minimalist life there's also slightly less readable but more compact
```javascript
if (!hasMessage()) {
    return false;
}
return shownOncePerSession() || hasNotSeenMessage();
```

andrewkfiedler: I wanted to avoid this: https://github.com/codice/ddf/blob/master/catalog/ui/search-ui/standard/src/main/webapp/js/controllers/SystemUsage.controller.js#L26

I like how it reads as is.

catalog/ui/catalog-ui-search/package.json
```diff
@@ -28,6 +28,7 @@
     "connect-livereload": "0.3.2",
     "console-polyfill": "0.2.2",
     "copy-webpack-plugin": "3.0.1",
+    "cors": "^2.8.1",
```
clockard: Do we need to rerun the shrinkwrap since we added this dependency?

andrewkfiedler:
Updated to add an acknowledge button and prevent continuing until it is clicked.  Going to look into shrinkwrap next.

[Here's what it currently looks like](https://webmshare.com/play/yA3yb)

catalog/ui/catalog-ui-search/package.json
shaundmorris: Should that be a specific version? maybe I'm reading it wrong :-)

shaundmorris:
thanks for the nice little video

andrewkfiedler:
Added the shrinkwrap for the new cors dev dependency.

PR #1559, opened 2017-01-11
DDF-2706 Add configuration option to hide editing capability in the Catalog UI

andrewkfiedler:
#### What does this PR do?
 - Adds an administrative option to hide editing capability in the Catalog UI from the catalog.ui.config.
 - Note that this doesn't do anything to restrict editing on the backend.
 - Default is to allow editing.

@clockard 
@pklinef
@rzwiefel 
@andrewkfiedler

Go to the Catalog UI and verify that editing is not allowed by default.   Editing should be hidden from the Summary, Details, History, & Associations views.  The Overwrite and Archive tabs should be hidden entirely.  When selecting multiple metacards as once, only the Details tab should be available and editing should be hidden from it.

Then go to the Admin UI and proceed to the org.codice.ddf.catalog.ui.config.  Turn the config for Allow Editing to true.  Return to the Catalog UI and verify the opposite of what you verified for when editing is not allowed.

Do note that this doesn't do anything to restrict editing on the backend.

https://codice.atlassian.net/browse/DDF-2706

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -74,6 +74,8 @@ public class ConfigurationApplication implements SparkApplication {
 
     private String terrainEndpoint;
 
+    private Boolean isEditingAllowed = false;
+
     private Boolean isSignIn = true;
 
     private Boolean isTask = false;
@@ -236,6 +238,7 @@ public class ConfigurationApplication implements SparkApplication {
         config.put("attributeAliases", attributeAliases);
         config.put("sourcePollInterval", sourcePollInterval);
         config.put("scheduleFrequencyList", scheduleFrequencyList);
+        config.put("isEditingAllowed", isEditingAllowed);
 
         return config;
     }
@@ -496,6 +499,14 @@ public class ConfigurationApplication implements SparkApplication {
 
     public void setIngest(Boolean isIngest) {
         this.isIngest = isIngest;
+    }
+
+    public Boolean getIsEditingAllowed() {
+        return this.isEditingAllowed;
+    }
+
+    public void setIsEditingAllowed(Boolean isEditingAllowed){
+        this.isEditingAllowed = isEditingAllowed;
     }
 
     public void setTypeNameMapping(String[] mappings) {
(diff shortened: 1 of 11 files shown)


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -498,6 +501,14 @@ public void setIngest(Boolean isIngest) {
         this.isIngest = isIngest;
     }
 
+    public Boolean getAllowEditing() {
```
rzwiefel: Not sure if its necessary or just good practice but for booleans I think the getter goes like `isBla` so this would be `isAllowEditing`

rzwiefel:
Any reason why editing should be **off** by default for DDF? It seems like it would be more useful if it was on by default.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
clockard: Maybe change the variable to `editingAllowed` and this method to `isEditingAllowed`

andrewkfiedler: Doesn't appear that way in this file.  I do see the variable names are like that though.

pklinef:
@andrewkfiedler Can you add some screenshots to show what it looks like when editing is restricted?

catalog/ui/catalog-ui-search/src/main/webapp/properties.js
```diff
@@ -81,13 +81,21 @@ define(function (require) {
                 throw new Error('Configuration could not be loaded: (status: ' + status + ', message: ' + errorThrown.message + ')');
             });
 
+            this.handleEditing();
+
             return props;
         },
+        handleEditing: function(){
+            $('html').toggleClass('is-editing-restricted', this.isEditingRestricted());
```
pklinef: @codice/security We need to talk about how we plan to enforce options like this on the backend as well.

rzwiefel:
I'm guessing the edit button just doesn't show up @pklinef

clockard:
Verified build and functionality
screen shot 2017-01-12 at 12 31 34 pm
screen shot 2017-01-12 at 12 30 42 pm

PR #1562, opened 2017-01-13
DDF-2711 Add exception for thumbnails to be shown in table visualization

andrewkfiedler:
#### What does this PR do?
 - All 'Binary' type attributes were being hidden from the table visualization, so an exception is now added for thumbnails.

@rzwiefel 
@djblue 
@jlcsmith 
@andrewkfiedler

Ingest data with thumbnails and use the table visualization.  Verify that the thumbnails show up in the table.  Also check the rearrange columns view and the hide / show columns view to make sure they show up there.

https://codice.atlassian.net/browse/DDF-2711

Table View
Table Settings View

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
@@ -24,6 +24,13 @@ define([
             this.getMetacardTypes();
             this.getDatatypeEnum();
         },
+        isHiddenTypeExceptThumbnail: function(id){
+            if (id === 'thumbnail'){
+                return false;
+            } else {
+                return this.isHiddenType(id);
+            }
+        },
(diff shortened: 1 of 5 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
```diff
@@ -24,6 +24,13 @@ define([
             this.getMetacardTypes();
             this.getDatatypeEnum();
         },
+        isHiddenTypeExceptThumbnail: function(id){
```
pklinef: Why is this not just added to `isHiddenType`?  This way requires someone to know to run both methods.  Are there cases where you want thumbnails to be hidden?

andrewkfiedler: Yeah, there are cases for both.

PR #1573, opened 2017-01-18
DDF-2722 Fixes update id ordering bug

rzwiefel:
#### What does this PR do?
- the catalog framework update request that takes an array of ids and list of metacards must have both sets in the correct respective order

this ensures the ids are in that correct order.

anyone

@andrewkfiedler
@coyotesqrl

ensure multi-select and edit does not throw an exception

https://codice.atlassian.net/browse/DDF-2722

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -109,8 +109,7 @@ public class MetacardApplication implements SparkApplication {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MetacardApplication.class);
 
-    private static final String UPDATE_ERROR_MESSAGE =
-            "Item is either restricted or not found.";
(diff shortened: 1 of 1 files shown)


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -624,8 +623,10 @@ protected UpdateResponse patchMetacards(List<MetacardChanges> metacardChanges)
                 .stream()
                 .map(Result::getMetacard)
                 .collect(Collectors.toList());
-        return catalogFramework.update(new UpdateRequestImpl(changedIds.toArray(new String[0]),
-                changedMetacards));
+        return catalogFramework.update(new UpdateRequestImpl(changedMetacards.stream()
+                .map(Metacard::getId)
+                .collect(Collectors.toList())
+                .toArray(new String[0]), changedMetacards));
```
rzwiefel: According to docs https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html

Streams that are sourced from things that define order will make the stream intrinsically ordered. So this means using `changedMetacards` for both arguments will result in the order of id's match.

PR #1586, opened 2017-01-20
DDF-2700 Add some check to fix NPE and add more checks to TestRegistry

clockard:
#### What does this PR do?
Added some more checks to ensure everything is set up in time for itests execution
@mcalcote @gordocanchola 
@shaundmorris
@stustison
Verify full build. Ideally the build would pass on jenkins
Previous commit fixed several issues in the itests but on very slow builds there still seems to be some issues even after the initial verification of state. This pr aims to verify state before each operation. 
https://codice.atlassian.net/browse/DDF-2700
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/FederationAdminServiceImpl.java
+++ b/catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/FederationAdminServiceImpl.java
@@ -626,7 +626,10 @@ public class FederationAdminServiceImpl implements FederationAdminService {
         for(ProcessingDetails detail : details){
             pw.append(detail.getSourceId());
             pw.append(":");
-            detail.getException().printStackTrace(pw);
+            if(detail.hasException()) {
+                detail.getException()
+                        .printStackTrace(pw);
+            }
             pw.append(System.lineSeparator());
         }
         return pw.toString();
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
@@ -87,6 +87,8 @@ public class TestRegistry extends AbstractIntegrationTest {
 
     private static final String REMOTE_METACARD_ID = "09876543210987654321098765432100";
 
+    private static final long SLEEP_TIME = 2000;
+
     private static String storeId;
 
     private static FederatedCswMockServer cswServer;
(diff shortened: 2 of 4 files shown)


catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/FederationAdminServiceImpl.java
```diff
@@ -626,7 +626,10 @@ private String stringifyProcessingErrors(Set<ProcessingDetails> details){
         for(ProcessingDetails detail : details){
             pw.append(detail.getSourceId());
             pw.append(":");
-            detail.getException().printStackTrace(pw);
+            if(detail.hasException()) {
+                detail.getException()
+                        .printStackTrace(pw);
```
vinamartin: Is this okay to do?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
```diff
@@ -585,5 +590,34 @@ private String createRegistryStoreEntry(String id, String regId, String remoteRe
...
+                        .statusCode(200)
+                        .when()
+                        .post(CSW_PATH.getUrl());
+                available = true;
+                LOGGER.info("Csw Mock Server is running");
+            } catch (Error e) {
+                if (System.currentTimeMillis() - startTime > timeout) {
+                    fail("Csw mock server didn't come up within allotted time.");
+                }
+                Thread.sleep(2000);
```
vinamartin: Should this sleep be a constant? How was it decided?

catalog/spatial/registry/registry-federation-admin-service-impl/src/main/java/org/codice/ddf/registry/federationadmin/service/impl/FederationAdminServiceImpl.java
clockard: We want to avoid printing stack traces to system out but this is just to a string so we don't lose it when creating a new exception.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
clockard: Changed to a constant. The amount of time is in line with other itests but still somewhat arbitrary. Lots of tests use 1 second but i chose 2 here so that there wouldn't be as much churn on slower systems.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestRegistry.java
```diff
@@ -87,6 +87,8 @@
 
     private static final String REMOTE_METACARD_ID = "09876543210987654321098765432100";
 
+    private static final long SLEEP_TIME = 2000;
```
shaundmorris: Just curious how you came up with the sleep time. In more resource constrained environments it shouldn't matter much since we have the timeout for 2 minutes, is that accurate?

clockard: Most places have the sleep time at 1 second but i went with 2 here to reduce thrashing. Figured if the system is really slow and our stuff isn't coming up in time it might be better to reduce the frequency of these checking threads.

clockard:
Full build with empty repo passed. :shipit:

PR #1589, opened 2017-01-21
DDF-2734 Remove incorrect falsy comparisons

andrewkfiedler:
#### What does this PR do?
 - Falsy comparisons were being performed on numbers, resulting in bad behavior at 0 (which is seen as false).

@djblue 
@bdeining 
@jlcsmith 
@rzwiefel 

Test out the location searches with 0 values for latitude and longitude.  Verify they update the drawing, as well as save correctly. 

https://codice.atlassian.net/browse/DDF-2734

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.hbs
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.hbs
@@ -24,7 +24,7 @@
         <i class="fa fa-circle-o"></i>
         Polygon
     </label>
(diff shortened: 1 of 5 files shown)


jlcsmith:
Built and tested.  Looks good.  The only issue I noticed which isn't precisely related to this is if bad values are used.  For example, entering bbox values of:
West = 0
South = 0
East = -10  <= invalid
North = 10
Results in a the "east" value being modified to "0.0001".  The resulting shape on the map is a line.

andrewkfiedler:
@jlcsmith That is the intended behavior.  I recently added validation to the inputs to make sure incorrect values don't get propagated where they will cause issues. 

See https://github.com/codice/ddf/commit/267a15da324a6231e69c9ede6c5dc7c20f7107f6 for more details on the validation.

PR #1599, opened 2017-01-24
DDF-2740 Update SaxEventHandlers to return attribute descriptors

gordocanchola:
#### What does this PR do?
This ticket moves the getSupportedAttributeDescriptors methods from the SaxEventHandlerFactories to the SaxEventHandlers. This move is needed in the case where handler data determines the attribute descriptors.
This ticket also changes the SaxEventHandlerDelegate to no longer put duplicate attributes on a metacard blindly. It now checks the AttributeDescriptor's multivalued flag to see if having multiple values is permitted. If multivalued is false it just keeps the first value, ignoring any following ones. 
This PR also splits the delegates read method into read and getMetacard methods.
@clockard @kcwire @mcalcote 
@pklinef
@shaundmorris
  Removing getSupportedAttributeDescriptors from the Factory classes and moving them to the Handlers.
  SaxEventHandlerDelegate
    Splitting read method into a read and getMetacard
    Changing getMetacard to not allow duplicate attribute values
    Updated tests.
[DDF-2740](https://codice.atlassian.net/browse/)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/transformer/catalog-transformer-streaming-api/src/main/java/org/codice/ddf/transformer/xml/streaming/SaxEventHandler.java
+++ b/catalog/transformer/catalog-transformer-streaming-api/src/main/java/org/codice/ddf/transformer/xml/streaming/SaxEventHandler.java
@@ -14,16 +14,22 @@
 package org.codice.ddf.transformer.xml.streaming;
 
 import java.util.List;
+import java.util.Set;
 
 import org.xml.sax.ContentHandler;
 
 import ddf.catalog.data.Attribute;
+import ddf.catalog.data.AttributeDescriptor;
 
 /**
  * Interface used to handle sax events fired from {@link org.codice.ddf.transformer.xml.streaming.lib.SaxEventHandlerDelegate} and {@link org.codice.ddf.transformer.xml.streaming.lib.XMLInputTransformer}.
  * At the end of parsing, it will have a list of {@link Attribute}s that can be used to populate a metacard.
  * A specific implementation can be used to handle specific XML elements and metacards. @see ddf.catalog.transformer.generic.xml.impl.XMLSaxEventHandlerImpl
  * {@inheritDoc}
+ * <p>
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
  */
 public interface SaxEventHandler extends ContentHandler {
 
@@ -32,4 +38,10 @@ public interface SaxEventHandler extends ContentHandler {
      */
     List<Attribute> getAttributes();
 
+    /**
+     * Get all the possible attribute types that can be returned by this factory's handler
+     *
+     * @return a set of attribute descriptors that can be returned by this handler
+     */
+    Set<AttributeDescriptor> getSupportedAttributeDescriptors();
 }
--- a/catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/GmlHandler.java
+++ b/catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/GmlHandler.java
@@ -15,13 +15,16 @@ package org.codice.ddf.transformer.xml.streaming.impl;
 
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.Stack;
 
 import javax.xml.stream.XMLStreamException;
 
 import org.codice.ddf.transformer.xml.streaming.AbstractSaxEventHandler;
 import org.codice.ddf.transformer.xml.streaming.Gml3ToWkt;
+import org.codice.ddf.transformer.xml.streaming.lib.SaxEventHandlerUtils;
 import org.codice.ddf.transformer.xml.streaming.lib.SaxEventToXmlElementConverter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -33,8 +36,11 @@ import com.vividsolutions.jts.io.WKTWriter;
 import com.vividsolutions.jts.io.gml2.GMLHandler;
 
 import ddf.catalog.data.Attribute;
+import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.impl.AttributeImpl;
+import ddf.catalog.data.impl.types.CoreAttributes;
+import ddf.catalog.data.impl.types.ValidationAttributes;
 import ddf.catalog.data.types.Validation;
 import ddf.catalog.validation.ValidationException;
 
@@ -47,6 +53,8 @@ import ddf.catalog.validation.ValidationException;
 public class GmlHandler extends AbstractSaxEventHandler {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GmlHandler.class);
+
+    private static Set<AttributeDescriptor> attributeDescriptors = new HashSet<>();
 
     GMLHandler gh;
 
@@ -63,6 +71,13 @@ public class GmlHandler extends AbstractSaxEventHandler {
     private Gml3ToWkt gml3Converter;
 
     private Stack<String> state;
+
+    private SaxEventHandlerUtils saxEventHandlerUtils = new SaxEventHandlerUtils();
+
+    static {
+        attributeDescriptors.addAll(new ValidationAttributes().getAttributeDescriptors());
+        attributeDescriptors.add(new CoreAttributes().getAttributeDescriptor(Metacard.GEOGRAPHY));
+    }
 
     public GmlHandler(GMLHandler gmlHandler, Gml3ToWkt gml3Converter) {
         this.gh = gmlHandler;
@@ -82,7 +97,14 @@ public class GmlHandler extends AbstractSaxEventHandler {
      */
     @Override
     public List<Attribute> getAttributes() {
-        return attributes;
+        return saxEventHandlerUtils.getCombinedMultiValuedAttributes(
+                getSupportedAttributeDescriptors(),
+                attributes);
+    }
+
+    @Override
+    public Set<AttributeDescriptor> getSupportedAttributeDescriptors() {
+        return attributeDescriptors;
     }
 
     @Override
--- a/catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/XmlSaxEventHandlerImpl.java
+++ b/catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/XmlSaxEventHandlerImpl.java
@@ -15,15 +15,21 @@ package org.codice.ddf.transformer.xml.streaming.impl;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.codice.ddf.transformer.xml.streaming.AbstractSaxEventHandler;
+import org.codice.ddf.transformer.xml.streaming.lib.SaxEventHandlerUtils;
 import org.xml.sax.Attributes;
 
 import ddf.catalog.data.Attribute;
+import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.impl.AttributeImpl;
+import ddf.catalog.data.impl.types.ContactAttributes;
+import ddf.catalog.data.impl.types.CoreAttributes;
 
 /**
  * A sax event handler used to parse urn:catalog:metacard {@link Metacard}s. By default, handles all elements defined in the {@link XmlSaxEventHandlerImpl#xmlToMetacard}
@@ -31,6 +37,8 @@ import ddf.catalog.data.impl.AttributeImpl;
  * {@inheritDoc}
  */
 public class XmlSaxEventHandlerImpl extends AbstractSaxEventHandler {
+
+    private static Set<AttributeDescriptor> attributeDescriptors = new HashSet<>();
 
     /*
      * A list of Attributes that is populated during parsing and then returned by getAttributes
@@ -52,9 +60,27 @@ public class XmlSaxEventHandlerImpl extends AbstractSaxEventHandler {
      */
     private Map<String, String> xmlToMetacard;
 
+    private SaxEventHandlerUtils saxEventHandlerUtils = new SaxEventHandlerUtils();
+
+    static {
+        CoreAttributes coreAttributes = new CoreAttributes();
+        attributeDescriptors.add(coreAttributes.getAttributeDescriptor(Metacard.ID));
+        attributeDescriptors.add(coreAttributes.getAttributeDescriptor(Metacard.TITLE));
+        attributeDescriptors.add(coreAttributes.getAttributeDescriptor(Metacard.METADATA));
+        attributeDescriptors.add(coreAttributes.getAttributeDescriptor(Metacard.DESCRIPTION));
+        attributeDescriptors.add(new ContactAttributes().getAttributeDescriptor(Metacard.POINT_OF_CONTACT));
+    }
+
     @Override
     public List<Attribute> getAttributes() {
-        return attributes;
+        return saxEventHandlerUtils.getCombinedMultiValuedAttributes(
+                getSupportedAttributeDescriptors(),
+                attributes);
+    }
+
+    @Override
+    public Set<AttributeDescriptor> getSupportedAttributeDescriptors() {
+        return attributeDescriptors;
     }
 
     protected XmlSaxEventHandlerImpl() {
--- a/catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
+++ b/catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
@@ -20,8 +20,14 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.apache.commons.collections.MapUtils;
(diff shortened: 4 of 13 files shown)


catalog/transformer/catalog-transformer-streaming-api/src/main/java/org/codice/ddf/transformer/xml/streaming/SaxEventHandler.java
```diff
@@ -32,4 +34,10 @@
      */
     List<Attribute> getAttributes();
 
+    /**
+     * Get all the possible attribute types that can be returned by this factory's handler
+     *
+     * @return a set of attribute descriptors that can be returned by this handler
+     */
+    Set<AttributeDescriptor> getSupportedAttributeDescriptors();
```
pklinef: Is this interface actually beta or internal?  This would be a breaking change if it is not.

gordocanchola: This is the interface change we discussed a little while ago.
It was suggested that if it wasn't marked as experimental, it could be modified if the package wasn't listed in the whitelist.

catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/GmlHandler.java
```diff
@@ -64,6 +69,30 @@
...
+
+    static {
+        attributeDescriptors.add(new AttributeDescriptorImpl(Metacard.GEOGRAPHY,
+                true /* indexed */,
+                true /* stored */,
+                false /* tokenized */,
+                false /* multivalued */,
+                BasicTypes.GEO_TYPE));
+
+        attributeDescriptors.add(new AttributeDescriptorImpl(Validation.VALIDATION_WARNINGS,
```
pklinef: I think all of these attribute descriptors are defined in `catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/`.  You should use the already defined versions.

Here is where the validation attributes are defined.
https://github.com/codice/ddf/blob/76975e51a8093358329ff828afbb7ae82a1adba7/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/ValidationAttributes.java#L37-L48

I think this location attribute descriptor is the same as the geography one above.
https://github.com/codice/ddf/blob/f8542dd272bd4950cbcd7dabea289709494fd622/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/types/CoreAttributes.java#L78-L83

catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/XmlSaxEventHandlerImpl.java
```diff
@@ -52,11 +57,51 @@
      */
     private Map<String, String> xmlToMetacard;
 
+    private static Set<AttributeDescriptor> attributeDescriptors = new HashSet<>();
+
+    static {
+        attributeDescriptors.add(new AttributeDescriptorImpl(Metacard.ID,
```
pklinef: Same comment as before to use the already defined attribute descriptors.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
```diff
@@ -259,6 +257,24 @@ InputTransformerErrorHandler getInputTransformerErrorHandler() {
...
+    public MetacardType getMetacardType(String id) {
+        Set<AttributeDescriptor> attributeDescriptors =
+                new HashSet<>(BasicTypes.BASIC_METACARD.getAttributeDescriptors());
+
+        attributeDescriptors.addAll(eventHandlers.stream()
+                .map(SaxEventHandler::getSupportedAttributeDescriptors)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet()));
+
+        return new DynamicMetacardType(attributeDescriptors, id);
```
pklinef: Looking at `DynamicMetacardType`, I think this will actually change the metacard type name.  I am not sure this is what we want to happen.  Why can we not use `MetacardTypeImpl` here?

catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/XmlSaxEventHandlerImpl.java
```diff
@@ -52,11 +57,27 @@
      */
     private Map<String, String> xmlToMetacard;
 
+    private static Set<AttributeDescriptor> attributeDescriptors = new HashSet<>();
+
+    static {
+        CoreAttributes coreAttributes = new CoreAttributes();
```
harrison-tarr: This is a great addition, really simplifies the code.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
```diff
@@ -89,12 +92,7 @@ public SaxEventHandlerDelegate(List<SaxEventHandler> eventHandlers) {
      * populated with all the {@link Attribute}s parsed by the {@link SaxEventHandlerDelegate#eventHandlers}
      * @throws CatalogTransformerException
      */
-    public Metacard read(InputStream inputStream) throws CatalogTransformerException {
-
-        /*
-         * Create a new MetacardImpl with the proper MetacardType
-         */
-        Metacard metacard = new MetacardImpl(metacardType);
+    public void read(InputStream inputStream) throws CatalogTransformerException {
```
harrison-tarr: if you change this to return a `SaxEventHandlerDelegate`, you can chain the calls. I don't know if it buys you anything, but smooths the code out a bit.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
```diff
@@ -137,19 +147,12 @@ public Metacard read(InputStream inputStream) throws CatalogTransformerException
...
             List<Attribute> attributes = eventHandler.getAttributes();
             for (Attribute attribute : attributes) {
-                Attribute tmpAttr;
 
                 /*
-                 * If metacard already has values in the attribute, put them together into a multivalued list,
+                 * If metacard already has values in the attribute, skip the attribute,
                  * instead of simply overwriting the existing values.
                  */
-                if ((tmpAttr = metacard.getAttribute(attribute.getName())) != null) {
```
harrison-tarr: Why do you do this change? I think this will prevent multivalued attributes, like if two handlers return the same kind of `Attribute`

gordocanchola: Yes. This was discussed because some attributes were being multivalued that shouldn't have been.  The change was made so that the first value set, wins. If this proves to be an issue we can revisit it.

pklinef: If using the GML streaming transformer, does it treat a list of values as the first value set or will it only take the first value in the list?

gordocanchola: I added a Utility class that will combine common attributes for a handler. The Delegate used to combine all common attributes across all handlers. Since this functionality was removed, common attributes will be combined within the handlers themselves. 
The Delegate will take the first attribute's value or list of values.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
gordocanchola: Modified as suggested. Good call.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import ddf.catalog.data.Attribute;
+import ddf.catalog.data.AttributeDescriptor;
+import ddf.catalog.data.impl.AttributeImpl;
+
+public class SaxEventHandlerUtils {
+
+    public List<Attribute> getCombinedMultiValuedAttributes(Set<AttributeDescriptor> descriptors,
```
clockard: Can you add some javadoc on how this method works?

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
```diff
@@ -0,0 +1,78 @@
...
+        Map<String, Boolean> multiValuedMap = getMultiValuedNameMap(descriptors);
+
+        List<String> validationWarnings = new ArrayList<>();
+        Map<String, Attribute> attributeMap = new HashMap<>();
+        for (Attribute attribute : attributes) {
+            String attributeName = attribute.getName();
+            Attribute addedAttribute = attributeMap.putIfAbsent(attributeName, attribute);
+
+            if (addedAttribute != null) {
+                Boolean addValue = multiValuedMap.get(attributeName);
```
clockard: Could do a `getOrDefault` so you don't need to do a null check later

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
```diff
@@ -0,0 +1,78 @@
...
+                if (addValue != null && addValue) {
+                    attributeMap.get(attributeName)
+                            .getValues()
+                            .addAll(attribute.getValues());
+                } else {
+                    validationWarnings.add(
+                            "Multiple values found for a non multi valued attribute. Attribute: "
```
clockard: Probably want to change this to a String.format.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
gordocanchola: Keeping the null default get. Modified to use the null check to add a validation-warning about not finding an attribute descriptor for a specific attribute.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
gordocanchola: Added javadoc.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
harrison-tarr: I thought some more about this. What if you check to see if the attribute is supposed to be multivalued or not? It should be one of the fields in the `AttributeDescriptor`. Then if it is, you can use all the values, and if it's not, skip the values, the way you are here.

gordocanchola: That was just added on a per handler basis, not across multiple handlers.

harrison-tarr: I meant here

gordocanchola: Made the changes here and verified itests passed

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
harrison-tarr: You construct one of these in the `impl` bundle and then inject it into all the handlers that need one, instead of newing one up in each handler. It doesn't look like this class uses state at all.

catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/GmlHandler.java
```diff
@@ -82,7 +97,12 @@ public GmlHandler(GMLHandler gmlHandler, Gml3ToWkt gml3Converter) {
      */
     @Override
     public List<Attribute> getAttributes() {
-        return attributes;
+        return saxEventHandlerUtils.getCombinedMultiValuedAttributes(getSupportedAttributeDescriptors(), attributes);
+    }
+
+    @Override
+    public Set<AttributeDescriptor> getSupportedAttributeDescriptors() {
```
harrison-tarr: You could probably add this as a method in the `AbstractSaxEventHandler`, then you would only have to implement this method if you don't want to `return the static set of attributeDescriptors`

harrison-tarr: while you're at it, you probably could even do a default implementation for `getAttributes`

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
gordocanchola: The handlers are newed by the factories. This would have to be injected into the factories just to be passed into the handlers.

catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/GmlHandler.java
gordocanchola: Moving getSupportedAttributeDescriptors to the abstract class. Leaving getAttributes alone for now because the default implementation would use the SaxEventHandlerUtils class which is in streaming-lib. The abstract class lives in the api. Don't want to tie the api to streaming-lib. The abstract class should find a new home outside the api.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
```diff
@@ -109,48 +111,58 @@ public Metacard read(InputStream inputStream) throws CatalogTransformerException
...
+            throw new CatalogTransformerException("Could not properly parse metacard", e);
+        }
+
+        return this;
+    }
 
+    public Metacard getMetacard(String id) {
+        MetacardType metacardType = getMetacardType(id);
+
+        if (metacardType == null) {
```
mcalcote: A debug level log statement inside this `if` branch saying that the metacard type was not dynamically found might be useful.

catalog/transformer/catalog-transformer-streaming-impl/src/main/java/org/codice/ddf/transformer/xml/streaming/impl/GmlHandler.java
gordocanchola: A ticket was created to address moving the abstract class out of the api

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerDelegate.java
gordocanchola: Added the debug statement.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventHandlerUtils.java
```diff
@@ -0,0 +1,110 @@
...
+                    addValue = false;
+                }
+
+                if (addValue) {
+                    attributeMap.get(attributeName)
+                            .getValues()
+                            .addAll(attribute.getValues());
+                } else {
+                    validationWarnings.add(String.format(
+                            "Multiple values found for a non multi valued attribute. Attribute: '%s', Existing value: '%s', New value: '%s'. Existing value will be kept, new value will be ignored.",
```
shaundmorris: should there be dashes in there for `non-multi-valued`?

gordocanchola: Added the dashes

clockard:
Hero build passed!

PR #1600, opened 2017-01-24
DDF-2744: Fix resource-uri loss during UpdateStorageRequest

peterhuffer:
#### What does this PR do?
This PR fixes the `RESOURCE_URI` missing from the updated `Metacard` during an `UpdateStorageRequest`, which is removed by the `OverrideAttributesSupport#addAttributes()` to prevent updates pointing `Metacard`s to content that they don't have permission to see. 

The fix was to add the `RESOURCE_URI`, gotten from the original queried `Metacard`, back to the updated `Metacard` before the update occurs.

Another option could be to remove the filtering and rely on the `ContentUriAccessPlugin` to stop processing of the incoming `UpdateRequest` if the URI is modified, but this would remove a level of security.

@troymohl @emmberk 

@pklinef
@stustison

Full build.

This was discovered when sending an `UpdateStorageRequest` back to the `Catalog` from the asynchronous `ProcessingFramework` in [DDF-2569](https://github.com/codice/ddf/pull/1495).

[DDF-2744](https://codice.atlassian.net/browse/DDF-2744)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
@@ -49,6 +49,7 @@ import ddf.catalog.data.Attribute;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
 import ddf.catalog.data.impl.AttributeImpl;
+import ddf.catalog.data.types.Core;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.history.Historian;
 import ddf.catalog.impl.FrameworkProperties;
@@ -612,6 +613,8 @@ public class UpdateOperations {
 
         updateRequest = rewriteRequestToAvoidHistoryConflicts(updateRequest, query);
 
+        updateRequest = populateResourceUris(updateRequest, query);
+
         HashMap<String, Metacard> metacardMap = new HashMap<>(query.getResults()
                 .stream()
                 .map(Result::getMetacard)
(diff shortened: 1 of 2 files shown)


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -733,4 +736,37 @@ private UpdateStorageRequest applyAttributeOverrides(UpdateStorageRequest update
                 metacardMap);
         return updateStorageRequest;
     }
+
+    private UpdateRequest populateResourceUris(UpdateRequest updateRequest, QueryResponse query) {
+        Set<Metacard> updateRequestMetacards = updateRequest.getUpdates()
+                .stream()
+                .map(Map.Entry::getValue)
+                .filter(metacard -> metacard.getResourceURI() == null)
```
brendan-hofmann: I would prefer to move away from these convenience methods in favor of getting attributes by their constants instead, as this results in more consistent code and API usage.

Also, I don't think checking for null is sufficient here. You probably want `StringUtils.isEmpty/isNotEmpty` instead.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -733,4 +736,37 @@ private UpdateStorageRequest applyAttributeOverrides(UpdateStorageRequest update
...
+                .filter(Objects::nonNull)
+                .filter(metacard -> metacard.getResourceURI() != null)
+                .collect(Collectors.toMap(Metacard::getId, Function.identity()));
+
+        for (Metacard updateMetacard : updateRequestMetacards) {
+            Metacard queriedMetacard = queryResultMetacards.get(updateMetacard.getId());
+            if (queriedMetacard != null && queriedMetacard.getId()
+                    .equals(updateMetacard.getId())) {
+                LOGGER.info(
+                        "Found URI \"{}\" missing from updated Metacard with ID \"{}\". Setting Metacard field \"{}\" with the value.",
```
brendan-hofmann: The specificity of this workaround concerns me. We have had multiple problems in the past with reconstructed metacards accidentally dropping attributes. I would prefer a more general solution, shouldn't we be checking the metacard's type to make sure we are not dropping anything?

stustison: This should also be a debug level log message and we shouldn't be logging from within a loop. Just do a single log message at the end of this once you have all the results.

pklinef: @brendan-hofmann It is possible to remove attributes as needed so I am not sure how you would tell the difference between dropping and removing currently.  Content resource URIs are a special case that we do not allow either way.  I do see things like this making it more difficult in the future to support partial updates but I don't think we need to worry about that use case for this PR.

brendan-hofmann: Agreed that it's out of scope for this PR, but I do wish we had a more explicit, or at least well-documented, behavior for removing attributes during an update operation. Ideally it would be part of the API, but that may not be feasible for backwards compatibility.

peterhuffer:
This PR is no longer valid. A different solution was found https://github.com/codice/ddf/pull/1602.

PR #1602, opened 2017-01-24
DDF-2745: Fix resource-uri being overriden by null update

peterhuffer:
#### What does this PR do?
- Fix resource-uri being overriden by null update

@emmberk @troymohl @brendan-hofmann 

@pklinef
@stustison

Full Build.

Existing `Metacard`'s URIs should not be able to be overridden in order to prevent resources with no references to it.

[DDF-2745](https://codice.atlassian.net/browse/DDF-2745)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
+++ b/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
@@ -63,7 +63,7 @@ public class OverrideAttributesSupportTest {
                 .getTitle(), is("updated"));
         assertThat(metacardMap.get("original")
                 .getResourceURI()
-                .toString(), is("content:stuff"));
+                .toString(), is("content:newstuff"));
         assertThat(metacardMap.get("original")
                 .getId(), is("original"));
         assertThat(metacardMap.get("original")
@@ -99,7 +99,7 @@ public class OverrideAttributesSupportTest {
                 .getTitle(), is("updated"));
         assertThat(metacardMap.get("original")
                 .getResourceURI()
-                .toString(), is("content:stuff"));
+                .toString(), is("content:newstuff"));
         assertThat(metacardMap.get("original")
                 .getId(), is("original"));
         assertThat(metacardMap.get("original")
@@ -128,7 +128,7 @@ public class OverrideAttributesSupportTest {
         assertThat(updatedMetacard.getMetadata(), is("updated"));
         assertThat(updatedMetacard.getTitle(), is("updated"));
         assertThat(updatedMetacard.getResourceURI()
-                .toString(), is("content:stuff"));
+                .toString(), is("content:newstuff"));
         assertThat(updatedMetacard.getId(), is("original"));
         assertThat(updatedMetacard.getMetacardType()
                 .getName(), is("ddf.metacard"));
@@ -155,7 +155,7 @@ public class OverrideAttributesSupportTest {
         assertThat(updatedMetacard.getMetadata(), is("updated"));
         assertThat(updatedMetacard.getTitle(), is("updated"));
         assertThat(updatedMetacard.getResourceURI()
-                .toString(), is("content:stuff"));
+                .toString(), is("content:newstuff"));
         assertThat(updatedMetacard.getId(), is("original"));
         assertThat(updatedMetacard.getMetacardType()
                 .getName(), is("other"));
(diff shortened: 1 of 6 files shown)


catalog/plugin/catalog-plugin-content-uri/src/main/java/org/codice/ddf/catalog/content/plugin/uri/ContentUriAccessPlugin.java
```diff
@@ -41,19 +45,31 @@ public UpdateRequest processPreUpdate(UpdateRequest input,
...
                     && ContentItem.CONTENT_SCHEME.equals(existingMetacard.getResourceURI()
                     .getScheme())) {
-                throw new StopProcessingException("Cannot overwrite resource URI in content store.");
+                throw new StopProcessingException(CANNOT_OVERWRITE_MESSAGE);
             }
         }
         return input;
     }
 
+    private boolean oneOrOtherUriNull(Metacard existingMetacard, Metacard updatedMetacard) {
```
brendan-hofmann: I think this check should also include the URI scheme check, as this plugin is only meant for content URIs.

pklinef: I would suggest doing the content scheme check for everything in the `for` loop which would include this method call.

peterhuffer: Went ahead and added this check at the beginning of the `for` loop before checking anything else. Also added an additional test where both existing and updated URIs are content schema and equal.

pklinef:
Make sure to either squash your second commit or add the JIRA ticket to the commit message.

peterhuffer:
Squashed commits.

peterhuffer:
I think merging this as it is now will break `UpdateStorageRequest`s for `Metacard`s that already have a URI because we remove the URI from the incoming `ContentItem`'s `Metacard` for security reasons, so it will always be null. Need to figure out what to do about this.

peterhuffer:
I have removed where we filter out the `RESOURCE_URI` inside the `OverrideAttributeSupport` and also added two integration tests, both for `CreateStorageRequests` and `UpdateStorageRequests` to ensure the `RESOURCE_URI` cannot be overridden.

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
```diff
@@ -715,7 +715,7 @@ public Response updateDocument(@PathParam("id") String id, @Context HttpHeaders
                 if (multipartBody != null) {
                     List<Attachment> contentParts = multipartBody.getAllAttachments();
                     if (contentParts != null && contentParts.size() > 0) {
-                        createInfo = parseAttachment(contentParts.get(0));
+                        createInfo = parseAttachments(contentParts, transformerParam);
```
peterhuffer: The `updateDocument` RESTEndpoint was not correctly parsing all attachments of the incoming request.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
```diff
@@ -63,7 +63,7 @@ public void testOverrideAttributesBasic() throws URISyntaxException {
                 .getTitle(), is("updated"));
         assertThat(metacardMap.get("original")
                 .getResourceURI()
-                .toString(), is("content:stuff"));
```
brendan-hofmann: These test lines are a little bit weird now. They would make more sense if they were something other than a content uri, like an ftp uri or something.

peterhuffer: It shouldn't matter what the URI is. It is asserting that the URI is unchanged.

brendan-hofmann: I understand it has no impact on the test results, but it is confusing/misleading to anyone who tries to look at our unit tests for examples of how the software should be used or should behave. 😺

PR #1605, opened 2017-01-25
DDF-2747 added check for bad characters

stustison:
@garrettfreibott 
@coyotesqrl
@pklinef
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
+++ b/platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
@@ -145,6 +145,8 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
         if (user == null) {
             return false;
         }
+        user = user.trim();
+        validateUsername(user);
 
         char[] tmpPassword = ((PasswordCallback) callbacks[1]).getPassword();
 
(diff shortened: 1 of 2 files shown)


platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -335,6 +337,42 @@ protected boolean doLogin() throws LoginException {
         return true;
     }
 
+    void validateUsername(String username) throws LoginException {
```
coyotesqrl: A little less efficient but easier to read would be:
```java
    private static final int[] BAD_CHARS = {'\\', ',', '+', '"', '<', '>', ';', '#'};

    void validateUsername(String username) throws LoginException {
        if (username.chars()
                .anyMatch(c -> Arrays.stream(BAD_CHARS)
                        .anyMatch(i -> i == c))) {
            throw new LoginException("Bad characters present in user name.");
        }
    }
```

Could also use a `Set<Integer>` and replace the nested stream/anyMatch with a `contains`.

pklinef: Another option would be to collapse all of the switch cases together since they do the same thing.

PR #1607, opened 2017-01-26
DDF-2723 added source id to metacard clone ctor

rzwiefel:
#### What does this PR do?
DDF-2723 added source id to metacard clone ctor
- The clone constructor on MetacardImpl(Metacard, MetacardType) will now
  copy the SourceID attribute to cloned metacard

anyone
anyone
ensure source id gets cloned

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
@@ -157,6 +157,9 @@ public class MetacardImpl implements Metacard {
             throw new IllegalArgumentException(
                     MetacardType.class.getName() + " instance should not be null.");
         }
+        if (metacard.getSourceId() != null) {
+            this.setSourceId(metacard.getSourceId());
+        }
         map = new HashMap<>();
         for (AttributeDescriptor descriptor : metacard.getMetacardType().getAttributeDescriptors()) {
             Attribute metacardAttribute = metacard.getAttribute(descriptor.getName());


shaundmorris:
Does it make sense to add a unit test to ensure we don't regress on the source id getting cloned?

rzwiefel:
SourceID will now always be cloned when using the clone ctor, So it is a true==true kinda test. What we could do (and could have caught this bug with) is itests of our cache source.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -157,6 +157,9 @@ public MetacardImpl(Metacard metacard, MetacardType type) {
             throw new IllegalArgumentException(
                     MetacardType.class.getName() + " instance should not be null.");
         }
+        if (metacard.getSourceId() != null) {
```
coyotesqrl: The `readObject` and `writeObject` methods should be analyzed to ensure they are not broken. Also, by serializing the cloned metacard you should be able to verify your change.

rzwiefel:
@shaundmorris Added unit test for clone ctor
@coyotesqrl updated readObject/writeObject methods and added unit test

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/data/impl/MetacardImplTest.java
```diff
@@ -562,4 +566,39 @@ public void testDeserializingHiddenAttribute() throws IOException, ClassNotFound
...
+            serializedMetacard = baos.toByteArray();
+        }
+
+        Metacard deserializedMetacard;
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(serializedMetacard);
+                ObjectInputStream ois = new ObjectInputStream(bais);) {
+            deserializedMetacard = (Metacard) ois.readObject();
+        }
+
+        assertEquals(deserializedMetacard.getSourceId(), sourceId);
```
bdeining: `assertThat`

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/data/impl/MetacardImplTest.java
```diff
@@ -562,4 +566,39 @@ public void testDeserializingHiddenAttribute() throws IOException, ClassNotFound
...
+    @Test
+    public void testMetacardCloneCtorCopiesSourceId() throws Exception {
+        String sourceId = "Luke-IAmYourSource";
+
+        MetacardImpl metacard = new MetacardImpl();
+        metacard.setSourceId(sourceId);
+
+        Metacard clonedMetacard = new MetacardImpl(metacard, metacard.getMetacardType());
+
+        assertEquals(clonedMetacard.getSourceId(), metacard.getSourceId());
```
bdeining: `assertThat`

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/data/impl/MetacardImplTest.java
rzwiefel: Currently the rest file is using the old assertEquals, so keeping it in the same format as the other tests

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/data/impl/MetacardImplTest.java
rzwiefel: Currently the rest file is using the old assertEquals, so keeping it in the same format as the other tests

brendan-hofmann:
@pklinef @kcwire I am wondering if changing the behavior of `source-id` could have any unwanted implications for caching or federation that might not be caught in our automated testing.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
brendan-hofmann: How confident are we that a null check is sufficient here, vs something like `StringUtils.isEmpty()`? If we *want* to allow zero-length/whitespace source ids, can we at least explain why?

rzwiefel: I doubt we need it at all. This is a clone constructor. Not setting it is the same as setting it to null.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
```diff
@@ -759,6 +762,8 @@ private void writeObject(ObjectOutputStream stream) throws IOException {
             stream.writeObject(mt);
         }
 
+        stream.writeObject(getSourceId());
```
pklinef: I am fairly confident this will break backwards compatibility with serialized metacards.  I do not think this should be included in this fix.

It was intentional to not include the source id when metacards are exported and imported.  The system that it is imported into will set its source id.

rzwiefel:
@pklinef Comments resolved.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardImpl.java
rzwiefel: Fixed.

PR #1609, opened 2017-01-26
DDF-2753 Fixed issue that the Lat & Lon were not verified on the Local Node UI modal

emmberk:
#### What does this PR do?
This PR fixes an issue where the Lat & Lon were not being verified on the Local Node UI modal.
@clockard @mcalcote
@shaundmorris
@stustison

1. Open an `Add Local Node` modal (`Admin Console` -> `Catalog` -> `Node Information` -> `Additional Local Nodes` -> :heavy_plus_sign:).
2. `Show Advanced` and add a `Custom Field` of type `point`.
Confirm that an error message is displayed for invalid values for `Lat` and/or `Lon`:
image
3. Confirm that the new `Local Node` cannot be saved by pressing the `Add` button. There should be an error message at the bottom:
image
4. Confirm that error checking for all other fields was not affected and behaves as expected.
[DDF-2753](https://codice.atlassian.net/browse/DDF-2753)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
+++ b/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
@@ -111,7 +111,7 @@ define([
                     return;
                 }
             } else {
-                if (this.isEmpty() && !this.get('regex') && !this.get('required')) {
+                if (!this.get('value') && !this.get('regex') && !this.get('required')) {
                     this.set('error', error);
                     return;
                 }


clockard:
Manual verification passed. :shipit:

andrewkfiedler:
Not familiar with this, but what is the regex?  

I ask because `!this.get('value')` will evaluate to true for the value 0.

andrewkfiedler:
Not familiar with this area, but what is the regex?  

I ask because `!this.get('value')` will evaluate to true for the value 0.

catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js
```diff
@@ -111,7 +111,7 @@ define([
                     return;
                 }
             } else {
-                if (this.isEmpty() && !this.get('regex') && !this.get('required')) {
+                if (!this.get('value') && !this.get('regex') && !this.get('required')) {
```
emmberk: @andrewkfiedler, not sure what you mean by regex, but I think that the value could be any of the fields in the `Add Local Node` modal. The value being 0 could very well have been an issue for the lat and lon fields, so maybe I can write up another ticket for that. What do you think?

andrewkfiedler: Ah, I didn't see that this was for an early return.  I'm a little interested in what the crux of the issue was, as that might clear things up.  Right now I'm guessing that lat / lon is of type Point, and based on the lines above this one, specifically this one: https://github.com/codice/ddf/blob/master/catalog/spatial/registry/registry-admin-modules/registry-admin-local-ui/src/main/webapp/js/model/Field.js#L103, I'm not sure how your change is changing anything (since those lines will set nonexistent point values to a blank array).

emmberk: `!this.get('value')` returns `false` after `this.set('value', []);`, so the early return is no longer called. @clockard, @mcalcote, and I found a few little issues with the error messages and have written up a ticket to refactor this code: [DDF-2758](https://codice.atlassian.net/browse/DDF-2758). This PR should serve as a quick fix until that ticket is completed.

andrewkfiedler: Sounds good.  I'm wondering if it might be worth it to split this into a separate `if` specifically for type 'point' so it doesn't affect the other types.

emmberk: I just completed some more thorough testing to make sure that this change does not affect the other types of fields.

PR #1610, opened 2017-01-30
DDF-2644 Adds cross browser support to Catalog UI

andrewkfiedler:
- Officially supports: IE10 and above, Chrome, Firefox ESR and above.
 - Upgrades openlayers dependency to 3.20.1 to address security issues.
 - Downgrades plotly dependency to address security issues.  Upgrading to address is not possible until they address this issue: https://github.com/plotly/plotly.js/issues/895
 -  Downgrades vis to 4.15.0 to address IE10 compatibility issues.  See this issue for details: https://github.com/almende/vis/issues/1839
 - Removes unused files from component.less.
 - Refactors the title bar for workspace name to be simpler and more compatible with IE10.  See content-title files for more details, as well as the workspace-menu files.
 - Updates anywhere max-height is being used with percentages (with dynamic content within them) to address IE10 issues.  Display block has rendering issues (it jumps to working, then not working, scrolls back to top randomly), but inline-block can handle it fine.  As a result, changed these block types to be inline-block with widths of 100%, so they’re effectively the same.  See editor.less, metacard-associations.less, metacard-history.less, query-advanced.less, query-basic.less, table-rearrange.less, table-visibility.less, editable-rows.less
 - Removes some extraneous floats in filter-comparator.less and query-src.less.
 - Updates help.view to be compatible with IE Edge and below.  See issue for details: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10574182/
 - Updates bulk-input view to better handle dates.  Issues were seen in all browsers when doing bulk updates on date fields and setting them to an existing value.  Now the value is converted to a proper date ahead of time, and the labels for the dates are more human readable.
 -  Changed almost everywhere that was using the in browser Date constructor to use Moment.  This helps ensure that our dates are properly formatted, and eliminates cross browser inconsistencies.  This led to updates in input-date.view, input-enum.view, metacard-history.view, query-feed.view, query-status.view, result-item.view, value.collection.view, QueryPolling.js, cql.js, Metacard.js, Query.js, Workspace.js, 
 - Removes unused files:  menu-item.action.hbs, menu-item.action.js, menu-item.action.less, menu-item.component.hbs, menu-item.component.js, menu-item.abstract.js, menu-item.collection.js, menu-item.collection.less, menu-item.collection.view.js, menu-item.less, menu-item.view.js, menu-item.submenus.js, query-federation.hbs, query-federation.less, query-federation.view.js, 
 - Update metacard-associations to handle another IE10 issue.  Specifically, they have a bug with line-height using ‘rem’ units for pseudo-elements.  See issue for details: https://connect.microsoft.com/IE/feedback/details/776744
 - Updates navigation.less to be more IE10 friendly by using a border instead of a box-shadow to achieve the same effect.
- Adds debounce to the paging.view updates to the selectionInterface.  This helps performance on all browsers, but especially in IE10 where unnecessary updates to visualizations can be particularly expensive.  The update is set to happen the first time it sees an update, and then at the end of updates that are happening faster than 200 ms.  
 - Updates paging.view to listen to reset event on the render rather than the paging state change, since unfortunately the state change happens before the collection page actually changes.  
 - Updates result-item.collection.view to debounce rather than throttle render updates.  This helps performance on all browsers, but especially in IE10.  Render method overwritten to account for possible destruction of the view before the render takes place.  
 - Updates autocomplete functionality in query-advanced.hbs, query-basic.hbs, and index.html to eliminate issues with unencrypted warnings in Firefox.  Instead of pointing to about:blank, which is seen as http, we now point to a blank page within our app that is https.  See https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox if you want more details.  
 - Adds type of button to cancel buttons on query-basic.hbs and query-advanced.hbs to avoid accidental cancellation issues in IE10.  Older IE tries to find the submit button when a user presses enter in an input, but has issues doing so.  See https://www.tjvantoll.com/2013/05/22/why-are-enter-keypresses-clicking-my-buttons-in-ie/ for details.  
 - Removes unnecessary styling from slideout.right.less.  This styling is already applied by slideout.less.  
 - Adds style in slideout.less to address IE10 issue where you can click through a transparent element.  See http://stackoverflow.com/questions/6480774/z-index-problem-in-ie-with-transparent-div for details.
 - Refactor how we do the system-usage view to support IE Edge and below.  Sandbox attribute of allow-same-origin is needed to allow us to write to the iframe. See http://stackoverflow.com/questions/13214419/alternatives-to-iframe-srcdoc for more details. Along the same vein, we wait for the user to be fetched before populating the iframe (so that his zoom preference is correct).  This necessitated a change in the user.js file as well to added a fetched attribute.  
 - Update theme-settings.view to provide better support for all browsers and below.  See http://stackoverflow.com/questions/18544890/onchange-event-on-input-type-range-is-not-triggering-in-firefox-while-dragging for details.
 - Refactored histogram.view to provide better performance on IE10, resulting in better performance all around.  Updates are now debounced instead of throttled, the timings are also lowered to 30ms to account for this.  There is now an updateHistogram method that is used when selections update, instead of reshowing the whole histogram.  The handleResize method now calles the resize method of plotly rather than the relayout method.  While the relayout method works well on modern browsers, it’s not efficient at all on IE10.  
 - Refactored how results are plotted on the maps.  There were a number of issues with the previous method on IE11 and below.  See https://github.com/openlayers/openlayers/issues/4090 for one of them.  Another issue is IE Edge and below don’t respect dominant baseline, so vertical centering of text is not so great.  See http://stackoverflow.com/questions/26722287/internet-explorer-10-doesnt-respect-svg-text-dominant-baseline-attribute and https://msdn.microsoft.com/en-us/library/gg558060(v=vs.85).aspx .  Luckily, our shapes are simple enough to draw on a canvas manually, so we can get around all of these issues.  Hence, DrawingUtility now exists to facilitate drawing result geometries on both cesium and openlayers.
 - Added fallback support for the cesium map.  If opengl fails, it will show a message and prompt them to switch to the openlayers map through either a button click, or simply waiting 10 seconds (thereafter it autoswitches).  See cesium.hbs, cesium.less, and cesium.view for those details.  Map.view and openlayers.view were also affected, since I needed to add a try catch around the map creation (try catch does not work on async events within the try catch).  
 - Updates map.cesium to fix an issue in IE11.  We weren’t properly destroying the default geocoder, which meant events still fired for it and threw errors (since the elements were detached from the dom).  Other browsers simply swallowed the error.  Regardless, the new method is more efficient.  
 - Added a special case to removeGeometry for unminified version of Cesium.  Still looking into why it only happens when using the unminified version, but this will be helpful to avoid red herrings during debugging.
 - Updates cluster.collection.view to debounce map clicks and prevent accidental clicks of clusters during drawing.
 - Adds debounce to cluster.view updateSelected method to allow better performance on IE10.  Adds logic to only updateCluster when change occurs.
 - Updates geometry.collection.view to ignore clicks on clusters.
 - Adds debounce to geometry.view updateSelected method to allow better performance on IE10.  Also adds logic to only updateGeometry when necessary (check if changed, or currently not visible).  
 - Updates map.view to properly destroy internal collections / views upon destruction.
 - Updates table.view to provide better performance / ux in browsers other than Chrome.  
 - Updates editable-rows.less to account for the header in max-height.
 - Splits main.js apart into ApplicationSetup and ApplicationStart.  ApplicationSetup contains the libraries, polyfills, and shims.  At the end, it calls ApplicationStart.
 - Disables cache for all ajax requests (added to ApplicationSetup).  IE10 aggressively caches anything if we don’t include this.  Things quickly go very wonky, especially when we’re making edits to metacards, associations, etc.
 - Eliminates unused getNiceDate in Common.js.
 - Eliminates unused methods in Query.js.
 - Updates User.js to allow others to determine if the User has been fetched yet.
 - Updates focus.less to account for IE11 and below behavior.  Had to remove opacity from being momentarily set to 1.  In most modern browsers, leaving it out of the next frame meant it transitioned back to its original opacity.  But in IE11 and below, it transitions to 0 opacity.
 - Updates input.less to account for IE10 issues.  IE10 does not take kindly to both padding and line-height being put on input fields, causing text within them to be offset.  Also added code to disable the clear button that IE pops up for input fields.  See http://stackoverflow.com/questions/23936236/input-text-at-the-bottom-on-ie10 for the line-height issue details.
 - Updates range-input.less to provide better cross browser experience.  Previously, in IE10 it would do strange things during different zooms.
 - Updates properties.js to use underscore for the match function.  The find method isn’t available on earlier versions of IE for arrays.
 -  Added two commented out lines to base.js to allow easier debugging of cesium and openlayers in the future.  

#### What does this PR do?
Officially adds support for: IE10 and above, Chrome, Firefox ESR and above.

@djblue 
@clockard 
@codice/ui 

@pklinef 
@andrewkfiedler

Open each officially supported browser and ensure that everything possible in one browser is possible in another, or that fallback (in the case of no webgl) takes place.  
I'm going to try and put together a session where we can each have different browsers open and go through a run of all the features to ensure everything is working.  

Previously, only Chrome was officially supported.

https://codice.atlassian.net/browse/DDF-2644


brendan-hofmann:
Does IE 10 and above include Edge? Does Chrome imply Safari, given that Blink is a fork of Webkit?

andrewkfiedler:
Verified with Edge.

andrewkfiedler:
Chrome implies only Chrome, not Safari.  And IE10 and above includes Edge.

PR #1613, opened 2017-01-31
Ddf 2698 export command

rzwiefel:
#### What does this PR do?
Causes workspace transformer to merge duplicate keys instead of throwing an exception. While our current and future system should try not to have duplicate keys, it was allowed in some legacy configurations so must be supported to deal with them. 
anyone
@jlcsmith
@kcwire

ingest a duplicate attribute descriptor. log in to ui. attempt to create workspace and ensure it doesn't fail.
Example Attribute descriptor that duplicates a builtin descriptor that can be put into `etc/definitions`
```json
{
  "inject": [
    {
      "attribute": "effective"
    }
  ],
  "attributeTypes": {
    "effective": {
      "type": "DATE_TYPE",
      "display": "Date Published",
      "stored": true,
      "indexed": true,
      "tokenized": false,
      "multivalued": true
    }
  }
}
```

https://codice.atlassian.net/browse/DDF-2755

### DIFF
--- a/catalog/core/catalog-core-commands/pom.xml
+++ b/catalog/core/catalog-core-commands/pom.xml
@@ -109,6 +109,16 @@
             <artifactId>platform-util-unavailableurls</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>net.lingala.zip4j</groupId>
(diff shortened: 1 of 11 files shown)


rzwiefel:
This was supposed to be something else.

PR #1616, opened 2017-02-01
DDF-2759: Feature configuration files are not always being processed

lessarderic:
Moved the WatchService directory registration in the `ConfigurationFilesPoller` `init()` method to ensure that no file creation and/or modification events are missed.

#### What does this PR do?
Fixes a race condition in the ConfigurationFilesPoller class that could cause some configuration files to not be read in at system startup.

@brjeter 
@oconnormi 

@figliold
@pklinef

Full builds including all integration tests.

[DDF-2759](https://codice.atlassian.net/browse/DDF-2759)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
@@ -72,10 +72,22 @@ public class ConfigurationFilesPoller implements Runnable {
         this.fileExtension = fileExtension;
     }
 
-    public void init() {
+    public void init() throws IOException {
         LOGGER.debug("Starting {}...",
                 this.getClass()
                         .getName());
+
+        try {
+            LOGGER.debug("Registering path [{}] with Watch Service.",
+                    configurationDirectoryPath.toString());
+            configurationDirectoryPath.register(watchService, ENTRY_CREATE);
+        } catch (IOException e) {
+            LOGGER.error("Unable to register path [{}] with Watch Service",
+                    configurationDirectoryPath.toString(),
+                    e);
+            throw e;
+        }
+
         executorService.execute(this);
     }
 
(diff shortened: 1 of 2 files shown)


platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -72,10 +72,22 @@ public ConfigurationFilesPoller(@NotNull Path configurationDirectoryPath,
...
+
+        try {
+            LOGGER.debug("Registering path [{}] with Watch Service.",
+                    configurationDirectoryPath.toString());
+            configurationDirectoryPath.register(watchService, ENTRY_CREATE);
+        } catch (IOException e) {
+            LOGGER.error("Unable to register path [{}] with Watch Service",
+                    configurationDirectoryPath.toString(),
+                    e);
+            throw e;
```
brjeter: ❓ This is probably the biggest change in behavior, since before it would just stop executing when an IOException would happen. Are we sure this is how we want the poller to behave now?

rzwiefel: I would guess so, since this is the init method vs the run method, returning will not work since it will just look like the init method completed successfully. So I'd guess the exception is necessary now. Though @lessarderic probably has the real answer.

brjeter: Yeah I was thinking Eric just wanted it to be more clear when this fails from a debugging standpoint, but I wanted to make sure that was his intent 😄

lessarderic: If the `ConfigurationFilesPoller` fails to properly initialize, we could end up with an application that's misconfigured or missing managed services. I felt that simply logging an error if we couldn't monitor the directory wasn't enough and that actually failing the container was a better option here (best effort vs. fail fast). But I'm open to other points of view :-)

PR #1618, opened 2017-02-01
DDF-2331 Made thread pool and readLockInterval configurable in Content Directory Monitor

bdeining:
#### What does this PR do?
This PR adds the ability to specify how many threads to use when configuring a Content Directory Monitor so files can be consumed in parallel.  It also reduces the default size of the readLockInterval to a more reasonable value. 
@rzwiefel @glenhein 
@brendan-hofmann 
@jlcsmith
@kcwire
Build / Install / configure CDM with multiple threads / Observe Parallel Processing 
[DDF-2331](https://codice.atlassian.net/browse/DDF-2331)
#### Checklist:
- [X] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
+++ b/catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
@@ -39,6 +39,7 @@ import org.slf4j.LoggerFactory;
 
 import ddf.catalog.Constants;
 import ddf.security.common.util.Security;
+
 import net.jodah.failsafe.Failsafe;
 import net.jodah.failsafe.RetryPolicy;
 
@@ -74,6 +75,16 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
 
     private final Executor configurationExecutor;
 
+    private static final int MAX_THREAD_SIZE = 8;
+
+    private static final int MIN_THREAD_SIZE = 1;
+
+    private static final int MIN_READLOCK_INTERVAL_MILLISECONDS = 100;
+
+    private Integer numThreads;
+
+    private Integer readLockIntervalMilliseconds;
+
     Processor systemSubjectBinder = new SystemSubjectBinder();
 
     /**
@@ -106,6 +117,36 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
         this.maxRetries = maxRetries;
         this.delayBetweenRetries = delayBetweenRetries;
         this.configurationExecutor = configurationExecutor;
+    }
+
+    /**
+     * Set the thread pool size with the given argument.  If the given argument is less than 1,
+     * numThreads is set to 1.  If the given argument is greater than the MAX_THREAD_SIZE, numThreads
+     * is set to MAX_THREAD_SIZE.
+     *
+     * @param numThreads - the specified size to make the thread pool
+     */
+    public void setNumThreads(Integer numThreads) {
+        this.numThreads = Math.max(numThreads, MIN_THREAD_SIZE);
+        this.numThreads = Math.min(this.numThreads, MAX_THREAD_SIZE);
+    }
+
+    public Integer getNumThreads() {
+        return numThreads;
+    }
+
+    /**
+     * Set the read lock interval for the Camel Route with the given argument.  If the readLockIntervalMilliseconds
+     * is less than 100, set it to 100.
+     *
+     * @param readLockIntervalMilliseconds
+     */
+    public void setReadLockIntervalMilliseconds(Integer readLockIntervalMilliseconds) {
+        this.readLockIntervalMilliseconds = Math.max(readLockIntervalMilliseconds, MIN_READLOCK_INTERVAL_MILLISECONDS);
+    }
+
+    public Integer getReadLockIntervalMilliseconds() {
+        return readLockIntervalMilliseconds;
     }
 
     /**
@@ -179,6 +220,8 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
         if (properties != null) {
             setMonitoredDirectoryPath((String) properties.get("monitoredDirectoryPath"));
             setProcessingMechanism((String) properties.get("processingMechanism"));
+            setNumThreads((Integer) properties.get("numThreads"));
+            setReadLockIntervalMilliseconds((Integer) properties.get("readLockIntervalMilliseconds"));
             String[] parameterArray = (String[]) properties.get(Constants.ATTRIBUTE_OVERRIDES_KEY);
             if (parameterArray != null) {
                 setAttributeOverrides(Arrays.asList(parameterArray));
@@ -270,11 +313,12 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
         return new RouteBuilder() {
             @Override
             public void configure() throws Exception {
-                // Configure the camel route to ignore changing file (larger files that are in the process of being copied)
+                // Configure the camel route to ignore changing files (larger files that are in the process of being copied)
                 // Set the readLockTimeout to continuously poll the directory so long as the directory monitor exists
-                // Set the readLockCheckInterval to check every 5 seconds
+                // Set the readLockCheckInterval to check every readLockIntervalMilliseconds
                 String inbox = "file:" + monitoredDirectory
-                        + "?recursive=true&moveFailed=.errors&readLock=changed&readLockTimeout=0&readLockCheckInterval=5000";
+                        + "?idempotent=true&readLockMinLength=0&recursive=true&moveFailed=.errors&readLock=changed&readLockTimeout=0&readLockCheckInterval="
+                        + readLockIntervalMilliseconds;
                 switch (processingMechanism) {
                 case DELETE:
                     inbox += "&delete=true";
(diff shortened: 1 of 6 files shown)


catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -109,6 +118,45 @@ public ContentDirectoryMonitor(CamelContext camelContext, int maxRetries,
...
 
     /**
+     * Set the thread pool size with the given argument.  If the given argument is less than 1,
+     * numThreads is set to 1.  If the given argument is greater than the MAX_THREAD_SIZE, numThreads
+     * is set to MAX_THREAD_SIZE.
+     *
+     * @param numThreads - the specified size to make the thread pool
+     */
+    public void setNumThreads(Integer numThreads) {
+        if (numThreads < 1) {
```
glenhein: Instead of the if-statements, how about:

this.numThreads = Math.max(numThreads, MIN_THREAD_SIZE);
this.numThreads = Math.min(this.numThreads, MAX_THREAD_SIZE);

That will reduce the branch-complexity.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -109,6 +118,45 @@ public ContentDirectoryMonitor(CamelContext camelContext, int maxRetries,
...
+    }
+
+    /**
+     * Set the read lock interval for the Camel Route with the given argument.  If the readLockIntervalMilliseconds
+     * is less than 100, set it to 100.
+     *
+     * @param readLockIntervalMilliseconds
+     */
+    public void setReadLockIntervalMilliseconds(Integer readLockIntervalMilliseconds) {
+        if (readLockIntervalMilliseconds < MIN_READLOCK_INTERVAL_MILLISECONDS) {
```
glenhein: you could use the same Math.min/Math.max trick:

this.readLockIntervalMilliseconds = Math.max(readLockIntervalMilliseconds, MIN_READLOCK_INTERVAL_MILLSECONDS);

catalog/core/catalog-core-directorymonitor/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -21,6 +21,14 @@
             name="Directory Path" id="monitoredDirectoryPath" required="true"
             type="String" default=""/>
 
+        <AD description="Specifies the number of threads to use when monitoring"
```
glenhein: Add a period at the end of the sentence.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -74,6 +75,14 @@
 
     private final Executor configurationExecutor;
 
+    private static final int MAX_THREAD_SIZE = 8;
```
brendan-hofmann: How was this maximum selected?

bdeining: This maximum was selected from the guidance in this ticket : https://codice.atlassian.net/browse/DDF-2756

brendan-hofmann:
That ticket mentions `Additionally, any pools in the system that are not using/injecting that system property to set their max pool size should be looked at and either fixed to do so or justified. In the latter case, comments should be left in code with the justification for using a custom value.` Should we consider using the system property or explaining why we are not?

bdeining:
@brendan-hofmann At first glance, 128 threads for disk i/o seems unnecessarily high.  I'm in the process of doing some benchmark testing to see what a reasonable maximum should be.  With regards to bottlenecks, one known bottleneck already is having a high readLockInterval.  Since camel will not consume files without a readLock, the interval between acquiring new ones will dictate the file ingestion speed.

catalog/core/catalog-core-directorymonitor/src/main/resources/OSGI-INF/metatype/metatype.xml
kcwire: This doesn't really tell me anything.  Does this mean that the number I put here will be the maximum number of concurrent files to be processed within a directory? 

An admin should not have to know about "threads" but rather what the result of changing this values means from a processing perspective. Also, if the maximum is 8 and it is hardcoded - I would at that to this description.

catalog/core/catalog-core-directorymonitor/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -21,6 +21,14 @@
...
+            name="Thread Pool Size" id="numThreads" required="true"
+            type="Integer" default="1"/>
+
+        <AD description="Specifies the time to wait (in milliseconds) before acquiring a lock on a file in the monitored directory.  This interval is used for sleeping between attempts to acquire the read lock on a file to be ingested.  The default value of 1 second is recommended."
```
kcwire: This is a great description btw. Although I would say a "default value of 1000 miliseconds" to keep the units consistent. Also fix the default value to match.

kcwire:
What testing has been done with this change? If you set the max number of threads and add several hundred files to a directory how well does the system behave? I'm curious if it's truly and improvement or if the results over a long enough timeline remain the same.

bdeining:
@kcwire @brendan-hofmann 
screen shot 2017-02-02 at 4 08 45 pm
After some testing, there is an obvious improvement in performance times when increasing the thread pool size.  It's worth noting that when the readLockInterval is set to 5000 milliseconds (the current default), that is what dictates the speed of ingest through the CDM (13 x 5 = 1 minute 5 seconds).  Going from 8 to 16 threads does not have a worthwhile performance increase (likely due to context switches).  Thus, a MAX_THREAD_SIZE of 8 seems appropriate.    The files tested were around 80mb each in size.  With 100 smaller files, the readLockInterval will dictate the speed since the files will be consumed very quickly (the number of threads may not have a significant impact).

catalog/core/catalog-core-directorymonitor/src/main/resources/OSGI-INF/metatype/metatype.xml
bdeining: done

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
bdeining: changed

catalog/core/catalog-core-directorymonitor/src/main/resources/OSGI-INF/metatype/metatype.xml
bdeining: fixed

brendan-hofmann:
Thanks for the detailed testing @bdeining! If I understand correctly, this means that the readLockInterval will need to be tuned for performance, likely contingent on the I/O performance of the target? Should we document this?

I don't think in-place monitoring currently has a readLockInterval. I'm not sure this really impacts your results, but worth noting.

bdeining:
@brendan-hofmann right.  This was the justification for making the readLockInterval configurable as well as the number of threads.  I agree that we should probably add this to the CDM documentation.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorTest.java
```diff
@@ -176,15 +178,53 @@ public void testDirectoryMonitorWithParameters() throws Exception {
...
                 .get(0);
-        assertThat(routeDefinition.toString(),
-                containsString("SetHeader[" + Constants.ATTRIBUTE_OVERRIDES_KEY
+        assertThat(routeDefinition.toString(), containsString(
+                "SetHeader[" + Constants.ATTRIBUTE_OVERRIDES_KEY
                         + ", simple{Simple: test1=someParameter1,test2=someParameter2}"));
     }
 
     @Test
+         public void testDirectoryMonitorThreadNumFallback() throws Exception {
```
bdeining: fixed

glenhein:
I built and installed the code. I configured a CCDM with 9 threads. I ingested 93 files. In the logs, I saw 8 threads (as expected) processing. Hero build was a success.

bdeining:
@brendan-hofmann Updated documentation to reflect changes

distribution/docs/src/main/resources/_contents/_resources/content-directory-monitor-contents.adoc
```diff
@@ -47,8 +47,16 @@ Regardless of the option, the ${branding} takes each file in a monitored directo
 ** If the original file is modified, the metacard is updated to reflect the new content.
 ** If the original file is renamed, the old metacard is deleted and a new metacard is created.
 
-.Atribute Mapper
-The Content Directory Monitor supports setting metacard attributes directly when ${branding} ingests a file. Custom mappings are entered in the form:
+.Parallel Processing
+The CDM supports parallel processing of files (up to 8 files processed concurrently).  This is configured by setting the number of *Maximum Concurrent Files* in the configuration.  A maximum of 8 is imposed to protect system resources.
+
+.Read Lock
```
brendan-hofmann: Is it worth noting that this does not apply to the `monitor in place` processing directive?

bdeining: Yes, added

PR #1620, opened 2017-02-01
DDF-2644 Fixes issue with datepicker dependency that was switched to npm

andrewkfiedler:
#### What does this PR do?
 - The datepicker was switched from being resolved through bower to being resolved through npm recently, with https://github.com/codice/ddf/commit/fc8a567a9e76f4020cbe7f2a92b1d37b3939d470
 - Unfortunately, the css has been missing since that commit.  When I added the css back in using webpack, I noticed the datepicker had some regressions.  As such, I rolled it back to the original bower version.

@andrewkfiedler
@pklinef 
@djblue 
@rzwiefel 
@tbatie 

Check out the Catalog UI and ensure the datepicker has the correct css.

### DIFF
--- a/catalog/ui/catalog-ui-search/package.json
+++ b/catalog/ui/catalog-ui-search/package.json
@@ -97,9 +97,9 @@
     "terraformer": "1.0.5",
     "terraformer-wkt-parser": "1.1.0",
     "uuid": "2.0.2",
-    "eonasdan-bootstrap-datetimepicker": "4.17.45",
     "openlayers": "3.20.1",
     "plotly.js": "1.10.5",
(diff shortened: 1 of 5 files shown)


vinamartin:
Built this change locally and attempted to test it.

screen shot 2017-02-01 at 2 47 35 pm

Here's what it looks like for me (this is Firefox, but it looks the same in Chrome). 
The only weird behavior I'm seeing is that when I hover over my selected date it unhighlights the date which seems a little strange. 

So, in the case where I have just selected a new date, I see it update in the textbox, however, I have to move my mouse off of the option I have just selected to actually see it highlighted. 

Behavior

PR #1622, opened 2017-02-01
DDF-2762 Update workspace sharing inputs to update on change rather than keydown

andrewkfiedler:
#### What does this PR do?
 - Previously they updated after each keypress, which causes the cursor to move to the end of the line during typing.

@pklinef 
@rzwiefel 
@djblue 
@garrettfreibott 

Go to the workspace sharing view.  Add an email.  Move the cursor to the center of the email and type.  Verify it doesn't move automatically to the end of the line.  Apply your change and verify the saving still works as it should.  

https://codice.atlassian.net/browse/DDF-2762


jlcsmith:
Built and tested successfully.

PR #1624, opened 2017-02-02
DDF-2763 Updates standing searches to retry when query status is unsuccessful

andrewkfiedler:
- Previously they only retried when the network request failed.  Now they will retry if the request succeeded, but the query status says unsuccessful.
 - Also updates the Sources polling to continueOnError rather than stop.
 - Updates Source polling to not begin immediately, since we fetch immediately.

@djblue 
@rzwiefel 
@jlcsmith 
@pklinef 

Add a standing search.  Cause it to fail but still return a 200.  Ensure that when queries can once again run successfully, a retry is made for the failed time range.

Previously the retries were only made when network requests outright failed.  Now they will happen if the status of a query is unsuccessful as well.

https://codice.atlassian.net/browse/DDF-2763

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/QueryPolling.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/QueryPolling.js
@@ -131,7 +131,7 @@ define([
 
     checkForFailures = function(responses, originalQuery, timeRange) {
         _.forEach(responses, function(response) {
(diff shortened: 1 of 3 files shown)


jlcsmith:
Built and tested.  Verified cases where the search returned an HTTP error code and when it returned 200 but was unsuccessful.  Notifications were received once the the source was reconnected - generally within 5 minutes.

PR #1629, opened 2017-02-03
DDF-2745: Fix itest request timing issue.

peterhuffer:
#### What does this PR do?
- Fixes a flaky test that attempted to update a Metacard immediately after ingesting it, which caused it to fail sometimes.

@emmberk @mweser

@coyotesqrl
@lessarderic

Run `TestCatalog` integration tests.

[DDF-2745](https://codice.atlassian.net/browse/DDF-2745)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
@@ -2962,14 +2962,19 @@ public class TestCatalog extends AbstractIntegrationTest {
                 .post(REST_PATH.getUrl())
                 .getHeader("id");
 
-        given().multiPart("parse.resource", tmpFile)
-                .multiPart(Core.RESOURCE_URI, overrideResourceUri)
-                .expect()
-                .log()
-                .headers()
-                .statusCode(HttpStatus.SC_BAD_REQUEST)
-                .when()
-                .put(REST_PATH.getUrl() + id);
+        expect("Sending bad HTTP PUT request with overridden resource-uri").within(1,
+                TimeUnit.MINUTES)
+                .until(() -> {
+                    Response response = given().multiPart("parse.resource", tmpFile)
+                            .multiPart(Core.RESOURCE_URI, overrideResourceUri)
+                            .expect()
+                            .log()
+                            .headers()
+                            .when()
+                            .put(REST_PATH.getUrl() + id);
+
+                    return response.getStatusCode() == HttpStatus.SC_BAD_REQUEST;
+                });
 
         deleteMetacard(id);
     }


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -2962,14 +2962,19 @@ public void testUpdateStorageCannotOverrideResourceUri()
...
 
-        given().multiPart("parse.resource", tmpFile)
-                .multiPart(Core.RESOURCE_URI, overrideResourceUri)
-                .expect()
-                .log()
-                .headers()
-                .statusCode(HttpStatus.SC_BAD_REQUEST)
-                .when()
-                .put(REST_PATH.getUrl() + id);
+        expect("Sending bad HTTP PUT request with overridden resource-uri").within(1,
```
lessarderic: ✏️  We may want to change the message to something that says what we're expecting instead of what we're doing, e.g., `HTTP PUT request with overridden resource-uri fails`.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -2962,14 +2962,19 @@ public void testUpdateStorageCannotOverrideResourceUri()
...
+                TimeUnit.MINUTES)
+                .until(() -> {
+                    Response response = given().multiPart("parse.resource", tmpFile)
+                            .multiPart(Core.RESOURCE_URI, overrideResourceUri)
+                            .expect()
+                            .log()
+                            .headers()
+                            .when()
+                            .put(REST_PATH.getUrl() + id);
```
lessarderic: ❗️ Instead of trying the PUT multiple times, it might be safer to do a GET until we know the metacard with the right resource-uri is available and then make one PUT request and check the status code. Calling PUT multiple times assumes that PUT will only fail if the resource-uri has changed which might not alway be the case.

emmberk:
Commit message could be more descriptive

peterhuffer:
This failure was unable to be reliably reproduced. The cause of the 200 response seen from the PUT request is not known as well. We are going to close this PR and leave it as is until this can be reliably reproduced.

We cannot do a GET request in the WaitCondition because a CREATEd metacard is temporarily stored in a cache before being committed to SOLR, so even if we got the metacard back from a GET, it would not guarantee that it is in the provider.

Doing multiple PUT requests does not seem like the correct answer either, since if we do get a 200 again, it will mean that the update was successful, fail the assertion, and then continue to issue more PUT requests. Since the metacard would have been updated, it would continually get 200 since the resource-uri in the catalog is now the resource-uri we are submitting as an update.

PR #1631, opened 2017-02-03
DDF-2612 Fixed CachingFederationStrategy source id issue and added itest to verify cache works

clockard:
#### What does this PR do?
Fixes CachingFederationStrategy so metacard cache queries will have the right source ids.
@bdeining @mcalcote 
@pklinef
@shaundmorris
Verify itests pass
[](https://codice.atlassian.net/browse/DDF-2612)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
@@ -14,7 +14,6 @@
 package ddf.catalog.cache.solr.impl;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -241,7 +240,7 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
 
                     QueryRequest sourceQueryRequest = new QueryRequestImpl(modifiedQuery,
                             queryRequest.isEnterprise(),
-                            Collections.singleton(source.getId()),
+                            new HashSet<>(queryRequest.getSourceIds()),
                             new HashMap<>(queryRequest.getProperties()));
                     try {
                         for (PreFederatedQueryPlugin service : preQuery) {
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
@@ -81,6 +81,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.karaf.bundle.core.BundleService;
 import org.codice.ddf.itests.common.AbstractIntegrationTest;
+import org.codice.ddf.itests.common.KarafConsole;
 import org.codice.ddf.itests.common.annotations.BeforeExam;
 import org.codice.ddf.itests.common.annotations.ConditionalIgnoreRule;
 import org.codice.ddf.itests.common.annotations.ConditionalIgnoreRule.ConditionalIgnore;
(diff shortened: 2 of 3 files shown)


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -1518,6 +1519,47 @@ public void testRetrievalReliabilityFails() throws Exception {
...
+                .body(srcRequest)
+                .when()
+                .post(cqlUrl)
+                .then()
+                .log()
+                .all()
+                .statusCode(200);
+
+        //CacheBulkProcessor could take up to 10 seconds to flush the cached results into solr
+        sleep(12000);
```
pklinef: This hard sleep might not work on all systems due to difference in cpu/io and become flaky.  I wonder if we could use the Solr client factory to get a Solrj client and query the metacard cache directly in a `waitFor`.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
```diff
@@ -241,7 +240,7 @@ private QueryResponse sourceFederate(List<Source> sources, final QueryRequest qu
 
                     QueryRequest sourceQueryRequest = new QueryRequestImpl(modifiedQuery,
                             queryRequest.isEnterprise(),
-                            Collections.singleton(source.getId()),
+                            queryRequest.getSourceIds(),
```
rzwiefel: Should we do `new HashSet<>(queryRequest.getSourceIds())` here (like below) to ensure there are no racy issues with mutability?

clockard: I don't think this should be an issue but I will create the new hash set to be on the safe side.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
clockard: Used console commands check the cache. Will wait up to 20 seconds for the cache to be populated.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -1518,6 +1530,79 @@ public void testRetrievalReliabilityFails() throws Exception {
...
+
+        //Start with a clean cache
+        console.runCommand(CLEAR_CACHE);
+
+        long start = System.currentTimeMillis();
+        long waitTime = 20000;
+        boolean resultsFound = true;
+        while (resultsFound && System.currentTimeMillis() - start < waitTime) {
+            resultsFound = !console.runCommand(QUERY_CACHE)
+                    .contains(" 0 result(s)");
```
pklinef: I think it would be less brittle if we injected the catalog framework and did a query on the cache that way instead of parsing the output of the cache command.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -1518,6 +1530,79 @@ public void testRetrievalReliabilityFails() throws Exception {
...
+    @Test
+    public void testMetacardCache() throws Exception {
+
+        //Start with a clean cache
+        console.runCommand(CLEAR_CACHE);
+
+        long start = System.currentTimeMillis();
+        long waitTime = 20000;
+        boolean resultsFound = true;
+        while (resultsFound && System.currentTimeMillis() - start < waitTime) {
```
pklinef: Is there a reason we cannot use the `waitFor` helper?

clockard: Updated to use the helper.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
clockard: I have updated this to not rely on the console command. It should be much less brittle now.

clockard:
Clean repo build passed and manually verified the fix.

clockard:
Clean repo build passed. :shipit:

PR #1524, opened 2017-01-04
DDF-2671: Added visibility of each sources active url and warnings for local addresses

oconnormi:
#### What does this PR do?

* Adds a column to the sources list to display the currently active URL
* Adds a warning when the currently active URL is pointing at `localhost` or `org.codice.ddf.system.hostname`

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?

@figliold 
@brjeter 
@Lambeaux 
@bantillo 
@Joy603 

#### Select at least one member from relevant component team(s) from below (at least one component team member needs to approve the PR).

@codice/ui 
@andrewkfiedler 

#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).

@lessarderic
@pklinef

#### What are the relevant tickets?

[DDF-2671](https://codice.atlassian.net/browse/DDF-2671)

#### Screenshots (if appropriate)

sources_loopback_warning
sources_loopback_warnings
sources_loopback_warning_tooltip

#### Checklist:
- [ ] Change Log Updated

PR #1632, opened 2017-02-03
DDF-2671: Added length limit to source bindingUrl column

oconnormi:
#### What does this PR do?

Limits the size of the bindingUrl column introduced in #1524 

If max size is exceeded overflow is hidden and an ellipsis is displayed.

@lessarderic 
@Lambeaux 

@codice/ui 
@andrewkfiedler 

@figliold 
@pklinef 

[DDF-2671](https://codice.atlassian.net/browse/DDF-2671)

image

### DIFF
--- a/catalog/admin/module/catalog-admin-module-sources/src/main/webapp/js/view/Source.view.js
+++ b/catalog/admin/module/catalog-admin-module-sources/src/main/webapp/js/view/Source.view.js
@@ -167,8 +167,8 @@ define([
         SourceView.SourcePage = Marionette.Layout.extend({
             template: 'sourcePage',
             events: {
-                'click .removeSourceLink': 'removeSource',
-                'click .addSourceLink': 'addSource'
+                'click #removeSourceButton': 'removeSource',
+                'click #addSourceButton': 'addSource'
             },
             initialize: function () {
                 _.bindAll(this);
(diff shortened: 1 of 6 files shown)


oconnormi:
Not sure if this is really the ideal solution. The max-width needs to be pretty small in order to not cause issues with the whole overall table and almost renders the column useless. The main purpose of the column is to provide a quick look at the value of the current url without having click into the source edit modal. Even with this limit it could be argued that an admin should still be able to see if anything is extremely off or unexpected. It might also be better in the future to look into a way of combining this and it's associated loopback warning into the status badge at some point instead

coyotesqrl:
@oconnormi, could you add a hover-over tooltip, maybe? One that displayed the full name?

oconnormi:
Alright, I actually originally had that but thought it looked sorta odd. I'll put up a screenshot for that as well

andrewkfiedler:
Unless there's a reason to leave them there, I think the solution should involve moving the buttons outside of the table.  On smaller screens, it's likely that the buttons will still get pushed off of the screen.

lessarderic:
@andrewkfiedler, are you suggesting to move the buttons _and_ truncate the URL, or just move the buttons?

oconnormi:
Added a tooltip to make it possible to see urls that are overflowing. Also moved the buttons out of the table

screenshots: 

image

lessarderic:
Is there an easy way to reduce the blank spaces on both sides of the Binding URLs?

oconnormi:
@lessarderic I had tried, but nothing seemed to work other than increasing the binding url column's max-width

lessarderic:
:shipit:

pklinef:
@oconnormi Can you please address @andrewkfiedler's comment about the buttons being pushed out of view?

lessarderic:
@pklinef, @oconnormi has already moved the buttons above the table (see most recent screenshot). We're just waiting on @andrewkfiedler's approval and @Joy603 to complete the heroing and we should be good to merge.

Lambeaux:
Hero was successful 👍

PR #1633, opened 2017-02-03
DDF 2698 export command + import command

rzwiefel:
#### What does this PR do?
Adds a new catalog Export command that includes history and its content
@bdeining 
notsure
@jlcsmith
@kcwire

https://codice.atlassian.net/browse/DDF-2698

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -45,7 +45,6 @@ import org.apache.karaf.shell.api.action.Command;
 import org.apache.karaf.shell.api.action.Option;
 import org.apache.karaf.shell.api.action.lifecycle.Reference;
 import org.apache.karaf.shell.api.action.lifecycle.Service;
-import org.apache.shiro.SecurityUtils;
 import org.codice.ddf.catalog.transformer.zip.JarSigner;
 import org.codice.ddf.commands.catalog.export.ExportItem;
 import org.codice.ddf.commands.catalog.export.IdAndUriMetacard;
@@ -74,7 +73,6 @@ import ddf.catalog.resource.ResourceNotSupportedException;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.transform.CatalogTransformerException;
 import ddf.catalog.transform.MetacardTransformer;
-import ddf.security.SubjectUtils;
 import ddf.security.common.audit.SecurityLogger;
 import net.lingala.zip4j.core.ZipFile;
 import net.lingala.zip4j.exception.ZipException;
@@ -211,14 +209,18 @@ public class ExportCommand extends CqlCommands {
         for (Result result : new QueryResulterable(catalogFramework,
                 (i) -> getQuery(filter, i, PAGE_SIZE),
                 PAGE_SIZE)) {
-            seenIds.add(result.getMetacard().getId());
-            writeToZip(zipFile, result);
-            exportedItems.add(new ExportItem(result.getMetacard()
-                    .getId(),
-                    getTag(result),
-                    result.getMetacard()
-                            .getResourceURI(),
-                    getDerivedResources(result)));
+            if (!seenIds.contains(result.getMetacard().getId())) {
+                writeToZip(zipFile, result);
+                exportedItems.add(new ExportItem(result.getMetacard()
+                        .getId(),
+                        getTag(result),
+                        result.getMetacard()
+                                .getResourceURI(),
+                        getDerivedResources(result)));
+                seenIds.add(result.getMetacard()
+                        .getId());
+            }
+
             // Fetch and export all history for each exported item
             for (Result revision : new QueryResulterable(catalogFramework,
                     (i) -> getQuery(getHistoryFilter(result), i, PAGE_SIZE),
@@ -233,6 +235,7 @@ public class ExportCommand extends CqlCommands {
                         revision.getMetacard()
                                 .getResourceURI(),
                         getDerivedResources(result)));
+                seenIds.add(revision.getMetacard().getId());
             }
         }
         return exportedItems;
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -20,7 +20,6 @@ import java.net.URISyntaxException;
 import java.nio.file.Paths;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -49,7 +49,7 @@ import org.codice.ddf.catalog.transformer.zip.JarSigner;
 import org.codice.ddf.commands.catalog.export.ExportItem;
 import org.codice.ddf.commands.catalog.export.IdAndUriMetacard;
 import org.codice.ddf.commands.util.CatalogCommandRuntimeException;
-import org.codice.ddf.commands.util.QueryResulterable;
+import org.codice.ddf.commands.util.QueryResultIterable;
 import org.fusesource.jansi.Ansi;
 import org.geotools.filter.text.cql2.CQLException;
 import org.opengis.filter.Filter;
@@ -80,7 +80,9 @@ import net.lingala.zip4j.exception.ZipException;
 import net.lingala.zip4j.model.ZipParameters;
 
 /**
- * Experimental.
+ * Exports Metacards, History, and their content into a zip file.
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
  */
 @Service
 @Command(scope = CatalogCommands.NAMESPACE, name = "export", description = "Exports Metacards and history from the current Catalog")
@@ -245,7 +247,7 @@ public class ExportCommand extends CqlCommands {
     private List<ExportItem> doMetacardExport(/*Mutable,IO*/ZipFile zipFile, Filter filter) {
         Set<String> seenIds = new HashSet<>(1024);
         List<ExportItem> exportedItems = new ArrayList<>();
-        for (Result result : new QueryResulterable(catalogFramework,
+        for (Result result : new QueryResultIterable(catalogFramework,
                 (i) -> getQuery(filter, i, PAGE_SIZE),
                 PAGE_SIZE)) {
             if (!seenIds.contains(result.getMetacard()
@@ -262,7 +264,7 @@ public class ExportCommand extends CqlCommands {
             }
 
             // Fetch and export all history for each exported item
-            for (Result revision : new QueryResulterable(catalogFramework,
+            for (Result revision : new QueryResultIterable(catalogFramework,
                     (i) -> getQuery(getHistoryFilter(result), i, PAGE_SIZE),
                     PAGE_SIZE)) {
                 if (seenIds.contains(revision.getMetacard()
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -228,29 +228,6 @@ public class ExportCommand extends CqlCommands {
         return null;
     }
 
-    private StringBuilder getUserInputModifiable() throws IOException {
-        int in;
-        StringBuilder builder = new StringBuilder();
-        while ((in = session.getKeyboard()
-                .read()) != '\r') {
-            if (in == 127) {
-                if (builder.length() > 0) {
-                    builder.deleteCharAt(builder.length() - 1);
-                }
-                console.print((char) 8);
-                console.print(' ');
-                console.print((char) 8);
-
-            } else {
-                builder.append((char) in);
-                console.print((char) in);
-            }
-            console.flush();
-        }
-        console.println();
-        return builder;
-    }
-
     private File initOutputFile(String output) {
         String resolvedOutput;
         File initialOutputFile = new File(output);
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -122,6 +122,9 @@ public class ExportCommand extends CqlCommands {
             "archived"}, multiValued = false, description = "Equivalent to --cql \"\\\"metacard-tags\\\" like 'deleted'\"")
     boolean archived = false;
 
+    @Option(name = "--UNSAFE-NO-VERIFY", required = false, multiValued = false, description = "Produces the export zip but does NOT sign the resulting zip file. This file will not be able to be verified on import for integrity and security.")
+    boolean unsafe = false;
+
     @Override
     protected Object executeWithSubject() throws Exception {
         Filter filter = getFilter();
@@ -188,18 +191,23 @@ public class ExportCommand extends CqlCommands {
             doDelete(exportedItems, exportedContentItems);
         }
 
-        console.println("Signing zip file...");
-        start = Instant.now();
-        jarSigner.signJar(zipFile.getFile(),
-                System.getProperty("org.codice.ddf.system.hostname"),
-                System.getProperty("javax.net.ssl.keyStorePassword"),
-                System.getProperty("javax.net.ssl.keyStore"),
-                System.getProperty("javax.net.ssl.keyStorePassword"));
-        console.println("zip file signed in: " + Duration.between(start, Instant.now())
-                .toString()
-                .substring(2)
-                .replaceAll("(\\d[HMS])(?!$)", "$1 ")
-                .toLowerCase());
+        if (!unsafe) {
+            SecurityLogger.audit("Signing exported data. file: [{}]",
+                    zipFile.getFile()
+                            .getName());
+            console.println("Signing zip file...");
+            start = Instant.now();
+            jarSigner.signJar(zipFile.getFile(),
+                    System.getProperty("org.codice.ddf.system.hostname"),
+                    System.getProperty("javax.net.ssl.keyStorePassword"),
+                    System.getProperty("javax.net.ssl.keyStore"),
+                    System.getProperty("javax.net.ssl.keyStorePassword"));
+            console.println("zip file signed in: " + Duration.between(start, Instant.now())
+                    .toString()
+                    .substring(2)
+                    .replaceAll("(\\d[HMS])(?!$)", "$1 ")
+                    .toLowerCase());
+        }
         console.println("Export complete.");
         console.println("Exported to: " + zipFile.getFile()
                 .getCanonicalPath());
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -55,7 +55,6 @@ import org.codice.ddf.commands.catalog.export.ExportItem;
 import org.codice.ddf.commands.catalog.export.IdAndUriMetacard;
 import org.codice.ddf.commands.util.CatalogCommandRuntimeException;
 import org.geotools.filter.text.cql2.CQLException;
-import org.opengis.filter.And;
 import org.opengis.filter.Filter;
 import org.opengis.filter.sort.SortBy;
 import org.osgi.framework.InvalidSyntaxException;
@@ -118,8 +117,7 @@ public class ExportCommand extends CqlCommands {
 
     private Filter revisionFilter;
 
-    @Reference
-    private JarSigner jarSigner;
+    private JarSigner jarSigner = new JarSigner();
 
     @Reference
     private StorageProvider storageProvider;
@@ -192,13 +190,14 @@ public class ExportCommand extends CqlCommands {
         }
 
         console.println("Signing zip file...");
+        start = Instant.now();
         jarSigner.signJar(zipFile.getFile(),
                 System.getProperty("org.codice.ddf.system.hostname"),
                 System.getProperty("javax.net.ssl.keyStorePassword"),
                 System.getProperty("javax.net.ssl.keyStore"),
                 System.getProperty("javax.net.ssl.keyStorePassword"));
-        console.println("zip file signed.");
-
+        console.println("zip file signed in: " + Duration.between(start, Instant.now()));
+        console.println("Export complete.");
         return null;
     }
 
@@ -217,7 +216,7 @@ public class ExportCommand extends CqlCommands {
     }
 
     private List<ExportItem> doMetacardExport(ZipFile zipFile, Filter filter) {
-        List<ExportItem> exportedItems = new ArrayList<>(1024);
+        List<ExportItem> exportedItems = new ArrayList<>();
         for (Result result : new QueryResulterable(catalogFramework,
                 (i) -> getQuery(filter, i, PAGE_SIZE),
                 PAGE_SIZE)) {
@@ -263,7 +262,7 @@ public class ExportCommand extends CqlCommands {
                         .getSchemeSpecificPart()))
                 .collect(Collectors.toList());
 
-        List<ExportItem> exportedContentItems = new ArrayList<>(1024);
+        List<ExportItem> exportedContentItems = new ArrayList<>();
         for (ExportItem contentItem : contentItemsToExport) {
             ResourceResponse resource;
             try {
@@ -286,6 +285,21 @@ public class ExportCommand extends CqlCommands {
         Instant start;
         console.println("Starting delete");
         start = Instant.now();
+        for (ExportItem exportedContentItem : exportedContentItems) {
+            try {
+                DeleteStorageRequestImpl deleteRequest =
+                        new DeleteStorageRequestImpl(Collections.singletonList(new IdAndUriMetacard(
+                                exportedContentItem.getId(),
+                                exportedContentItem.getResourceURI())),
+                                exportedContentItem.getId(),
+                                Collections.emptyMap());
+                storageProvider.delete(deleteRequest);
+                storageProvider.commit(deleteRequest);
+            } catch (StorageException e) {
+                printErrorMessage(
+                        "Could not content for metacard: " + exportedContentItem.toString());
+            }
+        }
         for (ExportItem exported : exportedItems) {
             try {
                 catalogProvider.delete(new DeleteRequestImpl(exported.getId()));
@@ -294,16 +308,6 @@ public class ExportCommand extends CqlCommands {
             }
         }
 
-        for (ExportItem exportedContentItem : exportedContentItems) {
-            try {
-                storageProvider.delete(new DeleteStorageRequestImpl(Collections.singletonList(new IdAndUriMetacard(
-                        exportedContentItem.getId(),
-                        exportedContentItem.getResourceURI())), Collections.emptyMap()));
-            } catch (StorageException e) {
-                printErrorMessage(
-                        "Could not content for metacard: " + exportedContentItem.toString());
-            }
-        }
         console.println(
                 "Metacards and Content deleted in: " + Duration.between(start, Instant.now()));
         console.println("Number of content deleted: " + exportedItems.size());
@@ -446,23 +450,11 @@ public class ExportCommand extends CqlCommands {
                 TimeUnit.MINUTES.toMillis(1)), new HashMap<>());
     }
 
-    private And getIdsFilter(List<String> ids, int pageSize, Integer index) {
-        return filterBuilder.allOf(filterBuilder.attribute(Metacard.TAGS)
-                        .is()
-                        .like()
-                        .text("*"),
-                filterBuilder.anyOf(ids.stream()
-                        .skip(Math.max(0, index - 1))
-                        .limit(pageSize)
-                        .map(id -> filterBuilder.attribute(Metacard.ID)
-                                .is()
-                                .equalTo()
-                                .text(id))
-                        .collect(Collectors.toList())));
-    }
-
     /**
      * Effectively a cursor over the results of a filter that automatically pages through all results
+     * <br/>
+     * Throws a {@link CatalogCommandRuntimeException} if anything goes wrong during iteration or
+     * querying
      */
     class QueryResulterable implements Iterable<Result> {
         private final CatalogFramework catalog;
@@ -475,7 +467,8 @@ public class ExportCommand extends CqlCommands {
          * For paging through a single filter with a default pageSize of 64
          *
          * @param catalog catalog to query
-         * @param filter  The filter to query with
+         * @param filter  A dynamic supplier of a filter that takes the current index such that
+         *                the caller can control iteration based on their own logic
          */
         public QueryResulterable(CatalogFramework catalog, Function<Integer, QueryRequest> filter) {
             this(catalog, filter, 64);
@@ -485,7 +478,8 @@ public class ExportCommand extends CqlCommands {
          * For paging through a single filter.
          *
          * @param catalog  catalog to query
-         * @param filter   The filter to query with
+         * @param filter   A dynamic supplier of a filter that takes the current index such that
+         *                 the caller can control iteration based on their own logic
          * @param pageSize How many results should each page hold
          */
         public QueryResulterable(CatalogFramework catalog, Function<Integer, QueryRequest> filter,
@@ -495,18 +489,6 @@ public class ExportCommand extends CqlCommands {
             this.pageSize = pageSize;
         }
 
-        //        /**
-        //         * For efficient batching and iterating by a given list of metacard ids.
-        //         *
-        //         * @param catalog  catalog to query
-        //         * @param ids      List of ids to query for and iterate through
-        //         * @param pageSize How many results should each page hold
-        //         */
-        //        QueryResulterable(CatalogFramework catalog, Function<Integer, QueryRequestImpl> filter, int pageSize) {
-        //            this.catalog = catalog;
-        //            this.filter = (i) -> getQuery(getIdsFilter(ids, pageSize, i), 1, pageSize + 1);
-        //            this.pageSize = pageSize;
-        //        }
         @Override
         public Iterator<Result> iterator() {
             return new ResultQueryIterator();
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -151,9 +152,8 @@ public class ExportCommand extends CqlCommands {
             return null;
         }
 
-        SecurityLogger.audit("Called catalog:export command with path : {}\nCurrent user: {}",
-                output,
-                SubjectUtils.getName(SecurityUtils.getSubject()));
+        SecurityLogger.audit("Called catalog:export command with path : {}",
+                output);
 
         // TODO (RCZ) - Warn if breaking associations
         ZipFile zipFile = new ZipFile(outputFile);
@@ -186,6 +186,8 @@ public class ExportCommand extends CqlCommands {
                 System.getProperty("javax.net.ssl.keyStorePassword"));
         console.println("zip file signed in: " + Duration.between(start, Instant.now()));
         console.println("Export complete.");
+        console.println("Exported to: " + zipFile.getFile()
+                .getCanonicalPath());
         return null;
     }
 
@@ -204,10 +206,12 @@ public class ExportCommand extends CqlCommands {
     }
 
     private List<ExportItem> doMetacardExport(/*Mutable,IO*/ZipFile zipFile, Filter filter) {
+        Set<String> seenIds = new HashSet<>(1024);
         List<ExportItem> exportedItems = new ArrayList<>();
         for (Result result : new QueryResulterable(catalogFramework,
                 (i) -> getQuery(filter, i, PAGE_SIZE),
                 PAGE_SIZE)) {
+            seenIds.add(result.getMetacard().getId());
             writeToZip(zipFile, result);
             exportedItems.add(new ExportItem(result.getMetacard()
                     .getId(),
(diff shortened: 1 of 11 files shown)


bdeining:
@rzwiefel The catalog-transformer-zip can zip up and sign binary content items.  Is there any way we can leverage that code instead (and/or JarSigner)?

rzwiefel:
@bdeining I couldn't because I needed more fine control over when we signed and such. I have to make multiple requests and would have called  the transformer a large number of times which means it would have updated the zip and kept signing it after each change. And currently it's taking almost 20 minutes to sign a 20GB Jar just once at the end.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+            "-o"})
+    String output = Paths.get(System.getProperty("ddf.home"), name.get())
+            .toString();
+
+    @Option(name = "--delete", required = true, aliases = {"-d",
+            "delete"}, multiValued = false, description = "Whether or not to delete metacards after export")
+    boolean delete = false;
+
+    @Option(name = "--archived", required = false, aliases = {"-a",
+            "archived"}, multiValued = false, description = "Equivalent to --cql \"\\\"metacard-tags\\\" like 'deleted'\"")
```
jlcsmith: Is this description true?  Won't it also include version history for the deleted items?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+    private final int PAGE_SIZE = 64;
+
+    private Filter revisionFilter;
+
+    private JarSigner jarSigner = new JarSigner();
+
+    @Reference
+    private StorageProvider storageProvider;
+
+    @Option(name = "--output", description = "Output file to export Metacards into. Paths are absolute and must be in quotes. Will default to auto generated name inside of ddf.home", multiValued = false, required = false, aliases = {
```
jlcsmith: The description should also say that content is exported.

jlcsmith:
@rzwiefel - there is a findbugs failure that needs to be addressed

jlcsmith:
There needs to be a "force" option if there are any prompts in the command so this can be run by the scheduler.  Also there needs to be a "CQL" option to export/delete a portion of the catalog

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        Filter filter = getFilter();
+        transformer = getTransformers(DEFAULT_TRANSFORMER_ID).stream()
+                .findFirst()
+                .orElseThrow(() -> new CatalogCommandRuntimeException(
+                        "Could not find transformer " + DEFAULT_TRANSFORMER_ID));
+        revisionFilter = initRevisionFilter();
+
+        final File outputFile = initOutputFile(output);
```
jlcsmith: If the output location isn't provided, we should provide a default so this can easily be run as a scheduled job

rzwiefel: There already is a default location and filename generated. Generates a file called export-timestamp.zip into the ddf.home for unless one is specified.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: Correct. The entire command always I searched and includes any history items for any command, whether  by using --archived or by specifying cql.

rzwiefel:
@jlcsmith there already is cql flag, it's inherited from a superclass.  Currently there's no prompts. But I might add project specific prompts in their respective downstream code bases.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+            return null;
+        }
+
+        String filename = FilenameUtils.getName(outputFile.getPath());
+        if (StringUtils.isBlank(filename) || !filename.endsWith(".zip")) {
+            console.println("Filename must end with '.zip' and not be blank");
+            return null;
+        }
+
+        SecurityLogger.audit("Called catalog:dump command with path : {}\nCurrent user: {}",
```
jlcsmith: Incorrect command name here "catalog;dump"

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+                        .getId(),
+                        getTag(revision),
+                        revision.getMetacard()
+                                .getResourceURI()));
+            }
+        }
+        return exportedItems;
+    }
+
+    private List<ExportItem> doContentExport(/*Mutable,IO*/ZipFile zipFile,
```
jlcsmith: I may be missing something, but I don't think this handles exporting derived content.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+
+        List<ExportItem> exportedContentItems = new ArrayList<>();
+        for (ExportItem contentItem : contentItemsToExport) {
+            ResourceResponse resource;
+            try {
+                resource =
+                        catalogFramework.getLocalResource(new ResourceRequestById(contentItem.getId()));
+            } catch (IOException | ResourceNotSupportedException e) {
+                throw new CatalogCommandRuntimeException(
+                        "Something went wrong while fetching resources",
```
jlcsmith: How about 
`throw new CatalogCommandRuntimeException("Unable to retrieve resource for {}.  Resource will not be included in export.", contentId.getId(), e);`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+        for (ExportItem exportedContentItem : exportedContentItems) {
+            try {
+                storageProvider.delete(new DeleteStorageRequestImpl(Collections.singletonList(new IdAndUriMetacard(
+                        exportedContentItem.getId(),
+                        exportedContentItem.getResourceURI())),
+                        exportedContentItem.getId(),
+                        Collections.emptyMap()));
+            } catch (StorageException e) {
+                printErrorMessage(
+                        "Could not content for metacard: " + exportedContentItem.toString());
```
jlcsmith: "Could not **delete** content for metacard"

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
jlcsmith: Also, I believe we need to audit at least the ids, whether they were just exported or also deleted, and the location of the export file.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,569 @@
...
+        console.println("Starting content export...");
+        start = Instant.now();
+        List<ExportItem> exportedContentItems = doContentExport(zipFile, exportedItems);
+        console.println("Content exported in: " + Duration.between(start, Instant.now()));
+        console.println("Number of content exported: " + exportedContentItems.size());
+
+        console.println();
+
+        if (delete) {
+            doDelete(exportedItems, exportedContentItems);
```
jlcsmith: I am wondering if it makes sense to thread off the delete from the jar signing.  Both of these could be long running tasks and they don't depend on each other.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
jlcsmith: Ah, thanks I missed that.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: Yup it does both. The list of items from the metacard export contains the history and original metacards

rzwiefel: Or do you mean the derived association???

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: I thought @pklinef said we shouldn't audit the id's , unless we set it up to another logger

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
jlcsmith: I specifically mean the resources pointed to by Metacard.DERIVED_RESOURCE_URI.  Are they included in the export?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: btw this exception halts the export. Though the `ResourceNotSupportedException` should probably go with the `ResourceNotFoundException` below in the catch block that continues on with the export

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: Updated to include derived resources

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
pklinef: I think it is ok to audit the list of IDs in the audit log.  In general I think we try to avoid auditing anything more than just the IDs so we would not include titles for example.

@codice/security

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,479 @@
...
+            return null;
+        }
+
+        String filename = FilenameUtils.getName(outputFile.getPath());
+        if (StringUtils.isBlank(filename) || !filename.endsWith(".zip")) {
+            console.println("Filename must end with '.zip' and not be blank");
+            return null;
+        }
+
+        SecurityLogger.audit("Called catalog:export command with path : {}\nCurrent user: {}",
```
stustison: If you need to pass the subject in for some reason (the audit command already pulls the current user), then you should use the correct method to do so: https://github.com/codice/ddf/blob/master/platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/SecurityLogger.java#L140

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/export/ExportItem.java
```diff
@@ -0,0 +1,67 @@
...
+import java.net.URI;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+public class ExportItem {
+    private String id = "";
+
+    private String metacardTag = "";
```
bdeining: Do these need to be initialized with empty strings?

glenhein:
I built and installed the code. I ingested approx 90 items. I then performed an export with:

  catalog:export --delete=true --cql "\"metacard-tags\" like '*'"

I then confirmed that there were no items in ddf. I imported with:

  catalog:import export-2017-02-06T21-47-37.551Z.zip

I checked the catalog and all of my items were present.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/export/ExportItem.java
rzwiefel: With the only constructor forcing them to be set, technically no. But if any are ever added that don't change those, I'd prefer them to be the default string value instead of a null value because of how they are used in the export command.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,496 @@
...
+    @Reference
+    private StorageProvider storageProvider;
+
+    @Option(name = "--output", description = "Output file to export Metacards and contents into. Paths are absolute and must be in quotes. Will default to auto generated name inside of ddf.home", multiValued = false, required = false, aliases = {
+            "-o"})
+    String output = Paths.get(System.getProperty("ddf.home"), name.get())
+            .toString();
+
+    @Option(name = "--delete", required = true, aliases = {"-d",
+            "delete"}, multiValued = false, description = "Whether or not to delete metacards after export")
```
bdeining: Alternative : `Flag to delete Metacards after export`

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
```diff
@@ -94,6 +98,22 @@
 
     private FilterBuilder filterBuilder;
 
+    static {
+        Bundle bundle = FrameworkUtil.getBundle(Historian.class);
+        BundleContext context = bundle == null ? null : bundle.getBundleContext();
+        if (bundle == null || context == null) {
+            LOGGER.error("Could not get bundle to register history metacard types!");
+        } else {
+            context.registerService(MetacardType.class,
```
bdeining: If the History MetacardType lives in catalog-core-versioning, should they be registered there instead?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
```diff
@@ -94,6 +98,22 @@
 
     private FilterBuilder filterBuilder;
 
+    static {
+        Bundle bundle = FrameworkUtil.getBundle(Historian.class);
+        BundleContext context = bundle == null ? null : bundle.getBundleContext();
+        if (bundle == null || context == null) {
+            LOGGER.error("Could not get bundle to register history metacard types!");
```
bdeining: and deleted metacard types

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,212 @@
...
+                    CreateStorageRequestImpl createStorageRequest = new CreateStorageRequestImpl(
+                            Collections.singletonList(contentItem),
+                            id,
+                            new HashMap<>());
+                    storageProvider.create(createStorageRequest);
+                    storageProvider.commit(createStorageRequest);
+                    break;
+                }
+                default: {
+                    LOGGER.debug("Cannot interpret type of " + type);
```
bdeining: `"Cannot interpret type of {}"`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,212 @@
...
+                    CreateStorageRequestImpl createStorageRequest = new CreateStorageRequestImpl(
+                            Collections.singletonList(contentItem),
+                            id,
+                            new HashMap<>());
+                    storageProvider.create(createStorageRequest);
+                    storageProvider.commit(createStorageRequest);
+                    break;
+                }
+                case "derived": {
+                    String qualifier = pathParts[NAME];
```
bdeining: `pathParts[NAME]` is used in all cases and could be moved out of the switch statement

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Reference
+    private StorageProvider storageProvider;
+
+    @Argument(name = "Import File", description = "The file to import", index = 0, multiValued = false, required = true)
+    String importFile;
+
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        ZipValidator zipValidator = initZipValidator();
```
bdeining: No need to init the zipValidator if initImportFile fails; you could switch these two lines

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
rzwiefel: Only problem is its not a bundle, just a jar. So technically if no one pulled them in they wouldn't start. The historian is the 'closest' respective bundle to them so thats why I chose here.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
rzwiefel: I wanted all initialization at the top of the method, then method logic.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
rzwiefel: But has differed meanings depending on which case. Which means i'd have to pick an ambiguous variable name which could obfuscate the code.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
rzwiefel: fixed

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: fixed, and added '..metacards `and content` after export'

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: until we absolutely need the speed, I'd rather not pre-optimize and muddy the code with concurrency.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: fixed

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: fixed

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,496 @@
...
+        }
+
+        SecurityLogger.audit("Called catalog:export command with path : {}", output);
+
+        ZipFile zipFile = new ZipFile(outputFile);
+
+        console.println("Starting metacard export...");
+        Instant start = Instant.now();
+        List<ExportItem> exportedItems = doMetacardExport(zipFile, filter);
+        console.println("Metacards exported in in: " + Duration.between(start, Instant.now()));
```
bdeining: Extra `in`.  Can we make these durations more human readable?  Right now `Metacards exported in in: PT1.267S` may be confusing to an administrator

jlcsmith: What does PT mean?

rzwiefel: Not sure. Java says this is the ISO-8601 compliant format.

rzwiefel: https://en.wikipedia.org/wiki/ISO_8601#Durations

rzwiefel: So the P signifies that a **Period** of Time elapsed. And the T is the separator between year/month/day and hour/minute/second.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,212 @@
...
+                }
+                default: {
+                    LOGGER.debug("Cannot interpret type of {}", type);
+                }
+                }
+
+                entry = zipInputStream.getNextEntry();
+            }
+        }
+        console.println("File imported successfully.");
```
bdeining: The export command logs the number of content items and metacards exported.  Ideally the import command would log the same.

bdeining: Does there need to be `SecurityLogger.audit`ing on the import command too?

rzwiefel: done

rzwiefel: done

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
jlcsmith: You could simply use `Duration.between(start, Instant.now()).getSeconds()` which would be easier to understand for an admin.

glenhein: "in in"? Should that be repeated?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,223 @@
...
+    private StorageProvider storageProvider;
+
+    @Argument(name = "Import File", description = "The file to import", index = 0, multiValued = false, required = true)
+    String importFile;
+
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        int metacards = 0;
+        int content = 0;
+        int derived_content = 0;
```
bdeining: nit: this should be camel case

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: I would disagree. `exported in: 18243s` in my opinion is much less readable than `exported in: PT5H4M3.00S`

rzwiefel: @glenhein https://github.com/codice/ddf/pull/1633#discussion_r99926240

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
rzwiefel: right. Must be dreaming of writing something other than java 😍

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
jlcsmith: I don't think anyone other than a developer is going to understand that format, at least without explanation.  Clearly some of us developers are confused as well.

rzwiefel: I added spaces and lowercased to make it more clear: `6h 50m 15s`

jlcsmith:
@rzwiefel - sorry for the late comment, but I am concerned about exception handling.  There are several cases that will cause the entire export to fail.  For example, a failed transform or failed resource retrieval.  Theoretically, these cases shouldn't occur, but it only takes a single bad transform and the admin will be unable to export.  It may be difficult to recover as well.  I think we need to continue the export on single metacard/resource failure and report the failures at the end.

rzwiefel:
@jlcsmith Which besides this struck you a too restrictive case?
https://github.com/codice/ddf/pull/1633/files#diff-05279c31c48661ae76fd9282fafea8baR305

jlcsmith:
@rzwiefel - Agree with you about the IOException.  Specifically ResourceNotSupportedException and the CatalogTransformerException when adding the metacard to the zip should not abort the entire export.  We'll need to report the failed metacards/resources at the end.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,242 @@
...
+
+    private static final int DERIVED_NAME = 5;
+
+    @Reference
+    private StorageProvider storageProvider;
+
+    @Argument(name = "Import File", description = "The file to import", index = 0, multiValued = false, required = true)
+    String importFile;
+
+    @Option(name = "--UNSAFE-NO-VERIFY", required = false, multiValued = false, description = "Exports the data but does NOT sign the resulting zip file. "
```
pklinef: I would suggest changing this to something like `--skip-signature-verification`.  Also the first sentence of the description needs to be updated for import instead of export.  Also include a `WARNING: ` before the second sentence.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,542 @@
...
+
+    @Option(name = "--delete", required = true, aliases = {"-d",
+            "delete"}, multiValued = false, description = "Flag to delete Metacards and content after export")
+    boolean delete = false;
+
+    @Option(name = "--archived", required = false, aliases = {"-a",
+            "archived"}, multiValued = false, description = "Equivalent to --cql \"\\\"metacard-tags\\\" like 'deleted'\"")
+    boolean archived = false;
+
+    @Option(name = "--UNSAFE-NO-VERIFY", required = false, multiValued = false, description = "Produces the export zip but does NOT sign the resulting zip file. This file will not be able to be verified on import for integrity and security.")
```
pklinef: I would suggest changing this to something like `--skip-signing` which is more explicit about what it is doing.  I think we should also add a `WARNING: ` before the second sentence of the description.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,242 @@
...
+        int derivedContent = 0;
+        ZipValidator zipValidator = initZipValidator();
+        File file = initImportFile(importFile);
+        InputTransformer transformer = getServiceByFilter(InputTransformer.class,
+                String.format("(%s=%s)",
+                        "id",
+                        DEFAULT_TRANSFORMER_ID)).orElseThrow(() -> new CatalogCommandRuntimeException(
+                "Could not get " + DEFAULT_TRANSFORMER_ID + " input transformer"));
+
+        if (unsafe) {
```
pklinef: I think `unsafe` should also require the user to `--force` the command similar to `rm -rf`.

rzwiefel: @pklinef Added --force command

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,256 @@
...
+    private static final int DERIVED_NAME = 5;
+
+    @Reference
+    private StorageProvider storageProvider;
+
+    @Argument(name = "Import File", description = "The file to import", index = 0, multiValued = false, required = true)
+    String importFile;
+
+    @Option(name = "--skip-signature-verification", required = false, multiValued = false, description =
+            "Exports the data but does NOT sign the resulting zip file. "
```
jlcsmith: this line should be removed from the description

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,256 @@
...
+
+            while (entry != null) {
+                String filename = entry.getName();
+
+                if (filename.startsWith("META-INF")) {
+                    entry = zipInputStream.getNextEntry();
+                    continue;
+                }
+
+                String[] pathParts = filename.split("/");
```
jlcsmith: I don't think this will work on windows

bdeining: `File.separator` should be platform specific

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -214,6 +227,29 @@ protected Object executeWithSubject() throws Exception {
         return null;
     }
 
+    private StringBuilder getUserInputModifiable() throws IOException {
+        int in;
+        StringBuilder builder = new StringBuilder();
+        while ((in = session.getKeyboard()
+                .read()) != '\r') {
```
ahoffer: Is `\r` portable? It probably is, but I think `\n` is supposed to represent newline for all OSes. I feel like I read somewhere that `\n` was supposed to change itself depending on the OS hosting the JVM. (I could be wrong).

coyotesqrl: Am I missing some great need to read character by character? Couldn't you just do something like...
```java
        try (BufferedReader reader = new BufferedReader(new InputStreamReader(session.getKeyboard()))) {
            String line = reader.readLine();
            getConsole().println(line);
            return new StringBuilder(line);
        }
```

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,256 @@
...
+        File file = initImportFile(importFile);
+        InputTransformer transformer = getServiceByFilter(InputTransformer.class,
+                String.format("(%s=%s)",
+                        "id",
+                        DEFAULT_TRANSFORMER_ID)).orElseThrow(() -> new CatalogCommandRuntimeException(
+                "Could not get " + DEFAULT_TRANSFORMER_ID + " input transformer"));
+
+        if (unsafe) {
+            SecurityLogger.audit("Skipping validation check of imported data. There are no "
+                    + "guarantees of integrity or authenticity of the imported data");
```
pklinef: I would suggest including the unsigned filename and path in the audit log (might need to escape it somehow to avoid log injection @codice/security ).

stustison: We should be better about this in general. The only way I can think of (off the top of my head anyways) to forge logs when using log4j would be to include CRLFs in the log message to make it look like a new log was written at the right time. Might be worth it to make sure we encode those correctly in the SecurityLogger so that folks don't need to worry about it. Might be worth a ticket, but I wouldn't try to fix it here.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: @ahoffer I will ensure it works on all OSes (though I did test on windows and i believe it worked)
@coyotesqrl Indeed, we must do this if we want to have an "editable" input field. Originally did have it simple like you had above, though if somone backspaces to correct an error for some reason the karaf shell/console does not support this by default making it uneditable. So this implementation allows editing of the input field (basically backspace support)

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
rzwiefel: done

rzwiefel:
Built successfully

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/export/ExportItem.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+package org.codice.ddf.commands.catalog.export;
+
+import java.net.URI;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+public class ExportItem {
```
kcwire: Javadoc please.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,269 @@
...
+import ddf.catalog.content.operation.impl.CreateStorageRequestImpl;
+import ddf.catalog.data.Metacard;
+import ddf.catalog.operation.impl.CreateRequestImpl;
+import ddf.catalog.transform.CatalogTransformerException;
+import ddf.catalog.transform.InputTransformer;
+import ddf.security.common.audit.SecurityLogger;
+
+@Service
+@Command(scope = CatalogCommands.NAMESPACE, name = "import", description = "Imports Metacards and history into the current Catalog")
+public class ImportCommand extends CatalogCommands {
```
kcwire: Javadoc.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,557 @@
...
+import ddf.catalog.source.IngestException;
+import ddf.catalog.transform.CatalogTransformerException;
+import ddf.catalog.transform.MetacardTransformer;
+import ddf.security.common.audit.SecurityLogger;
+import net.lingala.zip4j.core.ZipFile;
+import net.lingala.zip4j.exception.ZipException;
+import net.lingala.zip4j.model.ZipParameters;
+
+/**
+ * Experimental.
```
kcwire: We have a standard javadoc we use - also please provide class level Javadoc.

https://github.com/codice/ddf/blob/08651d5cc59e559b940b4faaec6334657948c18c/catalog/core/catalog-core-api/src/main/java/ddf/catalog/content/operation/CreateStorageRequest.java#L23

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/export/IdAndUriMetacard.java
```diff
@@ -0,0 +1,42 @@
...
+package org.codice.ddf.commands.catalog.export;
+
+import java.net.URI;
+
+import javax.annotation.concurrent.Immutable;
+
+import ddf.catalog.data.impl.MetacardImpl;
+
+@Immutable
+public class IdAndUriMetacard extends MetacardImpl {
```
kcwire: Javadoc.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/QueryResulterable.java
```diff
@@ -0,0 +1,154 @@
...
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
+
+/**
+ * Effectively a cursor over the results of a filter that automatically pages through all results
+ * <br/>
+ * Throws a {@link CatalogCommandRuntimeException} if anything goes wrong during iteration or
+ * querying
+ */
+public class QueryResulterable implements Iterable<Result> {
```
kcwire: how about `QueryResultIterable` to be clear what this thing is.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -0,0 +1,557 @@
...
+
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        Filter filter = getFilter();
+        transformer = getServiceByFilter(MetacardTransformer.class,
+                String.format("(%s=%s)",
+                        "id",
+                        DEFAULT_TRANSFORMER_ID)).orElseThrow(() -> new CatalogCommandRuntimeException(
+                "Could not get " + DEFAULT_TRANSFORMER_ID + " transformer"));
+        revisionFilter = initRevisionFilter();
```
rzwiefel: https://github.com/codice/ddf/pull/1633/files/eabdf3ff8610438d96fd83e29c4184db53acf507#diff-05279c31c48661ae76fd9282fafea8baR528

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
rzwiefel: added

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -0,0 +1,269 @@
...
+            while (entry != null) {
+                String filename = entry.getName();
+
+                if (filename.startsWith("META-INF")) {
+                    entry = zipInputStream.getNextEntry();
+                    continue;
+                }
+
+                String[] pathParts = filename.split("\\" + File.separator);
+                String id = pathParts[ID];
```
rzwiefel: I am not a fan of asserts, but since there isn't a way to guarantee they are importing a valid file i will add it.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/QueryResulterable.java
rzwiefel: done

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/export/IdAndUriMetacard.java
rzwiefel: done

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/export/ExportItem.java
rzwiefel: done

rzwiefel:
test this please

rzwiefel:
test this please

jlcsmith:
@rzwiefel - I'm seeing checkstyle failures after your last commit.

PR #1636, opened 2017-02-03
DDF-2768 Gracefully shutdown http proxy routes

brendan-hofmann:
#### What does this PR do?
Shuts down the http proxy routes gracefully instead of taking five minutes to time out.
@Lambeaux 
@vinamartin 
@shaundmorris
@stustison
Install DDF. Hit the search UI and do a search to bring up the globe. `shutdown -f` then `log:tail` and verify there's no camel barf and the system shuts down smoothly and quickly.
[DDF-2768](https://codice.atlassian.net/browse/DDF-2768)

### DIFF
--- a/libs/httpproxy/proxy-camel-route/src/main/java/org/codice/proxy/http/HttpProxyServiceImpl.java
+++ b/libs/httpproxy/proxy-camel-route/src/main/java/org/codice/proxy/http/HttpProxyServiceImpl.java
@@ -17,6 +17,9 @@ import java.io.File;
 import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.builder.RouteBuilder;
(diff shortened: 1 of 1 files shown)


libs/httpproxy/proxy-camel-route/src/main/java/org/codice/proxy/http/HttpProxyServiceImpl.java
```diff
@@ -184,6 +187,7 @@ public void configure() throws Exception {
         LOGGER.debug("Started proxy route at servlet endpoint: {}, routing to: {}",
                 endpointName,
                 targetUri);
+        endpointIds.add(endpointName);
```
stustison: Shouldn't we also be removing these endpoints from this array when the stop(...) method is called?

shaundmorris: Good question @stustison, I was wondering the same. Also Should we use a Set instead of a List for the endpointIds?

brendan-hofmann: Both correct, had the same thought on my way in. 😸

ahoffer:
Successfully heroed and manually tested this PR.

PR #1637, opened 2017-02-03
DDF-2769 Validation problems not shown in Search UI "Quality" tab

ahoffer:
The validation  errors and warnings do not show on the "Quality" tab of the catalog search UI because the schema and schematron validators do not implement the ReportingMetacardValidator interface.
#### What does this PR do?
Implements the ReportingMetacardValidator in several classes and updates their service definitions.
@brendan-hofmann 
@emanns95 
@AzGoalie 
@mweser 
#### Select at least one member from relevant component team(s) from below (at least one 
@rzwiefel 
@andrewkfiedler
@pklinef
Ingest files with schema or schematron warnings.
Retrieve them as metacards in the catalog search UI.
Navigate to the "Quality" tab on the results page.
The expected result is that the warnings are displayed in the tab.
https://codice.atlassian.net/browse/DDF-2769

### DIFF
--- a/catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
+++ b/catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
@@ -30,7 +30,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.stream.Collectors;
 
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.ErrorListener;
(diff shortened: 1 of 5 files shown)


catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
```diff
@@ -85,16 +94,17 @@
  * @author rodgersh
  * @see <a href="http://www.schematron.com">Schematron</a>
  */
-public class SchematronValidationService implements MetacardValidator, Describable {
+public class SchematronValidationService
+        implements MetacardValidator, Describable, ReportingMetacardValidator {
```
brendan-hofmann: Should we upgrade to the non-deprecated `Describable` while we're here? https://github.com/codice/ddf/blob/0a87e4db67061f90cc9ed49ad1648ed1ff29b755/platform/platform-api/src/main/java/org/codice/ddf/platform/services/common/Describable.java#L22-L22

catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
```diff
@@ -354,6 +364,39 @@ public String getOrganization() {
...
+            }
+            if (e.getErrors() != null) {
+                for (String each : e.getErrors()) {
+                    report.addMetacardViolation(new ValidationViolationImpl(ImmutableSet.of(metacard.getMetadata()),
+                            each,
+                            ValidationViolation.Severity.ERROR));
+                }
+            }
+        } catch (ValidationException e) {
+            LOGGER.warn("Exception validating metacard ID {}\n{}",
```
brendan-hofmann: Would prefer not to swallow the stacktrace here, maybe log it out at debug if it's not being logged elsewhere?

catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
ahoffer: Done.

catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
ahoffer: Added exception to logging statement.

brendan-hofmann:
Should we deprecate the old `MetacardValidator`? Or does keeping both make sense?

catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
```diff
@@ -354,6 +364,40 @@ public String getOrganization() {
         return null;
     }
 
+    @Override
+    public Optional<MetacardValidationReport> validateMetacard(Metacard metacard) {
+
+        MetacardValidationReportImpl report = null;
+        try {
+            validate(metacard);
+        } catch (SchematronValidationException e) {
```
pklinef: Instead of controlling flow using exceptions, I would suggest splitting `validate` into two methods.  One would have the current signature but the other would extract most of the code and return a `SchematronReport`.  This `validateMetacard` method would then just use the `SchematronReport` instead of parsing the thrown `SchematronValidationException`.

catalog/schematron/catalog-schematron-plugin/src/main/java/ddf/services/schematron/SchematronValidationService.java
```diff
@@ -254,39 +274,74 @@ public void setPriority(int priority) {
...
+                        warnings.add(violation.getMessage());
+                    }
+                });
+
+        SchematronValidationException exception = new SchematronValidationException(
+                "Schematron validation failed",
+                errors,
+                warnings);
+
+        if (!errors.isEmpty()) {
```
pklinef: Instead of these two `if` blocks, can you still just do a `if (!report.isValid(suppressWarnings)) {`?

ahoffer: The report was switched from the "old class", `SchemaReport` to the "enhanced class", `MetacardValidationReport`. Unfortunately, MetacardValidationReport`does not implement the `isValid`.

ahoffer:
Heroed by @brendan-hofmann

brendan-hofmann:
Successfully heroed.

ahoffer:
@brendan-hofmann heroed.

PR #1641, opened 2017-02-04
DDF-2774: client.bat script does not give access to the DDF console

lessarderic:
#### What does this PR do?
Updates the documentation to show that the `-h localhost` option must be used on Windows when running `client.bat`.
Also addressed a few other minor issues in the installation documentation related to Windows.

@Joy603 
@brjeter 
@ricklarsen 

[Docs](https://github.com/orgs/codice/teams/docs)

@figliold
@kcwire

Build the documentation and make sure information is accurate.

[DDF-2774](https://codice.atlassian.net/browse/DDF-2774)

#### Checklist:
- [x] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distribution/docs/src/main/resources/_contents/_running/starting-intro-contents.adoc
+++ b/distribution/docs/src/main/resources/_contents/_running/starting-intro-contents.adoc
@@ -31,7 +31,7 @@ Alternatively, to run ${branding} as a background service, run the `start` scrip
 
 [NOTE]
 ====
-If console access is needed while running as a service, run the `client` script:
+If console access is needed while running as a service, run the `client` script on the host where the ${branding} is running:
 
 .*NIX
 ----
@@ -42,6 +42,8 @@ If console access is needed while running as a service, run the `client` script:
 ----
 <${branding}_HOME>/bin/client.bat -h localhost
 ----
+
+If running the `client` script form a different host, use the `-h` option followed by the name or IP of the host where ${branding} is running.
 ====
 
 ==== Stopping ${branding}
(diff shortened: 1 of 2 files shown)


distribution/docs/src/main/resources/_contents/_running/starting-intro-contents.adoc
```diff
@@ -35,12 +35,12 @@ If console access is needed while running as a service, run the `client` script:
...
 .*NIX
 ----
-<${branding}_HOME/bin/client
+<${branding}_HOME>/bin/client
 ----
 
 .Windows
 ----
-<${branding}_HOME>/bin/client.bat
+<${branding}_HOME>/bin/client.bat -h localhost
```
shaundmorris: Should we add a note that this is an example? Basically we are specifying the host to connect to.

lessarderic: Good question. The reason I had to put `-h localhost` on Windows is because there's a bug in the Karaf's batch file and the script doesn't work with no arguments. If I add a note, I'd probably have to also explain how to use the argument on both platforms and mention that it is required on Windows or else it won't work. Just afraid that it'll make things more messy than really needed here. Thoughts?

shaundmorris: It should be fine, since that should work by default. I was wondering if they were on a separate machine if they would have to put in -h `hostname`. If so, we could just put a note to specify the hostname of the host running DDF. For example...

lessarderic: I like that. I'll add that tomorrow.

lessarderic: Addressed the comment @shaundmorris. Let me know what you think.

figliold:
Built docs and verified that they look good. 👍

PR #1642, opened 2017-02-05
DDF-2775 Update to ensure actions displayed in the catalog/search ui are appropriate for the metacards tag.

clockard:
#### What does this PR do?
Adds service property to the RegistryTransformer service so it doesn't get displayed as an action for normal metacards. Updated the AbstractMetacardActionProvider so it will only return true for canHandle if the metacard is a resource metacard. Non resource metacards would fail if these actions were attempted.
@mcalcote @gordocanchola 
@figliold
@pklinef
Install DDF, include the Registry in the  install.
In the Admin UI Standard Search UI->Catalog UI Search remove 'metacard-tags' from the Hidden Fields. 
Go to the Catalog UI and upload a file. 
Perform a search and go to the actions tab of that result and verify that Export as rim:RegistryPackage is not present.
Update your search by going to the advanced tab and adding another clause of metacard-tags contains registry.
Rerun the search. You should only get back one registry metacard. Verify that it only has one action and it works.
[DDF-2775](https://codice.atlassian.net/browse/)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-actions/src/main/java/org/codice/ddf/catalog/actions/AbstractMetacardActionProvider.java
+++ b/catalog/core/catalog-core-actions/src/main/java/org/codice/ddf/catalog/actions/AbstractMetacardActionProvider.java
@@ -111,7 +111,8 @@ public abstract class AbstractMetacardActionProvider implements ActionProvider {
      * {@inheritDoc}
      */
     public <T> boolean canHandle(T subject) {
-        return (subject instanceof Metacard) && (canHandleMetacard((Metacard) subject));
+        return (subject instanceof Metacard) && isResourceMetacard((Metacard) subject)
+                && canHandleMetacard((Metacard) subject);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


catalog/core/catalog-core-actions/src/main/java/org/codice/ddf/catalog/actions/AbstractMetacardActionProvider.java
```diff
@@ -111,7 +111,8 @@ public String getId() {
      * {@inheritDoc}
      */
     public <T> boolean canHandle(T subject) {
-        return (subject instanceof Metacard) && (canHandleMetacard((Metacard) subject));
+        return (subject instanceof Metacard) && ((Metacard) subject).getTags()
```
gordocanchola: Should the Metacard.getTags() check be moved to canHandleMetacard?

clockard: I put it here because all of the action providers that extend this can't handle anything but resource tagged metacards. Ideally each of the handlers would do a better job of checking if they can handle a metacard and this wouldn't be needed. There is another ticket out there to look at improvements to our action providers that could incorporate this kind of change.

brianfelix:
PR has passed manual testing. :+1:

clockard:
Hero build passed.

PR #1534, opened 2017-01-06
DDF-2687 Update vendored DrawHelper to not remove overlapping points

andrewkfiedler:
- See leforthomas/cesium-drawhelper#13 for details.
 - At lower zooms, points were being erroneously removed from drawings.

#### What does this PR do?
  Updates the vendored version of DrawHelper to pull in a fix for lower zoom drawings.

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@jlcsmith 
@kcwire 

#### Select at least one member from relevant component team(s) from below (at least one component team member needs to approve the PR).
[UI](https://github.com/orgs/codice/teams/ui)

#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@jlcsmith
@kcwire

#### How should this be tested? (List steps with links to updated documentation)
Zoom to street level somewhere and draw a line or a polygon. 

#### Any background context you want to provide?
This applies the fix to both the standard and the catalog ui.

#### What are the relevant tickets?
https://codice.atlassian.net/browse/DDF-2687

#### Screenshots (if appropriate)
Before: https://gfycat.com/SereneSelfassuredBoto
After: https://gfycat.com/ElectricLittleHummingbird

PR #1647, opened 2017-02-06
DDF-2779 Remove extra points from line / polygon drawings in Cesium

andrewkfiedler:
#### What does this PR do?
 - https://github.com/codice/ddf/pull/1534 made a change to DrawHelper which allowed drawings at higher zooms to work.  However, DrawHelper was relying on this to remove the extra points that occur when finishing a drawing.  As a result, we need to remove the final two points on all polygon / line drawings.

@djblue 
@garrettfreibott 
@andrewkfiedler
@pklinef 

Create line and polygon drawings in Cesium and verify that they contain only the number of points you that were drawn.  

https://codice.atlassian.net/browse/DDF-2779


codymacdonald:
Manually verified 👍

PR #1649, opened 2017-02-07
DDF-2781 Update app to start only after initial user fetch

andrewkfiedler:
#### What does this PR do?
 - Starting before then could cause issues with routing to notifications, or inconsistencies in the display username.
 - We were already waiting for the initial workspace fetch, so this adds on to that.  Now we wait for both the initial workspace and the initial user fetch before starting.

@djblue 
@rzwiefel 
@pklinef 
@andrewkfiedler

Cause a notification (such as by uploading).  Make sure your settings have persistent notifications. Go to it. Refresh. Test with network latency to see the improvement from the old way.

Another way is to add setTimeouts into the code around the initial fetches to see the app wait for them to complete.

https://codice.atlassian.net/browse/DDF-2781

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationStart.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationStart.js
@@ -16,22 +16,31 @@ require([
     'application',
     'properties',
     'store',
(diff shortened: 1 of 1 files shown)


codymacdonald:
Manually verified 👍

PR #1651, opened 2017-02-07
DDF-2759: Feature configuration files are not always being processed

lessarderic:
Added code to notify listener of existing configuration files.

#### What does this PR do?
Makes sure existing configuration files in the watched directory are processed.

@figliold 

@coyotesqrl
@stustison

Run all integration tests.

This is a follow on to https://github.com/codice/ddf/pull/1616.

[DDF-2759](https://codice.atlassian.net/browse/DDF-2759)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
@@ -22,12 +22,15 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.apache.commons.lang.Validate.notNull;
 
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.WatchEvent;
 import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -94,6 +97,7 @@ public void init() throws IOException {
     public void register(@NotNull ChangeListener listener) {
         notNull(listener, "ChangeListener cannot be null");
         changeListener = listener;
+        processExistingConfigurationFiles();
```
pklinef: How do we know the watch service is ready and receiving events at this point?

lessarderic: Yeah, good point. It's unlikely but not impossible. We could always add a wait here (countdown latch or something) to wait until the watch thread has started.

lessarderic: The plan is to replace this with a better file poller that takes care of any existing files and any new files that get added afterward without having to worry about potential timing or synchronization problems.

Having said that, would this fix be acceptable for now?

pklinef: This is better than it is currently for 2.10.  We should mark the fix for 2.10.1.

PR #1653, opened 2017-02-07
DDF-2612 Fix caching in CachingFederationStrategy.

clockard:
#### What does this PR do?
 Fix caching in CachingFederationStrategy for real this time without breaking enterprise query.
@brendan-hofmann 
@pklinef
@shaundmorris

Verify itests pass for ddf and upstream projects
[DDF-2612](https://codice.atlassian.net/browse/)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheQueryFactory.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheQueryFactory.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.cache.solr.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.lang.StringUtils;
+import org.opengis.filter.Filter;
+
+import ddf.catalog.filter.FilterBuilder;
+import ddf.catalog.operation.QueryRequest;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.source.solr.SchemaFields;
+
+public class CacheQueryFactory {
+
+    private FilterBuilder builder;
+
+    public CacheQueryFactory(FilterBuilder filterBuilder) {
+        builder = filterBuilder;
+    }
+
+    QueryRequest getQueryRequestWithSourcesFilter(QueryRequest input) {
+        QueryRequest queryWithSources = input;
+        if (input.getSourceIds() != null) {
+            List<Filter> sourceFilters = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheQueryFactory.java
```diff
@@ -0,0 +1,62 @@
...
+        if (input.getSourceIds() != null) {
+            for (String sourceId : input.getSourceIds()) {
+                sourceFilters.add(builder.attribute(StringUtils.removeEnd(SolrCache.METACARD_SOURCE_NAME,
+                        SchemaFields.TEXT_SUFFIX))
+                        .is()
+                        .equalTo()
+                        .text(sourceId));
+            }
+            QueryImpl sourceQuery = new QueryImpl(builder.allOf(input.getQuery(),
+                    builder.anyOf(sourceFilters)));
```
pklinef: The Catalog UI already has a temporary workaround to query by source.  Does this now duplicate the sources part of the query when using the Catalog UI?  If it does then we should consider removing the workaround on the frontend of the Catalog UI.

clockard: In the case of the catalog ui, yes the query will have duplicate source parameters but it will still work. I think the need for the workaround in the catalog ui still exists since the way the cache query is formed is the same as it was before, just its location has changed.

pklinef: https://github.com/codice/ddf/blob/752a1b0533d0e56a76f59cbb1c005e114d58b7d2/catalog/ui/catalog-ui-search/src/main/webapp/js/CacheSourceSelector.js

https://github.com/codice/ddf/blob/380ca0ddd533f37beb37ea12e75071f369f94146/catalog/ui/catalog-ui-search/src/main/webapp/js/model/Query.js#L388-L392

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -181,16 +181,6 @@
 
     private static final int NO_RETRIES = 0;
 
-    private static final int FAIL_RETRIES = 3;
```
brendan-hofmann: Guessing these fields were all unused?

clockard: Yes they were not being used so I removed them

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheQueryFactory.java
clockard: The catalog ui frontend change will be address in https://codice.atlassian.net/browse/DDF-2800

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheQueryFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private FilterBuilder builder;
+
+    public CacheQueryFactory(FilterBuilder filterBuilder) {
+        builder = filterBuilder;
+    }
+
+    QueryRequest getQueryRequestWithSourcesFilter(QueryRequest input) {
+        QueryRequest queryWithSources = input;
+        if (input.getSourceIds() != null) {
```
clockard: In general a query requests source ids can be null but probably not in the case of querying the cache. In the oddball case that it is null then queryWithSources variable will not be set in the if statement which is why I initialize it to the input and only have one return. Let me know if you want me to change this.

pklinef:
Is this being backported to 2.10.x?

clockard:
@pklinef yes it will be packported.

brianfelix:
Hero build passed :+1:

PR #1661, opened 2017-02-08
DDF-2785 Prevents duplicate points from being added onto the ends of polygons

andrewkfiedler:
#### What does this PR do?
 - Polygons are meant to be a closed shape, so we were adding on the beginning point to the end.  However, we didn't check to see if we had already done so in the past.

@emanns95 
@codymacdonald
@djblue 
@figliold 
@andrewkfiedler

Draw a polygon.  Save it.  Check how many points.  Edit, save, edit, save.  Check how many points.  Number of points should not change.  

Add an end or beginning point to an existing polygon.  Save.  Check to make sure we closed the polygon off.  

https://codice.atlassian.net/browse/DDF-2785

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/CQLUtils.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/CQLUtils.js
@@ -98,7 +98,9 @@ define([
             var cqlPolygon = model.map(function (point) {
                 return point[0] + ' ' + point[1];
             });
-            cqlPolygon.push(cqlPolygon[0]);
+            if (cqlPolygon[0] !== cqlPolygon[cqlPolygon.length - 1]) {
+                cqlPolygon.push(cqlPolygon[0]);
+            }
             return cqlPolygon;
         },
         lineToCQLLIne: function(model){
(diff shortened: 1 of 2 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/js/CQLUtils.js
```diff
@@ -98,7 +98,9 @@ define([
             var cqlPolygon = model.map(function (point) {
                 return point[0] + ' ' + point[1];
             });
-            cqlPolygon.push(cqlPolygon[0]);
+            if (cqlPolygon[0] !== cqlPolygon[cqlPolygon.length - 1]) {
+                cqlPolygon.push(cqlPolygon[0]);
```
emanns95: Does this remove the first point if it doesn't match the last point? I feel like the extra points were being generated at the end and we should be trimming duplicates off the end? Correct me if I'm wrong 😄

andrewkfiedler: No, it adds a duplicate of the first point on to the end if the first point does not match the last point.  This ensures the polygon is closed.

codymacdonald:
Manually tested & verified 👍

PR #1665, opened 2017-02-09
DDF-2788, DDF-2789

stustison:
@brendan-hofmann 
@coyotesqrl
@pklinef
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/CachedSource.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/CachedSource.java
@@ -20,8 +20,10 @@ package ddf.catalog.util.impl;
  */
 
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import org.slf4j.Logger;
@@ -196,5 +198,48 @@ public class CachedSource implements Source {
 
     private void clearContentTypes() {
         this.cachedContentTypes = Collections.<ContentType>emptySet();
+    }
+
+    public boolean equals(Object cachedSource) {
+        if (cachedSource == null) {
+            return false;
+        }
+        if (!(cachedSource instanceof CachedSource)) {
+            return false;
+        }
+        CachedSource tmpSource = (CachedSource) cachedSource;
+        Comparator<CachedSource> sourceKeyComparator = Comparator.comparing(
+                CachedSource::getId,
+                Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(CachedSource::getTitle,
+                        Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(CachedSource::getVersion,
+                        Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(CachedSource::getDescription,
+                        Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(CachedSource::getOrganization,
+                        Comparator.nullsLast(Comparator.naturalOrder()));
+        return sourceKeyComparator.compare(this, tmpSource) == 0;
+    }
+
+    public int hashCode() {
+        final int number = 31;
+        int code = 1;
+        code = code * number + Optional.ofNullable(getId())
+                .orElse("0")
+                .hashCode();
+        code = code * number + Optional.ofNullable(getTitle())
+                .orElse("0")
+                .hashCode();
+        code = code * number + Optional.ofNullable(getVersion())
+                .orElse("0")
+                .hashCode();
+        code = code * number + Optional.ofNullable(getDescription())
+                .orElse("0")
+                .hashCode();
+        code = code * number + Optional.ofNullable(getOrganization())
+                .orElse("0")
+                .hashCode();
+        return code;
     }
 }
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
@@ -33,11 +33,10 @@ import ddf.catalog.source.Source;
  * The poller to check the availability of all configured sources. This class is instantiated by the
  * CatalogFramework's blueprint and is scheduled by the {@link SourcePoller} to execute at a
  * configured rate, i.e., the polling interval.
- *
+ * <p>
  * This class maintains a list of all of the sources to be polled for their availability. Sources
  * are added to this list when they come online and when they are deleted. A cached map is
  * maintained of all the sources and their last availability states.
- *
  */
 public class SourcePollerRunner implements Runnable {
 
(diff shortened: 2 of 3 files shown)


platform/security/pep/security-pep-interceptor/src/main/java/ddf/security/pep/interceptor/PEPAuthorizingInterceptor.java
```diff
@@ -194,10 +194,13 @@ private String getActionUri(Message message) {
         if (StringUtils.isEmpty(actionURI)) {
             BindingOperationInfo bindingOpInfo = message.getExchange()
                     .get(BindingOperationInfo.class);
-            SoapOperationInfo soi = bindingOpInfo.getExtensor(SoapOperationInfo.class);
-            if (soi == null && bindingOpInfo.isUnwrapped()) {
-                soi = bindingOpInfo.getWrappedOperation()
-                        .getExtensor(SoapOperationInfo.class);
+            SoapOperationInfo soi = null;
+            if (bindingOpInfo != null) {
```
coyotesqrl: Is it worth DEBUG logging when this is null?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -96,41 +94,36 @@ public void run() {
...
-
-                    if (sourceStatusThreadLock.tryLock()) {
-                        LOGGER.debug("Acquired lock for Source [{}] with id [{}]",
+        final Runnable statusRunner = () -> {
+            final CachedSource cachedSource = cachedSources.get(getSourceKey(source));
+            if (cachedSource != null) {
+                Lock sourceStatusThreadLock = sourceStatusThreadLocks.get(cachedSource);
+
+                if (sourceStatusThreadLock.tryLock()) {
+                    LOGGER.debug("Acquired lock for Source [{}] with id [{}]",
```
coyotesqrl: This debug statement could still throw an exception if the underlying source is gone. Should probably use `getSourceKey(source).getId()`. Might also be best to move it inside the `try` block so it can't prevent the lock from being released.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -96,41 +94,36 @@ public void run() {
...
+                            source,
+                            source.getId());
+
+                    try {
+                        cachedSource.checkStatus();
+                    } finally {
+                        // release the lock acquired initially
+                        sourceStatusThreadLock.unlock();
+                        LOGGER.debug("Released lock for Source [{}] with id [{}]",
```
coyotesqrl: This debug statement could likewise throw an exception.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -96,41 +94,36 @@ public void run() {
...
-                            sourceStatusThreadLock.unlock();
-                            LOGGER.debug("Released lock for Source [{}] with id [{}]",
-                                    source,
-                                    source.getId());
-                        }
-                    } else {
-                        LOGGER.debug("Unable to get lock for Source [{}] with id [{}]."
-                                + "  A status thread is already running.", source, source.getId());
                     }
+                } else {
```
coyotesqrl: This one as well.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -33,34 +33,32 @@
...
+                            Comparator.nullsLast(Comparator.naturalOrder()))
+                    .thenComparing(SourceKey::getDescription,
+                            Comparator.nullsLast(Comparator.naturalOrder()))
+                    .thenComparing(SourceKey::getOrganization,
+                            Comparator.nullsLast(Comparator.naturalOrder())));
 
     private ExecutorService pool;
 
-    private Map<Source, Lock> sourceStatusThreadLocks = new ConcurrentHashMap<Source, Lock>();
+    private Map<CachedSource, Lock> sourceStatusThreadLocks = new ConcurrentHashMap<>();
```
brendan-hofmann: Why use `CachedSource` instead of `SourceKey` for this map here?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -198,4 +186,76 @@ public void shutdown() {
         }
         LOGGER.trace("Status threads shut down");
     }
+
+    private SourceKey getSourceKey(Source source) {
+        return new SourceKey(source);
+    }
+
+    private static class SourceKey {
```
brendan-hofmann: You would also want to generate `equals()` and `comparable()` for this class.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
stustison: Because the lock corresponds to the cached source object, not the source.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
coyotesqrl: Agree we should have `equals()` and `hashcode()`. Can get by without the `Comparable` interface because the `Comparator` on the map using it is explicitly defined.

stustison: It needs equals and hashcode. I will also add it to the CachedSource object because it is missing them.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
stustison: Fixed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
stustison: Fixed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
stustison: Fixed.

platform/security/pep/security-pep-interceptor/src/main/java/ddf/security/pep/interceptor/PEPAuthorizingInterceptor.java
stustison: We will already generate an audit for it in this case further down the line.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/CachedSource.java
```diff
@@ -197,4 +202,30 @@ public String getAttributeValue(String attr) {
...
+                        Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(CachedSource::getDescription,
+                        Comparator.nullsLast(Comparator.naturalOrder()))
+                .thenComparing(CachedSource::getOrganization,
+                        Comparator.nullsLast(Comparator.naturalOrder()));
+        return sourceKeyComparator.compare(this, tmpSource) == 0;
+    }
+
+    public int hashCode() {
+        return uniqueId.hashCode();
```
coyotesqrl: I believe this and your equals() method could give you different results. If you're going to use the UUID for its hashcode, your equals method should test that for equality.

pklinef: This will violate the hash code contract. 

http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()

PR #1664, opened 2017-02-09
DDF-2787- Updates how thumbnail action url is generated

vinamartin:
#### What does this PR do?
Old code assumed that there would always have a preceding parameter before adding the timestamp, this PR updates to remove that assumption.
 
#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@djblue @garrettfreibott 

#### Select at least one member from relevant component team(s) from below (at least one component team member needs to approve the PR).
[UI](https://github.com/orgs/codice/teams/ui)

#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@andrewkfiedler 
@pklinef 

#### How should this be tested? (List steps with links to updated documentation)
Build catalog-ui-search

#### What are the relevant tickets?
[](https://codice.atlassian.net/browse/DDF-2787)

PR #1666, opened 2017-02-09
DDF-2787- Updates how thumbnail action url is generated

vinamartin:
Backport of https://github.com/codice/ddf/pull/1664

@andrewkfiedler @pklinef


catalog/ui/catalog-ui-search/src/main/webapp/js/model/Metacard.js
```diff
@@ -32,6 +32,17 @@ define([
 
         var blacklist = [];
 
+        function generateThumbnailUrl(url){
+            if(url.indexOf("?") >= 0) {
```
andrewkfiedler: Can make a local variable to use instead of modifying the passed in url?

PR #1674, opened 2017-02-10
DDF-2792 Update client side sorting to handle null / blank values more consistently

andrewkfiedler:
#### What does this PR do?
 - Updates sorting to always push null or undefined values to the end of a sort (regardless of ascending or descending).
 - Blanks are handled as they should by the particular datatype.  Hence, for a String type ascending puts them at the top, and descending at the bottom.
 - Updated string comparisons to be case insensitive to better fit user expectations.
 - Fixed issues where certain attributes were sortable from the normal sort view, but not from the table view.

@andrewkfiedler
@jlcsmith 
@bdeining 
@rzwiefel 

Ingest some varied data.  Start testing various sorts.  Confirm that attributes with null or undefined values are always pushed to the end of a sort (regardless of ascending or descending).  

https://codice.atlassian.net/browse/DDF-2792


jlcsmith:
Built and verified sorting of null and empty values.  Verified case insensitive String sorting works as expected.  Verified table sorting matches list view sorting.

PR #1678, opened 2017-02-13
DDF-2794 Adjusted RegistryPolicyPlugin to set user defined security attributes on the item policy

mcalcote:
#### What does this PR do?
This change adds additional user defined security attributes from the Security Attributes field on the node information modal to the item policy that is generated in the RegistryPolicyPlugin. 

@clockard @emmberk @vinamartin @brianfelix @gordocanchola 

[Security](https://github.com/orgs/codice/teams/security) @adimka 

@coyotesqrl
@stustison

Set up a DDF Registry with the Security Attributes set on the General Information tab of the local node, then try to connect to it with a second DDF without the same security attributes. The registry node should not get pulled (though the connection should return a response).

Further considerations will have to be made for any other elements of a local node that can have Security Attributes assigned. 

[DDF-2794](https://codice.atlassian.net/browse/DDF-2794)

screen shot 2017-02-13 at 1 35 09 pm

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
+++ b/catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
@@ -20,11 +20,13 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.codice.ddf.configuration.SystemBaseUrl;
 import org.codice.ddf.registry.common.metacard.RegistryObjectMetacardType;
 import org.codice.ddf.registry.common.metacard.RegistryUtility;
 
+import ddf.catalog.data.Attribute;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
 import ddf.catalog.operation.Query;
@@ -35,6 +37,7 @@ import ddf.catalog.plugin.PolicyResponse;
 import ddf.catalog.plugin.StopProcessingException;
 import ddf.catalog.plugin.impl.PolicyResponseImpl;
 import ddf.catalog.util.impl.Requests;
+import ddf.security.permission.Permissions;
 
 public class RegistryPolicyPlugin implements PolicyPlugin {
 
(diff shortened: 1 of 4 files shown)


catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
```diff
@@ -146,10 +148,30 @@ private void parsePermissionsFromString(List<String> permStrings,
...
+        Map<String, Set<String>> operationPolicy = new HashMap<>();
+        Map<String, Set<String>> itemPolicy = new HashMap<>();
+        Map<String, Set<String>> securityAttributes = new HashMap<>();
+
         if (Requests.isLocal(properties) && input != null && (
                 RegistryUtility.isInternalRegistryMetacard(input)
                         || RegistryUtility.isRegistryMetacard(input))) {
+            Attribute securityLevel = input.getAttribute(RegistryObjectMetacardType.SECURITY_LEVEL);
+            if (securityLevel != null) {
+                securityLevel.getValues()
```
clockard: I think you could use the Permissions.parsePermissionsFromString helper to do this for you.

catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
```diff
@@ -160,7 +182,9 @@ private PolicyResponse getWritePolicy(Metacard input, Map<String, Serializable>
                 operationPolicy.putAll(policy);
             }
         }
-        return new PolicyResponseImpl(operationPolicy, operationPolicy);
+        itemPolicy.putAll(operationPolicy);
```
clockard: Why are we adding the operation policy to the item policy?

stustison: This line seems very strange to me.

mcalcote: Previously the operation policy was used for both, so I thought we would keep those policy settings as well as the ones we define ourselves.

stustison: I think it is redundant and was probably redundant before as well. If the user or the system cannot imply the operation policy, then the entire operation is dropped on the floor. Adding that same policy to the items would have no effect because you wouldn't even get to the point where you were considering them individually.

catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
mcalcote: Updated

catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
mcalcote: 👍  Will remove

catalog/spatial/registry/registry-policy-plugin/src/main/java/org/codice/ddf/registry/policy/RegistryPolicyPlugin.java
```diff
@@ -160,7 +175,13 @@ private PolicyResponse getWritePolicy(Metacard input, Map<String, Serializable>
                 operationPolicy.putAll(policy);
             }
         }
-        return new PolicyResponseImpl(operationPolicy, operationPolicy);
+        if (securityAttributes.isEmpty()) {
+            return new PolicyResponseImpl(operationPolicy, operationPolicy);
+        } else {
+            Map<String, Set<String>> itemPolicy = new HashMap<>();
+            itemPolicy.putAll(securityAttributes);
```
stustison: It looks like we create the securityAttributes map for every request anyways. Is there a reason why we dump it into a new map here? Couldn't we just pass securityAttributes as the item policy?

mcalcote: No, that is totally unnecessary with the latest changes. Good call, fixed.

clockard:
Hero build passed

PR #1679, opened 2017-02-13
DDF-2794 -> 2.10.x Adjusted RegistryPolicyPlugin to set user defined security attributes on the item policy

mcalcote:
Port of https://github.com/codice/ddf/pull/1678

#### What does this PR do?
This change adds additional user defined security attributes from the Security Attributes field on the node information modal to the item policy that is generated in the RegistryPolicyPlugin. 

Permissions.java was also updated to trim whitespace from security attributes.

@clockard @emmberk @vinamartin @brianfelix @gordocanchola 

[Security](https://github.com/orgs/codice/teams/security) @adimka 

@coyotesqrl
@stustison

Set up a DDF Registry with the Security Attributes set on the General Information tab of the local node, then try to connect to it with a second DDF without the same security attributes. The registry node should not get pulled (though the connection should return a response).

Further considerations will have to be made for any other elements of a local node that can have Security Attributes assigned. 

[DDF-2794](https://codice.atlassian.net/browse/DDF-2794)

screen shot 2017-02-13 at 1 35 09 pm

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/registry/registry-policy-plugin/src/test/java/org/codice/ddf/registry/policy/RegistryPolicyPluginTest.java
+++ b/catalog/spatial/registry/registry-policy-plugin/src/test/java/org/codice/ddf/registry/policy/RegistryPolicyPluginTest.java
@@ -18,8 +18,13 @@ import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.Assert.assertThat;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import org.codice.ddf.registry.common.RegistryConstants;
 import org.codice.ddf.registry.common.metacard.RegistryObjectMetacardType;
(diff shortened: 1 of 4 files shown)


platform/security/platform-security-core-api/src/main/java/ddf/security/permission/Permissions.java
```diff
@@ -47,10 +47,14 @@ private Permissions() {
                     String[] parts = perm.split("=");
                     if (parts.length == 2) {
                         String attributeName = parts[0];
-                        String attributeValue = parts[1];
-                        String[] attributeValues = attributeValue.split(",");
-                        permissions.put(attributeName,
-                                new HashSet<>(Arrays.asList(attributeValues)));
+                        Scanner s = new Scanner(parts[1]).useDelimiter(",");
```
coyotesqrl: ❗️  This works but it's less efficient than using a compiled `Pattern`.

I just ran a comparison of the `Scanner` code vs. the following code over 10K runs and the `Scanner` takes about three times longer. It's even more pronounced - about four-to-five times slower - when I make the `Pattern` a static final.
```java
        Pattern pattern = Pattern.compile(",");

        return pattern.splitAsStream(input)
                .map(String::trim)
                .collect(Collectors.toSet());
```

mcalcote: I was conflicted about adding it to be honest. Will update.

catalog/spatial/registry/registry-policy-plugin/src/test/java/org/codice/ddf/registry/policy/RegistryPolicyPluginTest.java
```diff
@@ -206,6 +211,39 @@ public void testNoRegistryBypassPermissions() throws Exception {
     }
 
     @Test
+    public void testSecurityValueSet() throws Exception {
```
mcalcote: testCudRegistryOperations calls into the method and does not generate a security result. However I did rely on the front end input verification as well as an attribute groomer to prevent an empty string from being passed in. I will add a test to make sure this case doesn't result in any NPEs should those elements change.

mcalcote:
Successfully Hero'd

PR #1672, opened 2017-02-09
DDF-2791 Added support for queries with empty values in SolrFilterDelegate

bdeining:
#### What does this PR do?
This PR adds support to match empty strings (search for metacards with an empty attribute)
#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@glenhein @jrnorth 
#### Select at least one member from relevant component team(s) from below (at least one component team member needs to approve the PR).
[Solr](https://github.com/orgs/codice/teams/solr)
@rzwiefel 
#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).
@jlcsmith
@pklinef
#### How should this be tested? (List steps with links to updated documentation)
Build / Install / Ingest Data / Verify queries
Note : There is no frontend support; so the query may have to be done through cURL or Postman
#### What are the relevant tickets?
[DDF-2791](https://codice.atlassian.net/browse/DDF-2791)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #1683, opened 2017-02-14
DDF-2799 Add support for empty string queries to Catalog UI

andrewkfiedler:
#### What does this PR do?
 - https://codice.atlassian.net/browse/DDF-2791 (https://github.com/codice/ddf/pull/1672) adds support for empty string queries on the server, so this adds support on the client.
 - Removes guards against empty string in filter.view and filter-builder.view.
 - This only adds support for empty string fields.
 - Updates how client side filtering is done on empty fields to match how the server side querying currently works.  As of now, searching for an empty string on a particular field will return all the null and blank valued results for that field.  On a side note, searching for wildcard (*) filters out all null and blank valued results (both client and server side).
 - Removes some unnecessary cruft from Metacard.  There's no need to listen to the unfiltered collection anymore (it already gets recomputed each time the original changes, partially due to side effects from the paginator), so this is wasted computation.

@andrewkfiedler
@bdeining 
@rzwiefel 
@jlcsmith
@pklinef 

Go to the Catalog UI.  Go to the advanced page and attempt to query for blanks on any string field.  

Note that I did not add support to the basic page, since that is anyText and it's assumed you typically want a wildcard search if you leave it blank.  

https://codice.atlassian.net/browse/DDF-2799

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/filter-builder/filter-builder.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/filter-builder/filter-builder.view.js
@@ -135,7 +135,7 @@ define([
                     filters: this.filterContents.currentView.children.map(function(childView){
                         return childView.getFilters();
                     }).filter(function(filter){
-                       return filter && filter.value !== "";
+                       return filter;
                     })
                 }
             }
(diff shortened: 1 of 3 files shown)


bdeining:
Hero Build Successful.  Verified that empty string queries return the proper results from the backend. 👍

bdeining:
depends on https://github.com/codice/ddf/pull/1672

PR #1688, opened 2017-02-14
DDF-2795 Fix attribute overrides in directory monitor causing an error if the value has a comma

emanns95:
#### What does this PR do?
Pass in map of attribute overrides to routeDefinition instead of comma-separated string to ensure attributes with commas do not cause an error

@brendan-hofmann 
@ryeats 

@coyotesqrl
@millerw8 

* Monitor a directory
* Set an attribute override of `location=Polygon((10 10, 10 20, 20 20, 20 10, 10 10 ))`
* Check that the override works for files placed in the directory (`catalog:inspect metacard`)

[DDF-2795](https://codice.atlassian.net/browse/DDF-2795)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
+++ b/catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
@@ -16,22 +16,19 @@ package ddf.camel.component.catalog.content;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.Serializable;
 import java.nio.file.Path;
 import java.nio.file.StandardWatchEventKinds;
 import java.nio.file.WatchEvent;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.apache.camel.Message;
 import org.apache.camel.component.file.GenericFile;
 import org.apache.camel.component.file.GenericFileMessage;
 import org.apache.commons.io.FilenameUtils;
-import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -192,25 +189,16 @@ public class ContentProducerDataAccessObject {
 
     public void processHeaders(Map<String, Object> headers, StorageRequest storageRequest,
             File ingestedFile) {
-        String attributeOverrideHeaders = (String) headers.get(Constants.ATTRIBUTE_OVERRIDES_KEY);
-        if (StringUtils.isNotEmpty(attributeOverrideHeaders)) {
+        Map<String, Serializable> attributeOverrideHeaders = new HashMap<>((Map) headers.get(
+                Constants.ATTRIBUTE_OVERRIDES_KEY));
+        if (!attributeOverrideHeaders.isEmpty()) {
             storageRequest.getProperties()
                     .put(Constants.ATTRIBUTE_OVERRIDES_KEY,
-                            createAttributeOverrideMapFromHeaders(attributeOverrideHeaders));
+                            (Serializable) attributeOverrideHeaders);
         }
         if (headers.containsKey(Constants.STORE_REFERENCE_KEY)) {
             storageRequest.getProperties()
                     .put(Constants.STORE_REFERENCE_KEY, ingestedFile.getAbsolutePath());
         }
-    }
-
-    public HashMap<String, List<String>> createAttributeOverrideMapFromHeaders(
-            String attributeOverrideHeaders) {
-        return Arrays.stream(attributeOverrideHeaders.split(","))
-                .collect(Collectors.toMap(s -> s.split("=")[0],
-                        s -> Collections.singletonList(s.split("=")[1]),
-                        (a, b) -> Stream.concat(a.stream(), b.stream())
-                                .collect(Collectors.toList()),
-                        HashMap::new));
     }
 }
--- a/catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
+++ b/catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
@@ -15,15 +15,16 @@ package org.codice.ddf.catalog.content.monitor;
 
 import static org.codice.ddf.security.common.Security.runAsAdmin;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 
 import org.apache.camel.CamelContext;
 import org.apache.camel.Exchange;
@@ -39,7 +40,6 @@ import org.slf4j.LoggerFactory;
 
 import ddf.catalog.Constants;
 import ddf.security.common.util.Security;
-
 import net.jodah.failsafe.Failsafe;
 import net.jodah.failsafe.RetryPolicy;
 
@@ -67,7 +67,7 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
 
     private List<RouteDefinition> routeCollection;
 
-    private List<String> attributeOverrides;
+    private Map<String, Serializable> attributeOverrides;
 
     private final int maxRetries;
 
@@ -142,7 +142,8 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
      * @param readLockIntervalMilliseconds
      */
     public void setReadLockIntervalMilliseconds(Integer readLockIntervalMilliseconds) {
-        this.readLockIntervalMilliseconds = Math.max(readLockIntervalMilliseconds, MIN_READLOCK_INTERVAL_MILLISECONDS);
+        this.readLockIntervalMilliseconds = Math.max(readLockIntervalMilliseconds,
+                MIN_READLOCK_INTERVAL_MILLISECONDS);
     }
 
     public Integer getReadLockIntervalMilliseconds() {
@@ -248,7 +249,23 @@ public class ContentDirectoryMonitor implements DirectoryMonitor {
      * @param attributeOverrides - a list of attributes to override
      */
     public void setAttributeOverrides(List<String> attributeOverrides) {
-        this.attributeOverrides = attributeOverrides;
+        Map<String, Serializable> attributeOverrideMap = new HashMap<>();
+        for (String s : attributeOverrides) {
+            String[] keyValue = s.split("=", 2);
+
+            if (keyValue.length < 2) {
+                LOGGER.error("Invalid attribute override configured for monitored directory");
+                throw new IllegalStateException(
+                        "Invalid attribute override configured for monitored directory");
+            }
+
+            if (attributeOverrideMap.get(keyValue[0]) == null) {
+                attributeOverrideMap.put(keyValue[0], new ArrayList<String>());
+            }
+            ArrayList valueList = (ArrayList) attributeOverrideMap.get(keyValue[0]);
+            valueList.add(keyValue[1]);
+        }
+        this.attributeOverrides = attributeOverrideMap;
     }
 
     public List<RouteDefinition> getRouteDefinitions() {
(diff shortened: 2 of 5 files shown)


catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -192,25 +189,15 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
 
     public void processHeaders(Map<String, Object> headers, StorageRequest storageRequest,
             File ingestedFile) {
-        String attributeOverrideHeaders = (String) headers.get(Constants.ATTRIBUTE_OVERRIDES_KEY);
-        if (StringUtils.isNotEmpty(attributeOverrideHeaders)) {
+        HashMap<String, Serializable> attributeOverrideHeaders =
```
brendan-hofmann: I would prefer casting to the interface `Map` rather than the implementation `HashMap`, just as a matter of good form.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -248,7 +249,13 @@ public void setProcessingMechanism(String processingMechanism) {
      * @param attributeOverrides - a list of attributes to override
      */
     public void setAttributeOverrides(List<String> attributeOverrides) {
-        this.attributeOverrides = attributeOverrides;
+        Map<String, Serializable> attributeOverrideMap = new HashMap<>();
+        for (String s : attributeOverrides) {
+            String[] keyValue = s.split("=");
+            attributeOverrideMap.put(keyValue[0], keyValue[1]);
```
jlcsmith: Might want to validate the keyValue array size before accessing elements within it

brendan-hofmann: I think this approach may break multivalued attributes. The documentation currently states that multiple override values are specified by repeating the `key=value` entries, so we would need to collect those values somehow. Perhaps a map of lists.

emanns95: Oh true. Didn't really consider those

brendan-hofmann: I am surprised no tests broke. Perhaps we need some more. ;)

emanns95: Sure thing

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorTest.java
```diff
@@ -186,7 +186,7 @@ public void testDirectoryMonitorWithParameters() throws Exception {
                 .get(0);
         assertThat(routeDefinition.toString(), containsString(
                 "SetHeader[" + Constants.ATTRIBUTE_OVERRIDES_KEY
-                        + ", simple{Simple: test1=someParameter1,test2=someParameter2}"));
+                        + ", {{test1=someParameter1, test2=someParameter2}}"));
```
emanns95: I'm fine with rewriting the test :+1:

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -192,25 +189,15 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
 
     public void processHeaders(Map<String, Object> headers, StorageRequest storageRequest,
             File ingestedFile) {
-        String attributeOverrideHeaders = (String) headers.get(Constants.ATTRIBUTE_OVERRIDES_KEY);
-        if (StringUtils.isNotEmpty(attributeOverrideHeaders)) {
+        HashMap<String, Serializable> attributeOverrideHeaders = new HashMap<>((Map) headers.get(
```
brendan-hofmann: I would still prefer to cast to the interface instead of the implementation here.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -248,7 +249,22 @@ public void setProcessingMechanism(String processingMechanism) {
      * @param attributeOverrides - a list of attributes to override
      */
     public void setAttributeOverrides(List<String> attributeOverrides) {
-        this.attributeOverrides = attributeOverrides;
+        Map<String, Serializable> attributeOverrideMap = new HashMap<>();
+        for (String s : attributeOverrides) {
+            String[] keyValue = s.split("=");
+
+            if (keyValue.length != 2) {
+                LOGGER.debug("Got the wrong number of arguments for an attribute override");
```
brendan-hofmann: I think this should be higher than debug, perhaps even refuse to create the route. A malformed override is definitely not going to do what an admin wants it to do, and they should be informed of that.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
emanns95: I am casting to the `Map` interface. Unless you mean to declare it as a `Map` (which isn't Serializable so I wouldn't be able to pass it into the storageRequest properties)

coyotesqrl: I believe Brendan is saying that he'd like the left side to be `Map<String, Serializable>`. As it's being used locally it's not super important to me, but in general I agree with that.

coyotesqrl: Is there a guarantee that there will be some non-null value in `headers.get(Constants.ATTRIBUTE_OVERRIDES_KEY)`? If not, you need to guard against that or you'll thrown an NPE here.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
emanns95: Thinking warn?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -335,15 +351,13 @@ public void configure() throws Exception {
                 RouteDefinition routeDefinition = from(inbox);
 
                 if (attributeOverrides != null) {
-                    String attributeOverrideString = attributeOverrides.stream()
-                            .map(String::trim)
-                            .collect(Collectors.joining(","));
-                    routeDefinition.setHeader(Constants.ATTRIBUTE_OVERRIDES_KEY, simple(
-                            attributeOverrideString));
+                    routeDefinition.setHeader(Constants.ATTRIBUTE_OVERRIDES_KEY)
+                            .constant(attributeOverrides);
```
coyotesqrl: This is my ignorance of Camel showing: why is this being set as `.constant()` on the header instead of calling `setHeader` with two params where the second is the output of a call to `.simple()`? That's how it appears it's being done everywhere else.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
emanns95: I'll check on that NPE.

But yeah as far as I know, since Java's `Map` interface doesn't implement `Serializable`, I won't be able to add it to the `StorageRequest` unless I cast it to a `Seriablizable`, which I can, if we're okay with that 👍

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -248,7 +249,22 @@ public void setProcessingMechanism(String processingMechanism) {
      * @param attributeOverrides - a list of attributes to override
      */
     public void setAttributeOverrides(List<String> attributeOverrides) {
-        this.attributeOverrides = attributeOverrides;
+        Map<String, Serializable> attributeOverrideMap = new HashMap<>();
+        for (String s : attributeOverrides) {
+            String[] keyValue = s.split("=");
```
rzwiefel: If we expect there to only be two, why not just add the limit to the split (`s.split("=",  2)`) such that we ignore the rest of equals and include them in the value.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
emanns95: `.simple()` can't take an `Object`, so it'd reject my `Map` here

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
emanns95: Because I didn't know we could! 😄  I'll do that

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
emanns95: As for the NPE, it looks like it'll either be a populated or an empty list. If nothing is given from the front it end it gives an empty list and it won't be used anywhere since they're all checking if it's empty or not :+1:

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorTest.java
emanns95: I dug as deep as I think I can and it just ends up with me relying on a `toString` in `SetHeaderDefinition` instead...

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: I see what you are saying now, my bad for not being more clear. It's all fine, nothing to see here. 😅

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
brendan-hofmann: I was actually thinking `ERROR` and don't create the route either. Imagine if one of the overrides was `security.access*`, an admin would end up incorrectly restricting the data from that route, or not restricting it at all.

emanns95:
Multi-value attribute support verified @brendan-hofmann

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -247,8 +251,24 @@ public void setProcessingMechanism(String processingMechanism) {
     /**
      * @param attributeOverrides - a list of attributes to override
      */
-    public void setAttributeOverrides(List<String> attributeOverrides) {
-        this.attributeOverrides = attributeOverrides;
+    public boolean setAttributeOverrides(List<String> attributeOverrides) {
```
emanns95: OSGI doesn't like me changing this return type, that's why the itests failed

emanns95: Fixed 👍

emanns95:
Passed full build (bamboo test failure verified working)

catalog/core/catalog-core-directorymonitor/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -25,7 +25,7 @@
             name="Maximum Concurrent Files" id="numThreads" required="true"
             type="Integer" default="1"/>
 
-        <AD description="Specifies the time to wait (in milliseconds) before acquiring a lock on a file in the monitored directory.  This interval is used for sleeping between attempts to acquire the read lock on a file to be ingested.  The default value of 1000 milliseconds is recommended."
+        <AD description="Specifies the time to wait (in milliseconds) before acquiring a lock on a file in the monitored directory.  This interval is used for sleeping between attempts to acquire the read lock on a file to be ingested.  The default value of 100 milliseconds is recommended."
```
coyotesqrl: You might want to add that if the value is set below 100ms, it will be bumped up automatically to that instead.

emanns95: Sure thing

stustison:
Did we actually add a test that proves that you can have a value with a comma in it, since that is what this is supposed to be solving? I didn't see one when I looked.

emanns95:
@stustison @jlcsmith Added a unit test to verify that an attribute override containing commas is preserved and added to the route definition

codymacdonald:
Manually verified 👍

PR #1702, opened 2017-02-17
	DDF-2809 fixed issue with extremely long paths or badly formed paths

stustison:
@coyotesqrl @brendan-hofmann 
@pklinef
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/security/policy/security-policy-context/src/main/java/org/codice/ddf/security/policy/context/impl/PolicyManager.java
+++ b/platform/security/policy/security-policy-context/src/main/java/org/codice/ddf/security/policy/context/impl/PolicyManager.java
@@ -58,16 +58,20 @@ public class PolicyManager implements ContextPolicyManager {
     private static final String[] DEFAULT_REALM_CONTEXTS =
             new String[] {"/=" + DEFAULT_REALM_CONTEXT_VALUE};
 
+    private static final int MAX_TRAVERSAL_DEPTH = 500;
+
     private Map<String, ContextPolicy> policyStore = new HashMap<>();
 
     private List<String> whiteListContexts = new ArrayList<>();
 
     private ContextPolicy defaultPolicy = new Policy("/",
             DEFAULT_REALM,
-            new ArrayList<String>(),
-            new ArrayList<ContextAttributeMapping>());
+            new ArrayList<>(),
+            new ArrayList<>());
 
     private Map<String, Object> policyProperties = new HashMap<>();
+
+    private int traversalDepth;
 
     public PolicyManager() {
         policyStore.put("/", defaultPolicy);
@@ -75,11 +79,11 @@ public class PolicyManager implements ContextPolicyManager {
 
     @Override
     public ContextPolicy getContextPolicy(String path) {
-        return getContextPolicy(path, getPolicyStore(), getWhiteListContexts());
+        return getContextPolicy(path, getPolicyStore(), getWhiteListContexts(), 0);
     }
 
     private ContextPolicy getContextPolicy(String path, Map<String, ContextPolicy> policyStore,
-            List<String> whiteListContexts) {
+            List<String> whiteListContexts, int depth) {
         ContextPolicy entry;
         entry = policyStore.get(path);
 
@@ -89,8 +93,8 @@ public class PolicyManager implements ContextPolicyManager {
             return null;
         } else {
             String pathFragment = rollbackPath(path);
-            if (StringUtils.isNotEmpty(pathFragment)) {
-                return getContextPolicy(pathFragment, policyStore, whiteListContexts);
+            if (StringUtils.isNotEmpty(pathFragment) && depth <= traversalDepth) {
+                return getContextPolicy(pathFragment, policyStore, whiteListContexts, ++depth);
             } else {
                 //this is just here for safety
                 //if we get down to the point where we can never get an entry, return the default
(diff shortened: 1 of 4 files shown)


platform/security/policy/security-policy-context/src/main/java/org/codice/ddf/security/policy/context/impl/PolicyManager.java
```diff
@@ -58,28 +58,32 @@
...
 
     @Override
     public ContextPolicy getContextPolicy(String path) {
-        return getContextPolicy(path, getPolicyStore(), getWhiteListContexts());
+        return getContextPolicy(path, getPolicyStore(), getWhiteListContexts(), 0);
     }
 
     private ContextPolicy getContextPolicy(String path, Map<String, ContextPolicy> policyStore,
-            List<String> whiteListContexts) {
+            List<String> whiteListContexts, int depth) {
```
coyotesqrl: Is this formatted?

platform/security/policy/security-policy-context/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -19,6 +19,11 @@
          name="Web Context Policy Manager"
          id="org.codice.ddf.security.policy.context.impl.PolicyManager">
 
+        <AD description="Depth to which paths will be traversed. Any value greater than 500 will be set to 500."
```
coyotesqrl: Should we even support a path this deep? If they're not trying to DDOS the system, passing in some garbage like hundreds of slashes (which you're resolving) and were using a normal path with actual subcomponents in it, it'd be highly unlikely to contain 500 subelements.

stustison: Yeah, I was just trying to come up with a value that was large enough to cover any web app that someone might want to deploy and also small enough that it wouldn't cause any issues. Plus we're consuming the slashes now instead of recursing on them, so those will just be tossed. So that is why I picked 500. I also figured 20 was plenty for a default value of anything we're likely to encounter.

platform/security/policy/security-policy-context/src/main/java/org/codice/ddf/security/policy/context/impl/PolicyManager.java
stustison: It is.

platform/security/policy/security-policy-context/src/main/resources/OSGI-INF/metatype/metatype.xml
brendan-hofmann: I am inclined to agree with @coyotesqrl . Not that it matters much, but `100` is a more conventional arbitrary limit in my experience as well.

platform/security/policy/security-policy-context/src/test/java/org/codice/security/policy/context/PolicyManagerTest.java
```diff
@@ -119,6 +124,37 @@ public void setup() {
         rollBackTestManager.setPolicies(contextPolicies);
     }
 
+    @Test
+    public void testBadTraversal() {
```
brendan-hofmann: Somewhat of a nit, but this could be difficult to debug a breaking change. A little commentary on these edge cases would help.

May also want to consider testing the `MAX_TRAVERSAL_DEPTH` upper bound.

platform/security/policy/security-policy-context/src/main/java/org/codice/ddf/security/policy/context/impl/PolicyManager.java
```diff
@@ -58,28 +58,32 @@
     private static final String[] DEFAULT_REALM_CONTEXTS =
             new String[] {"/=" + DEFAULT_REALM_CONTEXT_VALUE};
 
+    private static final int MAX_TRAVERSAL_DEPTH = 500;
```
brendan-hofmann: We should probably have a lower bound, and a test for it.

platform/security/policy/security-policy-context/src/test/java/org/codice/security/policy/context/PolicyManagerTest.java
```diff
@@ -108,6 +109,10 @@ public void setup() {
                 new Policy("/aaa/aaa", null, new ArrayList<String>(), null));
         manager.setContextPolicy("/foo/bar",
                 new Policy("/foo/bar", null, new ArrayList<String>(), null));
+        manager.setContextPolicy(
+                "/foo/bar/baz/blah/bing/bab/bozo/bongo/bingo/burp/blaze/blurb/blunt/block",
```
pklinef: While these are cute, it might be easier to double check if it was `/1/2/3/...` instead.

platform/security/policy/security-policy-context/src/main/resources/OSGI-INF/metatype/metatype.xml
pklinef: I would rather err on the side of caution with a value that cannot be overridden.  If we pick something too low and it breaks a web app, there is no way to workaround this.

platform/security/policy/security-policy-context/src/test/java/org/codice/security/policy/context/PolicyManagerTest.java
stustison: I agree. I'll change them around to make it easier for debugging.

PR #1704, opened 2017-02-17
DDF-2814 Fixed self-intersecting polygon searches

clockard:
#### What does this PR do?
Fixed self-intersecting polygon searches
@brendan-hofmann @emmberk @mcalcote 
@andrewkfiedler
@pklinef

Install DDF and go to the catalog ui. 
Create a custom local search for anyGeo with an intersecting polygon.
Verify that the search completes without any errors. 
https://codice.atlassian.net/browse/DDF-2814
screen shot 2017-02-17 at 3 26 22 pm

Results of using convex hull. Original search polygon in blue with actually searched polygon in red.
screen shot 2017-02-20 at 12 35 41 pm

#### Checklist:
- [X] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests


catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
```diff
@@ -87,6 +87,28 @@ public void intersectsWithValidJtsWkt() {
...
+        stub(mockResolver.getField("testProperty", AttributeFormat.GEOMETRY, false)).toReturn(
+                "testProperty_geohash_index");
+        SolrQuery query = toTest.contains("testProperty", wkt);
+        assertThat(query.getQuery(),
+                startsWith(
+                        "testProperty_geohash_index:\"Contains(POLYGON ((5 -5, 0 0, 0 20, 10 20, 10 0, 5 -5)))\""));
+    }
+
+    @Test
+    public void nonIntersectingPolygon() {
```
brendan-hofmann: If this is meant to assert that non-intersecting polygons are not modified, maybe we should throw some more edge cases in there?

catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
```diff
@@ -87,6 +87,28 @@ public void intersectsWithValidJtsWkt() {
...
 
     @Test
+    public void selfIntersectingPolygon() {
+        String wkt = "POLYGON((0 0, 10 0, 10 20, 5 -5, 0 20, 0 0))";
+        stub(mockResolver.getField("testProperty", AttributeFormat.GEOMETRY, false)).toReturn(
+                "testProperty_geohash_index");
+        SolrQuery query = toTest.contains("testProperty", wkt);
+        assertThat(query.getQuery(),
+                startsWith(
+                        "testProperty_geohash_index:\"Contains(POLYGON ((5 -5, 0 0, 0 20, 10 20, 10 0, 5 -5)))\""));
```
brendan-hofmann: Couldn't we also be more direct and assert that it blows up before the repair operation is run and does not blow up afterwards? The modified query is sort of a side effect of the actual goal.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlResult.java
```diff
@@ -63,6 +64,8 @@
     private static final Map<String, String> SPATIAL_CONTEXT_ARGUMENTS = ImmutableMap.of(
             "spatialContextFactory",
             JtsSpatialContextFactory.class.getName(),
+            "validationRule",
+            ValidationRule.repairConvexHull.name(),
```
brendan-hofmann: Do we need `MultiOverlap` here too?

We should probably note why we are performing `ConvexHull` here, and why we are not using `Buffer0`. There is some discussion on it here: https://github.com/locationtech/spatial4j/issues/34#issuecomment-12426283

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -631,13 +656,25 @@ private Geometry getGeometry(String wkt) {
...
     }
 
+    private String sanitizeGeometry(String wkt) {
+        try {
+            Shape wktShape = WKT_READER.read(wkt);
+            return SPATIAL_CONTEXT.getFormats()
+                    .getWktWriter()
+                    .toString(wktShape);
+        } catch (IOException | java.text.ParseException | InvalidShapeException e) {
+            LOGGER.info("Failed sanitize or read WKT: {}", wkt, e);
```
brendan-hofmann: Typo? Think it should be 'Failed `to` sanitize'.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -82,6 +92,21 @@
 
     private static final WKTWriter WKT_WRITER = new WKTWriter();
 
+    private static final Map<String, String> SPATIAL_CONTEXT_ARGUMENTS = ImmutableMap.of(
+            "spatialContextFactory",
+            JtsSpatialContextFactory.class.getName(),
+            "validationRule",
+            ValidationRule.repairConvexHull.name(),
+            "allowMultiOverlap",
```
brendan-hofmann: Would be good to comment on these options, as these geospatial tools can be somewhat opaque. There is some discussion on it here: https://github.com/locationtech/spatial4j/issues/34#issuecomment-12426283


Do the unit tests exercise both the `repair` and `MultiOverlap` modifications?

catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
clockard: I feel that testing other geometry would be more a test of the JTS code than our code.

catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
clockard: I think this is the most direct way we can test this. The repair logic is called within the `contains` method logic and output of that is the updated query.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
clockard: Fixed

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
clockard: I had included the allowMultiOverlap since the CqlRequest had it but now that I am looking at it I don't think we need this here. Removing allowMultiOverlap from the configuration.

jlcsmith:
Build and tested successfully. Note that since we are using the recommended Convex Hull fix for self-intersecting geos, it is very possible to get more results than are technically included in the provided geometry.  The "Convex Hull" of a geometry appears to essentially be the envelope polygon.

brianfelix:
Hero build passed and test steps were manually verified

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -82,6 +92,21 @@
 
     private static final WKTWriter WKT_WRITER = new WKTWriter();
 
+    // For queries we use repairConvexHull which my cause false positives to be returned but this
```
pklinef: Is there a place in the documentation where we can add a note or warning about this?

clockard: Added a note to the Catalog UI search documentation since that seems like the most likely place a user would run into this. 
screen shot 2017-02-21 at 9 30 54 am

brendan-hofmann: Love the visual aid.

lifenouveau:
Perhaps there is a more intuitive way we could solve this now (or in the future). 

- We could divide the shape into different shapes based on the intersection.

- The shape could be saved as is by making the intersection points not intersect (perhaps only get very close thus redrawing any "bowties" to create valid polygons instead).

PR #1705, opened 2017-02-17
DDF-2813 Added metadata backup plugin and itests

bdeining:
#### What does this PR do?
This PR adds a Metadata Backup Plugin to the Catalog App as well as itests.
@rzwiefel @troymohl @glenhein 
@peterhuffer 
@kcwire 
@millerw8
Build / Install / Turn on `catalog-metadata-backup` / Configure `Metadata Backup Plugin` / Ingest, Update, Delete data via Rest and CSW and verify appropriate behavior.
The existing backup feature stores the entire metacard in a file using an ObjectOutputStream. In order to leverage the existing code, a configuration parameter would have to be added to enforce only metadata be backed up. This is not ideal in situations where an administrator wants the metacard backed up as well as the raw metadata. Additionally, making the existing backup plugin configurable undermines what it was designed to do, which is to backup local catalog records. Thus, it makes more sense to create a separate plugin for metadata backup. This plugin will use the newly designed asynchronous functionality to handle i/o.
This PR depends on https://github.com/codice/ddf/pull/1495
[DDF-2813](https://codice.atlassian.net/browse/DDF-2813)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests


catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,203 @@
...
+        } catch (IOException e) {
+            LOGGER.debug("Unable to backup metadata {}. ", filename, e);
+        }
+        clearEmptyDirectories(parent);
+    }
+
+    private File getCompleteDirectory(int depth, String id) {
+        File parent = new File(getOutputDirectory());
+        for (int i = 0; i < depth && id.length() > (i * 2 + 2); i++) {
+            parent = new File(parent, id.substring(i * 2, i * 2 + 2));
```
troymohl: I think this is creating directories based off of the ID 2 characters at a time. I'm a little confused on the +2? If it's b/c we are starting the loop at 0, would it make it cleaner to just start with `i = 1`?
Also, if the depth is configurable should we change the directory logic to try to more evenly divide the id, rather than going 2 characters at a time until the depth is reached?

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,203 @@
...
+    @Override
+    public ProcessRequest<ProcessUpdateItem> processUpdate(
+            ProcessRequest<ProcessUpdateItem> processRequest) throws PluginExecutionException {
+        processRequest(processRequest);
+        return processRequest;
+    }
+
+    @Override
+    public ProcessRequest<ProcessDeleteItem> processDelete(
+            ProcessRequest<ProcessDeleteItem> processRequest) throws PluginExecutionException {
```
bdeining: I agree, I'll add a boolean to keep deleted metadata files

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: This is just following the same method that the FileSystemStorageProvider and CatalogBackupPlugin use.  We could tweak this to evenly divide the id, but its only purpose is to circumvent file system limits for a single directory

bdeining:
@pklinef changed to use a managed service factory and made the MetacardTransformer ID configurable

kcwire:
It would appear this depends on #1495 which has not been merged. It would be good to know if there are other tickets required merging in the PR.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+
+    private Boolean keepDeletedMetacards = false;
+
+    private String metacardTransformerId;
+
+    private MetacardTransformer metacardTransformer;
+
+    private String outputDirectory;
+
+    private Integer folderDepth;
```
rzwiefel: maybe initialize to a non-null value? If it remains null and is passed to the method below it will NPE.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        validateDirectory(parent);
+        File file = new File(parent.getPath() + File.separator + metacardId);
+        LOGGER.trace("Writing metadata from {} to file {}", metacardId, file.getPath());
+
+        try (OutputStream outputStream = new FileOutputStream(file)) {
+            IOUtils.write(content.getByteArray(), outputStream);
+        } catch (IOException e) {
+            LOGGER.debug("Unable to backup {}. ", metacardId, e);
```
rzwiefel: Should this log level be higher? This could be thrown in the scenario the directory runs out of room/etc which is something an admin could take action on.

catalog/plugin/catalog-plugin-metadatabackup/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -0,0 +1,47 @@
...
+                description="Output directory to place metadata files for backup."
+                name="Output Directory" id="outputDirectory" required="true" type="String"
+                default="data/backup/metadata"/>
+        <AD
+                description="The Folder Depth (for use with extremely large catalogs).
+                             This is used to prevent the system from hitting the file limit
+                             in the output directory (range 0-4).  For example, FAT32 has a
+                             maximum of 65,534 files permissible in a single folder."
+                name="Folder Depth" id="folderDepth" required="true" type="Integer"
+                default="0"/>
```
rzwiefel: Should we default to at least 1 folder deep? The overhead will be negligible (256 folders top level) but the will allow the default to suit a much larger backup directory in the default case.

troymohl:
Hero build passed.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+                    "Unable to backup ingested metadata; no outputDirectory.");
+        }
+
+        if (metacardTransformer == null) {
+            throw new PluginExecutionException(
+                    "Unable to backup ingested metadata; no Metacard Transformer found.");
+        }
+
+        List<? extends ProcessResourceItem> processResourceItems = processRequest.getProcessItems();
+        for (ProcessResourceItem processResourceItem : processResourceItems) {
```
glenhein: Can this be a 'processResourceItems.forEach(...)' instead? That would eliminate some branch complexity

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+
+    public void setMetacardTransformer(MetacardTransformer metacardTransformer) {
+        this.metacardTransformer = metacardTransformer;
+    }
+
+    public String getMetacardTransformerId() {
+        return metacardTransformerId;
+    }
+
+    public void refresh(Map<String, Object> properties) {
```
glenhein: This is a lot of conditional logic and type checking in one method. Is there a cleaner way of doing this?

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+        try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(dir)) {
+            return !dirStream.iterator()
+                    .hasNext();
+        } catch (IOException e) {
+            LOGGER.debug("Unable to open directory stream for {}", dir.toString(), e);
+        }
+        return false;
+    }
+
+    private void validateDirectory(File directory) {
```
glenhein: nit: the parameter name is 'directory', but then you check to see if it is a directory. Seems like the parameter name should be 'path'.

catalog/plugin/catalog-plugin-metadatabackup/src/test/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPluginTest.java
```diff
@@ -0,0 +1,423 @@
...
+            metadataBackupPlugin.setFolderDepth(depth);
+            ProcessRequest<ProcessUpdateItem> newUpdateRequest = metadataBackupPlugin.processUpdate(
+                    updateRequest);
+            assertFiles(newUpdateRequest.getProcessItems(), depth);
+        }
+    }
+
+    @Test
+    public void testDeleteRequestVaryingDepths() throws Exception {
+        for (int i = 0; i < 12; i++) {
```
glenhein: why '12'?

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Changed

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Changed

catalog/plugin/catalog-plugin-metadatabackup/src/test/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPluginTest.java
bdeining: 12 was just an arbitrary number above the MAX_FOLDER_DEPTH.  These tests are just to verify larger depths behave appropriately.  I've changed it to 5 instead (1 above MAX_FOLDER_DEPTH).

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Unfortunately, I haven't seen a cleaner way to do this.  We need to type check the parameters that come in from the properties map.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Changed to initialize to MIN_FOLDER_DEPTH

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Elevated the log level and provided the admin with information on what action to take

bdeining:
@kcwire / @millerw8 please review when possible

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
kcwire: should this be pulled out into a utility if we are applying the same logic in 2 places?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFtp.java
```diff
@@ -305,6 +314,34 @@ public void testFtpsPutStream() throws Exception {
         verifyIngest(1, METACARD_TITLE);
     }
 
+    /**
+     * Upload a file via FTPS for ingest using streaming method and verify that the file is backed
```
kcwire: It seems odd to spread the testing of this plugin across 2 itests.  Would it be more appropriate to use the REST endpoint in `TestCatalog` so the feature is tested in a single itest?

kcwire:
Have you tested re-ingesting the metadata that is backed up?

bdeining:
@kcwire : @troymohl did that during the hero

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFtp.java
bdeining: We can probably just remove the itest from `TestFtp` since it is redundant.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: It follows the same method in the sense that the folder structure is generated by using 2 characters of the id at each depth level, but the logic isn't exactly the same in the `FileSystemStorageProvider`

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFtp.java
bdeining: Removed `TestFtp` changes

bdeining:
full build successful after removing itest from `TestFtp.java`

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
stustison: This seems like it does more than you really need it to. By just creating 2 directories that are 3 characters each, that provides enough randomness to allow for 361,000,000,000 metacards before you ran into the NTFS file system limit. It seems to me that it'd make more sense to just go with the same exact logic that FileSystemStorageProvider uses (which is also simpler and easier to read) and then put that logic in some common location for both of these.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+        if (!parent.exists()) {
+            try {
+                FileUtils.forceMkdir(parent);
+            } catch (IOException e) {
+                LOGGER.debug("Unable to make directory {}", parent.getPath(), e);
+            }
+        }
+
+        validateDirectory(parent);
+        File file = new File(parent.getPath() + File.separator + metacardId);
```
stustison: You should really be using https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html and related classes for all of these operations to avoid running into file system issues.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+                } catch (IOException e) {
+                    LOGGER.debug("Unable to delete directory {}", file.toPath(), e);
+                }
+            }
+            file = file.getParentFile();
+        }
+    }
+
+    private boolean isDirectoryEmpty(Path dir) {
+        try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(dir)) {
```
stustison: We're mixing the "Path" way of doing things with the old "File" way of doing things. Like my previous comment says, you should probably switch everything over to using Paths (which this code is doing).

catalog/plugin/catalog-plugin-metadatabackup/src/main/resources/OSGI-INF/metatype/metatype.xml
stustison: If you just use the code from FileSystemStorageProvider, you can remove this option entirely and it will just always default to 2 directories for depth. At the point where you would run into a file system limit, you'd be at more records than Solr could handle anyways.

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
```diff
@@ -0,0 +1,313 @@
...
+            throw new PluginExecutionException(
+                    "Unable to backup ingested metadata; no Metacard Transformer found.");
+        }
+
+        List<? extends ProcessResourceItem> processResourceItems = processRequest.getProcessItems();
+        processResourceItems.forEach(processResourceItem -> {
+            Metacard metacard = processResourceItem.getMetacard();
+            try {
+                LOGGER.trace("Backing up metadata for metacard : {}", metacard.getId());
+                BinaryContent binaryContent = metacardTransformer.transform(metacard, null);
```
millerw8: I like the idea of backing up based on any configured transformer, but with that being the case, should we be adjusting our naming convention?  Since they may not be using the metadata transformer

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Refactored this to create 2 directories that are 3 characters each.  The logic for the `FileSystemStorageProvider` is a little different in that it uses URIs (and their fragments), so refactoring to make a common location for both of these may not be practical (although i'm not against it).

catalog/plugin/catalog-plugin-metadatabackup/src/main/resources/OSGI-INF/metatype/metatype.xml
bdeining: Removed folder depth from the configuration

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Done

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Done

catalog/plugin/catalog-plugin-metadatabackup/src/main/java/org/codice/ddf/catalog/plugin/metadata/backup/MetadataBackupPlugin.java
bdeining: Adjusted naming convention to omit references to `metadata`

PR #1712, opened 2017-02-21
DDF-2815: Add UTM coordinate support to CSW endpoint

troymohl:
#### What does this PR do?
Adds UTM coordinate support for OGC Filters in the CSW endpoint.

@bdeining 
@rzwiefel 
@ricklarsen 

[OGC](https://github.com/orgs/codice/teams/OGC)

@kcwire
@millerw8

Build & Run Unit Tests and Itests

With postman, POST the following to the CSW endpoint and verify no exceptions are returned:
`https://localhost:8993/services/csw`
`<?xml version="1.0" ?>
<GetRecords xmlns="http://www.opengis.net/cat/csw/2.0.2"
        xmlns:ogc="http://www.opengis.net/ogc"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns:gml="http://www.opengis.net/gml"
        service="CSW"
        version="2.0.2"
        maxRecords="4"
        startPosition="1"
        resultType="results"
        outputFormat="application/xml"
        outputSchema="http://www.opengis.net/cat/csw/2.0.2"
        xsi:schemaLocation="http://www.opengis.net/cat/csw/2.0.2 ../../../csw/2.0.2/CSW-discovery.xsd">
    <Query typeNames="Record">
        <ElementSetName>summary</ElementSetName>
        <Constraint version="1.1.0">
            <ogc:Filter>
                <ogc:Intersects>
                    <ogc:PropertyName>ows:BoundingBox</ogc:PropertyName>
                    <gml:Envelope srsName="urn:ogc:def:crs:EPSG:32636">
                        <gml:lowerCorner>171070 1106907</gml:lowerCorner>
                        <gml:upperCorner>225928 1106910</gml:upperCorner>
                    </gml:Envelope>
                </ogc:Intersects>
            </ogc:Filter>
        </Constraint>
    </Query>
</GetRecords>`

UTM Coordinates are handled with the following coordinate reference systems:
EPSG:326XX -> North UTM coordinates where XX is the UTM zone
EPSG:327XX -> South UTM coordinates where XX is the UTM zone

https://codice.atlassian.net/browse/DDF-2815

#### Checklist:
- [X] Documentation Updated
- [X] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
@@ -23,6 +23,8 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.lang.StringUtils;
+import org.codice.ddf.libs.geo.GeoFormatException;
+import org.codice.ddf.libs.geo.util.GeospatialUtil;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.CswConstants;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.GmdConstants;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.converter.DefaultCswRecordMap;
@@ -49,11 +51,22 @@ import org.opengis.filter.expression.Literal;
 import org.opengis.filter.expression.PropertyName;
 import org.opengis.filter.spatial.BBOX;
 import org.opengis.filter.spatial.Beyond;
+import org.opengis.filter.spatial.Contains;
+import org.opengis.filter.spatial.Crosses;
 import org.opengis.filter.spatial.DWithin;
+import org.opengis.filter.spatial.Disjoint;
+import org.opengis.filter.spatial.Equals;
+import org.opengis.filter.spatial.Intersects;
+import org.opengis.filter.spatial.Overlaps;
+import org.opengis.filter.spatial.Touches;
+import org.opengis.filter.spatial.Within;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import org.opengis.temporal.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.helpers.NamespaceSupport;
+
+import com.vividsolutions.jts.geom.Geometry;
 
 import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.AttributeType;
@@ -111,6 +124,8 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
     public Object visit(BBOX filter, Object extraData) {
         Expression geometry1 = visit(filter.getExpression1(), SPATIAL_QUERY_TAG);
         Expression geometry2 = visit(filter.getExpression2(), extraData);
+        convertGeometryExpressionToEpsg4326(geometry1);
+        convertGeometryExpressionToEpsg4326(geometry2);
         return getFactory(extraData).within(geometry1, geometry2);
     }
 
@@ -120,6 +135,8 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
 
         Expression geometry1 = visit(filter.getExpression1(), SPATIAL_QUERY_TAG);
         Expression geometry2 = visit(filter.getExpression2(), extraData);
+        convertGeometryExpressionToEpsg4326(geometry1);
+        convertGeometryExpressionToEpsg4326(geometry2);
 
         return getFactory(extraData).beyond(geometry1,
                 geometry2,
@@ -133,6 +150,8 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
 
         Expression geometry1 = visit(filter.getExpression1(), SPATIAL_QUERY_TAG);
         Expression geometry2 = visit(filter.getExpression2(), extraData);
+        convertGeometryExpressionToEpsg4326(geometry1);
+        convertGeometryExpressionToEpsg4326(geometry2);
 
         return getFactory(extraData).dwithin(geometry1,
                 geometry2,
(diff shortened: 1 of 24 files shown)


catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
```diff
@@ -503,5 +578,21 @@ public Object visit(Function function, Object extraData) {
             return super.visit(function, extraData);
         }
     }
+
+    private static void convertGeometryExpression(Expression expression) {
```
bdeining: nit : should this be renamed to convertGeometryExpressionToEPSG4326?

troymohl: Updated with the more descriptive name.

kcwire:
@troymohl I think this warrants an update to the change log. 

@ricklarsen Can you remind me where we have the guidelines for when and how to update the change log?

distribution/docs/src/main/resources/_contents/_endpoints/csw-endpoint-contents.adoc
```diff
@@ -1230,6 +1230,43 @@ It is possible to receive a response to a `GetRecords` query that conforms to th
 </csw:GetRecordsResponse>
 ----
 
+.`GetRecords` XML Request using UTM Coordinates
+UTM coordinates can be used when making a CSW GetRecords request using an `ogc:Filter`.
+UTM coordinates should use EPSG:326XX where XX is the zone within the northern hemisphere.
```
kcwire: I think we want to be more explicit here. The `EPSG:326XX` is the srsName.  Do we also require the exact urn that is in the sample below? Just want to remove any ambiguities.

libs/geospatial/src/main/java/org/codice/ddf/libs/geo/util/GeospatialUtil.java
```diff
@@ -13,9 +13,34 @@
...
+import org.opengis.referencing.crs.CRSAuthorityFactory;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.opengis.referencing.operation.MathTransform;
+import org.opengis.referencing.operation.TransformException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.vividsolutions.jts.geom.Geometry;
 
 public class GeospatialUtil {
```
kcwire: Can we get some glass level javadoc?

libs/geospatial/src/main/java/org/codice/ddf/libs/geo/util/GeospatialUtil.java
```diff
@@ -171,4 +196,73 @@ public static Double parseDMSLongitudeWithDecimalSeconds(String dmsLon)
...
+            CRSAuthorityFactory factory =
+                    ReferencingFactoryFinder.getCRSAuthorityFactory("EPSG", hints);
+            CoordinateReferenceSystem targetCRS = factory.createCoordinateReferenceSystem(
+                    EPSG_4326);
+            MathTransform transform = CRS.findMathTransform(sourceCrs, targetCRS);
+            return JTS.transform(geometry, transform);
+        } catch (FactoryException | TransformException e) {
+            LOGGER.debug("Unable to convert {} into {}", sourceCrsName, EPSG_4326, e);
+        }
+        return null;
```
kcwire: Should we throw instead of returning null? how would a caller know what went wrong without the exception?

libs/geospatial/src/main/java/org/codice/ddf/libs/geo/util/GeospatialUtil.java
```diff
@@ -171,4 +196,73 @@ public static Double parseDMSLongitudeWithDecimalSeconds(String dmsLon)
...
+                        ReferencingFactoryFinder.getCRSAuthorityFactory("EPSG", hints);
+                CoordinateReferenceSystem targetCRS = factory.createCoordinateReferenceSystem(
+                        EPSG_4326);
+
+                MathTransform transform = CRS.findMathTransform(sourceCrs, targetCRS);
+                transformedGeometry = JTS.transform(geometry, transform);
+                LOGGER.debug("Converted CRS {} into {} : {}", sourceCrs, EPSG_4326, geometry);
+            }
+        } catch (FactoryException | TransformException e) {
+            LOGGER.debug("Unable to convert {} into {}", sourceCrs, EPSG_4326, e);
```
kcwire: It appears we are hiding a failure case to the caller by not throwing this exception up.

jlcsmith:
@troymohl - it looks like you updated the Wfs source to transform geometries to 4236 in the [AbstractFeatureConverter](https://github.com/codice/ddf/pull/1712/files#diff-5e03baa5c2a95c8f23ea95c99d26ce05R288).  However, you didn't make a similar change on the Csw Source side.  Any reason for this?

troymohl:
@jlcsmith I really just extracted the Wfs code into our common GeospatialUtil class and added code for the CSW Endpoint to call to do a similar conversion. I also extracted some statics that would be useful in both methods. I left the existing WFS code (and CSW) intact except for this minor refactor of a method that wasn't necessarily WFS specific.

jlcsmith:
@troymohl - thanks for the explanation.  That makes sense.  I think we should make a similar change in the Csw Source but that sounds like a separate ticket

distribution/docs/src/main/resources/_contents/_endpoints/csw-endpoint-contents.adoc
troymohl: The full URN wasn't required, so I updated the doc and test to reflect that.

libs/geospatial/src/main/java/org/codice/ddf/libs/geo/util/GeospatialUtil.java
troymohl: Updated these methods to throw an exception rather than just returning null.

libs/geospatial/src/main/java/org/codice/ddf/libs/geo/util/GeospatialUtil.java
troymohl: Updated these methods to throw an exception rather than just returning null.

ricklarsen:
@kcwire We don't really have formal guidelines for the changelog, but I'll get some up on the documentation style guide wiki page.

troymohl:
@kcwire @ricklarsen I have updated the CHANGES.md file with this PR to add a 2.11 section and add my ticket in the `New Features` section

millerw8:
Can we add to this documentation that we have UTM support when using an OGC Filter but the CQL BNF doesn't support it: 
https://github.com/troymohl/ddf/blob/87d7a409173458bac2c61ceddd9cd5a8c6273fbd/distribution/docs/src/main/resources/_contents/_endpoints/csw-endpoint-contents.adoc

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
```diff
@@ -503,5 +579,26 @@ public Object visit(Function function, Object extraData) {
             return super.visit(function, extraData);
         }
     }
+
+    private static void convertGeometryExpressionToEpsg4326(Expression expression) {
+        if (expression instanceof LiteralExpressionImpl) {
```
rzwiefel: can't tell if this is better or worse. less complexity yes, but also i dont like the java not `!`. Anyways just posting it for comparison. No changes necessary
```java
if (!(expression instanceof LiteralExpressionImpl)) {
    return;
}

LiteralExpressionImpl literalExpression = (LiteralExpressionImpl) expression;
Object valueObj = literalExpression.getValue();
if (!(valueObj instanceof Geometry)) {
    return;
} 

Geometry geometry = (Geometry) valueObj;
Object userDataObj = geometry.getUserData();
if (!(userDataObj instanceof CoordinateReferenceSystem)) {
    return;
}

CoordinateReferenceSystem sourceCRS = (CoordinateReferenceSystem) userDataObj;
try {
    Geometry convertedGeometry = GeospatialUtil.transformToEPSG4326LonLatFormat(geometry, sourceCRS);
    literalExpression.setValue(convertedGeometry);
} catch (GeoFormatException e) {
    LOGGER.trace("Unable to convert geometry to EPSG:4326 format", e);
}
    
```

libs/geospatial/src/main/java/org/codice/ddf/libs/geo/util/GeospatialUtil.java
```diff
@@ -171,4 +199,75 @@ public static Double parseDMSLongitudeWithDecimalSeconds(String dmsLon)
...
+     * EPSG:4326
+     *
+     * @param geometry      - Geometry to transform
+     * @param sourceCrsName - Source geometry's coordinate reference system
+     * @return Geometry - Transformed geometry into EPSG:4326 lon/lat coordinate system
+     */
+    public static Geometry transformToEPSG4326LonLatFormat(Geometry geometry, String sourceCrsName)
+            throws GeoFormatException {
+        try {
+            CoordinateReferenceSystem sourceCrs = CRS.decode(sourceCrsName);
```
bdeining: sourceCrsName and geometry could be null here and throw an NPE

PR #1725, opened 2017-02-23
DDF-2831 Made retry intervals for Catalog Sources configurable

ahoffer:
#### What does this PR do?
 Change retry interval for unavailable endpoints
      The max retry interval can now be set in system properties and defaults to 300 seconds.
      The initial retry interval can be set in system properties and defaults to 10 seconds.
      The exponential behavior is noW doubling the interval instead of growing it by a factor of 10.
      The unit test includes a check to make sure no interval is longer than the max.
@stustison @brendan-hofmann @codymacdonald
@coyotesqrl
@stustison
Run unit and integration tests.

If a source is created and the endpoint is not available, DDF will periodically try to establish a connection. The problem is the retry period. The interval between retries grows exponentially, up to 420 minutes (7 hours).  Here is a table

||Retry #||Actual Timeout||
|1| 10 seconds|
|2| 2 minutes 45 seconds|
|3| 16m|
|4| 2 hours 45 minutes|
|5 |7 hours|

The system will re-try almost instantly (10 seconds), then retry after about 2 minutes, then 16 minutes, then again around 3 hours. After that, it retries the connection only every 7 hours. For example, if the endpoint for the source becomes available 20 minutes after the source is created, you will have to wait about 2 hours and 45 minutes for the endpoint to become available. 

The class that implements the re-try behavior is `UnavailableUrls`. It has two constant values, `INITIAL_TIMEOUT_SECONDS` and `MAX_TIMEOUT_SECONDS`. The most important one is `MAX_TIMEOUT_SECONDS`, because the exponential growth quickly reaches this value. Include  new system properties and let administrators change the values.

It is possible that this behavior is also responsible for the failures on Windows. If an itest is set to timeout at 10 minutes, but the endpoint is not responding 2 minutes and 45 seconds after the source is active, the test will always fail.

[DDF-2831](https://codice.atlassian.net/browse/DDF-2831)
N/A
#### Checklist:
- [x] Documentation Updated
- [x] Change Log Updated
- [x] Update / Add Unit Tests
- [x] Run OWASP

### DIFF
--- a/distribution/ddf-common/src/main/resources/etc/system.properties
+++ b/distribution/ddf-common/src/main/resources/etc/system.properties
@@ -143,6 +143,16 @@ org.codice.ddf.platform.requireAuditEncoding=false
 
 # Sets the default port number for the catalog-ftp feature FTP endpoint
 org.codice.ddf.catalog.ftp.port=8021
+#
+# Maximum Endpoint Contact Interval
+#
+# Max seconds between attempts to contact the endpoint of a source
+org.codice.ddf.platform.util.http.maxRetryInterval=300
+#
+# Initial Endpoint Contact Interval
+#
+# Initial seconds between attempts to contact the endpoint of a source
+org.codice.ddf.platform.util.http.initialRetryInterval=10
 
 
 #
--- a/distribution/docs/src/main/resources/_contents/_configuring/configuring-global-settings-contents.adoc
+++ b/distribution/docs/src/main/resources/_contents/_configuring/configuring-global-settings-contents.adoc
@@ -224,6 +224,29 @@ TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
 |org.apache.xerces.jaxp.DocumentBuilderFactoryImpl
 |Yes
 
+
+6+^h|Catalog Source Retry Interval
+
+|Initial Endpoint Contact Interval
+|org.codice.ddf.platform.util.http.initialRetryInterval
+|Integer
+|If a Catalog Source is unavailable, try to connect to it after the initial interval has elapsed.
+After every retry, the interval doubles, up to a given maximum interval.
+The interval is measured in seconds.
+|10
+|Yes
+
+|Maximum Endpoint Contact Interval
+|Maximum seconds between attempts to establish contact with unavailable Catalog Source.
+|Integer
+|Do not wait longer than the maximum interval to attempt to establish a connection with an
+unavailable Catalog Source. Smaller values result in more current information about the status
+ of Catalog Sources, but cause more network traffic.
+The interval is measured in seconds.
+|300
+|Yes
+
+
 6+^h|File Upload Settings
 
 |File extensions flagged as potentially dangerous to the host system or external clients
--- a/platform/util/platform-util-unavailableurls/pom.xml
+++ b/platform/util/platform-util-unavailableurls/pom.xml
@@ -12,7 +12,9 @@
  *
  **/
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>ddf.platform.util</groupId>
@@ -23,6 +25,12 @@
     <name>DDF :: Platform :: Util :: Unavailable Urls</name>
     <packaging>jar</packaging>
     <dependencies>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.16.1</version>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
--- a/platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
+++ b/platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
@@ -24,14 +24,24 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
+ * /**
  * A class to contain a set of unavailable URL's. They will be periodically checked
  * to determine if the URL has become available. If the URL is reachable, this implies
  * the web service is available, and the URL will be removed from the set.
  * <p>
- * NOTE: to avoid excessive network traffic, the period to wait before checking for a
+ * NOTE:
+ * to avoid excessive network traffic, the period to wait before checking for a
  * URL increases exponentially.
  */
 public class UnavailableUrls {
+
+    // Name of system property
+    static final String MAX_TIMEOUT_SECONDS_PROPERTY =
+            "org.codice.ddf.platform.util.http.maxRetryInterval";
+
+    // Name of system property
+    static final String INITIAL_TIMEOUT_SECONDS_PROPERTY =
+            "org.codice.ddf.platform.util.http.initialRetryInterval";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UnavailableUrls.class);
 
@@ -44,12 +54,6 @@ public class UnavailableUrls {
 
     // time to give the ping request before we stop trying
     private static final long PING_TIMEOUT_SECONDS = 10;
-
-    // initial timeout for checking a URL
-    private static final long INITIAL_TIMEOUT_SECONDS = 3;
-
-    // maximum timeout for checking a URL
-    private static final long MAX_TIMEOUT_SECONDS = TimeUnit.HOURS.toSeconds(7);
 
     private final ConcurrentHashMap<String, UrlChecker> checkers = new ConcurrentHashMap<>();
 
@@ -101,7 +105,7 @@ public class UnavailableUrls {
 
         private final String url;
 
-        private long timeout = INITIAL_TIMEOUT_SECONDS;
+        private long timeout = getInitialRetryInterval();
 
         /**
          * Let me check that URL for you!
@@ -116,15 +120,31 @@ public class UnavailableUrls {
         }
 
         /**
-         * Computes the next timeout, doesn't exceed MAX_TIMEOUT_SECONDS
+         * Computes the next timeout, doesn't exceed maximum internval
          */
         private void backoff() {
-            timeout = Math.min(timeout * INITIAL_TIMEOUT_SECONDS, MAX_TIMEOUT_SECONDS);
+            // Because the interval doubles, we check against getMaxRetryInterval() / 2.
+            // This prevents a timeout whose period is larger than MAX.
+            if (timeout >= getMaxRetryInterval() / 2) {
+                timeout = getMaxRetryInterval();
+            } else {
+                timeout += timeout;
+            }
         }
 
         private void schedule() {
             getScheduler().schedule(this, timeout, TimeUnit.SECONDS);
             backoff(); // the only reason to schedule is because of a failed ping
+        }
+
+        private int getMaxRetryInterval() {
+            String property = System.getProperty(MAX_TIMEOUT_SECONDS_PROPERTY);
+            return Integer.parseInt(property);
+        }
+
+        private int getInitialRetryInterval() {
+            String property = System.getProperty(INITIAL_TIMEOUT_SECONDS_PROPERTY);
+            return Integer.parseInt(property);
         }
 
         /**
--- a/platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
+++ b/platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
@@ -76,6 +76,23 @@ public class UnavailableUrls {
     }
 
     /**
+     * @return The maximum number of seconds to wait before pinging an unavailable URL
+     */
+    public int getMaxRetryInterval() {
+        String property = System.getProperty(MAX_TIMEOUT_SECONDS_PROPERTY, "300");
+        return Integer.parseInt(property);
+    }
+
+    /**
+     * @return The number of seconds to between the first and seconds attempts to
+     * contact the unavailable URL.
+     */
+    public int getInitialRetryInterval() {
+        String property = System.getProperty(INITIAL_TIMEOUT_SECONDS_PROPERTY, "10");
+        return Integer.parseInt(property);
+    }
+
+    /**
      * Add a URL to the set of unavailable URL's.
      * NOTE: the URL will automatically be removed when it can be reached.
      *
@@ -88,7 +105,7 @@ public class UnavailableUrls {
     /**
      * Does the set contain a specified URL.
      *
-     * @param url
+     * @param url of the endpoint to ping
      * @return the truth
      */
     public boolean contains(String url) {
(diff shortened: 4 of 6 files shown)


platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
```diff
@@ -116,17 +120,31 @@ public boolean contains(String url) {
...
-         * Computes the next timeout, doesn't exceed MAX_TIMEOUT_SECONDS
+         * Computes the next timeout, doesn't exceed maximum internval
          */
         private void backoff() {
-            timeout = Math.min(timeout * INITIAL_TIMEOUT_SECONDS, MAX_TIMEOUT_SECONDS);
+            if (timeout >= getMaxRetryInterval() / 2) {
+                timeout = getMaxRetryInterval();
+            } else {
+                timeout += timeout;
+            }
```
ahoffer: Change the algo a bit. The previous implementation multiplied the last timeout interval by 10 on each retry. This implementation doubles the interval every time.

platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
```diff
@@ -116,17 +120,31 @@ public boolean contains(String url) {
         }
 
         /**
-         * Computes the next timeout, doesn't exceed MAX_TIMEOUT_SECONDS
+         * Computes the next timeout, doesn't exceed maximum internval
          */
         private void backoff() {
-            timeout = Math.min(timeout * INITIAL_TIMEOUT_SECONDS, MAX_TIMEOUT_SECONDS);
+            if (timeout >= getMaxRetryInterval() / 2) {
```
ahoffer: Because the interval double, we check against `getMaxRetryInterval() / 2`. This prevents a timeout whose period is larger than MAX.

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
```diff
@@ -27,11 +32,17 @@
...
+
+    private static final String INITIAL_TIMEOUT_SECONDS = "10";
+
     private ScheduledExecutorService scheduler;
 
     @Before
     public void setUp() throws Exception {
         scheduler = Mockito.mock(ScheduledExecutorService.class);
+        System.setProperty(MAX_TIMEOUT_SECONDS_PROPERTY, MAX_TIMEOUT_SECONDS);
+        System.setProperty(INITIAL_TIMEOUT_SECONDS_PROPERTY, INITIAL_TIMEOUT_SECONDS);
```
ahoffer: Probably have to do this in some itests.

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
```diff
@@ -61,20 +72,6 @@ public void testInitialContains() {
         assertThat(set.contains("hello"), is(true));
     }
```
ahoffer: The code formatter moved the ScheduleCapture method around.

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
```diff
@@ -121,21 +118,21 @@ public void testUnsuccessfulPing() {
      * @param list - list of two or more positive integers
      * @return true if sequence is exponential, false otherwise
      */
-    private boolean isExponentialBackoff(long[] list) {
+    private boolean isExponentialBackoff(Long[] list) {
```
ahoffer: Switched over to object (Long) so I could use a better Hamcrest matcher. The matcher does not accept primitive arrays.

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
```diff
@@ -121,21 +118,21 @@ public void testUnsuccessfulPing() {
...
+        Long max = list[list.length - 1];
 
         if (!(base > 1)) {
             return false;
         }
 
         for (int i = 1; i < list.length - 1; i++) {
             double changeOfBase = Math.log(list[i]) / Math.log(base);
-            if (changeOfBase < i && list[i] != max) {
+            if (changeOfBase < i && !list[i].equals(max)) {
```
ahoffer: can't use `==` or `!=` with objects like Long.

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
```diff
@@ -181,6 +178,26 @@ public void testExponentialBackoff() {
         }
 
         assertThat(isExponentialBackoff(backoffs), is(true));
+        assertThat("List should not contain a value greater than the maximum cutoff value",
+                backoffs,
+                not(hasItemInArray(greaterThan(getMaxTimoutSeconds()))));
```
ahoffer: I had a bug where an interval larger than the max got into the array. This was a nice, non-intrusive place to stick a test for it.

distribution/docs/src/main/resources/_contents/_configuring/configuring-global-settings-contents.adoc
```diff
@@ -224,6 +224,30 @@ TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
...
+The interval doubles after each try, up to a given maximum interval.
+The interval is measured in seconds.
+|10
+|Yes
+
+|Maximum Endpoint Contact Interval
+|Maximum seconds between attempts to establish contact with unavailable Catalog Source.
+|Integer
+|Do not wait longer than the maximum interval to attempt to establish a connection with an
+unavailable Catalog Source. Large values result in more network traffic trying to contact
```
brendan-hofmann: This wording is confusing in my mind. It implies that a longer max retry interval will result in more total network traffic than a lower max retry interval.

platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
```diff
@@ -24,15 +24,25 @@
...
  * URL increases exponentially.
  */
 public class UnavailableUrls {
 
+    // Name of system property
+    static final String MAX_TIMEOUT_SECONDS_PROPERTY =
+            "org.codice.ddf.platform.util.http.maxRetryInterval";
+
+    // Name of system property
+    static final String INITIAL_TIMEOUT_SECONDS_PROPERTY =
```
brendan-hofmann: I'm not sure `min` is any more clear than `initial`, but I feel like symmetric wording helps make the relationship more clear to users.

platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
brendan-hofmann: This would be a good comment in the source code. 😉

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
brendan-hofmann: Modifying the environment in unit tests is not hermetic. You could get the initial values (if any) and reset them in your `@After`, but that's still not hermetic for the duration the test runs.

I would suggest refactoring the unit under test slightly so that you can inject a mock system properties map. This keeps the test hermetic without compromising the integrity of the production code.

platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
ahoffer: It really is _initial_. This is the interval before the first retry. The internal changes after every retry.

distribution/docs/src/main/resources/_contents/_configuring/configuring-global-settings-contents.adoc
ahoffer: That is beyond confusing. It looks just plain wrong.  I changed it.

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
ahoffer: I'd like to do that I looked around on the web and only found two options:
1. Use PowerMock to mock out the System class.
2. Clean up after each test with @After. 

I think option #2 is the lesser of two evils. Making the change now.

platform/util/platform-util-unavailableurls/pom.xml
```diff
@@ -24,6 +26,11 @@
     <packaging>jar</packaging>
     <dependencies>
         <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
```
coyotesqrl: ❓ What is this dependency being used for? Is it production or test code? If the latter, it should be marked as `scope` test.

❓ In either case, have we run OWASP over this module yet to ensure it has no vulnerabilities?

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -137,6 +137,16 @@ org.codice.allowBasicAuthOverHttp=false
 
 # Sets the default port number for the catalog-ftp feature FTP endpoint
 org.codice.ddf.catalog.ftp.port=8021
+#
+# Maximum Endpoint Contact Interval
+#
+# Max seconds between attempts to contact the endpoint of a source
+org.codice.ddf.platform.util.http.maxRetryInterval=300
```
coyotesqrl: ❗️  Based on your algorithm update, to double the initial retry value up to by not exceeding this max value, it would probably make more sense to set the default here to either 240 or 480 seconds (probably the former). We're never going to get to 300, even though I believe it sets an expectation we will based on this max value.

ahoffer: We can definitely set the max retry interval ot 480. Scott suggested 5 minutes and that is where 300 came from. I'm not sure what you mean that we are never going to get 300?  I ran the unit test `testIsExponentialBackoff` and the interval sequence looks like:
```
0 = {Long@1077} "10"
1 = {Long@1078} "20"
2 = {Long@1079} "40"
3 = {Long@1080} "80"
4 = {Long@1081} "160"
5 = {Long@1082} "300"
6 = {Long@1083} "300"
7 = {Long@1084} "300"
8 = {Long@1085} "300"
...
```

platform/util/platform-util-unavailableurls/pom.xml
ahoffer: The dependency is for testing. It lets me set system properties for a test, then restores the original values. I'll add the test scope directive.

I have not run OWASP. I'll do that right after I go through the comments.

ahoffer:
OWASP points out some access issues in the test class. But it is a test class and I'm not going to make the changes:

```
/Users/aaronhoffer/repos/connexta/ddf/ddf/platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java:203:31: warning: Variable 'runnable' must be private and have accessor methods.
/Users/aaronhoffer/repos/connexta/ddf/ddf/platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java:205:27: warning: Variable 'timeout' must be private and have accessor methods.
/Users/aaronhoffer/repos/connexta/ddf/ddf/platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java:207:31: warning: Variable 'unit' must be private and have accessor methods.
```

distribution/ddf-common/src/main/resources/etc/system.properties
coyotesqrl: My mistake. I read the increment code too quickly and thought it was taking _min_, not _max_. This is fine.

platform/util/platform-util-unavailableurls/pom.xml
ahoffer: OWASP is clean on the production code. It wants the Test class to use accessors in an inner class, but I don't feel it merits a code change.

```
/Users/aaronhoffer/repos/connexta/ddf/ddf/platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java:203:31: warning: Variable 'runnable' must be private and have accessor methods.
/Users/aaronhoffer/repos/connexta/ddf/ddf/platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java:205:27: warning: Variable 'timeout' must be private and have accessor methods.
/Users/aaronhoffer/repos/connexta/ddf/ddf/platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java:207:31: warning: Variable 'unit' must be private and have accessor methods.
```

platform/util/platform-util-unavailableurls/src/test/java/org/codice/ddf/platform/util/http/UnavailableUrlsTest.java
```diff
@@ -13,25 +13,48 @@
...
+            MAX_TIMEOUT_SECONDS_PROPERTY,
+            MAX_TIMEOUT_SECONDS);
+
     private ScheduledExecutorService scheduler;
 
     @Before
     public void setUp() throws Exception {
         scheduler = Mockito.mock(ScheduledExecutorService.class);
+        //        System.setProperty(MAX_TIMEOUT_SECONDS_PROPERTY, MAX_TIMEOUT_SECONDS);
+        //        System.setProperty(INITIAL_TIMEOUT_SECONDS_PROPERTY, INITIAL_TIMEOUT_SECONDS);
```
emanns95: Do we want to just delete this?

platform/util/platform-util-unavailableurls/pom.xml
clockard: Do we want to pull in a new dependency for this? Seems easy enough to set the system property in the setup method and then revert it in a teardown method after the test.

ahoffer: I can pull it out if it isn't in the best interest of the code base
I thought it was elegant, but it isn't necessary.

coyotesqrl: I'm personally indifferent to adding a new dependency that passes OWASP and is scoped only for testing. That is a very low-risk operation.

ahoffer: Leaving it in the POM.

platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
```diff
@@ -116,17 +120,33 @@ public boolean contains(String url) {
...
+            }
         }
 
         private void schedule() {
             getScheduler().schedule(this, timeout, TimeUnit.SECONDS);
             backoff(); // the only reason to schedule is because of a failed ping
         }
 
+        private int getMaxRetryInterval() {
+            String property = System.getProperty(MAX_TIMEOUT_SECONDS_PROPERTY);
```
clockard: Consider providing a default value with System.getProperty(MAX_TIMEOUT_SECONDS_PROPERTY, <default-time-out>) in case the system property isn't set.

platform/util/platform-util-unavailableurls/src/main/java/org/codice/ddf/platform/util/http/UnavailableUrls.java
```diff
@@ -116,17 +120,33 @@ public boolean contains(String url) {
...
+            return Integer.parseInt(property);
+        }
+
+        private int getInitialRetryInterval() {
+            String property = System.getProperty(INITIAL_TIMEOUT_SECONDS_PROPERTY);
```
clockard: Consider using a default here too.

emanns95:
Passed full build and tests 👍

PR #1727, opened 2017-02-23
DDF-2816 Update attribute views to convert link-like values to links

andrewkfiedler:
#### What does this PR do?
 - The details view in the non-editing state will now attempt to convert link-like text fields to links.
 - The table view will also do so.
 - Adds a handlebars helper for determining if a text field is a link.
 - Updates UX for multi-values on both the details view and the table view.  The table view now puts each individual value on it's own line.  It also sets a max width per line.  The details view now shows each value on their own line when in non-edit, and uses a seperator when editing.  Tooltips have been added to assist users in scenarios where overflow occurs.
 - Moves the custom selection to be close to the top for bulk input selects to facilitate visibility on edits with a large range of values.
 - Adds a style sheet for links.
 - Adds some help to the input-bulk view and support for help to the select view.
 - Removes unused css from searchPage.css

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@andrewkfiedler
@clockard 
@pklinef
@djblue 

#### How should this be tested? (List steps with links to updated documentation)
Ensure that link values become links in the UI.  Test out that the changes to multi valued attributes don't regress anything.  I've tested all the supported browsers, but it wouldn't hurt to have a second look by someone.  

#### What are the relevant tickets?
https://codice.atlassian.net/browse/DDF-2816

#### Screenshots
screen shot 2017-02-23 at 3 34 38 pm
screen shot 2017-02-23 at 3 34 32 pm
screen shot 2017-02-23 at 3 34 17 pm
screen shot 2017-02-23 at 3 34 05 pm
screen shot 2017-02-23 at 3 33 52 pm
screen shot 2017-02-23 at 3 33 37 pm
screen shot 2017-02-23 at 3 33 06 pm

PR #1728, opened 2017-02-23
DDF-2816 Update attribute views to convert link-like values to links

andrewkfiedler:
backport of https://github.com/codice/ddf/pull/1727


clockard:
Heroed and and manually verified :shipit:

PR #1729, opened 2017-02-26
DDF-2814 Fix SolrFilterDelegate to not convert polygons into simpler shapes

clockard:
#### What does this PR do?
Fixes a problem where simple polygons (like rectangles) were getting converted into simpler WKT shapes that were not recognized by other classes.
@brendan-hofmann 
@lessarderic
@shaundmorris
Verify unit and itests pass on DDF and downstream projects
[DDF-2814](https://codice.atlassian.net/browse/DDF-2814)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
@@ -38,6 +38,7 @@ import org.locationtech.spatial4j.distance.DistanceUtils;
 import org.locationtech.spatial4j.exception.InvalidShapeException;
 import org.locationtech.spatial4j.io.ShapeReader;
 import org.locationtech.spatial4j.shape.Shape;
+import org.locationtech.spatial4j.shape.jts.JtsGeometry;
 import org.opengis.filter.sort.SortBy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -666,6 +667,11 @@ private Geometry getGeometry(String wkt) {
     private String sanitizeGeometry(String wkt) {
         try {
             Shape wktShape = WKT_READER.read(wkt);
+            //All polygons will be an instance of JtsGeometry. If it is not a polygon we don't need
```
lessarderic: ❓ Should we rename the method or add a comment on it to make its purpose clearer?

catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
```diff
@@ -109,6 +120,29 @@ public void nonIntersectingPolygon() {
...
+        String wkt = "POLYGON ((35 10, 45 45, 15 40, 10 20, 35 10), (20 30, 35 35, 30 20, 20 30))";
+        stub(mockResolver.getField("testProperty", AttributeFormat.GEOMETRY, false)).toReturn(
+                "testProperty_geohash_index");
+        SolrQuery query = toTest.contains("testProperty", wkt);
+        assertThat(query.getQuery(),
+                startsWith(
+                        "testProperty_geohash_index:\"Contains(POLYGON ((35 10, 45 45, 15 40, 10 20, 35 10), (20 30, 35 35, 30 20, 20 30)))\""));
+    }
+
+    @Test
```
lessarderic: ❓ Should we also have tests for `MULTIPOLYGONS` that contain a self-intersecting polygon and a polygon with a hole?

clockard:
Local DDF build with all itests passed. Downstream itests passed as well.

catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
clockard: @lessarderic I tried to add a unit test for MULTIPOLYGON where one of the polygons was self intersecting but it looks like in this case the polygon is not fixed but instead cause a intersection error. Looks like the fix only applies to the POLYGON and not to the MULTIPOLYGON which is a little odd since you would think that the individual polygons of the MULTIPOLYGON would go through the same code as the POLYGON

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
brendan-hofmann: I am also wondering if the reverse is necessarily true? i.e. `All polygons are JtsGeometry` does not imply `All JtsGeometry are polygons`.

Is there a source for this fix, maybe on the mailing lists or something? Agree with @lessarderic that we should be as clear as possible here.

clockard: Renamed the method to make it clear what this method does

catalog/core/catalog-core-solr/src/test/java/ddf/catalog/source/solr/SolrFilterDelegateTest.java
clockard: Right now I don't think intersecting multi-polygons are too big of an issue since they can't be generated from the ui.

PR #1733, opened 2017-02-27
DDF-2807: Fixes documentation profile activation bug

vinamartin:
#### What does this PR do?
Changes the documentation profile to be activated when the property “skipDocs” is false or not provided. (Reference an example here: http://maven.apache.org/guides/introduction/introduction-to-profiles.html). This was done to prevent issues with profile activation that were happening. 
From the Maven website, emphasis mine:
_[A profile using `<activeByDefault>`] will automatically be active for all builds unless another profile **in the same POM** is activated using one of the previously described methods. All profiles that are active by default are **automatically deactivated** when a profile in the POM is activated on the command line or through its activation config._

By changing it to property based activation, the profile will still run by default without being deactivated by other profiles and can also be skipped with `-DskipDocs=true` when running a build. 

Backport of https://github.com/codice/ddf/pull/1696

@coyotesqrl
@lessarderic

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,13 +116,13 @@ The distribution will be available under "distribution/ddf/target" directory.
 To build DDF using the [parallel builds feature of maven](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3), Run the following command:
 
 ```
-mvn install -T 8 -P\!documentation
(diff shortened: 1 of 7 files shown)


clockard:
Verified build runs with all the new options on and with them off :shipit:

PR #1736, opened 2017-02-27
DDF-2835: Quick fix for CSW sortby

AzGoalie:
#### What does this PR do?
Quick fix to allow csw queries to search when the sortby property isn't defined in the csw output schema.
@ahoffer @emanns95 
@coyotesqrl
@pklinef
Ingest some data and create a CSW endpoint and query it using 'effective' as the sort by property.
Verify that results are returned.
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
@@ -51,6 +51,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.SAXException;
 
+import ddf.catalog.data.AttributeRegistry;
 import ddf.catalog.data.MetacardType;
 import ddf.catalog.data.types.Core;
 import ddf.catalog.filter.FilterAdapter;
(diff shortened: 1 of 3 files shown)


pklinef:
@AzGoalie Please backport this PR to 2.10.x.

AzGoalie:
Successfully heroed.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -120,7 +120,10 @@ public QueryRequest getQuery(GetRecordsType request) throws CswException {
 
         CswRecordMapperFilterVisitor filterVisitor = buildFilter(query.getConstraint());
         QueryImpl frameworkQuery = new QueryImpl(filterVisitor.getVisitedFilter());
-        frameworkQuery.setSortBy(buildSort(query.getSortBy()));
+
+        if (CswConstants.CSW_OUTPUT_SCHEMA.equals(request.getOutputSchema())) {
```
kcwire: Rather than check if this is the right output schema - I think it would be more precise to check if the AttributeRegistry.  The AttributeRegistry will contain all attributes available within the system. If it's not present we should probably return an error rather than fail quietly.

AzGoalie:
Successfully passed heroing.

PR #1747, opened 2017-03-02
DDF-2842 Fixed coordinate order for CSW Federated Sources

bdeining:
#### What does this PR do?
This PR fixes an issue with passing context information (coordinateOrder) to the CswRecordConverter and adds a unit test to prevent regression.  Additionally it adds more robust support for CRS conversions when parsing CSW responses.
@clockard @glenhein 
@troymohl 
@jlcsmith
@kcwire
Build / Install/ Configure various CSW sources ( some with LAT / LON vs LON / LAT) and verify results.
This PR relies on this PR for GeospatialUtils : https://github.com/codice/ddf/pull/1712
[DDF-2842](https://codice.atlassian.net/browse/DDF-2842)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
+++ b/catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
@@ -14,16 +14,24 @@
 package org.codice.ddf.spatial.ogc.csw.catalog.common;
 
 import org.apache.commons.lang.StringUtils;
+import org.codice.ddf.libs.geo.GeoFormatException;
+import org.codice.ddf.libs.geo.util.GeospatialUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.thoughtworks.xstream.io.HierarchicalStreamReader;
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.GeometryFactory;
+import com.vividsolutions.jts.io.ParseException;
+import com.vividsolutions.jts.io.WKTReader;
+import com.vividsolutions.jts.io.WKTWriter;
 
 /**
  * Parses OWS Bounding Box geometry XML and converts it to WKT.
- *
+ * <p>
  * Bounding Box XML is of the form:
- *
+ * <p>
  * <pre>
  *  {@code
  *  <ows:BoundingBox crs="urn:x-ogc:def:crs:EPSG:6.11:4326">
@@ -34,20 +42,32 @@ import com.thoughtworks.xstream.io.HierarchicalStreamReader;
  * </pre>
  *
  * @author rodgersh
- *
  */
 public class BoundingBoxReader {
     private static final transient Logger LOGGER = LoggerFactory.getLogger(BoundingBoxReader.class);
 
     private static final String SPACE = " ";
 
+    private static final GeometryFactory GEOMETRY_FACTORY = new GeometryFactory();
+
+    private static final ThreadLocal<WKTWriter> WKT_WRITER_THREAD_LOCAL =
+            new ThreadLocal<WKTWriter>() {
+                @Override
+                protected WKTWriter initialValue() {
+                    return new WKTWriter();
+                }
+            };
+
     private HierarchicalStreamReader reader;
 
     private CswAxisOrder cswAxisOrder;
 
+    private WKTReader wktReader;
+
     public BoundingBoxReader(HierarchicalStreamReader reader, CswAxisOrder cswAxisOrder) {
         this.reader = reader;
         this.cswAxisOrder = cswAxisOrder;
+        wktReader = new WKTReader(GEOMETRY_FACTORY);
     }
 
     public String getWkt() throws CswException {
@@ -62,11 +82,10 @@ public class BoundingBoxReader {
         }
 
         String crs = reader.getAttribute("crs");
-        // TODO: CRS value determines order of bounding box values, i.e.,
-        // WGS-84 specifies lon,lat order
-        // EPSG:4326 specifies lat,lon order
-        // For now we assume lat,lon order
-        LOGGER.debug("crs = {}", crs);
+        // For now, default an empty CRS to EPSG:4326
+        crs = StringUtils.isEmpty(crs) ? "urn:ogc:def:crs:EPSG:6.11:4326" : crs;
+
+        LOGGER.debug("crs = {}, coordinate order = {}", crs, this.cswAxisOrder);
 
         // Move down to the first child node of <BoundingBox>, which should
         // be the <LowerCorner> tag
@@ -77,7 +96,7 @@ public class BoundingBoxReader {
         if (reader.getNodeName()
                 .contains("LowerCorner")) {
             String value = reader.getValue();
-            lowerCornerPosition = getCoordinates(value, this.cswAxisOrder);
+            lowerCornerPosition = getCoordinates(value);
         }
 
         // Move back up to the <BoundingBox> parent tag
@@ -92,12 +111,11 @@ public class BoundingBoxReader {
         if (reader.getNodeName()
                 .contains("UpperCorner")) {
             String value = reader.getValue();
-            upperCornerPosition = getCoordinates(value, this.cswAxisOrder);
+            upperCornerPosition = getCoordinates(value);
         }
 
         // If both corner positions parsed, then compute other 2 corner
-        // positions of the
-        // bounding box and create the WKT for the bounding box.
+        // positions of the bounding box and create the WKT for the bounding box.
         if (lowerCornerPosition != null && lowerCornerPosition.length == 2
                 && upperCornerPosition != null && upperCornerPosition.length == 2) {
 
@@ -111,17 +129,19 @@ public class BoundingBoxReader {
                     "BoundingBoxReader.getWkt(): could not find either LowerCorner or UpperCorner tags.");
         }
 
-        // Move position back up to the parent <BoundingBox> tag, where we
-        // started
+        // Move position back up to the parent <BoundingBox> tag, where we started
         reader.moveUp();
 
-        LOGGER.debug("Returning WKT in LON/LAT coord order: {}.", wkt);
+        if (!isEPSG4326(crs) || (cswAxisOrder.equals(CswAxisOrder.LAT_LON) && isEPSG4326(crs))) {
+            wkt = convertToEPSG4326(wkt, crs);
+        }
+        LOGGER.debug("Returning WKT {}.", wkt);
         return wkt;
     }
 
     /**
      * We want to create WKT in LON/LAT order.
-     *
+     * <p>
      * Points for bounding box will be computed starting with the lower corner, and proceeding in a
      * clockwise rotation. So lower corner point would be point 1, upper corner point would be point
      * 3, and points 2 and 4 would be computed.
@@ -130,9 +150,9 @@ public class BoundingBoxReader {
         LOGGER.debug("Creating WKT in LON/LAT coordinate order.");
         if (upperCornerPosition[0].equals(lowerCornerPosition[0]) && upperCornerPosition[1].equals(
                 lowerCornerPosition[1])) {
-            return "POINT(" + lowerCornerPosition[0] + SPACE + lowerCornerPosition[1] + ")";
+            return "POINT (" + lowerCornerPosition[0] + SPACE + lowerCornerPosition[1] + ")";
         }
-        return "POLYGON((" + lowerCornerPosition[0] + SPACE + lowerCornerPosition[1] + ", "
+        return "POLYGON ((" + lowerCornerPosition[0] + SPACE + lowerCornerPosition[1] + ", "
                 + upperCornerPosition[0] + SPACE + lowerCornerPosition[1] + ", "
                 + upperCornerPosition[0] + SPACE + upperCornerPosition[1] + ", "
                 + lowerCornerPosition[0] + SPACE + upperCornerPosition[1] + ", "
@@ -140,30 +160,49 @@ public class BoundingBoxReader {
     }
 
     /**
-     * @param coords
-     *            The latitude and longitude coordinates (in no particular order).
-     * @param axisOrder
-     *            The order that the axes are represented.
-     * @return The coordinates in LON/LAT order.
+     * @param coords The latitude and longitude coordinates (in no particular order).
+     * @return The coordinates in an array
      */
-    private String[] getCoordinates(String coords, CswAxisOrder axisOrder) {
+    private String[] getCoordinates(String coords) {
+        return coords.split(SPACE);
+    }
 
-        switch (axisOrder) {
-        case LAT_LON: {
-            /**
-             * We want to create WKT in LON/LAT order. Since the response has the coords in LAT/LON
-             * order, we need to reverse them.
-             */
-            return StringUtils.reverseDelimited(coords, SPACE.charAt(0))
-                    .split(SPACE);
+    private String convertToEPSG4326(String wkt, String crs) throws CswException {
+        try {
+            Geometry geometry = wktReader.read(wkt);
+            Geometry convertedGeometry = GeospatialUtil.transformToEPSG4326LonLatFormat(geometry,
+                    crs);
+
+            if (convertedGeometry != null && withinEPSG4326Bounds(convertedGeometry)) {
+                return WKT_WRITER_THREAD_LOCAL.get()
+                        .write(convertedGeometry);
+            } else {
+                throw new GeoFormatException();
+            }
+        } catch (ParseException | GeoFormatException e) {
+            LOGGER.debug("Unable to convert {} from {} to EPSG:4326", wkt, crs, e);
+            throw new CswException(String.format("Unable to convert %s from %s to EPSG:4326",
+                    wkt,
+                    crs));
         }
-        default: {
-            /**
-             * We want to create WKT in LON/LAT order. Since this is the order of the coords in the
-             * response, we use them as-is.
-             */
-            return coords.split(SPACE);
+    }
+
+    private boolean isEPSG4326(String crs) {
+        return (crs.contains("4326") || crs.contains("CRS84"));
+    }
+
+    private boolean withinEPSG4326Bounds(Geometry geometry) {
+        Coordinate[] coordinates = geometry.getCoordinates();
+        for (Coordinate coordinate : coordinates) {
+            // Longitude
+            if (coordinate.x <= -180 || coordinate.x >= 180) {
+                return false;
+            }
+            // Latitude
+            if (coordinate.y <= -90 || coordinate.y >= 90) {
+                return false;
+            }
         }
-        }
+        return true;
     }
 }
--- a/catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
+++ b/catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
@@ -48,6 +48,8 @@ public class BoundingBoxReader {
 
     private static final String SPACE = " ";
 
+    private static final String DEFAULT_CRS = "urn:ogc:def:crs:EPSG:4326";
+
     private static final GeometryFactory GEOMETRY_FACTORY = new GeometryFactory();
 
     private static final ThreadLocal<WKTWriter> WKT_WRITER_THREAD_LOCAL =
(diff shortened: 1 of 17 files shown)


catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -115,7 +123,8 @@ public String getWkt() throws CswException {
         // started
         reader.moveUp();
 
-        LOGGER.debug("Returning WKT in LON/LAT coord order: {}.", wkt);
+        wkt = convertToEPSG4326(wkt, crs);
```
kcwire: I think we only want to convert the ordering *IF* we need to (don't bother reading & writing the WKT if it's already in lon/lat)

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -166,4 +175,19 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
...
+    private String convertToEPSG4326(String wkt, String crs) {
+
+        try {
+            Geometry geometry = wktReader.read(wkt);
+            Geometry convertedGeometry = GeospatialUtil.transformToEPSG4326LonLatFormat(geometry, crs);
+            if (convertedGeometry != null) {
+                return wktWriter.write(convertedGeometry);
+            }
+        } catch (ParseException | GeoFormatException e) {
+            LOGGER.debug("Unable to convert {} from {} to EPSG:4326", wkt, crs, e);
```
kcwire: if we can't parse the WKT what are the consequences? will this blow up in the CatalogProvider? - seems like we are failing quietly here which could push the issue further down the call stack (and be more difficult to debug)

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -166,4 +175,19 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
         }
         }
     }
+
+    private String convertToEPSG4326(String wkt, String crs) {
```
kcwire: It'd be nice to see some unit tests (positive and negative for the WKT parsing)

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
kcwire: I'm not sure how the configuration option in CSW config would cause a change in behavior, should we remove the config option and handle it all internally?

kcwire:
It would be nice if we could add an itest for this since it has regressed over time.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: There is no guarantee of the ordering for EPSG:4326 coming in, so we cannot handle this internally.  Thus, we should still allow the Admin to specify the ordering in the CSW Source config.  Here's a good read : http://docs.geotools.org/latest/userguide/library/referencing/order.html

bdeining: The only time we wouldn't want to convert the incoming geometry is when the CRS is EPSG:4326 and the ordering is lon / lat.  The issue here is that there are multiple CRSs for EPSG:4326 ( `urn:ogc:def:crs:OGC:1.3:CRS84`, `EPSG:4326`, `urn:x-ogc:def:crs:EPSG:6.11:4326`)

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: As it stands now, geometries that come in from a CSW Source that are not EPSG:4326 ( or the lat / lon is not properly ordered), there will be exceptions further down in the stack.  It seems like the best thing to do is to not set the geometry if the WKT cannot be parsed.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
troymohl: I think this should check the `cswAxisOrder` and only call that conversion method if it's in lat/lon.

bdeining: What about the case where we are converting from EPSG:XXXX that is in lon / lat to EPSG:4236 in lon / lat?

troymohl: The underlying conversion will swap the coordinates, so I think we should have some check to not call it if it's already EPSG:4326 & lon/lat ordered.

bdeining: Added checks for EPSG:4326 & lon/lat ordered

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: Done

bdeining:
@kcwire The added unit test should be sufficient to cover the case where the context is not passed correctly to the `CswRecordConverter` from the transform provider.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -45,9 +51,15 @@
...
     private CswAxisOrder cswAxisOrder;
 
+    private WKTReader wktReader;
+
+    private WKTWriter wktWriter;
+
     public BoundingBoxReader(HierarchicalStreamReader reader, CswAxisOrder cswAxisOrder) {
         this.reader = reader;
         this.cswAxisOrder = cswAxisOrder;
+        wktWriter = new WKTWriter();
```
jlcsmith: Don't we need to be careful about the (lack of) thread safety of the writer?  https://github.com/codice/ddf/blob/08651d5cc59e559b940b4faaec6334657948c18c/catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/transformer/GmdTransformer.java#L101

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -130,40 +144,56 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
...
+                throw new GeoFormatException();
+            }
+        } catch (ParseException | GeoFormatException e) {
+            LOGGER.debug("Unable to convert {} from {} to EPSG:4326", wkt, crs, e);
+            throw new CswException(String.format("Unable to convert %s from %s to EPSG:4326", wkt, crs));
         }
+    }
+
+    private boolean isEPSG4326(String crs) {
+        return (crs.equals("urn:x-ogc:def:crs:EPSG:6.11:4326") || crs.contains("urn:ogc:def:crs:OGC:1.3:CRS84"));
```
jlcsmith: We need to find a better way to check for EPSG 4326.  Your check here is too restrictive.  "EPSG:6.11" means version 6.11 of the EPSG database, and we don't really care about the epsg db version.  Also, I'm wondering if the CRS has to be URN?  Can it just be the code (EPSG:4326)?

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
pklinef: Here is an example of using WKTReader in a thread safe way.

https://github.com/codice/ddf/blob/08651d5cc59e559b940b4faaec6334657948c18c/catalog/core/catalog-core-commons/src/main/java/ddf/util/WktStandard.java#L68

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
jlcsmith: It is possible the JTS Crs/CrsAuthority utilities could be used here.
Otherwise, we might simply need to look for "4326" or "CRS84" strings to identify this as a 4326 CRS.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: Fixed

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: I wasn't able to find anything in JTS that would help us determine if a CRS is some form of EPSG:4326, so the check has been relaxed per @jlcsmith's guidance

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -62,11 +82,10 @@ public String getWkt() throws CswException {
         }
 
         String crs = reader.getAttribute("crs");
-        // TODO: CRS value determines order of bounding box values, i.e.,
-        // WGS-84 specifies lon,lat order
-        // EPSG:4326 specifies lat,lon order
-        // For now we assume lat,lon order
-        LOGGER.debug("crs = {}", crs);
+        // For now, default an empty CRS to EPSG:4326
+        crs = StringUtils.isEmpty(crs) ? "urn:x-ogc:def:crs:EPSG:6.11:4326" : crs;
```
jlcsmith: any idea what the difference is between x-ogc and ogc?  The EPSG database just uses "ogc".  See http://www.epsg-registry.org/report.htm?type=selection&entity=urn:ogc:def:crs:EPSG::4326&reportDetail=long&style=urn:uuid:report-style:default-with-urn&style_name=OGP%20Default%20With%20Urn&title=

bdeining: `x-ogc` is actually experimental.  Changed to regular `ogc`

jlcsmith: thanks for the clarification

jlcsmith:
build now

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -62,11 +82,10 @@ public String getWkt() throws CswException {
         }
 
         String crs = reader.getAttribute("crs");
-        // TODO: CRS value determines order of bounding box values, i.e.,
-        // WGS-84 specifies lon,lat order
-        // EPSG:4326 specifies lat,lon order
-        // For now we assume lat,lon order
-        LOGGER.debug("crs = {}", crs);
+        // For now, default an empty CRS to EPSG:4326
+        crs = StringUtils.isEmpty(crs) ? "urn:ogc:def:crs:EPSG:6.11:4326" : crs;
```
glenhein: Can you make this string a static final, like DEFAULT_CRS?

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -111,17 +129,19 @@ public String getWkt() throws CswException {
                     "BoundingBoxReader.getWkt(): could not find either LowerCorner or UpperCorner tags.");
         }
 
-        // Move position back up to the parent <BoundingBox> tag, where we
-        // started
+        // Move position back up to the parent <BoundingBox> tag, where we started
         reader.moveUp();
 
-        LOGGER.debug("Returning WKT in LON/LAT coord order: {}.", wkt);
+        if (!isEPSG4326(crs) || (cswAxisOrder.equals(CswAxisOrder.LAT_LON) && isEPSG4326(crs))) {
```
glenhein: Isn't the second `isEPSG4326(crs)` redundant? If it is FALSE, the first half of the OR is TRUE. If it is TRUE, then the second half of the OR evaluates to 'x && TRUE' which is 'x'.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -130,40 +150,59 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
...
+                    crs);
+
+            if (convertedGeometry != null && withinEPSG4326Bounds(convertedGeometry)) {
+                return WKT_WRITER_THREAD_LOCAL.get()
+                        .write(convertedGeometry);
+            } else {
+                throw new GeoFormatException();
+            }
+        } catch (ParseException | GeoFormatException e) {
+            LOGGER.debug("Unable to convert {} from {} to EPSG:4326", wkt, crs, e);
```
glenhein: It always feels weird to have a log and throw at the same time. Can you chain `e` to `CswException`?

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -130,40 +150,59 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
...
+            throw new CswException(String.format("Unable to convert %s from %s to EPSG:4326",
+                    wkt,
+                    crs));
         }
+    }
+
+    private boolean isEPSG4326(String crs) {
+        return (crs.contains("4326") || crs.contains("CRS84"));
```
glenhein: The outer parens are not needed.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -130,40 +150,59 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
...
+
+    private boolean withinEPSG4326Bounds(Geometry geometry) {
+        Coordinate[] coordinates = geometry.getCoordinates();
+        for (Coordinate coordinate : coordinates) {
+            // Longitude
+            if (coordinate.x <= -180 || coordinate.x >= 180) {
+                return false;
+            }
+            // Latitude
+            if (coordinate.y <= -90 || coordinate.y >= 90) {
```
glenhein: So, you can't have a point right at the north/south pole?

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
kcwire: You should remove version number `6.11` if we are making this a default.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
```diff
@@ -130,40 +150,59 @@ private String createWkt(String[] lowerCornerPosition, String[] upperCornerPosit
...
+        try {
+            Geometry geometry = wktReader.read(wkt);
+            Geometry convertedGeometry = GeospatialUtil.transformToEPSG4326LonLatFormat(geometry,
+                    crs);
+
+            if (convertedGeometry != null && withinEPSG4326Bounds(convertedGeometry)) {
+                return WKT_WRITER_THREAD_LOCAL.get()
+                        .write(convertedGeometry);
+            } else {
+                throw new GeoFormatException();
```
troymohl: Might be nice to include some details in the exception on why it's being thrown ("Converted Geometry is not set, or not within EPSG4326 bounds")

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: Done

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: removed redundancy

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: removed log statement

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: nice catch, took away `=` for these comparisons (as we should be able to have points there)

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: removed

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/BoundingBoxReader.java
bdeining: added some details

bdeining:
build now

PR #1748, opened 2017-03-02
DDF-2843: Add incremental builder extension to speed up PR builds

lessarderic:
#### What does this PR do?
Adds the `gitflow-incremental-builder` build extension to the root `pom.xml` to support local and PR incremental builds.

@oconnormi 
@LinkMJB 

@coyotesqrl
@shaundmorris

1. Run `mvn clean install` and make sure a normal build is executed.
1. Make a local change to a source file, run `mvn -Dgib.enaled=true clean install` and make sure only the module that was changed and all its dependent modules are built

[DDF-2843](https://codice.atlassian.net/browse/DDF-2843)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,12 @@
         <jodah-failsafe.version>0.9.5</jodah-failsafe.version>
         <apache.shiro.version>1.3.2</apache.shiro.version>
         <spark.version>2.5.5</spark.version>
+
+        <!-- gitflow-incremental-builder -->
+        <!-- See https://github.com/vackosar/gitflow-incremental-builder for a list of all the options -->
+        <gib.referenceBranch>refs/remotes/origin/master</gib.referenceBranch>
+        <gib.baseBranch>HEAD</gib.baseBranch>
+        <gib.enabled>false</gib.enabled>
     </properties>
     <!--
     NOTE: The properties ddf.scm.connection.url, snapshots.repository.url and releases.repository.url should be defined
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -176,6 +176,12 @@
         <jodah-failsafe.version>0.9.5</jodah-failsafe.version>
         <apache.shiro.version>1.3.2</apache.shiro.version>
         <spark.version>2.5.5</spark.version>
+
+        <!-- gitflow-incremental-builder -->
+        <!-- See https://github.com/vackosar/gitflow-incremental-builder for a list of all the options -->
+        <gib.referenceBranch>refs/remotes/origin/master</gib.referenceBranch>
```
coyotesqrl: ❓ How does this work with backport branches? We need simply to change the reference branch, right?

lessarderic: Yes. I used master as the reference branch by default so that most of the time people don't have to worry about it. But you can change it if you use `-Dgib.referenceBranch=...`, which would allow us to use this for release branches as well.

lessarderic: One thing though. This assumes that you use `origin` as your remote name. Don't know if there's a better way to refer to `codice/ddf/master` that doesn't use the remote name.

coyotesqrl: I do know several engineers who do things oddly, making their forks origin. I'm not sure if they at least stick with the name `origin`. This might require some documentation and a little dissemination of usage.

oconnormi: I almost never use origin :-) usually i have my remotes named for the org or user that the fork belongs to (i like to be able to see where i'm pushing immediately). It looks like this would be fine though since the reference branch property has the remote name in it

shaundmorris: Should we account for this somehow as we do minor releases? So that we remember to change it as we make new minor branches.

lessarderic: My thinking here @shaundmorris was to use a default that would work for people most of the time if and when they use the incremental builder.

For our Jenkins jobs/pipelines, the job should always specify this value on the command line and override this default value. This way we never have to update the pom file.

We could always consider using `libero` or our release scripts to update this value for us when we create other branches if we feel it would be useful. But that wouldn't help with the remote name so even that wouldn't be a 100% solution.

PR #1760, opened 2017-03-06
DDF-2849 CQL endpoint now respects Accept-Encoding header

rzwiefel:
#### What does this PR do?
see title
@djblue @garrettfreibott 
@kcwire
@pklinef 
send a request to cql endpoint and ensure the response is not gzipped
https://codice.atlassian.net/browse/DDF-2849

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -14,10 +14,12 @@
 package org.codice.ddf.catalog.ui.query;
 
 import static spark.Spark.after;
+import static spark.Spark.before;
 import static spark.Spark.exception;
 import static spark.Spark.post;
 import static spark.route.RouteOverview.enableRouteOverview;
 
+import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
@@ -27,6 +29,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.ws.rs.NotAcceptableException;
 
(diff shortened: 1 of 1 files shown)


pklinef:
It looks like Spark is suppose to be doing this.  Is this actually a big in Spark?

http://sparkjava.com/documentation.html#gzip
https://github.com/perwendel/spark/blob/4a6a3888d89c0070268d827d060f660209bdc07b/src/main/java/spark/utils/GzipUtils.java#L65-L74

pklinef:
Maybe this is the related issue on the Spark side.

https://github.com/perwendel/spark/issues/691

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -93,6 +123,62 @@ public void init() {
...
+     * @return Map of {Encoding, Q value}
+     */
+    private Map<String, String> parseAcceptEncodings(String header) {
+        if (StringUtils.isEmpty(header)) {
+            return Collections.emptyMap();
+        }
+        String[] encodings = commaSpace.split(header.trim());
+        return Arrays.stream(encodings)
+                .map(semicolon::split)
+                .collect(Collectors.toMap(v -> v[0],
```
pklinef: Do we need to check length before accessing the first index in case we are sent a malicious or malformed header?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -74,7 +86,25 @@ public void init() {
...
-            res.header("Content-Encoding", "gzip");
+
+            // Must manually check and set header for gzip because of spark issue
+            // https://github.com/perwendel/spark/issues/691
+            Map<String, String> acceptEncodings =
+                    parseAcceptEncodings(req.headers("Accept-Encoding"));
+
+            if (acceptEncodingSupports(acceptEncodings, "gzip")) {
+                res.header("Content-Encoding", "gzip");
+            } else if (acceptEncodingSupports(acceptEncodings, "identity")) {
```
pklinef: Technically `identity` could have a higher `qvalue` than `gzip` which would make this `else if` invalid.  Not sure why anyone would want to do that though.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
rzwiefel: I don't believe so because at this point we have confirmed the header is not empty (line 135) and the Pattern `split` method will always return an array with a single item in it even if there is no delimiter token found in the string. 

http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#split(java.lang.CharSequence,%20int)
>  If this pattern does not match any subsequence of the input then the resulting array has just one element, namely the input sequence in string form.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
rzwiefel: Implemented Based off Q-Values. yea or nay?
https://github.com/codice/ddf/pull/1760/commits/1b1728e16bddbb964ab6914c8916a18e95284e9b

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
stustison: I think Phil is asking how all of this might behave if someone were to provide an Accept-Encoding header with extremely large input, input with special characters that shouldn't be allowed, etc. If we get an Accept-Encoding header with 5000 characters in it, it is probably just malicious input, for example.

rzwiefel:
NOT ready for merge.

rzwiefel:
Closing for now since there needs to be updates done to this but is a non-critical priority. Will pickup later when time permits.

PR #1761, opened 2017-03-06
DDF-2494 create an email service

glenhein:
#### What does this PR do?
Create a centralized email service that handles the configuration details for connecting to an email server.

@bdeining 
@rzwiefel 
@michaelmenousek  [optional]

[Core APIs](https://github.com/orgs/codice/teams/core-apis) @brendan-hofmann 

@pklinef 
@lessarderic

For testing, you will need to know the hostname of your email server. If you don't know what it is, you can try asking your system administrator or run the un*x command "dig mx DOMAINNAME", where DOMAINNAME is the domain name portion of your email address.

There are two areas that need to be tested: non-authenticating and authenticating. A non-authenticating email server does not require a username/password. For the authenticating test, the only email server that I know of that uses authentication w/ TLS is gmail.

If you have a gmail account. The following should provide some help: https://support.google.com/mail/answer/7104828?ctx=gmail&hl=en&visit_id=1-636244123112925149-4219943305&rd=1
You may need to enable access from "less secure apps" in your gmail settings.

The only way to testing email from inside DDF is through the Catalog UI and standing query notifications.

1) Build and unzip DDF.

2) Add user:
  a) edit etc/user.properties, add the line:

     USERNAME=PASSWORD,group,admin,manager,viewer,system-admin,systembundles

     where USERNAME and PASSWORD are you preferred login credentials for DDF. These values are not related
     to the email server username and password mentioned earlier.

  b) edit etc/user.attributes, add the entry:

     "USERNAME" : { "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress" : "EMAILADDR" },

     where USERNAME is the value used in Step 2a, and EMAILADDR your email address that will receive the test email.

3) Install DDF.

4) Under 'Platform', go to 'Configuration' and select 'Email Service'. If you are performing a non-authenticating test, then set the Host and Port fields. If you are performing an authenticating test, then set the Host, Port, User Name and Password fields. Click 'Save changes'.

5) Under 'Standard Search UI', go to 'Configuration' and select 'Email Notifier'. The default values should be sufficient. Click 'Save changes'.

6) Under 'Standard Search UI', go to 'Configuration' and select 'Workspace Query Monitor'. Set the Email Subscription Interval field to '0 * * * * ?'. This will cause an email to be sent every minute. More information on this field's format can be found at http://www.quartz-scheduler.org/documentation/quartz-2.x/tutorials/crontrigger.html

7) Using a webbrowser, navigate to the Catalog Search UI: https://localhost:8993/search/catalog/

8) Sign-in using the username and password used in Step 2a.

9) Create a new workspace by clicking the 'Blank' icon.

10) Create a new search, the default parameters should be sufficient.

11) From the menu next to the workspace name, select 'Subscribe'.

12) From the menu next to the workspace name, select 'Save'. If you get a workspace error, then try to logout and login.

13) In about a minute, you should start receiving an email notification every minute.

[DDF-2494](https://codice.atlassian.net/browse/DDF-2494)
#### Checklist:
- [X] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
@@ -17,12 +17,10 @@ import static org.apache.commons.lang3.Validate.notBlank;
 import static org.apache.commons.lang3.Validate.notNull;
 
 import java.util.Map;
-import java.util.Properties;
 
 import javax.mail.Message;
 import javax.mail.MessagingException;
 import javax.mail.Session;
-import javax.mail.Transport;
 import javax.mail.internet.InternetAddress;
 import javax.mail.internet.MimeMessage;
 
@@ -31,6 +29,7 @@ import org.codice.ddf.catalog.ui.metacard.workspace.WorkspaceMetacardImpl;
 import org.codice.ddf.catalog.ui.query.monitor.api.MetacardFormatter;
 import org.codice.ddf.catalog.ui.query.monitor.api.QueryUpdateSubscriber;
 import org.codice.ddf.catalog.ui.query.monitor.api.SubscriptionsPersistentStore;
+import org.codice.ddf.platform.email.SmtpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -38,10 +37,6 @@ import org.slf4j.LoggerFactory;
  * Sends an email for each workspace to the owner of the workspace.
  */
 public class EmailNotifier implements QueryUpdateSubscriber {
-
-    private static final String SMTP_HOST_PROPERTY = "mail.smtp.host";
-
-    private static final String SMTP_AUTH_PROPERTY = "mail.smtp.auth";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(EmailNotifier.class);
 
@@ -53,48 +48,36 @@ public class EmailNotifier implements QueryUpdateSubscriber {
 
     private String fromEmail;
 
-    private String mailHost;
-
     private SubscriptionsPersistentStore subscriptionsPersistentStore;
+
+    private SmtpClient smtpClient;
 
     /**
      * The {@code bodyTemplate} and {@code subjectTemplate} may contain the tags supported by the
      * {@code metacardFormatter}.
      *
-     * @param bodyTemplate                 must be non-null
-     * @param subjectTemplate              must be non-null
-     * @param fromEmail                    must be non-null
-     * @param mailHost                     must be non-null
-     * @param metacardFormatter            must be non-null
-     * @param subscriptionsPersistentStore must be non-null
+     * @param bodyTemplate                 A string that represents the email body, will be passed to {@link MetacardFormatter}.
+     * @param subjectTemplate              A string that represents the subject line of the email, will be passed to {@link MetacardFormatter}.
+     * @param fromEmail                    The 'from' email address to be set in the email.
+     * @param metacardFormatter            Used to format the body and subject line.
+     * @param subscriptionsPersistentStore Used to find the email address associated with a workspace id.
+     * @param smtpClient                   The {@link SmtpClient} that will send the email.
      */
     public EmailNotifier(String bodyTemplate, String subjectTemplate, String fromEmail,
-            String mailHost, MetacardFormatter metacardFormatter,
-            SubscriptionsPersistentStore subscriptionsPersistentStore) {
+            MetacardFormatter metacardFormatter,
+            SubscriptionsPersistentStore subscriptionsPersistentStore, SmtpClient smtpClient) {
         notNull(bodyTemplate, "bodyTemplate must be non-null");
         notNull(subjectTemplate, "subjectTemplate must be non-null");
         notNull(fromEmail, "fromEmail must be non-null");
-        notNull(mailHost, "mailHost must be non-null");
         notNull(subscriptionsPersistentStore, "subscriptionsPersistentStore must be non-null");
+        notNull(smtpClient, "smtpClient must be non-null");
 
         this.bodyTemplate = bodyTemplate;
         this.subjectTemplate = subjectTemplate;
         this.fromEmail = fromEmail;
-        this.mailHost = mailHost;
         this.metacardFormatter = metacardFormatter;
         this.subscriptionsPersistentStore = subscriptionsPersistentStore;
-    }
-
-    /**
-     * The hostname of the mail server.
-     *
-     * @param mailHost must be non-blank
-     */
-    @SuppressWarnings("unused")
-    public void setMailHost(String mailHost) {
-        notBlank(mailHost, "mailHost must be non-blank");
-        LOGGER.debug("Setting mailHost : {}", mailHost);
-        this.mailHost = mailHost.trim();
+        this.smtpClient = smtpClient;
     }
 
     /**
@@ -152,9 +135,7 @@ public class EmailNotifier implements QueryUpdateSubscriber {
 
         String subject = metacardFormatter.format(subjectTemplate, workspaceMetacard, hitCount);
 
-        Properties properties = createSessionProperies();
-
-        Session session = Session.getDefaultInstance(properties);
+        Session session = smtpClient.createSession();
 
         try {
             MimeMessage mimeMessage = new MimeMessage(session);
@@ -167,9 +148,9 @@ public class EmailNotifier implements QueryUpdateSubscriber {
 
             mimeMessage.setText(emailBody);
 
-            LOGGER.debug("Attempting to send email to mailHost : {}", mailHost);
+            LOGGER.trace("Attempting to send email");
 
-            Transport.send(mimeMessage);
+            smtpClient.send(mimeMessage);
 
         } catch (MessagingException e) {
             LOGGER.warn("unable to send email to {}", email, e);
@@ -177,24 +158,11 @@ public class EmailNotifier implements QueryUpdateSubscriber {
 
     }
 
-    private Properties createSessionProperies() {
-        Properties properties = System.getProperties();
-
-        properties.setProperty(SMTP_HOST_PROPERTY, mailHost);
-        properties.setProperty(SMTP_AUTH_PROPERTY, "false");
-
-        return properties;
-    }
-
     @Override
     public String toString() {
-        return "EmailNotifier{" +
-                "metacardFormatter=" + metacardFormatter +
-                ", bodyTemplate='" + bodyTemplate + '\'' +
-                ", subjectTemplate='" + subjectTemplate + '\'' +
-                ", fromEmail='" + fromEmail + '\'' +
-                ", mailHost='" + mailHost + '\'' +
-                ", subscriptionsPersistentStore=" + subscriptionsPersistentStore +
-                '}';
+        return "EmailNotifier{" + "metacardFormatter=" + metacardFormatter + ", bodyTemplate='"
+                + bodyTemplate + '\'' + ", subjectTemplate='" + subjectTemplate + '\''
+                + ", fromEmail='" + fromEmail + '\'' + ", subscriptionsPersistentStore="
+                + subscriptionsPersistentStore + '}';
     }
 }
--- a/platform/email/platform-email-impl/pom.xml
+++ b/platform/email/platform-email-impl/pom.xml
@@ -1,0 +1,141 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<project
+    xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>email</artifactId>
+        <groupId>ddf.platform.email</groupId>
+        <version>2.11.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>platform-email-impl</artifactId>
+    <name>DDF :: Platform :: Email :: Impl</name>
+    <packaging>bundle</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>ddf.platform.email</groupId>
+            <artifactId>platform-email-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.simplejavamail</groupId>
+            <artifactId>simple-java-mail</artifactId>
+            <version>4.1.3</version>
+        </dependency>
+        <dependency>
+            <groupId>com.sun.mail</groupId>
+            <artifactId>javax.mail</artifactId>
+            <version>1.5.5</version>
+        </dependency>
+        <dependency>
+            <groupId>net.iharder</groupId>
+            <artifactId>base64</artifactId>
+            <version>2.3.9</version>
+        </dependency>
+        <dependency>
+            <groupId>net.markenwerk</groupId>
+            <artifactId>utils-data-fetcher</artifactId>
+            <version>4.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>net.markenwerk</groupId>
+            <artifactId>commons-nulls</artifactId>
+            <version>1.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>dumbster</groupId>
+            <artifactId>dumbster</artifactId>
+            <version>1.6</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
+            <version>1.3.9_1</version>
+        </dependency>
+        <dependency>
+            <groupId>ddf.security.core</groupId>
+            <artifactId>security-core-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Embed-Dependency>
+                            simple-java-mail,
+                            base64,
+                            utils-data-fetcher,
+                            commons-nulls,
+                            org.apache.servicemix.bundles.jsr305
+                        </Embed-Dependency>
+                        <Export-Package/>
+                    </instructions>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.95</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.75</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.8</minimum>
+                                        </limit>
+                                    </limits>
+                                </rule>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
+++ b/platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.platform.email.impl;
+
+import static org.apache.commons.lang.Validate.isTrue;
+import static org.apache.commons.lang.Validate.notEmpty;
+import static org.apache.commons.lang.Validate.notNull;
+
+import java.util.Arrays;
+import java.util.Properties;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import javax.annotation.Nullable;
+import javax.mail.Authenticator;
+import javax.mail.Message;
+import javax.mail.MessagingException;
+import javax.mail.PasswordAuthentication;
+import javax.mail.Session;
+import javax.mail.Transport;
+
+import org.apache.commons.lang.StringUtils;
+import org.codice.ddf.platform.email.SmtpClient;
+
+import ddf.security.common.audit.SecurityLogger;
+
+/**
+ * Supports unauthenticated connections to a mail server and username/password authenticated connections
+ * over TLS. If the username is not blank and the password is not empty, then auth/tls will be used.
+ * <p>
+ * The method {@link #send(Message)} will exceute asynchronously in a single thread. The number
+ * of operations awaiting execution is unbounded.
+ * <p>
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ */
+public class SmtpClientImpl implements SmtpClient {
+
+    private static final String SMTP_HOST_PROPERTY = "mail.smtp.host";
+
+    private static final String SMTP_PORT_PROPERTY = "mail.smtp.port";
+
+    private static final String SMTP_AUTH_PROPERTY = "mail.smtp.auth";
+
+    private static final String SMTP_START_TLS_ENABLE_PROPERTY = "mail.smtp.starttls.enable";
+
+    private static final String TRUE = Boolean.TRUE.toString();
+
+    private static final String FALSE = Boolean.FALSE.toString();
+
+    private final ExecutorService executorService = Executors.newFixedThreadPool(1);
+
+    private String hostName;
+
+    private Integer portNumber;
+
+    private String userName;
+
+    private String password;
+
+    /**
+     * Set the username for the email server.
+     *
+     * @param userName the username of the email server
+     */
+    public void setUserName(@Nullable String userName) {
+        this.userName = userName;
+    }
+
+    /**
+     * Set the password for the email server.
+     *
+     * @param password the password for the email server
+     */
+    public void setPassword(@Nullable String password) {
+        this.password = password;
+    }
+
+    /**
+     * @param hostName must be non-empty
+     */
+    public void setHostName(String hostName) {
+        notEmpty(hostName, "hostName must be non-empty");
+        this.hostName = hostName;
+    }
+
+    /**
+     * @param portNumber must be non-null and >0
+     */
+    public void setPortNumber(Integer portNumber) {
+        notNull(portNumber, "portNumber must be non-null");
+        isTrue(portNumber > 0, "portNumber must be >0");
+        this.portNumber = portNumber;
+    }
+
+    @Override
+    public Session createSession() {
+
+        Properties properties = new Properties();
+
+        properties.setProperty(SMTP_HOST_PROPERTY, hostName);
+        properties.setProperty(SMTP_PORT_PROPERTY, portNumber.toString());
+
+        if (StringUtils.isNotBlank(userName)) {
+            properties.put(SMTP_AUTH_PROPERTY, TRUE);
+            properties.put(SMTP_START_TLS_ENABLE_PROPERTY, TRUE);
+
+            return Session.getInstance(properties, createAuthenticator());
+        } else {
+            properties.setProperty(SMTP_AUTH_PROPERTY, FALSE);
+            return Session.getInstance(properties);
+        }
+    }
+
+    @Override
+    public Future<Void> send(Message message) throws MessagingException {
+        notNull(message, "message must be non-null");
+        return executorService.submit(new Callable<Void>() {
+            @Override
+            public Void call() throws Exception {
+                Transport.send(message);
+                SecurityLogger.audit("Sent an email: recipient={} subject={}",
+                        Arrays.toString(message.getAllRecipients()),
+                        message.getSubject());
+                return null;
+            }
+        });
+    }
+
+    Authenticator createAuthenticator() {
+        return new Authenticator() {
+            protected PasswordAuthentication getPasswordAuthentication() {
+                return new PasswordAuthentication(userName, password);
+            }
+        };
+    }
+
+}
--- a/platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
+++ b/platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<metatype:MetaData xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xmlns:metatype="http://www.osgi.org/xmlns/metatype/v1.2.0"
+                   xsi:schemaLocation="http://www.osgi.org/xmlns/metatype/v1.2.0 http://www.osgi.org/xmlns/metatype/v1.2.0">
+
+    <OCD name="Email Service"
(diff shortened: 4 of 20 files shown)


shaundmorris:
@glenhein are you planning on adding to the DDF documentation with this PR?

platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -0,0 +1,47 @@
...
+
+        <AD
+                description="Mail server user name."
+                name="User Name" id="userName" required="false"
+                type="String" default=""/>
+
+        <AD
+                description="Mail server password."
+                name="Password" id="password" required="false"
+                type="String" default=""/>
```
stustison: This should be type="Password". You need to change the metatype schema to version 1.2.0

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/EmailServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private static final String HOSTNAME = "127.0.0.1";
+
+    private static final String ATTACHMENT_TEXT = "text text text";
+
+    private static final String ATTACHMENT_FILENAME = "myFile.txt";
+
+    private static final String SUBJECT = "sample subject";
+
+    private static final String TO_ADDR = "glen.hein@connexta.com";
```
bdeining: We should use `example.com` instead

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/EmailServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private static final String FROM_ADDR = "nobody@connexta.com";
```
bdeining: `example.com`

platform/email/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>platform</artifactId>
+        <groupId>ddf.platform</groupId>
+        <version>2.11.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>ddf.platform.email</groupId>
+    <artifactId>email</artifactId>
+    <name>DDF Platform Email</name>
```
bdeining: I think the convention here is `DDF :: Platform :: Email`

platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
glenhein: thanks, changed it

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/EmailServiceImplTest.java
glenhein: changed

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/EmailServiceImplTest.java
glenhein: changed

platform/email/pom.xml
glenhein: changed

platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
glenhein: @stustison  I changed the version (http://www.osgi.org/xmlns/metatype/v1.2.0), but now the default for username and password show up as "admin", even though my default is explicitly set to "".

stustison: @glenhein You can find an example of how to do it here: https://github.com/codice/ddf/blob/master/platform/security/sts/security-sts-ldaplogin/src/main/resources/OSGI-INF/metatype/metatype.xml

glenhein:
@stustison here's a screenshot of the metatype dialog after changing it to v1.2. The username and password fields should be blank, but are not.

screen shot 2017-03-07 at 10 40 29 am

platform/email/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>ddf.platform.email</groupId>
+    <artifactId>email</artifactId>
+    <name>DDF :: Platform :: Email</name>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
+                <version>3.5</version>
```
bdeining: The catalog uses commons-lang3 version 3.4.  https://github.com/codice/ddf/blob/master/catalog/pom.xml#L32.  We should move this dependency up to the root pom and keep commons-lang3 on the same version throughout DDF.

glenhein: moved to root pom, and adjusted the platform-app to use the correct commons-lang3 version

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
```diff
@@ -0,0 +1,65 @@
...
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.platform.email.api;
```
pklinef: Do not include `api` in the package.  It should just be `org.codice.ddf.platform.email`.  Any implementation related to this package should be in a `impl` folder.

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
```diff
@@ -0,0 +1,65 @@
...
+ * multipart.addBodyPart(messageBodyPart);
+ *
+ * mimeMessage.setContent(multipart);
+ *
+ * emailService.send(mimeMessage);
+ *
+ * }
+ * </pre>
+ */
+public interface EmailService {
```
pklinef: Why not just `Email` or `Emails`?  That would help avoid the use of service which does not add information about this object type.

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/EmailServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+    private static final String SUBJECT_HEADER = "Subject";
+
+    private static final String FROM_HEADER = "From";
+
+    private static final String TO_HEADER = "To";
+
+    @Test
+    public void testSendWithAttachments() throws IOException, MessagingException {
+
+        int port = findAvailablePort();
```
pklinef: I always get concerned when I see ports being opened in unit tests.  Can we use a mock `Session` instead to avoid needing to startup a SMTP server?

glenhein: I wanted to be able test that the email was actually getting sent. I could mock Session, but then I'm not testing that the correct parameters are being sent to javax.mail.

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
glenhein: I'll rename it.

platform/platform-app/src/main/resources/features.xml
```diff
@@ -52,6 +52,13 @@
             <!--mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.jaxb-impl/2.2.6_1-->
         <!--</bundle>-->
     <!--</feature>-->
+    <feature name='platform-email' install="automatic" version='${project.version}' start-level='50'>
+        <bundle dependency="true">mvn:org.apache.commons/commons-lang3/${commons-lang3.version}</bundle>
```
brendan-hofmann: Do we need lang3? I don't see any usages that aren't also available in 2.6 which is already available in the container.

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
brendan-hofmann: I believe our convention is to not put api in the package, and put impl in the implementation package.

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
brendan-hofmann: I think we should include the experimental header since this is a new interface.

platform/email/platform-email-impl/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Embed-Dependency>
+                            commons-lang3,
```
brendan-hofmann: See my other comment about commons-lang, but if we do need lang3, we should not be both embedding it and including it as a bundle dependency in the feature.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/EmailServiceImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        Properties properties = System.getProperties();
+
+        properties.setProperty(SMTP_HOST_PROPERTY, hostName);
+        properties.setProperty(SMTP_PORT_PROPERTY, portNumber.toString());
+
+        if (StringUtils.isNotBlank(userName) && StringUtils.isNotEmpty(password)) {
+            properties.put(SMTP_AUTH_PROPERTY, TRUE);
+            properties.put(SMTP_START_TLS_ENABLE_PROPERTY, TRUE);
+
+            return Session.getInstance(properties, new Authenticator() {
```
brendan-hofmann: What happens in the log if the hostname is unreachable here? Should we do a little more proactive validation so we can tell admins how to fix the problem?

glenhein: I'll run a test to see what happens, but DNS can change at any time. An admin could even enter a hostname that does not currently resolve knowing that DNS will eventually update.

platform/platform-app/src/main/resources/features.xml
glenhein: I can try switching over to commons-lang

platform/email/platform-email-impl/pom.xml
glenhein: removed

platform/platform-app/src/main/resources/features.xml
glenhein: I can try switching over to commons-lang

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/EmailServiceImpl.java
glenhein: So, If a bad hostname is entered, an error appears when the email is about to be sent. Example:

10:42:00,296 | WARN  | heduler_Worker-2 | log.ui.query.monitor.email.EmailNotifier  156 | catalog-ui-search    | unable to send email to admin@localhost.local
com.sun.mail.util.MailConnectException: Couldn't connect to host, port: connexta-com.mail.protection.outlook.co, 25; timeout -1;
  nested exception is:
	java.net.UnknownHostException: connexta-com.mail.protection.outlook.co

The error message includes the hostname, so an admin should be able to diagnose the problem with default log settings.

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/EmailServiceImplTest.java
pklinef: I wonder if we need to run this using failsafe maven plugin so it executes in the integration phase.

glenhein:
test this please

glenhein:
@stustison did you have any more requests before approving this PR?

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
glenhein: this package has been renamed

platform/email/platform-email-api/src/main/java/org/codice/ddf/platform/email/api/EmailService.java
glenhein: Added: "This code is experimental. While this interface is functional and tested, it may change or be removed in a future version of the library."

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/SmtpClientImplITCase.java
```diff
@@ -0,0 +1,219 @@
...
+
+        emailService.setHostName("host.com");
+        emailService.setPortNumber(25);
+        emailService.setUserName(username);
+        emailService.setPassword(password);
+
+        Session session = emailService.createSession();
+
+        PasswordAuthentication actual = session.requestPasswordAuthentication(InetAddress.getByName(
+                "127.0.0.1"), 25, "smptp", "prompt", "defaultUserName");
```
bdeining: is this supposed to be `smtp`?

glenhein: fixed

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    private String userName;
+
+    private String password;
+
+    /**
+     * Set the username for the email server.
+     *
+     * @param userName may be null or blank
+     */
+    public synchronized void setUserName(String userName) {
```
lessarderic: ❗️We should put `@Nullable` on `userName`.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        this.userName = userName;
+    }
+
+    /**
+     * Set the password for the email server.
+     *
+     * @param password may be null or blank
+     */
+    public synchronized void setPassword(String password) {
+        this.password = password;
```
lessarderic: ❗️We should put `@Nullable` on `userName`.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    public synchronized void setHostName(String hostName) {
+        notEmpty(hostName, "hostName must be non-empty");
+        this.hostName = hostName;
+    }
+
+    /**
+     * @param portNumber must be non-null and >0
+     */
+    public synchronized void setPortNumber(Integer portNumber) {
```
lessarderic: ❓Could `portNumber` be an `int` since it can't be `null`?

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
```diff
@@ -0,0 +1,132 @@
...
+            return Session.getInstance(properties, createAuthenticator());
+        } else {
+            properties.setProperty(SMTP_AUTH_PROPERTY, FALSE);
+            return Session.getInstance(properties);
+        }
+    }
+
+    @Override
+    public synchronized void send(Message message) throws MessagingException {
+        Transport.send(message);
```
lessarderic: ❓ Should we check that `message` isn't `null`?

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/SmtpClientImplITCase.java
```diff
@@ -0,0 +1,219 @@
...
+
+    }
+
+    private int findAvailablePort() throws IOException {
+        try (ServerSocket s = new ServerSocket(0)) {
+            return s.getLocalPort();
+        }
+    }
+
+    private static class DeletableFile implements Closeable {
```
lessarderic: ❗️ We should be able to use http://junit.org/junit4/javadoc/4.12/index.html?org/junit/rules/TemporaryFolder.html instead of creating our own `DeletableFile`.

platform/email/platform-email-impl/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Embed-Dependency>
+                            simple-java-mail,
+                            base64,
```
brendan-hofmann: Is this a transitive dependency? I don't see it used anywhere. Curious because there's a `Base64` in `java.util` that we normally use.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    public synchronized void setPortNumber(Integer portNumber) {
+        notNull(portNumber, "portNumber must be non-null");
+        isTrue(portNumber > 0, "portNumber must be >0");
+        this.portNumber = portNumber;
+    }
+
+    @Override
+    public synchronized Session createSession() {
+
+        Properties properties = System.getProperties();
```
brendan-hofmann: Since this is initialized from the system properties, I think we should either check for existing values or explicitly state that we will not use them. I could reasonably see someone expect to be able to configure this once in their environment and be done.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static final String SMTP_HOST_PROPERTY = "mail.smtp.host";
+
+    private static final String SMTP_PORT_PROPERTY = "mail.smtp.port";
+
+    private static final String SMTP_AUTH_PROPERTY = "mail.smtp.auth";
+
+    private static final String SMTP_START_TLS_ENABLE_PROPERTY = "mail.smtp.starttls.enable";
+
+    private static final String TRUE = "true";
```
brendan-hofmann: Nit, but could just use `Boolean.{FALSE|TRUE}.toString()` instead of creating our own constants.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
```diff
@@ -53,48 +48,36 @@
...
      * The {@code bodyTemplate} and {@code subjectTemplate} may contain the tags supported by the
      * {@code metacardFormatter}.
      *
      * @param bodyTemplate                 must be non-null
      * @param subjectTemplate              must be non-null
      * @param fromEmail                    must be non-null
-     * @param mailHost                     must be non-null
      * @param metacardFormatter            must be non-null
      * @param subscriptionsPersistentStore must be non-null
+     * @param smtpClient                 must be non-null
```
glenhein: I'll update the javadocs.

platform/email/platform-email-impl/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.99</minimum>
```
glenhein: done, changed it .95

platform/email/platform-email-impl/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.6</minimum>
```
glenhein: Added a test, but also fixed a bug in the code that improved the branch complexity

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: Added @Nullable

glenhein: The synchronization was done at @pklinef request. He was concerned about heavily loading the javax.mail code with too many concurrent requests. I also added synchronization to to the setters to make findbugs happy. But I could switch over to an explicit lock on createSession() and send().

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: Added @Nullable to password.

glenhein: My understanding is that the only way to store an encrypted password is through the karaf command-line, and @millerw8 was opposed to that solution

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: My understanding has been to always use objects when receiving data from metatype. Is that not correct?

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: I didn't think so originally, since the Transport.send javadoc doesn't state that it must be non-null. But I can add a check for it.

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/SmtpClientImplITCase.java
```diff
@@ -0,0 +1,219 @@
...
+import javax.mail.internet.MimeBodyPart;
+import javax.mail.internet.MimeMessage;
+import javax.mail.internet.MimeMultipart;
+
+import org.junit.Test;
+
+import com.dumbster.smtp.SimpleSmtpServer;
+import com.dumbster.smtp.SmtpMessage;
+
+public class SmtpClientImplITCase {
```
glenhein: Because it is an integration phase test using the failsafe (instead of surefire) plugin. Done at the request of @pklinef .

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/SmtpClientImplITCase.java
glenhein: changed to TemporaryFolder

platform/email/platform-email-impl/pom.xml
glenhein: I originally needed it, but that may have changed. I'll try taking it out.

glenhein: So, without 'base64', I get the error: missing requirement [platform-email-impl/2.11.0.SNAPSHOT] osgi.wiring.package; filter:="(&(osgi.wiring.package=net.iharder)(version>=2.3.0)(!(version>=3.0.0)))"]]]]

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: changed it as requested

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: If we never tell the user that we are initializing from the system properties, then why would they try to use the system properties?

brendan-hofmann: We document that this is a lightweight interface to `javax.mail`. It would be reasonable to expect it to behave like `javax.mail`.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
```diff
@@ -53,48 +48,36 @@
...
      * The {@code bodyTemplate} and {@code subjectTemplate} may contain the tags supported by the
      * {@code metacardFormatter}.
      *
-     * @param bodyTemplate                 must be non-null
-     * @param subjectTemplate              must be non-null
-     * @param fromEmail                    must be non-null
-     * @param mailHost                     must be non-null
-     * @param metacardFormatter            must be non-null
-     * @param subscriptionsPersistentStore must be non-null
+     * @param bodyTemplate                 A string that represents the email body, will be passed to MetacardFormatter.
```
bdeining: Might be useful to link the MetacardFormatter in the javadoc
`{@link org.codice.ddf.catalog.ui.query.monitor.api.MetacardFormatter}`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
```diff
@@ -53,48 +48,36 @@
...
+     * @param subjectTemplate              A string that represents the subject line of the email, will be passed to MetacardFormatter.
```
bdeining: `{@link org.codice.ddf.catalog.ui.query.monitor.api.MetacardFormatter}`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
```diff
@@ -167,34 +148,21 @@ private void sendEmailForWorkspace(WorkspaceMetacardImpl workspaceMetacard, Long
 
             mimeMessage.setText(emailBody);
 
-            LOGGER.debug("Attempting to send email to mailHost : {}", mailHost);
+            LOGGER.debug("Attempting to send email");
```
bdeining: This seems more appropriate as trace

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
```diff
@@ -53,48 +48,36 @@
...
-     * @param fromEmail                    must be non-null
-     * @param mailHost                     must be non-null
-     * @param metacardFormatter            must be non-null
-     * @param subscriptionsPersistentStore must be non-null
+     * @param bodyTemplate                 A string that represents the email body, will be passed to MetacardFormatter.
+     * @param subjectTemplate              A string that represents the subject line of the email, will be passed to MetacardFormatter.
+     * @param fromEmail                    The 'from' email address to be set in the email.
+     * @param metacardFormatter            Used to format the body and subject line.
+     * @param subscriptionsPersistentStore Used to find the email address associated with a workspace id.
+     * @param smtpClient                   The SmtpClient that will send the email.
```
bdeining: Might be useful to link to the SmtpClient class in javadoc

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: I just did an experiment where I removed System.getProperties and the email was delivered as expected. I don't know why the javax.mail examples include it, but it doesn't seem to be required for our purposes.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
glenhein: changed debug level

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
lessarderic: So was the synchronization on the class a way to throttle requests? If so, I'm not sure this is the right way to achieve that.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
lessarderic: Pretty sure we could use `int` and the container would take care of the conversion, but it's not a big deal.

platform/email/platform-email-impl/src/test/java/org/codice/ddf/platform/email/impl/SmtpClientImplITCase.java
lessarderic: If we're going to use failsafe for our components level testing (which I like), we may want to update the https://codice.atlassian.net/wiki/display/DDF/Unit+Testing wiki to provide some basic guidelines and make sure things are consistent.

bdeining:
Full build successful.  Verified non-authenticating email capability by following testing procedures.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: @lessarderic let's figure out a solution that you are happy with. I really need to get this ticket wrapped up. If you like, I can schedule a meeting so we can resolved any questions you may have.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
```diff
@@ -167,34 +148,21 @@ private void sendEmailForWorkspace(WorkspaceMetacardImpl workspaceMetacard, Long
 
             mimeMessage.setText(emailBody);
 
-            LOGGER.debug("Attempting to send email to mailHost : {}", mailHost);
+            LOGGER.trace("Attempting to send email");
 
-            Transport.send(mimeMessage);
+            smtpClient.send(mimeMessage);
```
stustison: An audit should be written anytime an email message is sent. It should include who the recipient is.

platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ricklarsen: This should probably go into the documentation. See `distribution/docs/src/main/resources/_contents/_tables` for examples of the tables that map to the metatype configuration. 

Unless we put it somewhere else, that table should be included in the `managing-platform-contents.adoc` file

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
lessarderic: I guess my concern is more that adding synchronization to throttle clients so they don't overflow the javax.mail code just pushes the problem up to the clients and forces them to handle the scenario where those calls could hang for a potentially long time.

One option would be to have an internal queue and have one or more threads that consume from it and push out to the external mail service at a pre-established well-known rate.

No matter what solution we pick, we should clearly document in the interface what the expected implementation behavior needs to be (blocking or not) so all the implementations are consistent and clients know what to expect.

millerw8:
build now

platform/email/platform-email-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
glenhein: @ricklarsen I added the documentation. Please review it.

platform/email/platform-email-impl/src/main/java/org/codice/ddf/platform/email/impl/SmtpClientImpl.java
glenhein: @lessarderic I switched the code over to using a executor service with one worker thread and updated the javadoc for the class. The API javadoc was updated to indicate that the email may be send asynchronously.

glenhein:
@bdeining @brendan-hofmann  can you take a look at this PR? I'm trying to get the final sign-off for merging, thanks!

platform/email/platform-email-impl/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Embed-Dependency>
+                            simple-java-mail,
+                            base64,
+                            utils-data-fetcher,
+                            commons-nulls,
+                            org.apache.servicemix.bundles.jsr305
```
bdeining: Is there a reason why this bundle is being embedded?

pom.xml
```diff
@@ -214,6 +215,16 @@
                 <artifactId>shiro-core</artifactId>
                 <version>${apache.shiro.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
+                <version>${commons-lang3.version}</version>
```
bdeining: We may need a tech debt ticket to clean up commons-lang3 dependencies in platform & lib with this change

pom.xml
```diff
@@ -214,6 +215,16 @@
...
+            </dependency>
+            <dependency>
+                <groupId>commons-lang</groupId>
+                <artifactId>commons-lang</artifactId>
+                <version>${commons-lang.version}</version>
```
bdeining: We may need a tech debt ticket to clean up commons-lang dependencies with this change

platform/email/platform-email-impl/pom.xml
glenhein: because if was getting errors about the Nullable annotation. Is there a better way or a better jar to embed?

bdeining: I think you should be able to add it to your email feature in the platform features.xml file instead

glenhein: ok, i'll try that

pom.xml
glenhein: https://codice.atlassian.net/browse/DDF-2924

pom.xml
glenhein: https://codice.atlassian.net/browse/DDF-2924

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/monitor/email/EmailNotifier.java
millerw8: Looks like this has been addressed in the smtpClient.  This is pushing the onus of the audit to the implementation of the interface.  This seems reasonable in this case.  @stustison does this address your remaining issues?

stustison: Yes, that looks fine to me, thanks!

PR #1787, opened 2017-03-15
DDF-2862 Added configurable default QueryResponse queue timeout

emanns95:
#### What does this PR do?
Ensures that a default timeout is attached to query requests so that they cannot sit on the queue forever in the case of system load slowing the system.

@rzwiefel 
@brendan-hofmann 

@pklinef
@stustison

A solr update seems to have resolved us being able to replicate the system freezing, but the timeout is still being added.

[DDF-2862](https://codice.atlassian.net/browse/DDF-2862)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests


emanns95:
I'll update the docs now, slipped my mind

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
```diff
@@ -204,6 +213,23 @@ QueryResponse doQuery(QueryRequest queryRequest, FederationStrategy strategy)
 
         LOGGER.debug("Calling strategy.federate()");
 
+        Query originalQuery = queryRequest.getQuery();
+
+        long timeout = originalQuery.getTimeoutMillis();
+
+        if (timeout == 0) {
+            timeout = TimeUnit.MINUTES.toMillis(queryTimeout);
```
brendan-hofmann: If  `queryTimeout` is already in milliseconds, I do not think you want this conversion here.

emanns95: Oh jeez you're right 😂

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/QueryResponseImpl.java
```diff
@@ -50,6 +50,8 @@
 
     List<Result> resultList = null;
 
+    private long timeout;
```
pklinef: It can be useful to include the units in the variable name like `timeoutMillis`.

pklinef: I would recommend setting a default value in the class to avoid initialization race conditions.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
pklinef: This is why it is useful to include the units in the variable name.  :wink:

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
```diff
@@ -91,6 +94,8 @@
 
     private List<String> fanoutProxyTagBlacklist = new ArrayList<>();
 
+    private long queryTimeout;
```
pklinef: Same comment about setting a default in the class to avoid potential initialization race conditions.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/QueryResponseImpl.java
rzwiefel: @pklinef, they are set explicitly as a TimeUnit below. Do we really want 'timeoutMillis, TimeUnit.MILLISECONDS`?

rzwiefel: meh nevermind. I guess never hurts to be extra careful. And we are in java so verbosity doesn't matter ¯\\_(ツ)_/¯

emanns95:
@pklinef I agree no harm in making the names specific, it's done 😃 
@brendan-hofmann Added unit tests and some documentation

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
```diff
@@ -204,6 +212,26 @@ QueryResponse doQuery(QueryRequest queryRequest, FederationStrategy strategy)
 
         LOGGER.debug("Calling strategy.federate()");
 
+        Query originalQuery = queryRequest.getQuery();
+
+        long timeout = originalQuery.getTimeoutMillis();
+
+        if (timeout == 0) {
+            timeout = queryTimeoutMillis;
```
rzwiefel: This is breaking the contract of the query timeout: https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api/src/main/java/ddf/catalog/operation/Query.java#L76

emanns95: Yeah it is, but I don't see another way to guarantee that they'll timeout, though. We could set a maximum timeout, but nothing will be able to stop that no timeout.

rzwiefel: Removing infinite timeouts is *way* outside the scope of just adding the current timeout to the queue. So I would push for this to be removed until we have discussion about it. ( We need to make sure that we're fine with removing the ability to do a timeout-less query completely. (@pklinef / @lessarderic))

If we do change it we would need to update the javadoc there to reflect its current behavior and ensure the defaults are still what we want them to be. And if we want to completely disable infinite timeout we would have to make this logic `timeout <= 0`.

emanns95: If that's the case I must've misunderstood the scope. I thought the purpose was to enforce timeouts altogether, not just enforce any timeouts already included in the query. I'll remove this logic then

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
```diff
@@ -204,6 +206,22 @@ QueryResponse doQuery(QueryRequest queryRequest, FederationStrategy strategy)
 
         LOGGER.debug("Calling strategy.federate()");
 
+        Query originalQuery = queryRequest.getQuery();
+
+        long timeout = originalQuery.getTimeoutMillis();
+
+        Query modifiedQuery = new QueryImpl(originalQuery,
```
codymacdonald: Why create new instances of Query and QueryRequest? If we really need to do this, can we make copy constructors on QueryRequest and Query?

emanns95: Nope. This actually needs to be removed. It was leftover when I was adding in a default timeout. My bad

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/QueryResponseImpl.java
```diff
@@ -285,7 +290,11 @@ public Result take() {
     private Result handleTake() {
         Result result = null;
         try {
-            result = queue.take();
+            if (timeoutMillis <= 0) {
+                result = queue.take();
```
emanns95: Needed this here because the poll method will immediately return without waiting if given a timeout of 0

pklinef: This will allow the current bug to happen if developers send 0 or less..  I think if we  should override to 5 minutes or some other sensible default to avoid a potential indefinite timeout.

emanns95: That's what I originally had implemented but @rzwiefel pointed out that it does indeed break the contract of the Query interface. I think you were tagged in an earlier comment about it. We should discuss it

emanns95:
build now

emanns95:
build now

emanns95:
build now

emanns95:
build now

emanns95:
build now

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java
pklinef: The goal was to be able to control a maximum timeout if an unlimited timeout was chosen.  We do not want to bugs to allow queries to stay open forever and allow leaks.  There should still be a maximum timeout for all queries in the system.

rzwiefel: As long as we are fine with changing the interface, :+1:

emanns95:
build now

emanns95:
build now

emanns95:
CI build failed due to a networking issue. Passed full build with exams on a linux environment.

PR #1793, opened 2017-03-18
DDF-2869 Rework ConfigurationFilesPoller to be more reliable

brendan-hofmann:
#### What does this PR do?
Replace nio with commons-io's file watcher, and replace the one second sleep with a poll for the file size to become stable.

These changes should mitigate itest stability problems in CI environments that are prone to IOWait spikes.

Also includes some improvements to itest logging that were needed to debug this solution anyways.
@ahoffer @clockard 
[Continuous Integration](https://github.com/orgs/codice/teams/continuous-integration)
@coyotesqrl
@lessarderic
[DDF-2869](https://codice.atlassian.net/browse/DDF-2869)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/annotations/PaxExamRule.java
+++ b/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/annotations/PaxExamRule.java
@@ -17,6 +17,7 @@ import static org.junit.Assert.fail;
 
 import java.lang.annotation.Annotation;
 
+import org.codice.ddf.itests.common.utils.LoggingUtils;
 import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.rules.TestRule;
@@ -41,10 +42,10 @@ public class PaxExamRule implements TestRule {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(PaxExamRule.class);
 
-    public static final String BEFORE_EXAM_FAILURE_MESSAGE = "Failed to setup exam for %s: %s";
+    public static final String BEFORE_EXAM_FAILURE_MESSAGE = "Failed to setup exam for %s: ";
 
     public static final String AFTER_EXAM_FAILURE_MESSAGE =
-            "Failed to run AfterExam annotation(s) for %s: %s";
+            "Failed to run AfterExam annotation(s) for %s: ";
 
     public static final String EXAM_SETUP_FAILED_MESSAGE = "Exam setup failed";
 
@@ -90,7 +91,7 @@ public class PaxExamRule implements TestRule {
             runAnnotations(PostTestConstruct.class, testClass);
         } catch (Throwable throwable) {
             setupFailed = true;
-            fail(String.format(BEFORE_EXAM_FAILURE_MESSAGE, testClassName, throwable.getMessage()));
+            failRule(testClassName, throwable, BEFORE_EXAM_FAILURE_MESSAGE);
         }
 
         if (firstRun) {
@@ -103,13 +104,16 @@ public class PaxExamRule implements TestRule {
             } catch (Throwable throwable) {
                 setupFailed = true;
                 LOGGER.error("Failed to setup " + testClassName, throwable);
-                fail(String.format(BEFORE_EXAM_FAILURE_MESSAGE,
-                        testClassName,
-                        throwable.getMessage()));
+                failRule(testClassName, throwable, BEFORE_EXAM_FAILURE_MESSAGE);
             }
         }
 
         LOGGER.info("Starting {} ({}/{})", description.getMethodName(), testsExecuted, testCount);
+    }
+
+    private void failRule(String testClassName, Throwable throwable, String failureMessage) {
+        LoggingUtils.failWithThrowableStacktrace(throwable,
+                String.format(failureMessage, testClassName));
     }
 
     private void finished(Description description) {
@@ -123,9 +127,7 @@ public class PaxExamRule implements TestRule {
             try {
                 runAnnotations(AfterExam.class, new TestClass(description.getTestClass()));
             } catch (Throwable throwable) {
-                fail(String.format(AFTER_EXAM_FAILURE_MESSAGE,
-                        testClassName,
-                        throwable.getMessage()));
+                failRule(testClassName, throwable, AFTER_EXAM_FAILURE_MESSAGE);
             }
             LOGGER.info("Finished test(s) for {}", testClassName);
         }
--- a/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/utils/LoggingUtils.java
+++ b/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/utils/LoggingUtils.java
@@ -1,0 +1,34 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.itests.common.utils;
+
+import static org.junit.Assert.fail;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public class LoggingUtils {
+    /**
+     * The stacktrace of {@code throwable} will be appended to {@code message}.
+     * It is the responsibility of the caller to end their message with a space or newline.
+     *
+     * @param throwable contains the stacktrace to be logged
+     * @param message a descriptive message for the stacktrace
+     */
+    public static void failWithThrowableStacktrace(Throwable throwable, String message) {
+        StringWriter errors = new StringWriter();
+        throwable.printStackTrace(new PrintWriter(errors));
+        fail(message + errors.toString());
+    }
+}
--- a/distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/PaxExamRuleTest.java
+++ b/distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/PaxExamRuleTest.java
@@ -38,14 +38,15 @@ import org.ops4j.pax.exam.junit.PaxExam;
 import org.ops4j.pax.exam.spi.reactors.ExamReactorStrategy;
 import org.ops4j.pax.exam.spi.reactors.PerClass;
 
-
 public class PaxExamRuleTest {
 
     public static final String FAILING_TEST_MESSAGE = "test failed";
 
-    public static final String BEFORE_EXAM_EXCEPTION_MESSAGE = "BeforeExam exception";
+    public static final String BEFORE_EXAM_EXCEPTION_MESSAGE =
+            "java.lang.RuntimeException: BeforeExam exception";
 
-    public static final String AFTER_EXAM_EXCEPTION_MESSAGE = "AfterExam exception";
+    public static final String AFTER_EXAM_EXCEPTION_MESSAGE =
+            "java.lang.RuntimeException: AfterExam exception";
 
     public static final String EXPECTED_BEFORE_EXAM_ERROR_MESSAGE =
             String.format(PaxExamRule.BEFORE_EXAM_FAILURE_MESSAGE,
@@ -175,10 +176,16 @@ public class PaxExamRuleTest {
         JUnitCore core = new JUnitCore();
         Result result = core.run(FailingBeforeExamTest.class);
 
-        assertThat(result.getFailures()).extracting("message")
-                .containsOnly(EXPECTED_BEFORE_EXAM_ERROR_MESSAGE,
-                        PaxExamRule.EXAM_SETUP_FAILED_MESSAGE);
+        boolean examFail = result.getFailures()
+                .stream()
+                .anyMatch(failure -> failure.getMessage()
+                        .equals(PaxExamRule.EXAM_SETUP_FAILED_MESSAGE));
+        boolean beforeExamFail = result.getFailures()
+                .stream()
+                .anyMatch(failure -> failure.getMessage()
+                        .contains(EXPECTED_BEFORE_EXAM_ERROR_MESSAGE));
 
+        assertThat(examFail && beforeExamFail);
         assertResultCounts(result, 4);
     }
 
@@ -187,9 +194,16 @@ public class PaxExamRuleTest {
         JUnitCore core = new JUnitCore();
         Result result = core.run(FailingAfterExamTest.class);
 
-        assertThat(result.getFailures()).extracting("message")
-                .contains(EXPECTED_AFTER_EXAM_ERROR_MESSAGE, FAILING_TEST_MESSAGE);
+        boolean examFail = result.getFailures()
+                .stream()
+                .anyMatch(failure -> failure.getMessage()
+                        .equals(PaxExamRule.EXAM_SETUP_FAILED_MESSAGE));
+        boolean afterExamFail = result.getFailures()
+                .stream()
+                .anyMatch(failure -> failure.getMessage()
+                        .contains(EXPECTED_AFTER_EXAM_ERROR_MESSAGE));
 
+        assertThat(examFail && afterExamFail);
         assertResultCounts(result, 2);
     }
 
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
@@ -92,6 +92,7 @@ import org.codice.ddf.itests.common.annotations.SkipUnstableTest;
 import org.codice.ddf.itests.common.catalog.CatalogTestCommons;
 import org.codice.ddf.itests.common.config.UrlResourceReaderConfigurator;
 import org.codice.ddf.itests.common.csw.CswQueryBuilder;
+import org.codice.ddf.itests.common.utils.LoggingUtils;
 import org.codice.ddf.persistence.PersistentItem;
 import org.codice.ddf.persistence.PersistentStore;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.CswConstants;
@@ -179,8 +180,7 @@ public class TestCatalog extends AbstractIntegrationTest {
             configureRestForGuest();
             getSecurityPolicy().waitForGuestAuthReady(REST_PATH.getUrl() + "?_wadl");
         } catch (Exception e) {
-            LOGGER.error("Failed in @BeforeExam: ", e);
-            fail("Failed in @BeforeExam: " + e.getMessage());
+            LoggingUtils.failWithThrowableStacktrace(e, "Failed in @BeforeExam: ");
         }
     }
 
@@ -198,8 +198,8 @@ public class TestCatalog extends AbstractIntegrationTest {
     @Test
     public void testCreateStorage() throws IOException {
         String fileName = testName.getMethodName() + ".jpg";
-        File tmpFile = createTemporaryFile(fileName, IOUtils.toInputStream(getFileContent(
-                SAMPLE_IMAGE)));
+        File tmpFile = createTemporaryFile(fileName,
+                IOUtils.toInputStream(getFileContent(SAMPLE_IMAGE)));
         String id = given().multiPart(tmpFile)
                 .expect()
                 .log()
@@ -215,8 +215,8 @@ public class TestCatalog extends AbstractIntegrationTest {
     @Test
     public void testReadStorage() throws IOException {
         String fileName = testName.getMethodName() + ".jpg";
-        File tmpFile = createTemporaryFile(fileName, IOUtils.toInputStream(getFileContent(
-                SAMPLE_IMAGE)));
+        File tmpFile = createTemporaryFile(fileName,
+                IOUtils.toInputStream(getFileContent(SAMPLE_IMAGE)));
         String id = given().multiPart(tmpFile)
                 .expect()
                 .log()
@@ -442,16 +442,16 @@ public class TestCatalog extends AbstractIntegrationTest {
                 .replaceAll("-", "");
 
         return given().header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_XML)
-                .body(getCswInsertRequest("csw:Record", getFileContent(
-                                CSW_RECORD_RESOURCE_PATH + "/CswRecord", ImmutableMap.of("id",
-                                uuid))))
+                .body(getCswInsertRequest("csw:Record",
+                        getFileContent(CSW_RECORD_RESOURCE_PATH + "/CswRecord",
+                                ImmutableMap.of("id", uuid))))
                 .post(CSW_PATH.getUrl());
     }
 
     private Response ingestXmlViaCsw() {
         return given().header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_XML)
-                .body(getCswInsertRequest("xml", getFileContent(
-                                XML_RECORD_RESOURCE_PATH + "/SimpleXmlNoDecMetacard",
+                .body(getCswInsertRequest("xml",
+                        getFileContent(XML_RECORD_RESOURCE_PATH + "/SimpleXmlNoDecMetacard",
                                 ImmutableMap.of("uri", "http://example.com"))))
                 .post(CSW_PATH.getUrl());
     }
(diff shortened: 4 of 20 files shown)


distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/ServiceManagerImpl.java
```diff
@@ -681,27 +680,36 @@ public void printInactiveBundlesInfo() {
...
-                Enumeration<String> keys = headers.keys();
-
-                while (keys.hasMoreElements()) {
-                    String key = keys.nextElement();
-                    headerString.append(key)
-                            .append("=")
-                            .append(headers.get(key))
-                            .append(", ");
-                }
+                if (bundleService.getInfo(bundle)
```
brendan-hofmann: This section was donated by @coyotesqrl to improve bundle fragment logging.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1907,6 +1918,7 @@ public void testContentDirectoryMonitor() throws Exception {
         cdmProperties.putAll(getServiceManager().getMetatypeDefaults("content-core-directorymonitor",
                 "org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor"));
         cdmProperties.put("monitoredDirectoryPath", tmpDir.toString() + "/");
+        cdmProperties.put("processingMechanism", "delete");
```
brendan-hofmann: Interestingly enough, I found this one by accident. When the default processing mechanism was changed we didn't update the itests, so as a result we had no itest coverage for this mechanism.

clockard:
Looks like there are some unused imports that is causing the build to fail on checkstyle.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/annotations/PaxExamRule.java
```diff
@@ -90,7 +98,9 @@ private void starting(Description description) {
             runAnnotations(PostTestConstruct.class, testClass);
         } catch (Throwable throwable) {
             setupFailed = true;
-            fail(String.format(BEFORE_EXAM_FAILURE_MESSAGE, testClassName, throwable.getMessage()));
+            StringWriter errors = new StringWriter();
```
coyotesqrl: Could refactor this block as it's duped three times.

distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/PaxExamRuleTest.java
```diff
@@ -175,9 +176,13 @@ public void failingBeforeExam() {
         JUnitCore core = new JUnitCore();
         Result result = core.run(FailingBeforeExamTest.class);
 
-        assertThat(result.getFailures()).extracting("message")
-                .containsOnly(EXPECTED_BEFORE_EXAM_ERROR_MESSAGE,
-                        PaxExamRule.EXAM_SETUP_FAILED_MESSAGE);
+        assertThat(result.getFailures()
```
coyotesqrl: This formatting is terrible. Maybe extract the two boolean terms into variables so it becomes more readable?

distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/PaxExamRuleTest.java
```diff
@@ -187,9 +192,13 @@ public void failingAfterExam() {
         JUnitCore core = new JUnitCore();
         Result result = core.run(FailingAfterExamTest.class);
 
-        assertThat(result.getFailures()).extracting("message")
-                .contains(EXPECTED_AFTER_EXAM_ERROR_MESSAGE, FAILING_TEST_MESSAGE);
-
+        assertThat(result.getFailures()
```
coyotesqrl: Same formatting content. It would be fine if you could move the second boolean term to start on the line after the first so it lined up better but in the absence of that, I think two terms will be easier to read.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/annotations/PaxExamRule.java
brendan-hofmann: Got it, guess that didn't get refactored because of the `String.format()`.

distribution/test/itests/test-itests-common/src/test/java/ddf/common/test/PaxExamRuleTest.java
brendan-hofmann: Can do.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -198,8 +198,8 @@ public void tearDown() throws IOException {
     @Test
     public void testCreateStorage() throws IOException {
         String fileName = testName.getMethodName() + ".jpg";
-        File tmpFile = createTemporaryFile(fileName, IOUtils.toInputStream(getFileContent(
-                SAMPLE_IMAGE)));
+        File tmpFile = createTemporaryFile(fileName,
+                IOUtils.toInputStream(getFileContent(SAMPLE_IMAGE)));
```
lessarderic: ❓ There is a lot of "reformatting noise" in this diff. Could we just have the actual changes in this diff and keep the reformatting changes in a separate commit/PR?

brendan-hofmann: They're pretty entangled now. You can tell github to ignore whitespace in the diff and that helps.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/annotations/PaxExamRule.java
```diff
@@ -62,6 +64,12 @@ public PaxExamRule(Object testInstance) {
         this.testInstance = testInstance;
     }
 
+    public static void failWithThrowableStacktrace(Throwable t, String message) {
```
brendan-hofmann: I botched it a bit in my last refactoring, but the intent was to avoid coupling in the other direction, since `PaxExamRule` also should use this helper method, so I was trying to avoid a circular dependency. I can create a separate class just for this method if you'd prefer.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/ServiceManagerImpl.java
```diff
@@ -681,27 +680,36 @@ public void printInactiveBundlesInfo() {
 
     private void printInactiveBundles(Consumer<String> headerConsumer,
             BiConsumer<String, Object[]> logConsumer) {
+        if (bundleService == null) {
```
brendan-hofmann: Those print statements are called fairly often during itest setup. I think the member variable makes sense to avoid having to retrieve the reference for every call. I don't think this class is touched very often, but since there are only two instances I am comfortable with leaving it explicit for now unless you feel strongly otherwise.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -80,16 +72,7 @@ public void init() throws IOException {
...
-            LOGGER.debug("Registering path [{}] with Watch Service.",
-                    configurationDirectoryPath.toString());
-            configurationDirectoryPath.register(watchService, ENTRY_CREATE);
-        } catch (IOException e) {
-            LOGGER.error("Unable to register path [{}] with Watch Service",
-                    configurationDirectoryPath.toString(),
-                    e);
-            throw e;
-        }
+        watchService.addListener(new ListenerImpl());
```
lessarderic: ❗️ Does this get the polling started or does the polling start the first time `checkAndNotify()` is called on L90? I'm asking because we need to make sure we don't process the same configuration file twice, once because `processExistingConfigurationFiles()` is called in the `register()` method and once because `ListenerImpl` gets notified.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -102,67 +85,15 @@ public void register(@NotNull ChangeListener listener) {
...
-        } catch (InterruptedException | RuntimeException e) {
-            LOGGER.debug("The [{}] was interrupted.",
-                    this.getClass()
-                            .getName(),
-                    e);
-            Thread.currentThread()
-                    .interrupt();
+        while (!Thread.currentThread()
+                .isInterrupted()) {
+            watchService.checkAndNotify(); // blocking
```
lessarderic: ❓ Is this really a blocking call? It's not clear from the documentation. If not, we may want to have some kind of sleep between `checkAndNotify` calls to prevent a busy loop.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
lessarderic: Actually, I wonder if we shouldn't just move the `addListener` call and `executorService.execute()` call inside the `register()` method and use the listener's `onStart` method to process existing files. This would guarantee that the processing of existing files and the notifications are all done on the same thread and should eliminate the risk of having a file being processed twice.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -58,19 +53,16 @@
...
-            @NotNull ExecutorService executorService) {
+            @NotNull String fileExtension, @NotNull ExecutorService executorService) {
         notNull(configurationDirectoryPath, "configurationDirectoryPath cannot be null");
         notNull(fileExtension, "fileExtension cannot be null");
-        notNull(watchService, "watchService cannot be null");
         notNull(executorService, "executorService cannot be null");
 
         this.configurationDirectoryPath = configurationDirectoryPath;
-        this.watchService = watchService;
+        this.watchService = new FileAlterationObserver(configurationDirectoryPath.toString());
```
lessarderic: ❓ Do you think we could have used the `FileAlterationMonitor` to replace this whole class?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -199,4 +130,82 @@ private void processExistingConfigurationFiles() {
...
+        public void onDirectoryDelete(File file) {
+        }
+
+        @Override
+        public void onFileCreate(File file) {
+            try {
+                String filename = file.getName();
+                LOGGER.debug("Watcher has been notified. Handling event for [{}].", filename);
+
+                if (!filename.endsWith(fileExtension)) {
```
lessarderic: ❓ Couldn't we use a `IOFileFilter` when we create the `FileAlterationObserver` instead of doing this every time this method gets called?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -199,4 +130,82 @@ private void processExistingConfigurationFiles() {
...
+                if (changeListener != null) {
+                    waitForFileToBeCompletelyWritten(file);
+                    LOGGER.debug("Notifying [{}] of creation for file [{}].",
+                            changeListener.getClass()
+                                    .getName(),
+                            configurationDirectoryPath.resolve(filename));
+                    changeListener.notify(configurationDirectoryPath.resolve(filename));
+                }
+            } catch (InterruptedException e) {
+                LOGGER.debug("The [{}] was interrupted.",
```
lessarderic: ❗️ This is pretty bad if it happens because configuration files will not be processed anymore. We may want to bump up the logging level here (WARN?) and say that the application/bundle needs be restarted.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -102,67 +85,15 @@ public void register(@NotNull ChangeListener listener) {
...
-            LOGGER.debug("The [{}] was interrupted.",
-                    this.getClass()
-                            .getName(),
-                    e);
-            Thread.currentThread()
-                    .interrupt();
+        while (!Thread.currentThread()
+                .isInterrupted()) {
+            watchService.checkAndNotify(); // blocking
         }
```
lessarderic: ❓ Wonder if we should log something when we exit this loop to let admins know that file polling has stopped and that the application/bundle needs to be restarted if the loop stopped because the thread was interrupted but the application is still running.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -199,4 +130,82 @@ private void processExistingConfigurationFiles() {
...
+                        .interrupt();
+            }
+        }
+
+        private void waitForFileToBeCompletelyWritten(File file) throws InterruptedException {
+            Long fileSizeBefore = 0L;
+            Long fileSizeAfter = 0L;
+            do {
+                fileSizeBefore = file.length();
+                Thread.sleep(2000);
```
lessarderic: ❗️ We should use a constant instead of hardcoding `2000` here and on L192.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -199,4 +130,82 @@ private void processExistingConfigurationFiles() {
...
+                        this.getClass()
+                                .getName(),
+                        e);
+                Thread.currentThread()
+                        .interrupt();
+            }
+        }
+
+        private void waitForFileToBeCompletelyWritten(File file) throws InterruptedException {
+            Long fileSizeBefore = 0L;
```
lessarderic: ❓ `long` instead of `Long`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
brendan-hofmann: I'll see if we can, I think it should be possible.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/annotations/PaxExamRule.java
lessarderic: Good point, didn't think of that. A new common logging class seems like a good way to go. Pretty sure we'll add new logging related methods in there before long :-)

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestConfiguration.java
```diff
@@ -680,16 +679,17 @@ public void testConfigStatusFailedImports() throws Exception {
...
-                getFileContentAsStream(INVALID_CONFIG_FILE_2));
         String output = console.runCommand(STATUS_COMMAND);
-
         assertThat(output,
                 containsString(String.format(FAILED_IMPORT_MESSAGE, INVALID_CONFIG_FILE_1)));
-        assertThat(output,
-                containsString(String.format(FAILED_IMPORT_MESSAGE, INVALID_CONFIG_FILE_2)));
         assertThat(Files.exists(getPathToFailedDirectory().resolve(INVALID_CONFIG_FILE_1)),
                 is(true));
+
```
brendan-hofmann: I did not know the ordering was significant, so I reordered them to make the test pass. I will revert this change and investigate further.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
brendan-hofmann: `SuffixFileFilter` is even more convenient. 😉

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -14,38 +14,34 @@
...
 /**
  * Class that monitors changes to files with a specific extension in a directory.
  */
-public class ConfigurationFilesPoller implements Runnable {
+public class ConfigurationFilesPoller implements FileAlterationListener {
+
     private static final Logger LOGGER = LoggerFactory.getLogger(ConfigurationFilesPoller.class);
 
-    private final WatchService watchService;
+    public static final long INTERVAL = 2000L;
```
lessarderic: ✏️ `POLLING_INTERVAL`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -58,100 +54,65 @@
...
-            configurationDirectoryPath.register(watchService, ENTRY_CREATE);
-        } catch (IOException e) {
-            LOGGER.error("Unable to register path [{}] with Watch Service",
-                    configurationDirectoryPath.toString(),
-                    e);
-            throw e;
+            watchService.start();
+        } catch (Exception e) {
+            String msg = "Failed to start, 'Platform :: Migration' must be restarted: ";
+            LOGGER.info(msg, e);
```
lessarderic: ❓If we really want the stack trace at a different log level, should we use `DEBUG` instead of `INFO`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -58,100 +54,65 @@
...
-                // (possibly deleted)
-                if (!key.reset()) {
-                    LOGGER.warn("Configurations in [{}] are no longer able to be observed.",
-                            configurationDirectoryPath.toString());
-                    break;
-                }
+            String filename = file.getName();
+            LOGGER.debug("Watcher has been notified. Handling event for [{}].", filename);
+
+            if (changeListener != null) {
```
lessarderic: ❓Since these methods will only be called when `register` is called and `changeListener` is set, do we still need this check?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -58,100 +54,65 @@
...
+                waitForFileToBeCompletelyWritten(file);
+                LOGGER.debug("Notifying [{}] of creation for file [{}].",
+                        changeListener.getClass()
+                                .getName(),
+                        file.getCanonicalPath());
+                changeListener.notify(Paths.get(file.getCanonicalPath()));
             }
-        } catch (InterruptedException | RuntimeException e) {
-            LOGGER.debug("The [{}] was interrupted.",
+        } catch (InterruptedException | IOException e) {
```
lessarderic: ❓Wonder if we should split this catch into `InterruptedException` and `IOException`. In the former we could do what we do now, but in the latter we should be able to log something and keep going. We should also consider what to do with `RuntimeException`. Thoughts?

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationFilesPollerTest.java
```diff
@@ -86,234 +77,40 @@ public void setup() {
...
-        configurationFilesPoller.register(mockChangeListener);
+        Path mockPath = Files.createTempDirectory("foo");
+        ConfigurationFilesPoller configurationFilesPoller = new ConfigurationFilesPoller(mockPath,
+                FILE_EXT);
 
         // Perform Test
-        configurationFilesPoller.run();
+        configurationFilesPoller.register(mockChangeListener);
+        Files.createTempFile(mockPath, "foo", FILE_EXT);
+        Thread.sleep(5000);
```
lessarderic: ❗️Instead of testing with real files and having to deal with sleeps and such, couldn't we simply test the individual notification methods and make sure they do what they're supposed to? That may require injecting and mocking out the `FileAlterationMonitor` object, but that sounds like a better option imho.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/utils/LoggingUtils.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.junit.Assert.fail;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public class LoggingUtils {
+    public static void failWithThrowableStacktrace(Throwable t, String message) {
+        StringWriter errors = new StringWriter();
+        t.printStackTrace(new PrintWriter(errors));
+        fail(message + errors.toString());
```
brendan-hofmann: I don't think there's a one-size-fits-all answer, so I'd rather leave that up to the caller.

lessarderic: I'm ok with leaving that up to the caller as long as we (JavaDoc) tell them they need to do it. Otherwise we'll end up with a bunch of log messages that don't have a space between `message` and `errors`.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ConfigurationFilesPoller.java
```diff
@@ -58,145 +53,96 @@
...
-        this.executorService = executorService;
         this.fileExtension = fileExtension;
     }
 
-    public void init() throws IOException {
-        LOGGER.debug("Starting {}...",
-                this.getClass()
-                        .getName());
-
+    public void register(@NotNull ChangeListener listener) {
```
lessarderic: ✏️  Just noticed that this class uses the `@NotNull` annotation. We changed our convention since we wrote this class to assume nothing should be `null` unless we use `@Nullable`. Feel free to remove those if you want.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationFilesPollerTest.java
```diff
@@ -86,362 +76,198 @@ public void setup() {
...
 
         // Perform Test
         configurationFilesPoller.register(mockChangeListener);
+        configurationFilesPoller.onFileCreate(file);
 
         // Verify
-        verify(mockChangeListener).notify(configFilePath);
-        verify(mockChangeListener, never()).notify(directoryPath);
-        verify(mockChangeListener, never()).notify(nonConfigFilePath);
+        verify(mockChangeListener).notify(any());
```
lessarderic: ❗️ We should avoid using `any()` here and use a specific matcher instead, otherwise we're not validating that the listener is called with the proper file.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationFilesPollerTest.java
```diff
@@ -86,362 +76,198 @@ public void setup() {
...
-        List<WatchEvent<?>> watchEvents = getSingleMockWatchEvent(mockWatchEvent);
-        WatchKey mockWatchKey = getMockWatchKey(watchEvents);
-        WatchService mockWatchService = getMockWatchService(mockWatchKey);
-        Path mockConfigurationDirectory = getMockConfigurationDirectoryPath(mockPath,
-                mock(Path.class));
-        ConfigurationFilesPoller configurationFilesPoller = createConfigurationFilesPoller(
-                mockWatchService,
-                mockConfigurationDirectory);
-        configurationFilesPoller.register(mockChangeListener);
+        ConfigurationFilesPoller configurationFilesPoller = new ConfigurationFilesPoller(
```
lessarderic: ✏️  Instead of inlining this class everywhere, could we create a static inner class and use that instead?

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationFilesPollerTest.java
```diff
@@ -86,362 +76,198 @@ public void setup() {
...
+                configFilePath,
+                FILE_EXT) {
+            @Override
+            void waitForFileToBeCompletelyWritten(File file) throws InterruptedException {
+                return;
+            }
+        };
+        when(configFilePath.toAbsolutePath()).thenReturn(Paths.get("foo"));
+        File file = Files.createTempFile("foo", "bar")
```
lessarderic: ✏️  Could we move this in a `@Before` method?

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationFilesPollerTest.java
```diff
@@ -86,362 +76,198 @@ public void setup() {
...
-                mock(Path.class));
         ConfigurationFilesPoller configurationFilesPoller = new ConfigurationFilesPoller(
-                mockConfigurationDirectory,
-                FILE_EXT,
-                mockWatchService,
-                mockExecutorService);
+                configFilePath,
+                FILE_EXT);
+        when(configFilePath.toAbsolutePath()).thenReturn(Paths.get("foo"));
+        // will cause an NPE in Paths.get()
```
lessarderic: ❓ Could we do `when(file.getPath()).thenThrow(new RuntimeException())` instead of relying on `Paths.get()` throwing an exception? Another option would be to overwrite `waitForFileToBeCompletelyWritten()` and have it throw the `RuntimeException`.

brendan-hofmann:
@lessarderic I liked the template approach better. I refactored the implementation and added unit tests accordingly.

PR #1795, opened 2017-03-20
DDF-2860 Enhance LDAP tests

ahoffer:
#### What does this PR do?
Adds unit tests for SslLdaploginModule to test against a running LDAP server.
@brendan-hofmann @codymacdonald 
@pklinef
@stustison
Run the unit tests.
Goal is better coverage of DDF's interfaces to LDAP providers.
[DDF-2860](https://codice.atlassian.net/browse/DDF-2860)
#### Checklist:
- [N/A] Documentation Updated
- [N/A] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/platform/security/sts/security-sts-ldaplogin/pom.xml
+++ b/platform/security/sts/security-sts-ldaplogin/pom.xml
@@ -12,7 +12,9 @@
  *
  **/
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <artifactId>security-sts-pom</artifactId>
@@ -80,17 +82,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.32</minimum>
+                                            <minimum>0.72</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.05</minimum>
+                                            <minimum>0.45</minimum>
                                         </limit>
                                         <limit>
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.30</minimum>
+                                            <minimum>0.45</minimum>
                                         </limit>
 
                                     </limits>
@@ -169,11 +171,24 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.github.trevershick</groupId>
+            <artifactId>ldap-test-utils</artifactId>
+            <version>2.0.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-api-mockito</artifactId>
             <version>${powermock.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.16.1</version>
+            <scope>test</scope>
+        </dependency>
+
         <dependency>
             <groupId>org.powermock</groupId>
             <artifactId>powermock-module-junit4-rule-agent</artifactId>
@@ -191,6 +206,15 @@
         <dependency>
             <groupId>org.apache.felix</groupId>
             <artifactId>org.apache.felix.utils</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>ddf.security.encryption</groupId>
+            <artifactId>security-encryption-impl</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>
--- a/platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
+++ b/platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
@@ -135,10 +135,6 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
 
     private ServiceReference serviceReference;
 
-    public SslLdapLoginModule() {
-        initialize();
-    }
-
     protected boolean doLogin() throws LoginException {
 
         //--------- EXTRACT USERNAME AND PASSWORD FOR LDAP LOOKUP -------------
@@ -410,6 +406,7 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
     public void initialize(Subject subject, CallbackHandler callbackHandler,
             Map<String, ?> sharedState, Map<String, ?> options) {
         super.initialize(subject, callbackHandler, options);
+        installEncryptionService();
         connectionURL = (String) options.get(CONNECTION_URL);
         connectionUsername = (String) options.get(CONNECTION_USERNAME);
         connectionPassword = getDecryptedPassword((String) options.get(CONNECTION_PASSWORD));
@@ -521,7 +518,7 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
         return ImmutableSet.copyOf(principals);
     }
 
-    private void initialize() {
+    private void installEncryptionService() {
 
         BundleContext bundleContext = getContext();
         if (null != bundleContext) {
--- a/platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
+++ b/platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
@@ -16,8 +16,10 @@ package ddf.ldap.ldaplogin;
 
 import java.io.IOException;
 import java.security.GeneralSecurityException;
+import java.security.Principal;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -55,6 +57,8 @@ import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.ServiceReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableSet;
 
 import ddf.security.common.audit.SecurityLogger;
 import ddf.security.encryption.EncryptionService;
@@ -105,7 +109,7 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
 
     private String connectionUsername;
 
-    private String connectionPassword;
+    private char[] connectionPassword;
 
     private String userBaseDN;
 
@@ -114,6 +118,8 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
     private boolean userSearchSubtree = true;
 
     private String roleBaseDN;
+
+    private EncryptionService encryptionService;
 
     private String roleFilter;
 
@@ -127,7 +133,15 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
 
     private LDAPConnectionFactory ldapConnectionFactory;
 
+    private ServiceReference serviceReference;
+
+    public SslLdapLoginModule() {
+        initialize();
+    }
+
     protected boolean doLogin() throws LoginException {
+
+        //--------- EXTRACT USERNAME AND PASSWORD FOR LDAP LOOKUP -------------
         Callback[] callbacks = new Callback[2];
         callbacks[0] = new NameCallback("Username: ");
         callbacks[1] = new PasswordCallback("Password: ", false);
@@ -135,8 +149,10 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
         try {
             callbackHandler.handle(callbacks);
         } catch (IOException ioException) {
+            boolean result;
             throw new LoginException(ioException.getMessage());
         } catch (UnsupportedCallbackException unsupportedCallbackException) {
+            boolean result;
             throw new LoginException(unsupportedCallbackException.getMessage()
                     + " not available to obtain information from user.");
         }
@@ -164,10 +180,15 @@ public class SslLdapLoginModule extends AbstractKarafLoginModule {
         if (tmpPassword == null) {
             tmpPassword = new char[0];
         }
-        principals = new HashSet<>();
 
-        Connection connection = null;
-        String userDn = null;
+        //---------------------------------------------------------------------
+        // RESET OBJECT STATE AND DECLARE LOCAL VARS
+        principals = new HashSet<>();
+        Connection connection;
+        String userDn;
+        //---------------------------------------------------------------------
+
+        //------------- CREATE CONNECTION #1 ----------------------------------
         try {
             connection = ldapConnectionFactory.getConnection();
         } catch (LdapException e) {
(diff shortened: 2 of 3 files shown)


platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -337,28 +372,6 @@ protected boolean doLogin() throws LoginException {
         return true;
     }
 
-    void validateUsername(String username) throws LoginException {
```
codymacdonald: This method was moved below the public methods.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -477,22 +479,61 @@ protected LDAPConnectionFactory createLdapConnectionFactory(String url, Boolean
...
+            if (hasBadCharacters) {
+                throw new LoginException("Bad characters present in user name.");
+            }
+        }
+    }
 
-        } catch (Exception le) {
-            throw new LoginException("Username [" + user
-                    + "] could not log in successfuly using LDAP authentication due to an exception");
+    Set<Principal> getPrincipals() {
```
codymacdonald: Added this method so unit test could test the result.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -477,22 +479,61 @@ protected LDAPConnectionFactory createLdapConnectionFactory(String url, Boolean
...
+        return ImmutableSet.copyOf(principals);
+    }
+
+    protected char[] getDecryptedPassword(String encryptedPassword) {
```
codymacdonald: Created a method for decrypting the password so that it can be overridden by the test.

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.junit.contrib.java.lang.system.ProvideSystemProperty;
+
+import com.github.trevershick.test.ldap.LdapServerResource;
+import com.github.trevershick.test.ldap.annotations.LdapAttribute;
+import com.github.trevershick.test.ldap.annotations.LdapConfiguration;
+import com.github.trevershick.test.ldap.annotations.LdapEntry;
+
+//Create and populate LDAP server
+// @formatter:off
+@LdapConfiguration(
```
codymacdonald: Creates an instance of an UnboundID LDAP server and populates it with test data.

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
```diff
@@ -0,0 +1,214 @@
...
+    @Test
+    public void testBadUserPassword() throws LoginException {
+        CallbackHandler mockHandler = getNamePasswordHandler("tstark",
+                "Ce n'est pas un mot de passe");
+        SslLdapLoginModule module = getTestModule();
+        module.initialize(new Subject(), mockHandler, new HashMap<String, String>(), options);
+        module.login();
+        assertThat("Expected no principals assigned to user", module.getPrincipals(), empty());
+    }
+
```
codymacdonald: Created three tickets, two shown below and DDF-2875.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -477,22 +479,61 @@ protected LDAPConnectionFactory createLdapConnectionFactory(String url, Boolean
...
-            boolean isLoggedIn = doLogin();
-
-            if (!isLoggedIn) {
-                SecurityLogger.audit("Username [" + user + "] failed LDAP authentication.");
+    void validateUsername(String username) throws LoginException {
+        boolean hasBadCharacters = false;
+        for (int i = 0; i < username.length(); i++) {
+            char curChar = username.charAt(i);
+            switch (curChar) {
+            case '\\':
```
AzGoalie: Should we move this into an array in case need to add more later? Could also use white list instead (only allow a-z A-Z 0-9 etc).

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -477,22 +479,61 @@ protected LDAPConnectionFactory createLdapConnectionFactory(String url, Boolean
...
+            case '<':
+            case '>':
+            case ';':
+            case '#':
+                hasBadCharacters = true;
+                break;
             }
-            return isLoggedIn;
+            if (hasBadCharacters) {
+                throw new LoginException("Bad characters present in user name.");
```
AzGoalie: Should we be less specific on why the login failed?

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -477,22 +479,61 @@ protected LDAPConnectionFactory createLdapConnectionFactory(String url, Boolean
...
+    }
+
+    protected char[] getDecryptedPassword(String encryptedPassword) {
+
+        char[] decryptedPassword = null;
+        BundleContext bundleContext = getContext();
+
+        if (null != bundleContext) {
+            ServiceReference ref;
+            ref = bundleContext.getServiceReference(EncryptionService.class.getName());
```
AzGoalie: Should this be passed in / set from blueprint instead of manually using osgi?

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
AzGoalie: Does this library have an object based builder instead of using annotations and having to disable the formatter?

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -386,60 +399,50 @@ protected BundleContext getContext() {
...
 
-                if (encryptionService != null) {
-                    String decryptedPassword = encryptionService.decryptValue(
-                            option.get(CONNECTION_PASSWORD));
-                    option.put(CONNECTION_PASSWORD, decryptedPassword);
-                } else {
-                    LOGGER.info("Encryption service reference for ldap was null.");
-                }
+    /**
+     * Added additional logging to the security LOGGER.
```
emanns95: Weird verb tense. Feel like it should say "adds"

ahoffer: Do we even need this comment? Does anyone mind if I just delete it?

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
ahoffer: Hmm. We should probably ask Scott. It's his code. I read somewhere that this is fastest way to do it, but I don't know if that was the reason. I'll tag @stustison in the comment to give him a chance to chime in.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
ahoffer: I'll have to defer to Scott again. @stustison

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
ahoffer: I wondered that myself. I fond examples where inject the `blueprintBundleContext`. Is that the same `BundleContext` returned by `getContext` @stustison do you know if we should be injecting the bundle context via blueprint?

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
ahoffer: Sure, you can build it with Java statements:
```// Create the configuration to use for the server. It looks like this.
InMemoryDirectoryServerConfig config =
     new InMemoryDirectoryServerConfig("dc=example,dc=com");
config.addAdditionalBindCredentials("cn=Directory Manager", "password");```

It has some advantages over the annotations, like being able to use string constants to replace duplicated strings. I just liked using declarative syntax to create the server. It felt more like test scaffolding to me. I'm willing to entertain the idea of pulling out the annotations and replacing them with Java statements. @brendan-hofmann @emanns95, how do you feel?

File: platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java (hunk 6 above)
Comment (stustison) [inline platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java]: Yeah, I did this because it was the most performant way of completing the check and it is the OWASP recommendation. We could change it to a whitelist instead, but there are a lot more characters that we'd need to potentially allow. You could use a regex for most of them, but that will slow things down a bit. I'm good with either, if you want to change it. Whitelisting is certainly a security best practice (as long as we don't overly restrict it) and what is already here is the OWASP recommendation, so either is good with me.

File: platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java (hunk 7 above)
Comment (stustison) [inline platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java]: This exception message is actually tossed on the floor and never returned to the client. We are actually missing an audit for this though down in the `login()` method for the exception case.

File: platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java (hunk 8 above)
Comment (stustison) [inline platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java]: You should never inject the bundle context via blueprint. So, no. This is done this way because it is a JAAS service.

File: platform/security/sts/security-sts-ldaplogin/pom.xml [hunk 10]
```diff
@@ -169,12 +170,24 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.github.trevershick</groupId>
+            <artifactId>ldap-test-utils</artifactId>
+            <version>2.0.0</version>
```
ahoffer: Yes. Thanks.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
ahoffer: Unless there is much ardor about it, I plan to leave the method as-it.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
AzGoalie: @stustison isn't the EncryptionService able to be injected though?

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
brendan-hofmann: I didn't get tagged because it's in the code block for some reason. 😄 

I do not like disabling the formatter. I would prefer to reenable it, and if you feel the annotations look too ugly after they have been formatted, that's your answer.

ahoffer: So much love for the formatter?

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -386,60 +402,53 @@ protected BundleContext getContext() {
...
+                    + "] could not log in successfuly using LDAP authentication due to an exception";
 
-            try {
-                ldapConnectionFactory = createLdapConnectionFactory(connectionURL, startTls);
-            } catch (LdapException e) {
-                LOGGER.info(
-                        "Unable to create LDAP Connection Factory. LDAP log in will not be possible.",
-                        e);
+            if (!isLoggedIn) {
+                SecurityLogger.audit("Username [" + user + "] failed LDAP authentication.");
```
pklinef: It think this allows the potential for log injection from user input of their username.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
pklinef: I agree with @AzGoalie.  It looks like we could have blueprint inject the EcryptionService and then we could inject a mock EcryptionService instead of overriding this method.

Here is an example.
https://github.com/codice/ddf/blob/17f05d54571d3b6e89146b990cc29d9a03fefb4e/platform/security/idp/security-idp-client/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L36

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
ahoffer: `SecurityLogger` now has a method `cleanAndEncode` that scrubs newlines from the message to prevent log injection attacks. So we should be good on that score.

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
ahoffer: Just for the heck of it, I tried injecting the EncryptionService. Blueprint does set the encryption service in an instance of SslLdapLoginModule, but a different instance is called when user attempts to authenticate. The encryption service is not set (null) in that instance.  

Encryption service set

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
```diff
@@ -0,0 +1,208 @@
...
+    public void testDigestMd5Sasl() {
+        // TODO: DDF-2877 - Enhance LDAP test for alt protocols
+
+    }
+
+    private SslLdapLoginModule getTestModule() {
+        SslLdapLoginModule module = new SslLdapLoginModule();
+        module.setEncryptionService(new EncryptionServiceImpl());
+        return module;
+    }
```
ahoffer: Refactored tested-class and the test to inject the encryption service. The test no longer creates an anonymous class that overrides the getDecryptedPassword() method. @pklinef

platform/security/sts/security-sts-ldaplogin/src/test/java/ddf/ldap/ldaplogin/LdapModuleTest.java
```diff
@@ -0,0 +1,208 @@
...
+import org.junit.contrib.java.lang.system.ProvideSystemProperty;
+
+import com.github.trevershick.test.ldap.LdapServerResource;
+import com.github.trevershick.test.ldap.annotations.LdapAttribute;
+import com.github.trevershick.test.ldap.annotations.LdapConfiguration;
+import com.github.trevershick.test.ldap.annotations.LdapEntry;
+
+import ddf.security.encryption.impl.EncryptionServiceImpl;
+
+//Create and populate LDAP server
```
ahoffer: I'm disinclined rewrite the LDAP configuration at this point. I'm willing to revisit that position when we add the LDAP tests for the claim handlers.

brendan-hofmann:
Heroed on windows. 👍

platform/security/sts/security-sts-ldaplogin/src/main/java/ddf/ldap/ldaplogin/SslLdapLoginModule.java
```diff
@@ -127,16 +133,26 @@
 
     private LDAPConnectionFactory ldapConnectionFactory;
 
+    private ServiceReference serviceReference;
+
+    public SslLdapLoginModule() {
```
stustison: Didn't notice this before, but why did we add this? This is going to cause initialize to get called twice.

ahoffer: Added it on Phil's recommendation to inject the EncryptionService instead of overriding the LoginModule's decrypt-password method. I didn't realize someone was already calling initialize. I'll delete the constructor and re-test.

platform/security/sts/security-sts-ldaplogin/pom.xml
ahoffer: fixed

PR #1798, opened 2017-03-21
DDF-2893 Refactor response filter to support cache-control

brendan-hofmann:
#### What does this PR do?
Adds cache-control to the response filter to fix browser caching.
@mcalcote 
@AzGoalie 
@andrewkfiedler
@pklinef
Install with `?dev=true` so the cert hostname matches your machine. Trust the new demo-CA in your browser. In the admin ui, verify that the platform-response module has the new cache-control configuration. Go to various pages such as admin or the catalog ui and verify that resources are shown as `cached` or `loaded from disk` instead of being fetched each time. You can check this using the network tab of the browser developer tools. Test this in Chrome and Firefox. Note that Firefox has an unrelated bug with url fragments, so it may not cache unless the url navigated to does not have a blank fragment in it.
[DDF-2893](https://codice.atlassian.net/browse/DDF-2893)
image
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests


platform/platform-filter-response/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -25,6 +25,9 @@
         <AD description="The HTTP X-XSS-Protection response header is a feature that stops pages from loading when they detect reflected cross-site scripting (XSS) attacks."
             name="X-XSS-Protection" id="xXssProtection" required="true" type="String"
             default="1; mode=block"/>
+        <AD description="The Cache-Control header specifies the behavior for browsers and proxies to cache web resources."
+            name="Cache-Control" id="cacheControl" required="true" type="String"
+            default="Cache-Control: \"public, max-age=604800, immutable\""/>
```
brendan-hofmann: Should `public` be the default, or should `private`? Does this need to be in the hardening guide?

platform/platform-filter-response/pom.xml
```diff
@@ -60,7 +60,7 @@
                                     <element>BUNDLE</element>
                                     <limits>
                                         <limit>
-                                            <!--The ResponseSecurityFilter consists of several setters and a doFilter method that sets several strings and performs no logic-->
+                                            <!--ResponseFilterlter consists of several setters and a doFilter method that sets several strings and performs no logic-->
```
codymacdonald: ResponseFilterlter looks like a misspelling.

platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
```diff
@@ -29,29 +29,36 @@
...
+
     public static final String DEFAULT_XSS_PROTECTION_VALUE = "1; mode=block";
 
     public static final String DEFAULT_X_FRAME_OPTIONS_VALUE = "SAMEORIGIN";
 
     public static final String DEFAULT_CONTENT_SECURITY_POLICY =
             "default-src 'none'; connect-src 'self'; script-src 'self'; style-src 'self'; img-src 'self'";
 
+    public static final String DEFAULT_CACHE_CONTROL_VALUE =
+            "Cache-Control: \"public, max-age=604800, immutable\"";
```
codymacdonald: How did you decide on the value for max-age?

platform/platform-filter-response/src/main/resources/OSGI-INF/metatype/metatype.xml
codymacdonald: I'm going to guess "private". @stustison 

_public
Indicates that the response MAY be cached by any cache, even if it would normally be non-cacheable or cacheable only within a non- shared cache. (See also Authorization, section 14.8, for additional details.)
private
Indicates that all or part of the response message is intended for a single user and MUST NOT be cached by a shared cache. This allows an origin server to state that the specified parts of the
response are intended for only one user and are not a valid response for requests by other users. A private (non-shared) cache MAY cache the response._

stustison: Yeah the default should be private. We aim for more secure out of the box, generally.

platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
brendan-hofmann: Arbitrarily chose one week as a reasonable default. 604800 is one week in seconds.

rzwiefel: That would be a great comment to add in right here 😄

brendan-hofmann:
@rzwiefel It is. I would argue that the caching is working as intended though. If there is an expectation that that endpoint be 'fresh', there are a few different ways to work around that.

brendan-hofmann:
@mcalcote good catch on the docs.

rzwiefel:
@brendan-hofmann But doesn't that mean you could theoretically be caching data of _varying confidentialities_ on a users local system.

brendan-hofmann:
Yes, but nothing that should exceed the permissions of the user or the machine they are using, if everything is configured correctly. I'm not sure I am following your concern correctly, maybe we should sync up today.

platform/platform-filter-response/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -25,10 +26,13 @@
         <AD description="The HTTP X-XSS-Protection response header is a feature that stops pages from loading when they detect reflected cross-site scripting (XSS) attacks."
             name="X-XSS-Protection" id="xXssProtection" required="true" type="String"
             default="1; mode=block"/>
+        <AD description="The Cache-Control header specifies the behavior for browsers and proxies to cache web resources."
+            name="Cache-Control" id="cacheControl" required="true" type="String"
+            default="private, max-age=604800, immutable"/>
```
brendan-hofmann: The admin UI does not correctly render this default. It drops everything after the first comma. I think there might be another bugfix that needs to be made or get backported along with this one.

pklinef:
I kind of wish we had a way to control when to inject headers based on context path kind of like the web context policy.  I could see wanting to disable caching for some context paths or specific resources.

platform/platform-filter-response/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -25,10 +26,13 @@
         <AD description="The HTTP X-XSS-Protection response header is a feature that stops pages from loading when they detect reflected cross-site scripting (XSS) attacks."
             name="X-XSS-Protection" id="xXssProtection" required="true" type="String"
             default="1; mode=block"/>
+        <AD description="The Cache-Control header specifies the behavior for browsers and proxies to cache web resources."
```
pklinef: I thought this configuration would be more generic where it would be a list of `header=value` that should be injected into all responses.

brendan-hofmann: I debated that (and can still make that change if necessary). It is a bit weird because this is not the only filter that adds header values. Would we collapse everything into one list configuration? Only some things? Just this plugin? I'm also not aware of any compelling reasons for being able to add arbitrary header values, but could easily be persuaded otherwise. It just doesn't seem like the kind of thing an admin should be doing, or if they really want to add arbitrary header values, they're probably capable of writing and deploying their own filter.

platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
andrewkfiedler: Anyway we can make it so the cache is automatically invalidated for different versions?

emanns95:
Manually verified the changes work as specified :+1:

platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
brendan-hofmann: You mean different versions of DDF? I can't think of an easy way, though I'm open to suggestions. However, the different versions would have to be at the same location from the browser's perspective. I think that is unlikely to be a pain point outside of a dev environment, where we are used to clearing our cache anyways.

LinkMJB:
build now

PR #1802, opened 2017-03-22
DDF-2825 [2.10.x] Contributes Configurator service and transactional implementation

coyotesqrl:
#### What does this PR do?
This pulls the Configurator service from the [admin-beta-console Connexta repository](https://github.com/connexta/admin-console-beta).

History on the service prior to this contribution can be found from this location: https://github.com/connexta/admin-console-beta/tree/f910923b5ff2b326cfb5654fb954c8ac851d9068/configurator

@stustison @paouelle 

@lessarderic
@pklinef

This is a new service with no callers in DDF. To properly test, a caller can be written to utilize the Configurator to update/create services, properties files, etc.

This functionality begins the process of extracting configuration away from low-level constructs to support simplified administration and node setup.

[DDF-2825](https://codice.atlassian.net/browse/DDF-2825)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/platform/configurator/configurator-api/pom.xml
+++ b/platform/configurator/configurator-api/pom.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>ddf.platform</groupId>
+        <artifactId>configurator</artifactId>
+        <version>2.10.2-SNAPSHOT</version>
+    </parent>
+
+    <name>DDF :: Configurator :: API</name>
+    <artifactId>configurator-api</artifactId>
+    <packaging>bundle</packaging>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                    </instructions>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.00</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.00</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.00</minimum>
+                                        </limit>
+                                    </limits>
+                                </rule>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
+++ b/platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
@@ -1,0 +1,195 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface Configurator {
+    /**
+     * Sequentially invokes all the {@link Operation}s, committing their changes. If a failure
+     * occurs during the processing, a rollback is attempted of those handlers that had already been
+     * committed.
+     * <p>
+     * In the case of a successful commit, changes should be logged.
+     *
+     * @param auditMessage In the case of a successful commit, the message to pass to a
+     *                     {@code SecurityLogger}
+     * @param auditParams  In the case of a successful commit, the optional parameters to pass to a
+     *                     {@code SecurityLogger} to be interpolated into the message
+     * @return report of the commit status, whether successful, successfully rolled back, or partially
+     * rolled back with errors
+     */
+    OperationReport commit(String auditMessage, String... auditParams);
+
+    /**
+     * Sequentially invokes all the {@link Operation}s, committing their changes. If a failure
+     * occurs during the processing, a rollback is attempted of those handlers that had already been
+     * committed.
+     *
+     * @return report of the commit status, whether successful, successfully rolled back, or partially
+     * rolled back with errors
+     */
+    OperationReport commit();
+
+    /**
+     * Starts the bundle with the given name.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String startBundle(String bundleSymName);
+
+    /**
+     * Stops the bundle with the given name.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String stopBundle(String bundleSymName);
+
+    /**
+     * Determines if the bundle with the given name is started.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return true if started; else, false
+     */
+    boolean isBundleStarted(String bundleSymName);
+
+    /**
+     * Installs and starts the feature with the given name.
+     *
+     * @param featureName the name of the feature
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String startFeature(String featureName);
+
+    /**
+     * Stops the feature with the given name.
+     *
+     * @param featureName the name of the feature
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String stopFeature(String featureName);
+
+    /**
+     * Determines if the feature with the given name is started.
+     *
+     * @param featureName the name of the feature
+     * @return true if started; else, false
+     */
+    boolean isFeatureStarted(String featureName);
+
+    /**
+     * Creates a property file in the system with the given set of new key:value pairs.
+     *
+     * @param propFile   the property file to create
+     * @param properties the set of key:value pairs to save to the property file
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String createPropertyFile(Path propFile, Map<String, String> properties);
+
+    /**
+     * Deletes a property file in the system.
+     *
+     * @param propFile the property file to delete
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String deletePropertyFile(Path propFile);
+
+    /**
+     * Updates a property file in the system with the given set of new key:value pairs.
+     *
+     * @param propFile    the property file to update
+     * @param properties  the set of key:value pairs to save to the property file
+     * @param keepIgnored if true, then any keys already in the property file will retain their
+     *                    initial values if they are excluded from the {@code properties} param; if
+     *                    false, then the only properties that will be in the updated file are those
+     *                    provided by the {@code properties} param
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String updatePropertyFile(Path propFile, Map<String, String> properties, boolean keepIgnored);
+
+    /**
+     * Gets the current key:value pairs set in the given property file.
+     *
+     * @param propFile the property file to query
+     * @return the current set of key:value pairs
+     */
+    Map<String, String> getProperties(Path propFile);
+
+    /**
+     * Updates a bundle configuration file in the system with the given set of new key:value pairs.
+     *
+     * @param configPid   the configId of the bundle configuration file to update
+     * @param configs     the set of key:value pairs to save in the configuration
+     * @param keepIgnored if true, then any keys already in the config file will retain their
+     *                    initial values if they are excluded from the {@code properties} param; if
+     *                    false, then the only config entires that will be in the updated file are those
+     *                    provided by the {@code properties} param
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String updateConfigFile(String configPid, Map<String, Object> configs, boolean keepIgnored);
+
+    /**
+     * Gets the current key:value pairs set in the given configuration file.
+     *
+     * @param configPid the configId of the bundle configuration file to query
+     * @return the current set of key:value pairs
+     */
+    Map<String, Object> getConfig(String configPid);
+
+    /**
+     * Creates a new managed service for the given factory.
+     *
+     * @param factoryPid the factoryPid of the service to create
+     * @param configs    the set of key:value pairs to save in the new managed service's configuration
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String createManagedService(String factoryPid, Map<String, Object> configs);
+
+    /**
+     * Deletes a managed service.
+     *
+     * @param configPid the configPid of the instance of the service to delete
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String deleteManagedService(String configPid);
+
+    Map<String, Map<String, Object>> getManagedServiceConfigs(String factoryPid);
+
+    /**
+     * Retrieves the service reference. The reference should only be used for reading purposes,
+     * any changes should be done through a commit
+     *
+     * @param serviceClass - Class of service to retrieve
+     * @return first found service reference of serviceClass
+     * @throws ConfiguratorException if any errors occur
+     */
+    <S> S getServiceReference(Class<S> serviceClass) throws ConfiguratorException;
+}
--- a/platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Operation.java
+++ b/platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Operation.java
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+/**
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface Operation<T, S> {
+    T commit() throws ConfiguratorException;
+
+    T rollback() throws ConfiguratorException;
+
+    S readState() throws ConfiguratorException;
+}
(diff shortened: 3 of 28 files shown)


platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Operation.java
```diff
@@ -0,0 +1,26 @@
...
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+/**
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface Operation<T, S> {
```
rzwiefel: Assuming you know about the other `Operation` interface.
Also is this not more of a 'Transaction' than an 'Operation' given the interface methods?

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,32 @@
...
+package org.codice.ddf.admin.configurator;
+
+import java.util.List;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface OperationReport {
+    boolean txactSucceeded();
```
rzwiefel: `txact`? Usually transaction is abbreviated as `tx`

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Result.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface Result {
+    boolean isTxactSucceeded();
+
+    Status getStatus();
+
+    Optional<Throwable> getBadOutcome();
```
rzwiefel: lol

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Status.java
```diff
@@ -0,0 +1,22 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public enum Status {
+    COMMIT_PASSED, COMMIT_FAILED, SKIPPED, ROLLBACK_PASSED, ROLLBACK_FAILED;
```
rzwiefel: Assuming there is no need for a state in which none of these has happened.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
rzwiefel: or just let the java flow through you and go full verbose -- `transactionSucceeded`

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Status.java
coyotesqrl: Correct. This is the complete set of possible states.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
coyotesqrl: `tx` is usually an abbreviation for "transmit" and its siblings (or for Texas). This is a fairly common abbreviation for `trans` `act`.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Operation.java
coyotesqrl: It's not an uncommon noun for interface/class names. I'm not especially worried that it and this will be used in the same context so there should be no name collisions requiring disambiguation.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
```diff
@@ -0,0 +1,203 @@
...
+            boolean loadCurrentProps) {
+        this.configFile = configFile.toFile();
+        this.configs = new HashMap<>(configs);
+
+        currentProperties = new HashMap<>();
+        if (loadCurrentProps) {
+            Properties result;
+            try {
+                Properties props = new Properties();
+                try (FileInputStream in = new FileInputStream(configFile.toFile())) {
```
paouelle: Here is a question, typically for reading a file, I never liked using the try() approach as it will throw an exception if closing the file fails. This is crucial when writing to a file as you want to make sure it is available after but for reading ... not so much. Wouldn't dealing with the input stream manually using something like apache.io IOUtils.closeQuiet() (don't recall the exact name) which basically ignores the closing exception be more appropriate?

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
```diff
@@ -0,0 +1,203 @@
...
+    @Override
+    public Void rollback() throws ConfiguratorException {
+        saveProperties(currentProperties);
+
+        return null;
+    }
+
+    @Override
+    public Map<String, String> readState() throws ConfiguratorException {
+        return currentProperties;
```
paouelle: Isn't it dangerous to return a reference to the internal maps? Do we account for outside change to it?

platform/configurator/configurator-api/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>ddf.platform</groupId>
+        <artifactId>configurator</artifactId>
+        <version>2.10.2-SNAPSHOT</version>
+    </parent>
+
+    <name>DDF :: Configurator :: API</name>
```
stustison: DDF :: Platform :: Configurator :: API

platform/configurator/configurator-api/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <artifactId>configurator-api</artifactId>
```
stustison: platform-configurator-api

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
stustison: Full words would make more sense to me here since all the other methods use full words. Should be consistent.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Result.java
stustison: Hm... I'll second the "lol" and say that this seems kind of odd to me from an API standpoint. I'd expect something more like `getOutcome` and that `Outcome` could then have an `isGood` or `isBad` method on it. Just feels a little weird to me.

platform/configurator/configurator-impl/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+        <dependency>
+            <groupId>ddf.admin.core</groupId>
+            <artifactId>admin-core-api</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.karaf.bundle</groupId>
+            <artifactId>org.apache.karaf.bundle.core</artifactId>
+            <version>${karaf.version}</version>
```
coyotesqrl: We actually tend not to use dependency management quite as much in our poms because we've had the odd issue in the past. Our version properties are generally defined in the root pom and managed there so this will get us a consistent version.

platform/configurator/configurator-impl/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>ddf.platform</groupId>
+        <artifactId>configurator</artifactId>
+        <version>2.10.2-SNAPSHOT</version>
+    </parent>
+
+    <name>DDF :: Configurator :: Transactional Implementation</name>
```
stustison: Same comment here about the name and artifact id

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/AdminOperation.java
```diff
@@ -0,0 +1,150 @@
...
+        this.pid = pid;
+        this.configs = new HashMap<>(configs);
+        this.keepIgnored = keepIgnored;
+        this.cfgAdmMbean = cfgAdmMbean;
+
+        try {
+            currentProperties = cfgAdmMbean.getProperties(pid);
+        } catch (IOException e) {
+            LOGGER.debug("Error getting current configuration for pid {}", pid, e);
+            throw new ConfiguratorException("Internal error");
```
stustison: @paouelle Generally, we throw a generic exception if it is one we expect to be returned to a client for security reasons. I'm not sure if this is the case here. We usually expect the generic messages to be returned (with the actuals logged) from wherever the actual endpoint is and then flow stack traces up from below that.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private final BundleContext bundleContext;
+
+    private final Bundle bundle;
+
+    private final boolean initActivationState;
+
+    private BundleOperation(String bundleSymName, boolean activate, BundleContext bundleContext) {
+        this.newState = activate;
+        this.bundleContext = bundleContext;
```
stustison: You should never inject and save the BundleContext. I'm not sure if it is grabbed and used just in this class and then this class is tossed, because I don't understand the full structure, yet. However, if you need the bundle context, you really need to do it using FrameworkUtil. It is ok to inject the bundle context via blueprint to a constructor, do something with it, and then throw it away, but hanging onto it will cause an exception once the bundle context goes stale.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    /**
+     * Starts the bundle with the given name.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    public String startBundle(String bundleSymName) {
```
stustison: Should all of these methods be validating their input?

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
```diff
@@ -0,0 +1,425 @@
...
+
+    /**
+     * Creates a property file in the system with the given set of new key:value pairs.
+     *
+     * @param propFile   the property file to create
+     * @param properties the set of key:value pairs to save to the property file
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    public String createPropertyFile(Path propFile, Map<String, String> properties) {
```
stustison: When creating the file, do we check that it is within the valid locations constrained to just our system. In other words, make sure that we can't write files outside of DDF.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
```diff
@@ -0,0 +1,425 @@
...
+    }
+
+    /**
+     * Deletes a property file in the system.
+     *
+     * @param propFile the property file to delete
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    public String deletePropertyFile(Path propFile) {
```
stustison: Same question here about deleting. We should validate that they are trying to delete a property file and that it is within the bounds of the paths we deem acceptable.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/FeatureOperation.java
```diff
@@ -0,0 +1,135 @@
...
+    private final boolean newState;
+
+    private final BundleContext bundleContext;
+
+    private final boolean initActivationState;
+
+    private FeatureOperation(String featureName, boolean newState, BundleContext bundleContext) {
+        this.featureName = featureName;
+        this.newState = newState;
+        this.bundleContext = bundleContext;
```
stustison: Same comment here on bundle context, just want to make sure this is dynamic and not being injected via blueprint or only set once (even if you're using FrameworkUtil somewhere).

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/AdminOperation.java
coyotesqrl: This one would potentially bubble all the way up to the end user, exposing internals.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
coyotesqrl: Definitely not being injected. It's retrieved from `FrameworkUtil` in the `ConfiguratorImpl` when it constructs this class and this class is short-lived.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
coyotesqrl: Good catch.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/AdminOperation.java
paouelle: Does our log have a way to track threads or request id to help correlate them? Caused exceptions helps simplify that but I definitely agree with not exposing internals.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Operation.java
coyotesqrl: This interface actually doesn't need to be in the public api; it is an implementation detail and will be moved in a subsequent commit on this PR. As such, I'm less concerned about the documentation.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
coyotesqrl: Will rename method to `isTransactionSucceeded`.

platform/configurator/configurator-api/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.00</minimum>
```
coyotesqrl: As an api bundle, this only contains interfaces. Unit tests are a bit of overkill for them, but it means the coverage in this bundle is 0.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Deletes a managed service.
+     *
+     * @param configPid the configPid of the instance of the service to delete
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String deleteManagedService(String configPid);
+
+    Map<String, Map<String, Object>> getManagedServiceConfigs(String factoryPid);
```
coyotesqrl: Absolutely. Oversight on my part.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Retrieves the service reference. The reference should only be used for reading purposes,
+     * any changes should be done through a commit
+     *
+     * @param serviceClass - Class of service to retrieve
+     * @return first found service reference of serviceClass
+     * @throws ConfiguratorException if any errors occur
+     */
+    <S> S getServiceReference(Class<S> serviceClass) throws ConfiguratorException;
```
coyotesqrl: I'm following Karaf's underlying implementation here. While technically it does return an implementation class and not an actual `S`, for the purposes of callers it should be treated as an `S`.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,32 @@
...
+
+    Result getResult(String key);
+
+    List<Result> getFailedResults();
```
coyotesqrl: I can provide documentation (or even annotation) guidance here, but I don't know that I want to lock any other implementations that may be created into an unmodifiable collection for this.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ManagedServiceOperation.java
```diff
@@ -0,0 +1,197 @@
...
+        @Override
+        public String rollback() throws ConfiguratorException {
+            deleteByPid(newConfigPid);
+            return null;
+        }
+    }
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ManagedServiceOperation.class);
+
+    protected String factoryPid;
```
coyotesqrl: The problem with moving the `factoryPid` initialization to the base class is that in the case of the `DeleteHandler` implementor I have to query to find the `factoryPid` and the query can throw a checked exception. I could get around that but it would be a lot of extra code that I’m not sure I see the value of in an internal implementation.

I will play around with it a bit more to see if I can come up with something that's not overly complex, but right now my feeling is that the exposure is low.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/OperationReportImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        return results.values()
+                .stream()
+                .allMatch(Result::isTxactSucceeded);
+    }
+
+    public Result getResult(String key) {
+        return results.get(key);
+    }
+
+    public List<Result> getFailedResults() {
```
coyotesqrl: I agree in principal, but in this case I have some idea of what callers will want to do with it (and that it will be a small collection, regardless). Handing a stream off would just push the requirement to convert to a collection up a level and not buy a lot.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
coyotesqrl: That's an interesting point. I'm not sure what I would make of a file that couldn't be closed correctly after being opened, so I'm not sure if it should be treated as an error or not here.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
coyotesqrl: Not especially. Good catch.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Result.java
coyotesqrl: The problem I have is that the `Status` is _almost_ orthogonal to this failure exception. It's an enum indicating what happened; in the case of failure, this field will have a value in it. The `Outcome` you suggest is what this `Result` class _is_. `Result.getStatus()` returns the good/bad/etc. and this method is here for extracting an exception if available.

I'm wide open to suggestions on naming.

platform/configurator/configurator-api/pom.xml
paouelle: Can we put a comment in to that regard?

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
paouelle: I get that. putting a restriction that the only classes that can be passed in extends the interface re-enforces that.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
paouelle: My concern here is that a user of this interface should expect one thing and not something based on the implementation. I have another comment below that actually suggest to change it to a stream instead of a list. But if you keep it as a list, I think we would want to indicate what the user can and cannot do with the list.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
coyotesqrl: Ah. Actually, the only thing that should be passed in _is_ the interface's class, not any of its implementors. The service interface behaves similarly to a key for a lookup.

platform/configurator/configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
coyotesqrl: In a break from my usual feelings about immutability, I'm actually indifferent as to how the caller uses this list. But it's not a big effort for me to indicate it should be unmodifiable and ensure that our implementation makes it so.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/AdminOperation.java
coyotesqrl: We do not currently tag our log messages in a way that makes them easy to correlate but it is something we intend to address in the medium-term.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
coyotesqrl: Were you thinking of something beyond null/blank tests?

platform/configurator/configurator-api/pom.xml
coyotesqrl: We haven't traditionally, so this would be the only bundle with that comment. I'm not opposed, but I think after a while we all just become aware that in that context zero-coverage makes sense.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
paouelle: I wouldn't treat it as an error. As for the state of the input stream, there is no difference either way. The important part is that you were able to read from it and able to go on normally.

platform/configurator/configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ManagedServiceOperation.java
paouelle: two approach to this:
1- do the query and the exception handling in a static method that returns the factoryPid and pass the result to the super() call.
2- create an abstract protected method getFactoryPid() in the base class and let the subclass handle its initialization and keeping track of it. Your code in the base class, simply gets it from the getter whenever it needs it.

PR #1806, opened 2017-03-23
DDF-2825 Contributes Configurator service and transactional implementation

coyotesqrl:
#### What does this PR do?
This pulls the Configurator service from the [admin-beta-console Connexta repository](https://github.com/connexta/admin-console-beta).

History on the service prior to this contribution can be found from this location: https://github.com/connexta/admin-console-beta/configurator

@stustison @paouelle 

@lessarderic
@pklinef

This is a new service with no callers in DDF. To properly test, a caller can be written to utilize the Configurator to update/create services, properties files, etc.

This functionality begins the process of extracting configuration away from low-level constructs to support simplified administration and node setup.

[DDF-2825](https://codice.atlassian.net/browse/DDF-2825)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests


platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Result.java
```diff
@@ -0,0 +1,30 @@
...
+package org.codice.ddf.admin.configurator;
+
+import java.util.Optional;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface Result {
+    boolean isTxactSucceeded();
```
stustison: Looks like the boolean in the OperationReport got changed to `isTransactionSucceeded` but this one is still using Txact

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfigValidator.java
```diff
@@ -0,0 +1,71 @@
...
+                    ddfHomePath.toString()));
+        }
+
+        if (propFile.getParent()
+                .equals(ddfHomePath)) {
+            throw new IllegalArgumentException("Invalid attempt to edit file in ddf.home directory");
+        }
+
+        if (propFile.getParent()
+                .endsWith("etc")) {
```
stustison: So we're just excluding files directly placed in etc, do we need to worry about any files in directories under etc?

coyotesqrl: Security configuration files are in `etc/ws-security`. Based on that, I decided it was probably unwise to try to restrict much beneath there.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/OperationReportImpl.java
```diff
@@ -0,0 +1,54 @@
...
+                .stream()
+                .allMatch(Result::isTransactionSucceeded);
+    }
+
+    public Result getResult(String key) {
+        return results.get(key);
+    }
+
+    public List<Result> getFailedResults() {
+        return Collections.unmodifiableList(results.values()
```
paouelle: I would be fine here not wrapping it as modifiable. My previous comment was just documenting at the top that the returned list would not impact internals of the report implementation and since you are collecting them in a new list, there won't be any possibilities to mess up the internals of this class with a non-modifable list returned this way.

coyotesqrl: It is a bit of overkill. But it's only a decorator so it's cheap overkill at this point.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,204 @@
...
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
```
lessarderic: ❗️Since this is meant to be an external API (at least based on the package name), we should add a description explaining what this interface is and when and how it should be used. For instance, it may be useful to clarify that an "operation" is a call to one of the other methods and that `commit` should be called once all the desired method have been called.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,204 @@
...
+ * removed in a future version of the library. </b>
+ */
+public interface Configurator {
```
lessarderic: ❓I'm not sure how much flexibility we have to change this interface at this time, but if we do should we consider making the concept of transactions, operations and commits a little bit clearer by splitting it into multiple interfaces? For instance, we could have interfaces such as `ConfigurationTransactionFactory`, `ConfigurationTransaction`, `Operation` and multiple specific `Operation` sub-interfaces.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * Sequentially invokes all the transaction's operations, committing their changes. If a failure
+     * occurs during the processing, a rollback is attempted of those handlers that had already been
+     * committed.
+     * <p>
+     * In the case of a successful commit, changes should be logged.
+     *
+     * @param auditMessage In the case of a successful commit, the message to pass to a
```
lessarderic: ❓Should we mention audit messages and security loggers here or should we keep things more generic and simply use concepts like transaction description and parameters? If we were to have a `ConfigurationTransactionFactory`, those could actually be passed in when a `ConfigurationTransaction` is created and used internally however the implementation sees fit.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,204 @@
...
+     */
+    String stopBundle(String bundleSymName);
+
+    /**
+     * Determines if the bundle with the given name is started.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return true if started; else, false
+     */
+    boolean isBundleStarted(String bundleSymName);
```
lessarderic: ❓These "read" methods (`isBundleStarted`, `isFeatureStarted`, `getProperties`, `getConfig`, `getManagedServiceConfigs` and `getServiceReference`) do not fit very well with the other concepts exposed by this interface (transactions, operations, commit/rollback). For instance, would calling the `isBundleStarted("bundle")` after calling `startBundle("bundle")` but before calling `commit()` return `true` or `false`?

To make this a little clearer, could we move them to separate interfaces? For instance, could we define something like a `BundleController` interface that allows a client to start, stop and check the status of a bundle, `StartBundleOperation` and `StopBundleOperation` interfaces and classes that use the `BundleController` when `commit` is called and know how to rollback if something fails?

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
coyotesqrl: As discussed offline, splitting out the read operations (that muddy up the transactional purpose of the interface) into their own service interface.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/ConfigReader.java
```diff
@@ -0,0 +1,78 @@
...
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * Provides reader methods to system configuration items.
+ * <p>
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface ConfigReader {
```
peterhuffer: Should we consider supporting getting services with a classname as a String? If so, do we want to want to supporting writing services this way as well?

coyotesqrl: I would not be opposed to supporting that model. @pklinef and @lessarderic, what are your thoughts on me overloading the `getServiceReference` method?

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Status.java
```diff
@@ -0,0 +1,22 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public enum Status {
+    COMMIT_PASSED, COMMIT_FAILED, SKIPPED, ROLLBACK_PASSED, ROLLBACK_FAILED;
```
lessarderic: ❓What sequence of events could cause an operation to get in the `COMMIT_FAILED` state and remain there?
❓Instead of using an `enum`, could we create sub-classes of `Result` and `ConfigurationException` to capture an operation's result and reason for failure? This would allow operations to create new exceptions as needed and add more useful information regarding the failure such as a factory PID, file path, etc.

lessarderic:
@coyotesqrl, if we're not planning on merging this, should we capture the comments/feedback and close the PR?

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * Provides pseudo-transactional semantics to system configuration tasks.
+ * <p>
+ * In order to use, invoke the various {@code start}, {@code stop}, {@code create}, {@code delete},
+ * {@code update}, methods in the intended order of operation, then invoke one of the {@code commit}
+ * methods to complete the transaction.
+ * <p>
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
```
coyotesqrl: I'm favoring consistency here. This is the same text we've used for experimental code everywhere.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+ */
+public interface Configurator {
```
coyotesqrl: There might be value in that, but I think it's too early to add that much extra flexibility for growth. If I turn out to be wrong, the fact that this is still experimental will give us the freedom to make those changes.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Sequentially invokes all the transaction's operations, committing their changes. If a failure
+     * occurs during the processing, a rollback is attempted of those handlers that had already been
+     * committed.
+     * <p>
+     * In the case of a successful commit, changes should be logged.
```
coyotesqrl: It should be used only once. I'll make that clear in the javadocs. I'm not sure it's worth trying to enforce that in code, though. I'll think about it.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+    OperationReport commit();
+
+    /**
+     * Starts the bundle with the given name.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String startBundle(String bundleSymName);
```
coyotesqrl: I agree in principle, but in practice that would mean creating a `OperationId` interface in this bundle and an implementation of it in an impl bundle where that class would then really just be wrapping a String or UUID or something similar.  I could, however, switch it out to be a UUID; I think that would be far less likely to be abused and provides the same semantics.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param featureName the name of the feature
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String stopFeature(String featureName);
+
+    /**
+     * Creates a property file in the system with the given set of new key:value pairs.
+     *
+     * @param propFile   the property file to create
```
coyotesqrl: Absolute path; however, in our implementation `ConfigValidator.validatePropertiesPath` ensures that it is beneath `$ddf.home`, not located _in_ that directory, and not located in the _etc_ directory (though may be beneath it).

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+     * @param properties the set of key:value pairs to save to the property file
```
coyotesqrl: No. It'll end up throwing an NPE on save. I'll update the javadoc to indicate that.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
```diff
@@ -0,0 +1,156 @@
...
+     * @return a lookup key that can be used to correlate this operation in the
+     * final {@link OperationReport}
+     */
+    String deletePropertyFile(Path propFile);
+
+    /**
+     * Updates a property file in the system with the given set of new key:value pairs.
+     *
+     * @param propFile    the property file to update
+     * @param properties  the set of key:value pairs to save to the property file
```
coyotesqrl: No. Will update javadocs.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,38 @@
...
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+import java.util.List;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
```
coyotesqrl: Will javadoc.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,38 @@
...
+ * removed in a future version of the library. </b>
+ */
+public interface OperationReport {
+    boolean isTransactionSucceeded();
```
coyotesqrl: Will rename and javadoc should clarify.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,38 @@
...
+
+    Result getResult(String key);
```
coyotesqrl: Updating to UUID keys.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * An immutable list of Results that failed.
+     *
+     * @return an immutable list of failed results.
+     * It is an implementor's responsibility to ensure that the collection is unmodifiable.
+     */
+    List<Result> getFailedResults();
+
+    boolean containsFailedResults();
```
coyotesqrl: Not really, but it will be a common call from clients in boolean tests.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/OperationReport.java
```diff
@@ -0,0 +1,38 @@
...
+
+    void putResult(String key, Result result);
```
coyotesqrl: I think the increased complexity to the Configurator implementation provided would be pretty bad. It's not undoable, but I'm not sure it it's a valuable enough change.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/ConfigReader.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Determines if the bundle with the given name is started.
+     *
+     * @param bundleSymName the symbolic name of the bundle
+     * @return true if started; else, false
+     */
+    boolean isBundleStarted(String bundleSymName);
```
coyotesqrl: Will update docs and signatures to add the specific runtime exception that is thrown.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/ConfigReader.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * For the given managed service factory, retrieves the full complement of configuration properties.
+     * <p>
+     * This will get all the key:value pairs for each available configuration.
+     *
+     * @param factoryPid the factoryPid of the service to query
+     * @return the the current sets of key:value pairs, in a map keyed on {@code configId}
+     */
+    Map<String, Map<String, Object>> getManagedServiceConfigs(String factoryPid);
```
coyotesqrl: I think that's best as a future enhancement.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/ConfiguratorFactory.java
```diff
@@ -0,0 +1,24 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.admin.configurator;
+
+/**
+ * <b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface ConfiguratorFactory {
+    Configurator getConfigurator();
```
coyotesqrl: I'll update the docs to indicate that it should be considered "used up" after a call to commit, but I think I'd rather hold off on updating the implementation to enforce that for now. If we determine we need it, it'll just be a check and runtime exception thrown everywhere.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/ConfiguratorFactory.java
```diff
@@ -0,0 +1,24 @@
...
+
+    ConfigReader getConfigReader();
```
coyotesqrl: I chose to be symmetric between the reader and writer. It wasn't necessary to do it that way, but seemed logical when doing the refactor.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Status.java
coyotesqrl: An operation would get to commit failed in the following example:
Given [op1, op2, op3, op4]
1. op1 commits successfully
2. op2 commits successfully
3. op3 fails
4. op2 rolls back successfully
5. op1 fails to rollback
the final results would be:
[op1: ROLLBACK_FAILED, op2: ROLLBACK_PASSED, op3: COMMIT_FAILED, op4: SKIPPED]

Any operation that fails to commit will go to the COMMIT_FAILED state and remain there; any subsequent operations will be marked as SKIPPED; all prior operations will be unrolled and rollback will be attempted. Rollback may then pass or fail on each of those.

coyotesqrl:
@lessarderic Upon further consideration, I believe there is value in merging the Configurator at this time. I've pushed my changes subsequent to your recent review.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
lessarderic: I'd rather have a different type to make things really clear, but `UUID` would probably be good enough for now.

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Configurator.java
lessarderic: Should we add those constraints to the JavaDoc?

platform/admin/configurator/admin-configurator-api/src/main/java/org/codice/ddf/admin/configurator/Status.java
lessarderic: Thanks for the clarification! That'd be a great example to add to the JavaDoc ;-)

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
```diff
@@ -0,0 +1,465 @@
...
+     * @return report of the commit status, whether successful, successfully rolled back, or partially
+     * rolled back with errors
+     */
+    public OperationReport commit() {
+        OperationReport configReport = new OperationReportImpl();
+        for (Map.Entry<UUID, Operation> row : configHandlers.entrySet()) {
+            try {
+                Object commitResult = row.getValue()
+                        .commit();
+                if (commitResult instanceof String) {
```
pklinef: This smells of tight coupling and leaking implementation details that a String response is a passing managed service operation.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
```diff
@@ -0,0 +1,465 @@
...
+                undoStack.push(row);
+            } else if (!failedStep.equals(row.getKey())) {
+                configReport.putResult(row.getKey(), ResultImpl.skip());
+            }
+        }
+
+        for (Map.Entry<UUID, Operation> row : undoStack) {
+            try {
+                row.getValue()
+                        .rollback();
```
pklinef: A rollback could return a value which is not captured in the config report.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/Operation.java
```diff
@@ -0,0 +1,28 @@
...
+package org.codice.ddf.admin.configurator.impl;
+
+import org.codice.ddf.admin.configurator.ConfiguratorException;
+
+/**
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+interface Operation<T, S> {
+    T commit() throws ConfiguratorException;
```
pklinef: There seems to be one case where T is not Void and in that case commit always returns null and rollback actually returns a String for ManagedServiceOperation.   Based on that I am not sure if commit and rollback should share the same return.

It might make more sense to have commit return a Result and rollback to return some kind of RollbackResult.  This would also remove the need for create Results and for control flow by exception.  Result and RollbackResult could then store transaction result data.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ManagedServiceOperation.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        @Override
+        public String commit() throws ConfiguratorException {
+            deleteByPid(configPid);
+            return null;
+        }
+
+        @Override
+        public String rollback() throws ConfiguratorException {
```
pklinef: Instead of returning a raw string, I would suggest returning something like CreatedManagedServicePid that is backed by a String.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ResultImpl.java
```diff
@@ -0,0 +1,82 @@
...
+import org.codice.ddf.admin.configurator.Result;
+import org.codice.ddf.admin.configurator.Status;
+
+public class ResultImpl implements Result {
+
+    private final Status status;
+
+    private final Throwable badOutcome;
+
+    private final String configId;
```
pklinef: Config ID seems to only be related to managed services.  Can we extract operation specific data into something more generic?  The cop out would be a map but I could also see something like Optional< T > data where T would be a generic for some kind of operation specific result data type.  For example I mentioned the CreatedManagedServicePid type.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/ConfiguratorImpl.java
```diff
@@ -0,0 +1,465 @@
...
+     * committed.
+     *
+     * @return report of the commit status, whether successful, successfully rolled back, or partially
+     * rolled back with errors
+     */
+    public OperationReport commit() {
+        OperationReport configReport = new OperationReportImpl();
+        for (Map.Entry<UUID, Operation> row : configHandlers.entrySet()) {
+            try {
+                Object commitResult = row.getValue()
```
pklinef: If commit returned a Result, then you could add the result to the config report as long as the result does not contain an error.

coyotesqrl:
build now

coyotesqrl:
I've reviewed the Codacy findings; they are all false positives; furthermore, I am actually concerned about one of the findings in that it implies Codacy might not quite understand Java scoping rules.

coyotesqrl:
build now

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
```diff
@@ -0,0 +1,206 @@
...
+        saveProperties(currentProperties);
+
+        return ResultImpl.rollback();
+    }
+
+    Map<String, String> readState() throws ConfiguratorException {
+        return Collections.unmodifiableMap(currentProperties);
+    }
+
+    void saveProperties(Map<String, String> propertyMap) throws ConfiguratorException {
```
pklinef: All of the uses of this method are internal to this class or its nested classes.  Is there a reason this could not be private?  I think this is also done in all of the other Operations too.

coyotesqrl: The nested classes are static. They don't have access to the private non-static methods. Java scoping rules are fun and this one befuddled me for a second.

It's a bit twisty in there, as the `PropertyOperation.Actions.create` method can happily invoke the private ctor of `PropertyOperation.CreateHandler` because it too is a static class but can't access `saveProperties` because it's in the outer class.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/KarafUtils.java
```diff
@@ -0,0 +1,67 @@
...
+import org.codice.ddf.ui.admin.api.ConfigurationAdmin;
+import org.codice.ddf.ui.admin.api.ConfigurationAdminMBean;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.ServiceReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class KarafUtils {
```
coyotesqrl: Will rename.

platform/admin/configurator/admin-configurator-actions-api/src/main/java/org/codice/ddf/internal/admin/configurator/actions/ServiceReader.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Set;
+
+import org.codice.ddf.admin.configurator.ConfiguratorException;
+
+/**
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface ServiceReader {
```
coyotesqrl: I'm going to whistle nonchalantly as I walk right past this comment...

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/PropertyOperation.java
pklinef: Missed the static part.  That is annoying.

coyotesqrl:
Build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

PR #1823, opened 2017-03-27
DDF-2910 added an exception to the IdP Handler to allow it to log in legacy clients

stustison:
@brendan-hofmann @clockard @pklinef 
@coyotesqrl
@stustison
#### Checklist:
- [x] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSingleSignOn.java
```diff
@@ -226,6 +226,7 @@ private ResponseHelper getSearchResponse(boolean isPassiveRequest, String url)
         // @formatter:off
         Response searchResponse =
                 given().
+                        header("User-Agent", "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36").
```
brendan-hofmann: How about a constant instead?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
```diff
@@ -245,6 +257,51 @@ public Object getAttribute(String name) {
             return doPaosRequest(request, response);
         }
 
+        if (userAgentCheck && userAgentIsNotBrowser(httpRequest)) {
+            SecurityLogger.audit("Attempting to log client in as a legacy system.");
+            //if we get here, it is most likely an older DDF that is federating
+            //it isn't going to understand the redirect to the IdP and it doesn't support ECP
+            //so we need to fall back to other handlers to allow it to log in using PKI, Basic or Guest
```
brendan-hofmann: I would have expected any other handlers for the realm to get called anyways. I am wondering if there is a way to do this check that just picks up the configured authentication handlers for the target realm and gives them a chance to answer, rather than hardcoding PKI/Basic/Guest. Doesn't this also exclude falling back to anything that's not one of those three? Technically it is possible to deploy and configure additional authentication mechanisms.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
```diff
@@ -245,6 +257,51 @@ public Object getAttribute(String name) {
...
+
+            //check for basic auth first
+            BasicAuthenticationHandler basicAuthenticationHandler = new BasicAuthenticationHandler();
```
coyotesqrl: Is it worth configuring the order in which the fallback auth methods will be checked?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
stustison: You don't configure the IdP with any other handlers, which is why I went this route. It is an interesting thought though. I could make it fall through to any handlers that might come after the IdpHandler, but the system would also need to be configured that way in order for any legacy clients to fall back appropriately. Also, this is for legacy clients, so while it is possible for some third party to deploy their own authentication mechanism, I don't expect them to need that to work with our IdP. I'd expect them to just change the auth method over to their custom one.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
stustison: Yeah, I'm mulling over Brendan's comment. If I were to go the direction of using any additionally configured handlers, then it would automatically allow you to do this.

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/IdpHandlerTest.java
```diff
@@ -90,6 +90,8 @@ public void setUp() throws Exception {
 
     @Test
     public void testGetNormalizedToken() throws Exception {
+        when(httpRequest.getHeader("User-Agent")).thenReturn(
```
coyotesqrl: Same re: Brendan's suggestion for a const.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
```diff
@@ -329,7 +329,7 @@ private HandlerResult doPaosRequest(ServletRequest request, ServletResponse resp
             }
             authnRequest = createAndSignAuthnRequest(true,
                     wantSigned && idpssoDescriptor.getWantAuthnRequestsSigned());
-            paosRequest = createPaosRequest();
+            paosRequest = createPaosRequest((HttpServletRequest) request);
```
pklinef: Is it ever possible that the `ServletRequest` would be implemented as something other than `HttpServletRequest` in the future?

platform/security/rest/security-rest-cxfwrapper/src/main/java/org/codice/ddf/cxf/SecureCxfClientFactory.java
```diff
@@ -306,6 +313,80 @@ private void configureConduit(ClientConfiguration clientConfig) {
...
+        TLSClientParameters tlsParams = httpConduit.getTlsClientParameters();
+        if (tlsParams == null) {
+            tlsParams = new TLSClientParameters();
+        }
+
+        tlsParams.setDisableCNCheck(disableCnCheck);
+
+        tlsParams.setUseHttpsURLConnectionDefaultHostnameVerifier(true);
+        tlsParams.setUseHttpsURLConnectionDefaultSslSocketFactory(true);
+        tlsParams.setCipherSuites(Arrays.asList(System.getProperty("https.cipherSuites")
```
pklinef: I am surprised CXF did not read these system properties already.

platform/security/rest/security-rest-cxfwrapper/src/main/java/org/codice/ddf/cxf/paos/PaosInInterceptor.java
```diff
@@ -286,7 +288,7 @@ private String createToken(String authorization) throws IOException {
 
     HttpResponseWrapper getHttpResponse(String responseConsumerURL, String soapResponse)
             throws IOException {
-        HttpTransport httpTransport = new ApacheHttpTransport();
+        HttpTransport httpTransport = new NetHttpTransport();
```
pklinef: Should we add a comment about why we need this particular implementation for this to work?

platform/security/rest/security-rest-cxfwrapper/src/main/java/org/codice/ddf/cxf/SecureCxfClientFactory.java
stustison: Me too! I'm fairly certain that it did at one point. I'm like 99% sure all of this worked just fine without us setting any of this stuff. I'm not sure when it broke, though.

platform/security/rest/security-rest-cxfwrapper/src/main/java/org/codice/ddf/cxf/paos/PaosInInterceptor.java
stustison: Sure.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
stustison: Hm... I'm not sure. There are some ServletRequest wrapper classes that are concrete and not of the HttpServletRequest type. I think literally everything else extends the HttpServletRequest class. We do make the assumption up at the beginning of the getNormalizedToken method that everything is HTTP though. If we wanted to fix this, I think we'd need to go through all of our handlers and filters everywhere. This is probably fine for now, but maybe one of those tech debt-ish type things.

platform/security/rest/security-rest-cxfwrapper/src/main/java/org/codice/ddf/cxf/SecureCxfClientFactory.java
```diff
@@ -306,6 +313,80 @@ private void configureConduit(ClientConfiguration clientConfig) {
...
+                .split(",")));
+
+        KeyStore keyStore = null;
```
coyotesqrl: I hate this block of code - I wish it could be extracted to methods for `getKeyManagers()` and `getTrustManagers()` so the sets on `tlsParams` would be more clearly delineated - but then you'd end up with two nearly identical private methods and that's no better.

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

PR #1825, opened 2017-03-28
DDF-2911 Prevent directory monitor from creating extra DurableFileConsumer directory 

codymacdonald:
#### What does this PR do?
Replaced String concatenation with calls to Paths.get when constructing the path to the persistence store to prevent extra DurableFileConsumer directories from being mistakenly created.
@emanns95 @ahoffer @brendan-hofmann @AzGoalie @mweser 

@coyotesqrl
@shaundmorris

1. Start and install a DDF instance
2. From the admin console, navigate to Catalog -> Configuration -> Catalog Content Directory Monitor
3. In the Directory Path field, add a local path to a directory from your machine for testing
4. Navigate to the `DDF_HOME/data` directory on your machine 
5. In the data directory, make sure there is only one `DurableFileConsumer` directory and one `processed` directory.
6. The `DurableFileConsumer` directory should contain only one .ser file, and .ser files for files in your test directory from step 3 should exist in `processed`.

[](https://codice.atlassian.net/browse/DDF-2911/)

### DIFF
--- a/catalog/core/catalog-core-camelcomponent/src/test/java/ddf/camel/component/catalog/content/FileSystemPersistenceProviderTest.java
+++ b/catalog/core/catalog-core-camelcomponent/src/test/java/ddf/camel/component/catalog/content/FileSystemPersistenceProviderTest.java
@@ -23,6 +23,7 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 import java.io.File;
+import java.nio.file.Paths;
 import java.util.Map;
 import java.util.Set;
 
@@ -56,7 +57,8 @@ public class FileSystemPersistenceProviderTest {
         fileSystemPersistenceProvider = new FileSystemPersistenceProvider("test");
         fileSystemPersistenceProvider.fileSystemDataAccessObject = mockFileSystemDao;
         assertThat(fileSystemPersistenceProvider.getMapStorePath(),
-                is(fileSystemPersistenceProvider.getPersistencePath() + "test" + File.separator));
+                is(Paths.get(fileSystemPersistenceProvider.getPersistencePath(), "test")
+                        .toString()));
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


catalog/core/catalog-core-camelcomponent/src/test/java/ddf/camel/component/catalog/content/FileSystemPersistenceProviderTest.java
```diff
@@ -56,7 +57,8 @@ public void setup() {
         fileSystemPersistenceProvider = new FileSystemPersistenceProvider("test");
         fileSystemPersistenceProvider.fileSystemDataAccessObject = mockFileSystemDao;
         assertThat(fileSystemPersistenceProvider.getMapStorePath(),
-                is(fileSystemPersistenceProvider.getPersistencePath() + "test" + File.separator));
+                is(Paths.get(fileSystemPersistenceProvider.getPersistencePath(), "test")
```
coyotesqrl: This unit test _should_ be failing. Your implementation of `getMapStorePath` is still tacking on `File.separator` to the end (right or wrong, I'm not clear at the moment); however, here, you're comparing it against a path without the separator character appended.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/FileSystemPersistenceProvider.java
```diff
@@ -95,7 +96,8 @@ String getPersistencePath() {
      * @return
      */
     String getMapStorePath() {
-        return getPersistencePath() + mapName + File.separator;
```
coyotesqrl: Shouldn't the `getPersistencePath()` method in this implementation also be using `Paths` concatenation instead of String concatenation?

Is there a reason it can't use the `AbsolutePathResolver` as the other implementations do?

codymacdonald: There doesn't seem to be any reason for not using `AbsolutePathResolver`. I will replace this `getPersistencePath()` with that implementation for consistency.

PR #488, opened 2016-01-25
DDF-1825 Adds way to export configs from Admin UI

andrewkfiedler:
- Updated the Gruntfile to run only unit tests when calling the test task.  The other tests have been turned off for so long that they no longer work, and they were more integration tests than unit tests.  As such I've commented them out, so the unit test can at least run.

 - Updated the package.json to add grunt-simple-mocha (for unit tests), requirejs, sinon (for mocking in unit tests), and squirejs (for mocking require dependencies in unit tests), and jsdom (to enabled us to use jQuery in unit tests which allows mocking $.ajax).  These all lay the foundation for our future UI unit testing.

- Updated the pom.xml to remove the —force on the grunt test.  Also added an execution to install the latest node.  We might want to revisit the topic of how to ensure the latest node version is installed.

- Updated style.css to remove the background from the alerts, delegating the background color back to bootstrap.

- Updated Alerts.js to add a default for type (‘danger’).

- Added module/Export.js to the javascript models.  It has only the public API of export, but views can also hook into changes to the model.  The model has: url, path, warnings, errors, inProgress.  I chose to use the inProgress model property over just a callback from export, so if multiple views hook into the model they can all update based on the state of the export operation.  A view can still use the deferred returned from the export method if they so choose.  Also, we decided to use _ as a standard for private methods.  If someone or something outside the model relies on these methods, they do so at their own risk.  We couldn’t find a more appealing way of making things private while also sharing them between the various instantiations of the model.

- Updated ModuleDetail.layout.js to add a region for the export modal, and to hook the ExportModal view up to that region through a click event on the export button.

- Updated common.less to remove some superfluous style given the updates to mixins.less.

- Updated mixins.less to make panels clear floats, and to add a margin to the bottom of all headers (see common.less update).

- Updated alerts.handlebars to infer the type of the alert from the model, and to allow html in the banner message.

- Updated ModuleDetail.layout.handlebars to remove some cruft, and to add the export button as well as the region for the export modal.

- Added unit/Export.js to the tests.  It tests the behavior of the Export modal, particularly the export method (its public API).  Since it doesn’t need a browser to run, it’s super quick.

- Added ExportModal.view.js. It hooks into the Export model, which for now is instantiated within the view.  In the future, this model may live somewhere else.  The hooks of note are updates to the warnings, errors, and inProgress.  The former two update the alerts in the view, while the latter updates the spinner.

[...]

PR #1829, opened 2017-03-29
DDF-2912 Implemented System Administrator Alerts

emmberk:
**NOTE: This implementation will not be pushed to master.** The implementation for 2.11.0 will be completed under [DDF-2878](https://codice.atlassian.net/browse/DDF-2878). This PR is meant to address a security concern and will only be merged in the 2.10.x branch.

#### What does this PR do?
This PR adds alerts to the SA when the audit log fails.

_Summary of changes:_
- Added 2 new modules,`admin-core-alertservice` and `admin-core-alertservice-data`
- Modified the jolokia endpoint (that used to be in the `admin-core-insecuredefaults` but is now in `admin-core-alertservice`) to return all of the alerts, not just the insecure defaults alerts
- Modified the front-end banner code to retrieve multiple alerts from the modified jolokia endpoint and to be able to dismiss alerts
- Added a failover logger called `AdminAlertAppender`

@clockard @mcalcote @brianfelix @vinamartin @ricklarsen @ahoffer
[UI](https://github.com/orgs/codice/teams/ui) @djblue
@andrewkfiedler
@lessarderic

1. Full build.
2. Before enabling the new failover logger:
    - In different browsers, confirm that all alerts appear and behave correctly on the Admin Console.
    - Confirm that audit logging behaves as expected.
    - Confirm that the logviewer at `admin/logviewer/index.html` still works.
    - Using a debugger, manually test adding more alerts (with different levels) to the `AlertServiceBean`. Confirm that the alerts appear and behave correctly on the Admin Console. Test viewing the details of and dismissing alerts.
    - Test that changes to [ExportModal.view.js](https://github.com/codice/ddf/blob/master/platform/admin/ui/src/main/webapp/js/views/module/ExportModal.view.js) did not cause regression with the `Export Configuration` banners (see https://github.com/codice/ddf/pull/488).
      image
3. Now test the new failover logger with a new distribution:
    - Follow the [Enabling Fallback Audit Logging section in the documentation](https://github.com/codice/ddf/blob/master/distribution/docs/src/main/resources/_contents/_securing/auditing-contents.adoc) to configure the system to use the fallback log and appender.
    - Confirm that audit logging behaves as expected.
    - Confirm that the logviewer at `admin/logviewer/index.html` works.
    - One was to manually test the logic in the `AdminAlertAppender` is to modify [log4j2.config.xml](https://github.com/codice/ddf/blob/master/distribution/ddf-common/src/main/resources/etc/log4j2.config.xml) to use the `osgi-adminAlertAppender` as an actual logger, not a failover logger. Confirm that only one logging failed banner appears on the Admin Console. (Move line 79 `<AppenderRef ref="osgi-adminAlertAppender"/>` to after line 98.)

[DDF-2912](https://codice.atlassian.net/browse/DDF-2912)
banners
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests


platform/admin/core/admin-core-alertservice-data/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.96</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>1.00</minimum>
```
stustison: Give some buffer space. Things set to the maximum amount will generally fail to build on other platforms, such as Windows. Allow at least 3%

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,148 @@
...
+        this.mBeanServer = mBeanServer;
+    }
+
+    public void init() {
+        try {
+            objectName = new ObjectName(
+                    AlertServiceBean.class.getName() + ":service=alert-service");
+            mBeanServer.registerMBean(this, objectName);
+        } catch (Exception e) {
+            LOGGER.error("Could not register AlertServiceMBean.", e);
```
stustison: Anything at warn or above should include steps that an admin can take to resolve the situation or if there isn't anything they can do to resolve it, they should be informed of the consequences.

emmberk: Updated mbean registration to follow pattern in [ResourceDownload](https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/org/codice/ddf/catalog/resource/download/impl/ResourceDownload.java) where logging statements are `debug`.

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -28,7 +28,7 @@
 # org.ops4j.pax.logging.log4j2.config.file = ${karaf.etc}/log4j2.config.xml
 
 # Root logger
-log4j.rootLogger=INFO, out, osgi:*
+log4j.rootLogger=INFO, out, osgi:VmLogAppender, osgi:org.codice.ddf.platform.logging.LoggingService
```
emmberk: This change to specify exactly what osgi loggers should receive audit messages was necessary because the new `AdminAlertAppender` was included in `osgi:*`.

platform/admin/core/admin-core-insecuredefaults/pom.xml
```diff
@@ -22,39 +22,8 @@
     <artifactId>admin-core-insecuredefaults</artifactId>
     <name>DDF :: Admin :: Core :: Insecure Defaults</name>
     <description>Verifies insecure default configurations have been changed.</description>
-    <packaging>bundle</packaging>
```
emmberk: Because the `InsecureDefaultsService` is now instantiated in the `AlertServiceBean` this module no longer needs to be a bundle. It is now embedded in `admin-core-alert-service`.

platform/admin/core/admin-core-insecuredefaults/src/main/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsService.java
```diff
@@ -13,26 +13,20 @@
...
-import javax.management.MalformedObjectNameException;
-import javax.management.NotCompliantMBeanException;
-import javax.management.ObjectName;
+import javax.annotation.Nonnull;
 
 import org.codice.ddf.configuration.AbsolutePathResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class InsecureDefaultsServiceBean implements InsecureDefaultsServiceBeanMBean {
```
emmberk: The mbean registration in this class was removed and is now part of `AlertServiceBean` instead.

platform/admin/core/admin-core-insecuredefaults/src/main/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsServiceBeanMBean.java
```diff
@@ -1,22 +0,0 @@
...
- * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
- * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
- * is distributed along with this program and can be found at
- * <http://www.gnu.org/licenses/lgpl.html>.
- */
-package org.codice.ddf.admin.insecure.defaults.service;
-
-import java.util.List;
-
-public interface InsecureDefaultsServiceBeanMBean {
```
emmberk: The mbean for the InsecureDefaultsService was removed and is now part of `AlertServiceBean` instead.

platform/admin/core/admin-core-insecuredefaults/src/test/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsServiceBeanTest.java
```diff
@@ -1,177 +0,0 @@
...
-
-import javax.management.MalformedObjectNameException;
-import javax.management.ObjectName;
-
-import org.codice.ddf.admin.insecure.defaults.service.Alert.Level;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Test;
-
-public class InsecureDefaultsServiceBeanTest {
```
emmberk: This test class was renamed to `InsecureDefaultsServiceTest`, and the tests having to do with the mbean were removed.

platform/admin/core/admin-core-alertservice/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <artifactId>admin-core-alertservice</artifactId>
+    <name>DDF :: Admin :: Core :: Alert Service</name>
+    <packaging>bundle</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
```
vinamartin: Out of curiosity, how are you using this dependency?

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,47 @@
...
+    public AdminAlertAppender(EventAdmin eventAdmin) {
+        notNull(eventAdmin, "eventAdmin may not be null");
+        this.eventAdmin = eventAdmin;
+    }
+
+    @Override
+    public void doAppend(PaxLoggingEvent event) {
+        // TODO better banner message
+        eventAdmin.postEvent(new Event(Alert.ALERT_EVENT_TOPIC_VALUE,
+                new Alert(Alert.Level.ERROR, "Logging has failed.", new ArrayList<>())));
```
vinamartin: I feel like this message should be a little stronger.

platform/admin/ui/src/main/webapp/main.js
```diff
@@ -156,33 +156,35 @@
...
+                itemView: AlertsView.View
+            });
+
+            Application.App.alertsRegion.show(new AlertsCollectionView({
+                collection: alerts
+            }));
+        });
+
+        // refresh the session if keypress  or mouse click events occur
+        SessionRefresherUtil(60000);
```
clockard: is this needed?

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
emmberk: I have a TODO to change this. Any suggestions, @ahoffer?

clockard: Maybe "Audit logging has failed"

platform/admin/core/admin-core-alertservice/pom.xml
emmberk: I'm using it for the `notNull` method in the constructor.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,200 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(AlertServiceBean.class);
+
+    private static final String CLASS_NAME = AlertServiceBean.class.getName();
+
+    protected static final String INSECURE_DEFAULTS_ALERT_TITLE =
+            "The system is insecure because default configuration values are in use.";
+
+    private final MBeanServer mBeanServer;
+
+    private final InsecureDefaultsService insecureDefaultsService = new InsecureDefaultsService();
```
emmberk: Are there any advantages to injecting it instead?

platform/admin/ui/src/main/webapp/main.js
emmberk: I don't know, but I didn't modify this line. It was already there: https://github.com/codice/ddf/blob/master/platform/admin/ui/src/main/webapp/main.js#L175.

platform/admin/core/admin-core-alertservice/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.25</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.37</minimum>
```
emmberk: These ratios will be increased once the TODO to add more tests to `AlertServiceBeanTest` is completed.

platform/admin/ui/src/main/webapp/templates/alerts.handlebars
```diff
@@ -11,31 +11,37 @@
...
-    <div class="panel panel-{{type}} accordion">
-        <div class="panel-heading">
-            <h4 class="panel-title">
-                <i class="fa fa-exclamation-triangle"></i>
-                {{{banner}}}
-                <button type="button" class="btn btn-link details" data-toggle="collapse"
-                        data-parent=".accordion" data-target="#{{collapseId}}">{{button}} details
+<div class="panel panel-{{level}} accordion">
+    <div class="panel-heading">
+        <h4 class="panel-title">
```
andrewkfiedler: Add `clearfix` class to this to make the wrapping work.

emmberk: Added:+1:
Before:
image
After:
image

emmberk:
build now

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,202 @@
...
+            return true;
+        }
+    }
+
+    /**
+     * Adds an {@link Alert} to the collection of current {@link Alert}s. The even must have topic
+     * {@link Alert#ALERT_EVENT_TOPIC_VALUE} and
+     * event.getProperty({@link Alert#ALERT_EVENT_ALERT_PROPERTY}) is the {@link Alert} to add.
+     * Already-added {@link Alert}s with the same {@value Alert#KEY} as the {@link Alert} in the event will be
+     * overriden.
```
ahoffer: What's the thinking behind overriding alerts with the same key? Does it serve a purpose of fulfill a use case? Just asking.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    public void handleEvent(Event event) {
+        if (event != null && ALERT_EVENT_TOPIC_VALUE.equals(event.getTopic())) {
+            final Object alertObject = event.getProperty(ALERT_EVENT_ALERT_PROPERTY);
+            if (alertObject instanceof Alert) {
+                final Alert alert = (Alert) alertObject;
+                String key = alert.getKey();
+
+                // added {@link Alert}s must have a {@value Alert#KEY} because they must be dismissable
+                if (key == null) {
```
ahoffer: I suggest`if (StringUtils.isEmpty(key))` instead. If will that method will return true is the value is null or if the string is empty.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,202 @@
...
+                    key = UUID.randomUUID()
+                            .toString()
+                            .replaceAll("-", "");
+                    alert.setKey(key);
```
ahoffer: What is the use for letting event generators define a key? Why not just assign a key to every Alert?

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+import static org.apache.commons.lang.Validate.notEmpty;
+import static org.apache.commons.lang.Validate.notNull;
+
+import java.util.HashMap;
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import net.shibboleth.utilities.java.support.annotation.constraint.NotEmpty;
```
ahoffer: That's pretty cool. I did not know about `javax.validation.constraints` until now.

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library.</b>
+ * </p>
+ * <p>
+ * An {@link Alert} holds the content for a system administrator alert.
+ */
+public class Alert extends HashMap<String, Object> {
```
ahoffer: @lessarderic Do we need a map? Could we just use properties with getters and setters? See my comment above.

lessarderic: Yeah, couldn't find any place where a `Map` is actually needed so I agree that a simple bean would be better. @emmberk, any specific reason to keep a `Map` around?

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final String TITLE = "title";
+
+    private static final String DETAILS = "details";
+
+    private static final String KEY = "key";
+
+    /**
+     * Constructs an {@link Alert} with a {@param key}
+     *
+     * @param key     non-null, non-empty {@value #KEY}
```
Lambeaux: Expounding on @lessarderic's comment above, we always assume `null` *cannot* be returned unless the `@Nullable` annotation is present.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
emmberk: Yes. For example, the AdminAlertAppender adds an alert with a specific key so that it may overwrite the alert that has already been added. The AdminAlertAppender only knows how to push alerts, and we wouldn't want many of the same alert banners to show up. We only want to see one. Instead of checking if an alert with all of the same values already exists in the current alerts, I can specify that I want to overwrite the alert with this specific key if there is one already.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
emmberk: The Alert#getKey method is defined to never be empty if it is not null. The Alert class may be modified after talking with eric, so I will reconsider this suggestion.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
emmberk: See response above to your question "What's the thinking behind overriding alerts with the same key? Does it serve a purpose of fulfill a use case? Just asking."

Basically, using this key allows me to not have to implement an equals method for Alert to check if the alert already exists in the map before adding.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
ahoffer: Thanks. I think it is worth a short comment to explain why we replace an object if it has the same key.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
ahoffer: I understand now. Thanks.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
ahoffer: Thanks.

platform/admin/core/admin-core-alertservice/src/test/groovy/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBeanTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+
+    final mbeanName = new ObjectName(AlertServiceBean.class.getName() + ":service=alert-service")
+
+    def alertServiceBean
+
+    void setup() throws Exception {
+        // the keystore path must be set for the InsecureDefaultsService to use
+        System.setProperty("javax.net.ssl.keyStore", "TestKeystorePath")
+        alertServiceBean = new AlertServiceBean(ManagementFactory.getPlatformMBeanServer())
+        assert !isRegistered()
```
lessarderic: ❗️ It's usually better not to have any assertions in the setup method and instead have a separate test method to cover those assertions. Makes it easier to figure out what's being tested and what's broken when an assertion fails.

platform/admin/core/admin-core-alertservice/src/test/groovy/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBeanTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+
+        then:
+            isRegistered()
+    }
+
+    def 'test init when the bean has already been registered'() throws Exception {
+        when:
+            alertServiceBean.init()
+
+        then:
```
lessarderic: ❗️ This is already tested in the previous test so we probably could remove this `then` and combine the two `when` blocks, i.e., call `init()` twice and and make sure nothing breaks and the MBean is still registered.

platform/admin/core/admin-core-alertservice/src/test/groovy/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBeanTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+            isRegistered()
+    }
+
+    def 'test destroy'() throws Exception {
+        when:
+            alertServiceBean.init()
+
+        then:
```
lessarderic: ❗️ Same comment as above, i.e., `then` not needed.

platform/admin/core/admin-core-alertservice/src/test/groovy/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBeanTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+
+    def 'test destroy when the bean has not already been registered'() throws Exception {
+        when:
+            alertServiceBean.destroy()
+
+        then:
+            !isRegistered()
+    }
+
+    private boolean isRegistered() throws MalformedObjectNameException {
```
lessarderic: ❓ Should we keep all the tests at the top and any private utility method at the bottom to avoid any confusion?

platform/admin/core/admin-core-alertservice/src/test/groovy/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBeanTest.groovy
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    def 'test handling an invalid alert event'(event) {
+        when:
+            alertServiceBean.handleEvent(event)
+
+        then:
+            alertServiceBean.addedAlerts.isEmpty()
+
+        when:
```
lessarderic: ❗️ As a general rule, tests shouldn't repeat the `when` and `then` blocks. When they do, it usually means that one of the pairs can be extracted to a separate test. In this case, I don't think we need to put the `getAlerts` in a `when` since it's really just code used to test the state of the object, not the operation being tested. See http://spockframework.org/spock/docs/1.0/spock_primer.htm for more on this.

platform/admin/ui/src/main/webapp/js/models/Alerts.js
```diff
@@ -16,36 +16,39 @@ define([
...
 
-    AlertsModel.AlertsDefaults = Backbone.Model.extend({
+    AlertsModel.Alert = Backbone.Model.extend({
         defaults: {
-            'banner': 'You should look at this.',
-            'button': 'Show',
+            'details-button-action': 'Show',
             'collapse': 'out',
-            'type': 'danger'
+            'level': 'danger'
```
lessarderic: ❓ `'error'` instead?

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
lessarderic: Makes it possible to mock out in unit tests.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
emmberk: 👍  Added
>Already-added {@link Alert}s with the same key as the {@link Alert} in the event will be overridden. For example, an application may overwrite an existing Alert with the timestamp of when an issue most-recently occurred by specifying the key of the {@link Alert} to overwrite.

distribution/docs/src/main/resources/_contents/_running/monitoring-contents.adoc
```diff
@@ -36,6 +36,11 @@ A useful way to debug and detect issues is to change the log level:
 
 `log4j.rootLogger=DEBUG, out, osgi:VmLogAppender`
 
+[NOTE]
```
emmberk: After discussing this suggestion with you, we decided that no other documentation is needed right now.

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
emmberk: From Eric's suggestion, Alert was refactored to not extend HashMap, so there are now variables for key, level, title, and details.

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+                return "info";
+            }
+        }, WARN {
+            public String toString() {
+                return "warning";
+            }
+        }, ERROR {
+            public String toString() {
+                return "danger";
+            }
```
emmberk: I extracted a method and added a comment in AlertServiceBean to explain why the enums are mapped to different strings (e.g. "ERROR" -> "danger").
```java
    /**
     * Maps the enums to existing Admin Console styles
     */
    private static String getLevelString(Alert alert) {
    ...
```

emmberk: That is, there are already-defined css styles named "danger", "warning", and "info".

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
emmberk: Replaced @NotEmpty (and it's pom dependency) with `javax.validation.constraints.Size`, and I kept the note in the javadoc that those values should not be empty (which is more restraining than just a minimum size of 1).

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.List;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import net.shibboleth.utilities.java.support.annotation.constraint.NotEmpty;
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
```
emmberk: Removed

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+ * removed in a future version of the library.</b>
+ * </p>
+ * <p>
+ * An {@link Alert} holds the content for a system administrator alert.
```
emmberk: Removed the experimental note

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
emmberk: The original motivation for the map was so that it would follow the model of Notification.java and it could easily be used by the jolokia endpoint. However, Alert has now been refactored to not extend HashMap, and the Alert is explicitly mapped to something that the jolokia endpoint likes in the AlertServiceBean.

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
emmberk: Added some more descriptors to what key, level, title, and details are

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Constructs an {@link Alert} with a {@param key}
+     *
+     * @param key     non-null, non-empty {@value #KEY}
+     * @param level   non-null {@value #LEVEL}
+     * @param title   non-null, not empty {@value #TITLE}
+     * @param details non-null (may be empty) list of String {@value #DETAILS}
+     */
+    public Alert(@Nonnull @NotEmpty String key, @Nonnull Level level,
```
emmberk: Removed `@Nonnull` annotations

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * @return non-null (may be empty) list of String {@value #DETAILS} of the {@link Alert}
+     */
+    @Nonnull
+    public List<String> getDetails() {
+        return (List<String>) this.get(DETAILS);
+    }
+
+    public enum Level {
```
emmberk: Moved

platform/admin/core/admin-core-alertservice-data/src/main/java/org/codice/ddf/admin/core/alert/service/data/internal/Alert.java
```diff
@@ -0,0 +1,168 @@
...
+            public String toString() {
+                return "info";
+            }
+        }, WARN {
+            public String toString() {
+                return "warning";
+            }
+        }, ERROR {
+            public String toString() {
+                return "danger";
```
emmberk: See reply to @ahoffer's same question

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,202 @@
...
+
+import org.codice.ddf.admin.core.alert.service.data.internal.Alert;
+import org.codice.ddf.admin.insecure.defaults.service.InsecureDefaultsService;
+import org.osgi.service.event.Event;
+import org.osgi.service.event.EventHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * <p>
```
emmberk: Removed experimental comment

platform/admin/ui/src/main/webapp/js/models/Alerts.js
emmberk: See reply to @ahoffer's question

emmberk:
build now

mcalcote:
Confirmed Alerts appear as expected in Chrome and Firefox. :+1:

emmberk:
I tested the banners on IE with @AzGoalie

emmberk:
build now

vinamartin:
Tested everything except using the debugger + dismiss. 
Docs look great and tested those instructions as well! :shipit:

emmberk:
@vinamartin and I tested adding the "Audit logging has failed alert" using the AdminAlertAppender by modifying the configuration in <INSTALL_HOME>/etc/log4j2.config.xml

emmberk:
build now

emmberk:
build now

emmberk:
I successfully heroed a full build locally

platform/admin/core/admin-core-alertservice/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>findbugs</artifactId>
+            <version>${findbugs.version}</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.4</version>
```
bdeining: ${commons-lang3.version}

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
```diff
@@ -0,0 +1,227 @@
...
+                mBeanServer.unregisterMBean(objectName);
+            } else {
+                LOGGER.debug(
+                        "Unable to unregister [{}] MBean under object name: [{}]. It is not registered.",
+                        CLASS_NAME,
+                        objectName.toString());
+            }
+        } catch (MBeanRegistrationException | InstanceNotFoundException e) {
+            String message = String.format(
+                    "Unable to unregistering [%s] MBean under object name: [%s].",
```
bdeining: nit : `unable to unregister`

platform/admin/core/admin-core-alertservice/pom.xml
emmberk: I had the same idea, but the commons-lang3 isn't standardized in this branch.

platform/admin/core/admin-core-alertservice/src/main/java/org/codice/ddf/admin/core/alert/service/internal/AlertServiceBean.java
emmberk: Good catch! FIxed.

platform/admin/core/admin-core-alertservice/pom.xml
bdeining: Ah, my mistake

PR #1840, opened 2017-03-30
DDF-2917 (2.10.x) Update caching to exclude base app pages and html

andrewkfiedler:
#### What does this PR do?
 - As of DDF-2893, we're caching everything far too aggressively.
 - This update allows apps to perform their own cache busting between versions, see DDF-2916.
 - Without this, a user will need to clear their cache (or wait a week) when upgrading to new versions.

@andrewkfiedler
@djblue 
@rzwiefel 
@kcwire
@pklinef

Go to the admin application.  Verify that the main app page and any html pages within do not get cached.  Verify that other assets do.

https://codice.atlassian.net/browse/DDF-2917

### DIFF
--- a/platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
+++ b/platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
@@ -23,6 +23,7 @@ import javax.servlet.FilterConfig;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


platform/platform-filter-response/src/main/java/org/codice/ddf/platform/response/filter/ResponseFilter.java
```diff
@@ -59,16 +60,17 @@
             X_CONTENT_SECURITY_POLICY + "=" + DEFAULT_CONTENT_SECURITY_POLICY,
             CACHE_CONTROL + "=" + DEFAULT_CACHE_CONTROL_VALUE);
 
-    @Override
-    public void init(FilterConfig filterConfig) throws ServletException {
-        LOGGER.debug("Initializing Response Security Filter.");
+    // Index paths (such as /search/catalog/ or /admin/) should never be cached
+    private void disableCachingForHTML(HttpServletRequest request, HttpServletResponse response) {
```
brendan-hofmann: I think this invalidates the comment in the pom. 😄 We should unit test each case handled here.

We should also probably update the documentation for this component, to reflect that we automatically disable caching for html and index paths, regardless of what is specified in the configuration.

andrewkfiedler: Sounds good, can you work with me on this update?  Trying to get it in for the 2.10.2 release.

pklinef: I agree that in general this should be configurable but that can maybe wait to be done on master with #1841 so this is at least fixed for 2.10.2.

brendan-hofmann: Given the upcoming release I agree those can wait for the master PR, like @pklinef said.

rzwiefel:
build now

rzwiefel:
Heroed.
Confirmed cache busting works on all root pages and html pages.

PR #1870, opened 2017-04-10
DDF-2920 Added initial jenkinsfile for Nightly Linux DDF pipelines build

mcalcote:
#### What does this PR do?
This initial PR is a WIP that's goal is to add the Jenkinsfile to the build and setup the format for improvements for the following tickets.

@clockard @oconnormi @LinkMJB 

**I will add more reviewers/committers after initial feedback or close and reopen when this is in a more finished state**

[DDF-2920](https://codice.atlassian.net/browse/DDF-2920)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,82 @@
+pipeline {
+    agent none
+    environment {
+        projectName = 'DDF'
+    }
+    stages {
+        stage('Setup') {
+            steps{
+                notifySlack('good', "STARTED: ${JOB_NAME} ${BUILD_NUMBER} ${BUILD_URL}")
+            }
+        }
+        stage('Parallel Build') {
+            //TODO Parallel Stage Syntax may soon be changing for declaritive Jenkinsfiles - https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel(
+                    linux: {
+                        node('linux-small') {
+                            timeout(time: 3, unit: 'HOURS') {
+                                checkout scm
+                                withMaven(maven: 'M3', mavenSettingsConfig: 'codice-maven-settings') {
+                                    sh 'mvn clean install'
+                                }
+                            }
+                        }
+                    }, windows: {
+                        node('proxmox-windows'){
+                            timeout(time: 3, unit: 'HOURS') {
+                                checkout scm
+                                withMaven(maven: 'M3', mavenSettingsConfig: 'codice-maven-settings') {
+                                    bat 'mvn clean install'
+                                }
+                            }
+                        }
+                    }
+                )
+            }
+        }
+        stage('Static Analysis') {
+            agent { label 'linux-small' }
+            steps{
+                timeout(time: 2, unit: 'MINUTES') {
+                    sh """echo Static Analysis run of "$projectName" """
+                    //TODO 6147 Add SA tools
+                }
+            }
+            post {
+                failure {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+    environment {
+        // What variables do we need here
+        projectName = 'DDF'
+    }
+    stages {
+        stage('Build') {
+            // Parallel Stage Syntax may soon be changing for declaritive Jenkinsfile(s)
+            // https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel (
```
mcalcote: pretty sure this isn't the correct usage

Jenkinsfile
```diff
@@ -0,0 +1,91 @@
+pipeline {
+    agent {
+        label 'linux-large'
+        //TODO 6146 windows builds ?using label 'proxmox-windows'?
```
clockard: might be able to use 'node' syntax so you can have the windows and linux work in parallel.

Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+    }
+    environment {
+        // What variables do we need here
+        projectName = 'DDF'
```
clockard: In the future I think this is where we would set variables to indicate if this is a pr build, master build or a nightly build.

Jenkinsfile
clockard: I think the parallel part might work at a higher level so we could run linux and windows builds in parallel

Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+                    "linuxBuild" : {
+                        steps {
+                            slackSend color: 'good', message: "STARTED: Job ${BUILD_NUMBER}"
+                            sh """echo Full build of the "$projectName" "$GIT_BRANCH" branch"""
+                            withMaven(maven: 'M3', mavenSettingsConfig: 'codice-maven-settings') {
+                                //TODO update mavenSettingsConfig with a valid config
+                                //sh 'mvn clean install -T 16 -P!documentation -P!findbugs -DskipTests=true -Dpmd.skip=true -Djacoco.skip=true -Dfindbugs.skip=true -Dcheckstyle.skip=true'
+
+                                //run full build here until we can clean the maven phases further
+                                sh 'mvn clean install'
```
clockard: I think initially all we want to skip itest here using something like ` -pl \!:test-itests-ddf` and then only build/run the itests in the Integration Test stage.

Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+                            }
+                        }
+                    }
+                )
+            }
+        }
+        stage('Unit Test') {
```
clockard: Until we get our poms/dependencies worked out we can't really run this separately.

Jenkinsfile
mcalcote: We might need to fallback to the 'node' scripted syntax. I will see how possible it is to manage agents in the declarative pipelines format.

Jenkinsfile
mcalcote: We can pull a little from the pre-exisiting pipelines environment variables. For the nightly(s) we could use the project names and that will likely be descriptive enough.

Jenkinsfile
mcalcote: Yeah, I wonder how far off we are....

mcalcote:
Closing to reopen and a full PR

PR #1883, opened 2017-04-12
DDF-2943 Fixes Spock tests running

rzwiefel:
#### What does this PR do?
Previously spock unit tests would not be ran unless they were suffixed with *Test. 
anyone

[Test](https://github.com/orgs/codice/teams/test)

@coyotesqrl 
@lessarderic

Ensure spock tests not ending in `Test` are ran and included in coverage

https://codice.atlassian.net/browse/DDF-2943

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -276,6 +276,12 @@
                     <version>2.8.1</version>
                     <configuration>
                         <argLine>${argLine} -Djava.awt.headless=true -noverify</argLine>
+                        <includes>
+                            <include>**/*.java</include>
+                            <!-- if the above include changes to **/*Test.java,
+                                 uncomment the line below -->
+                            <!--<include>**/*Spec.java</include>-->
+                        </includes>
                     </configuration>
                 </plugin>
                 <plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -277,10 +277,10 @@
                     <configuration>
                         <argLine>${argLine} -Djava.awt.headless=true -noverify</argLine>
                         <includes>
-                            <include>**/*.java</include>
-                            <!-- if the above include changes to **/*Test.java,
-                                 uncomment the line below -->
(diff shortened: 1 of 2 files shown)


rzwiefel:
build now

vinamartin:
Are the tests that weren't running passing?
That's a miracle! 🎉

pom.xml
```diff
@@ -276,6 +276,12 @@
                     <version>2.8.1</version>
                     <configuration>
                         <argLine>${argLine} -Djava.awt.headless=true -noverify</argLine>
+                        <includes>
+                            <include>**/*.java</include>
```
coyotesqrl: Spoke with @rzwiefel about this. It's confounding. Not clear how it doesn't now skip all Spock tests; not clear how it _fails_ if he includes `**/*.groovy` as an include pattern.

@lessarderic do you have some insight? It's not like the groovyc compiler is generating java files as an interim step...

lessarderic: Looks like this is a known issues: http://stackoverflow.com/a/34308282.

Could we do this instead? I tried and it seems to work:
```
<includes>
  <include>**/*Test.java</include>
  <include>**/Test*.java</include>
  <include>**/*TestCase.java</include>
  <include>**/*Spec.class</include>
```

Don't know if that's the plan but we should update all the other projects as well (e.g., Alliance).

rzwiefel: Hmmm I thought that we had test classes without Test on them at all (which then failed coverages checks). Will try this though and let you know :+1: Good find.

lessarderic: Shouldn't be the case. The default inclusion list only includes the first three lines I put in my comment.

rzwiefel:
build now

pom.xml
rzwiefel: Updated to the above list.

PR #1884, opened 2017-04-13
DDF-2945 Added unit tests for Historian

AzGoalie:
#### What does this PR do?
Adds unit test coverage to the Historian class.
@rzwiefel @ahoffer @codymacdonald 
@lessarderic [Test](https://github.com/orgs/codice/teams/test)
@coyotesqrl
@pklinef
#### Checklist:
- [x] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
@@ -98,6 +98,8 @@ public class Historian {
 
     private FilterBuilder filterBuilder;
 
+    private Security security;
+
     static {
         Bundle bundle = FrameworkUtil.getBundle(Historian.class);
         BundleContext context = bundle == null ? null : bundle.getBundleContext();
@@ -149,11 +151,13 @@ public class Historian {
     /**
      * Versions updated {@link Metacard}s and {@link ContentItem}s.
      *
-     * @param streamUpdateRequest   Needed to pass {@link MetacardVersion#SKIP_VERSIONING}
+     * @param streamUpdateRequest   Needed to pass {@link ddf.catalog.core.versioning.MetacardVersion#SKIP_VERSIONING}
      *                              flag into downstream update
      * @param updateStorageResponse Versions this response's updated items
      * @return the update response originally passed in
-     * @throws IOException
+     * @throws UnsupportedQueryException
+     * @throws SourceUnavailableException
+     * @throws IngestException
      */
     public UpdateStorageResponse version(UpdateStorageRequest streamUpdateRequest,
             UpdateStorageResponse updateStorageResponse, UpdateResponse updateResponse)
@@ -442,8 +446,11 @@ public class Historian {
      * @return result of the callable func
      */
     private <T> T executeAsSystem(Callable<T> func) {
-        Subject systemSubject = Security.runAsAdmin(() -> Security.getInstance()
-                .getSystemSubject());
+        if (security == null) {
+            security = Security.getInstance();
+        }
+
+        Subject systemSubject = Security.runAsAdmin(() -> security.getSystemSubject());
         if (systemSubject == null) {
             throw new RuntimeException("Could not get systemSubject to version metacards.");
         }
@@ -497,6 +504,10 @@ public class Historian {
 
     public void setMetacardTypes(List<MetacardType> metacardTypes) {
         this.metacardTypes = metacardTypes;
+    }
+
+    public void setSecurity(Security security) {
+        this.security = security;
     }
 
     private static class WrappedByteSource extends ByteSource {
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
@@ -332,6 +332,7 @@ public class Historian {
                 .filter(ci -> ci.getQualifier() == null || ci.getQualifier()
                         .equals(""))
                 .map(ContentItem::getMetacard)
+                .filter(Objects::nonNull)
                 .collect(Collectors.toList()));
     }
 
@@ -506,7 +507,7 @@ public class Historian {
         this.metacardTypes = metacardTypes;
     }
 
-    public void setSecurity(Security security) {
+    void setSecurity(Security security) {
         this.security = security;
     }
 
--- a/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
+++ b/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
@@ -19,8 +19,10 @@ import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
 import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyObject;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
@@ -408,19 +410,14 @@ public class HistorianTest {
         UpdateStorageResponse storageResponse = mock(UpdateStorageResponse.class);
         UpdateResponse updateResponse = mock(UpdateResponse.class);
 
-        storeMetacard(metacard);
-
-        // send a request to update the metacard
+        // create the update request
         updateMetacard(storageRequest, storageResponse, metacard);
-        storageProvider.update(storageRequest);
 
         mockQuery(metacard);
         historian.version(storageRequest, storageResponse, updateResponse);
 
         // Verify that it wasn't updated
-        Metacard update = readMetacard();
-
-        assertThat(update, equalTo(metacard));
+        verify(catalogProvider, times(0)).create(anyObject());
     }
 
     @Test(expected = IngestException.class)
@@ -433,6 +430,45 @@ public class HistorianTest {
         StorageProvider exceptionStorageProvider = mock(StorageProvider.class);
         doThrow(StorageException.class).when(exceptionStorageProvider)
                 .commit(any());
+
+        ContentItem item = mock(ContentItem.class);
+        when(item.getId()).thenReturn(METACARD_ID);
+        when(item.getUri()).thenReturn(RESOURCE_URI);
+        when(item.getMetacard()).thenReturn(metacards.get(0));
+
+        ReadStorageResponse readStorageResponse = mock(ReadStorageResponse.class);
+        when(readStorageResponse.getContentItem()).thenReturn(item);
+        when(exceptionStorageProvider.read(any())).thenReturn(readStorageResponse);
+
+        when(exceptionStorageProvider.create(any())).thenReturn(mock(CreateStorageResponse.class));
+
+        historian.setStorageProviders(Collections.singletonList(exceptionStorageProvider));
+
+        // Parameters for historian
+        UpdateStorageRequest storageRequest = mock(UpdateStorageRequest.class);
+        UpdateStorageResponse storageResponse = mock(UpdateStorageResponse.class);
+        UpdateResponse updateResponse = mock(UpdateResponse.class);
+
+        // send a request to update the metacard
+        updateMetacard(storageRequest, storageResponse, metacards.get(1));
+
+        mockQuery(metacards.get(1));
+        historian.version(storageRequest, storageResponse, updateResponse);
+        verify(exceptionStorageProvider).rollback(any(StorageRequest.class));
+    }
+
+    @Test(expected = IngestException.class)
+    public void testRollbackFailed()
+            throws StorageException, UnsupportedQueryException, SourceUnavailableException,
+            IngestException {
+        List<Metacard> metacards = getMetacardUpdatePair();
+
+        // Mock out a bad storage provider
+        StorageProvider exceptionStorageProvider = mock(StorageProvider.class);
+        doThrow(StorageException.class).when(exceptionStorageProvider)
+                .commit(any());
+        doThrow(StorageException.class).when(exceptionStorageProvider)
+                .rollback(any());
 
         ContentItem item = mock(ContentItem.class);
         when(item.getId()).thenReturn(METACARD_ID);
--- a/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
+++ b/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
@@ -16,12 +16,15 @@ package ddf.catalog.history;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasEntry;
+import static org.hamcrest.Matchers.nullValue;
 import static org.mockito.Matchers.any;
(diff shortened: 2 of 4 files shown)


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
```diff
@@ -499,6 +506,10 @@ public void setMetacardTypes(List<MetacardType> metacardTypes) {
         this.metacardTypes = metacardTypes;
     }
 
+    public void setSecurity(Security security) {
```
coyotesqrl: Not a fan of methods added solely to support unit testing, but as you've few alternatives, at least drop it to package-private.

rzwiefel:
@lessarderic (@azgoalie) I'm going to need to be working on that static block where we register metacard types so I can cover moving those out to an init method and getting coverage on them.

AzGoalie:
@lessarderic 
- Looks like we don't have a test that covers L323-324
- - L323-324 are covered, jacoco and intellij just don't show that it is :(
- We should have a test for the case where we fail to get the input stream of a ContentItem (L522)
- - Wouldn't it be covered by tests of `ContentItemImpl`?
- We need a test to cover the path where a metacard cannot be found in the list of created content items of a CreateStorageResponse (L482)
- - From what i can tell, it's not possible for it to get here without being able to find the metacard.

AzGoalie:
build now

lessarderic:
6: 👍

8: @rzwiefel, any way this could ever happen? If not, do we need to `null` check?

9: Not really. What we need to test here is the case where `WrappedByteSource.openStream()` throws an `IOException`. We should be able to test that by having the `storageProvider().read()` on L352 return a `ReadStorageResponse` that contains a mock `ContentItem` that throws an exception when its `getInputStream()` method is called.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+
+        mockQuery(metacards.get(1));
+        historian.version(storageRequest, storageResponse, updateResponse);
+
+        // Verify that the metacard updated
+        Metacard update = readMetacard();
+
+        assertThat(update, equalTo(metacards.get(1)));
+    }
+
```
lessarderic: ❓Do we have any assertions in the tests to verify that the resource URI are properly set on the metacards (set by `setResourceUriForContent()` method)?

rzwiefel:
@lessarderic (I'm assuming you mean inside the `setResourceUriForContent`method? line numbers are a bit off so just clarifying). I don't believe that case is possible.  (line numbers based off current master verison)

so we fetched this list of created content items based on the deleted metacards, the map of `contentItems` returned is a map of [ContentItem ID: ContentItem]. [code](https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java#L208)

Then we create the versioned metacards based of those same deleted metacards we just fetched the content for. This `versionedMap` returned is a map of [MetacardVersion.VERSION_OF_ID: MetacardVersion]. (the MetacardVersion.VERSION_OF_ID means the ID of the original metacard that was versioned). This `version_of_id` should correlate to the ContentItem::getId used in the above map of content items. [code](https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java#L211-L216)

We then go and create the Content Items that go along with the versioned metacard (based off that 'primary key' we've been using of the original id/ContentItem id) and then insert those via the StorageProvider which returns us the `createStorageResponse`. [code](https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java#L218-L219)

So that means going into the setResourceUriForContent method we have the `versionedMap` of [original id: version metacard] and the createStorageResponse of all content items that we just created and stored. [code](https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java#L220-L222)

In the setResourceUriForContent method that means that the list of created content items we are iterating through was created based on the same list of deleted metacards that the `versionMetacards` list we are looking for a match in was made from. So there should always be a match to correspond to the content item. If for whatever reason that metacard **isn't** there (which shouldn't happen), we do check for null and continue without setting the resource Uri on the metacard.

(*note*: If given `n` deleted metacards , not all of them may have content.. which would result in some number less than `n` content items being found. However all n deleted metacards should have been versioned, meaning there should always **at least** be a content item for every metacard. but not vice versa). 

However one thing that isnt' being accounted for (which I will fix and put up a PR for ) is I have not filtered out the versioned/deleted metacards from the deleteResponse before fetching the content items (we don't ever want to version a version/delete) . So right now it could technically have searched for versioned metacards content items (even though we will never version them).  (though that is definitely not pertinent to the specific method in question, it still does not need to be tested). though erics comment 
>❗️We may want to add non-versioned and deleted metacards to that list to cover L435-436, and verify that those were properly excluded.

should be a good test to ensure this doesn't happen. I would actually expect this test to fail right now, though its a true positive failure. So i will get up a fix PR asap :) 

(but regarding the original method, the failure case is covered even if it does somehow happen)

rzwiefel:
I will also take some of the previous comment and add some as documentation in these areas so its a little clearer what is happening :+1:

rzwiefel:
@lessarderic @AzGoalie I can also update the behavior of fetching the first catalog/storage provider to throw an exception instead of returning null.

rzwiefel:
Addressed in https://github.com/codice/ddf/pull/1895

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
```diff
@@ -332,6 +332,7 @@ private Filter forIds(List<String> ids) {
                 .filter(ci -> ci.getQualifier() == null || ci.getQualifier()
                         .equals(""))
                 .map(ContentItem::getMetacard)
+                .filter(Objects::nonNull)
```
rzwiefel: Since I refactored this method in my PR to fix this, I went ahead and added this line to where I refactored it to. So if you wish you can drop this to avoid conflicts.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+        Security security = mock(Security.class);
+        Subject subject = mock(MockSubject.class);
+        when(subject.execute(any(Callable.class))).thenCallRealMethod();
+        when(security.getSystemSubject()).thenReturn(subject);
+        historian.setSecurity(security);
+    }
+
+    @Test
+    public void testDoSkip() throws SourceUnavailableException, IngestException {
```
AzGoalie: Done

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+        historian.setHistoryEnabled(false);
+        historian.version(mock(UpdateResponse.class));
+
+        verifyZeroInteractions(catalogProvider);
+    }
+
+    @Test
+    public void testSetSkipFlag() throws SourceUnavailableException, IngestException {
```
AzGoalie: Done

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+        Map<String, Serializable> properties = new HashMap<>();
+        UpdateResponse updateResponse = createUpdateResponse(properties);
+
+        historian.version(updateResponse);
+
+        assertThat(properties, hasEntry(MetacardVersion.SKIP_VERSIONING, true));
+    }
+
+    @Test
+    public void testSkipProperty() throws SourceUnavailableException, IngestException {
```
AzGoalie: Done

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+        Map<String, Serializable> properties = new HashMap<>();
+        properties.put(MetacardVersion.SKIP_VERSIONING, true);
+
+        UpdateResponse updateResponse = createUpdateResponse(properties);
+
+        historian.version(updateResponse);
+        // Called once for skip check
+        verify(updateResponse).getProperties();
```
AzGoalie: Updated

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    @Test
+    public void testUpdateResponse() throws Exception {
+        UpdateResponse updateResponse = createUpdateResponse(null);
+        List<Update> updateList = createUpdatedMetacardList();
```
AzGoalie: Added deleted and non versioned metacards to the mock query function, this method is used to just get an updated pair of metacards.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+
+        Metacard versionedMetacard = createRequest.getValue()
+                .getMetacards()
+                .get(0);
+        assertThat(versionedMetacard.getAttribute(MetacardVersion.VERSION_OF_ID)
+                .getValue(), equalTo(METACARD_ID));
+    }
+
+    @Test
+    public void testUpdateStorageResponseSkip()
```
AzGoalie: Updated to `testUpdateStorageResponseEmptyStorageRequest`

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+            throws UnsupportedQueryException, SourceUnavailableException, IngestException {
+        UpdateStorageRequest storageRequest = mock(UpdateStorageRequest.class);
+        UpdateStorageResponse storageResponse = mock(UpdateStorageResponse.class);
+        UpdateResponse updateResponse = mock(UpdateResponse.class);
+
+        historian.version(storageRequest, storageResponse, updateResponse);
+
+        assertThat(storageProvider.storageMap.size(), equalTo(0));
```
AzGoalie: Can't verify zero interactions with the storage provider (not a mock), but added a check for that catalog provider.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+        historian.version(storageRequest, storageResponse, updateResponse);
+
+        // Verify that it wasn't updated
+        Metacard update = readMetacard();
+
+        assertThat(update, equalTo(metacard));
+    }
+
+    @Test(expected = IngestException.class)
+    public void testTryCommitStorageException()
```
AzGoalie: What will this test assert on? A failure on the rollback just prints a log message and continues off the previous storage exception which is covered by this test.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+        // Parameters for historian
+        UpdateStorageRequest storageRequest = mock(UpdateStorageRequest.class);
+        UpdateStorageResponse storageResponse = mock(UpdateStorageResponse.class);
+        UpdateResponse updateResponse = mock(UpdateResponse.class);
+
+        // send a request to update the metacard
+        updateMetacard(storageRequest, storageResponse, metacards.get(1));
+
+        mockQuery(metacards.get(1));
+        historian.version(storageRequest, storageResponse, updateResponse);
```
AzGoalie: Updated

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/history/Historian.java
AzGoalie: I need this as this was added to fix a bug that a test caught. I'll fix the collision when it happens.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,462 @@
...
+
+        // Verify that the metacard updated
+        Metacard update = readMetacard();
+
+        assertThat(update, equalTo(metacards.get(1)));
+    }
+
+    @Test
+    public void testUpdateStorageResponseNoContentItems()
```
AzGoalie: Added empty and non-null qualifiers to `updateMetacard` method.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
AzGoalie: Added

AzGoalie:
Pushed changes addressing comments.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
lessarderic: We basically need to make sure the `StorageException` is caught and ignored when `rollback` fails, but that a `IngestException` still gets thrown. The flow and expectation would be very similar to this test but with one more step, i.e., `commit` fails then `rollback` also fails.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,602 @@
...
+    private Historian historian;
+
+    @Before
+    public void setup() {
+        historian = new Historian();
+
+        catalogProvider = mock(CatalogProvider.class);
+        historian.setCatalogProviders(Collections.singletonList(catalogProvider));
+
+        storageProvider = new InMemoryStorageProvider();
```
AzGoalie: It was a lot easier to assert on the adding/removing of objects in this than trying to get a mock to use its own map.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
```diff
@@ -0,0 +1,602 @@
...
+        assertThat(item, nullValue());
+    }
+
+    @Test
+    public void testReadStorageException()
+            throws StorageException, SourceUnavailableException, IngestException,
+            UnsupportedQueryException, URISyntaxException {
+        StorageProvider exceptionStorageProvider = mock(StorageProvider.class);
+        when(exceptionStorageProvider.read(any())).thenThrow(StorageException.class);
+        historian.setStorageProviders(Collections.singletonList(exceptionStorageProvider));
```
AzGoalie: No, fixed and changed the assertion to make sure that `catalogProvider.create` was never called

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/history/HistorianTest.java
AzGoalie: Added

AzGoalie:
build now

PR #1887, opened 2017-04-13
DDF-2920 Added initial jenkinsfile for pipelines build

mcalcote:
#### What does this PR do?
The goal of this change is to add the initial Jenkinsfile to the project repo and establish the format for follow-on work. 

@oconnormi @LinkMJB @emmberk

[Build](https://github.com/orgs/codice/teams/build) @brendan-hofmann 

@lessarderic 
@clockard 

This job is currently running in the test Jenkins environment. After this ticket is merged we can set up a job for it on the standard jenkins to eventually migrate our nightly (and beyond) to.

This jenkinsfile utilizes the declarative syntax. This is currently in its 1.0 state and will likely have refinements and improvements rolling out in the near future.

[DDF-2920](https://codice.atlassian.net/browse/DDF-2920)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,64 @@
+//"Jenkins Pipeline is a suite of plugins which supports implementing and integrating continuous delivery pipelines into Jenkins. Pipeline provides an extensible set of tools for modeling delivery pipelines "as code" via the Pipeline DSL."
+//More information can be found on the Jenkins Documentation page https://jenkins.io/doc/
+pipeline {
+    agent none
+    options {
+        buildDiscarder(logRotator(numToKeepStr:'25'))
+    }
+    stages {
+        stage('Setup') {
+            steps{
+                slackSend color: 'good', message: "STARTED: ${JOB_NAME} ${BUILD_NUMBER} ${BUILD_URL}"
+            }
+        }
+        stage('Parallel Build') {
+            // TODO refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel(
+                    linux: {
+                        node('linux-large') {
+                            timeout(time: 3, unit: 'HOURS') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    sh 'mvn clean install -pl !distribution/test/itests/test-itests-ddf'
+                                    sh 'mvn install -pl distribution/test/itests/test-itests-ddf'
+                                }
+                            }
+                        }
+                    }, windows: {
+                        node('proxmox-windows'){
+                            timeout(time: 3, unit: 'HOURS') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    bat 'mvn clean install -pl !distribution/test/itests/test-itests-ddf'
+                                    bat 'mvn install -pl distribution/test/itests/test-itests-ddf'
+                                }
+                            }
+                        }
+                    }
+                )
+            }
+        }
+        stage('Deploy') {
+            agent { label 'linux-small' }
+            steps{
+                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice_settings.xml') {
+                    checkout scm
+                    sh 'mvn javadoc:aggregate -Pjavadoc -DskipStatic=true -DskipTests=true'
+                    sh 'mvn deploy -DskipStatic=true -DskipTests=true'
+                }
+            }
+        }
+    }
+    post {
+        success {
+            slackSend color: 'good', message: "SUCCESS: ${JOB_NAME} ${BUILD_NUMBER}"
+        }
+        failure {
+            slackSend color: '#ea0017', message: "FAILURE: ${JOB_NAME} ${BUILD_NUMBER}. See the results here: ${BUILD_URL}"
+        }
+        unstable {
+            slackSend color: '#ffb600', message: "UNSTABLE: ${JOB_NAME} ${BUILD_NUMBER}. See the results here: ${BUILD_URL}"
+        }
+    }
+}


Jenkinsfile
```diff
@@ -0,0 +1,82 @@
...
+            steps{
+                timeout(time: 2, unit: 'MINUTES') {
+                    sh """echo Deploy the artifacts of "$projectName" """
+                }
+            }
+        }
+    }
+    post {
+        success {
+            notifySlack('good', "SUCCESS: ${JOB_NAME} ${BUILD_NUMBER}")
```
emmberk: How do these slack notifications behave right now?

LinkMJB: See: https://connexta.slack.com/archives/C4T1TAGC9/p1492110209515483

Jenkinsfile
```diff
@@ -0,0 +1,82 @@
...
+            steps{
+                notifySlack('good', "STARTED: ${JOB_NAME} ${BUILD_NUMBER} ${BUILD_URL}")
+            }
+        }
+        stage('Parallel Build') {
+            //TODO Parallel Stage Syntax may soon be changing for declaritive Jenkinsfiles - https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel(
+                    linux: {
+                        node('linux-small') {
```
emmberk: Why is there a mix between using `agent` and `node`?

mcalcote: Agent is the preferred syntax used in a declarative model, but the parallel functionality still has 'scripted' syntax elements. This might change as part of the improvement to the parallel syntax for declarative jenkinsfiles.

Jenkinsfile
```diff
@@ -0,0 +1,82 @@
...
+        projectName = 'DDF'
+    }
+    stages {
+        stage('Setup') {
+            steps{
+                notifySlack('good', "STARTED: ${JOB_NAME} ${BUILD_NUMBER} ${BUILD_URL}")
+            }
+        }
+        stage('Parallel Build') {
+            //TODO Parallel Stage Syntax may soon be changing for declaritive Jenkinsfiles - https://issues.jenkins-ci.org/browse/JENKINS-41334
```
emmberk: Good find on this ticket!

Since I had some confusion understanding what you meant, this TODO could be improved to something like
> // TODO refactor this stage from scripted syntax to declarative syntax to match the rest of this file - https://issues.jenkins-ci.org/browse/JENKINS-41334

mcalcote: Updated 👍

Jenkinsfile
```diff
@@ -0,0 +1,82 @@
...
+            steps{
+                notifySlack('good', "STARTED: ${JOB_NAME} ${BUILD_NUMBER} ${BUILD_URL}")
+            }
+        }
+        stage('Parallel Build') {
+            // TODO refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel(
+                    linux: {
+                        node('linux-small') {
```
clockard: this should be linux-large

Jenkinsfile
```diff
@@ -0,0 +1,82 @@
...
+                }
+            }
+            post {
+                failure {
+                    sh """Static Analysis failure"""
+                }
+            }
+        }
+        stage('Deploy') {
+            agent { label 'linux-small' }
```
clockard: I realize there is nothing really in this method but right now since this is done with a separate agent, a deploy would have to rebuild everything that was previously built in the "Parallel Build" section

mcalcote: True, we don't have a good solution to passing the artifacts to this stage... Possible solutions include:
A) generate them again with a quick build (redundant/wasteful) 
B) move the deploy stage into the parallel block (this will be better with the declarative update to parallel syntax)

I can remove the placeholder stage for now, but we will have to see where the dust settles on the update to the stages. I imagine we will end up with one uber stage with the single parallel step and nested inside will be the stages for each build. Then we will have more granular control.

Jenkinsfile
mcalcote: super-sized

Jenkinsfile
oconnormi: FYI, I started looking at how to put notifications into a shared library so that we can have a bit more flexibility. Things like having multiple places that notifications are sent to (slack, email, ...), being able to override the default slack channel and a few other things.

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+                                withMaven(maven: 'M3', mavenSettingsConfig: 'codice-maven-settings') {
+                                    sh 'mvn clean install'
+                                }
+                            }
+                        }
+                    }, windows: {
+                        node('proxmox-windows'){
+                            timeout(time: 3, unit: 'HOURS') {
+                                checkout scm
+                                withMaven(maven: 'M3', mavenSettingsConfig: 'codice-maven-settings') {
```
oconnormi: we should add a global settings file called something generic like `default-global-settings` that contains stuff like any maven mirrors and other things specific to the build environment since the Maven Pipeline Plugin doesn't honor the default settings files provided by jenkins master until [JENKINS-39407](https://issues.jenkins-ci.org/browse/JENKINS-39407) is resolved

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+        failure {
+            notifySlack('#ea0017', "FAILURE: ${JOB_NAME} ${BUILD_NUMBER}. See the results here: ${BUILD_URL}")
+        }
+        unstable {
+            notifySlack('#ffb600', "UNSTABLE: ${JOB_NAME} ${BUILD_NUMBER}. See the results here: ${BUILD_URL}")
+        }
+    }
+}
+
+def notifySlack(String color, String message) {
```
oconnormi: we can probably just change to calling `slackSend` directly since all this does at the moment is proxy that call. I think originally this method had more logic in it to determine the message that should be sent based on the build status

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
+pipeline {
```
oconnormi: we should probably add some [options](https://jenkins.io/doc/book/pipeline/syntax/#options) to this for stuff like `buildDiscarder`

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+    agent none
+    environment {
+        projectName = 'DDF'
```
oconnormi: Is this being used anywhere? If not we should probably remove it

Jenkinsfile
```diff
@@ -0,0 +1,54 @@
...
+            steps{
+                notifySlack('good', "STARTED: ${JOB_NAME} ${BUILD_NUMBER} ${BUILD_URL}")
+            }
+        }
+        stage('Parallel Build') {
+            // TODO refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel(
+                    linux: {
+                        node('linux-large') {
```
oconnormi: I think these should be `agent { label 'linux-large' }` according to the jenkins declarative pipeline model

oconnormi: Nevermind, just read through the comments related to this above :-)

mcalcote: When declaring the agent at the steps level it is that format, but inside the parallel block it uses the scripted `node(` syntax. This should change when they get the `parallelSteps` command change merge in.

Jenkinsfile
mcalcote: I deleted out the placeholder stages that were using it. Removed.

Jenkinsfile
oconnormi: Yea I only saw the comment regarding that after I wrote my review

Jenkinsfile
mcalcote: Replaced

Jenkinsfile
mcalcote: Thought initially it is called notifySlack, I thought that method could be adapted to send an email as well, but I will remove it for now until that is sorted out.

Jenkinsfile
mcalcote: Thanks for adding it to Jenkins, I update the withMaven calls.

Jenkinsfile
mcalcote: added, with the initial amount of 5

oconnormi: I think when this replaces the legacy jobs we'll probably want more than 5 but for now that should be fine 😄

oconnormi:
Build Now

clockard:
I was looking back at the ticket for this and it is missing the deploy step. 
The current jenkins job runs these two maven targets after a successful build
```mvn javadoc:aggregate -Pjavadoc -DskipTests=true -Dpmd.skip=true -Djacoco.skip=true -Dfindbugs.skip=true -Dcheckstyle.skip=true -nsu
mvn deploy -nsu -DskipTests=true -Dpmd.skip=true -Dfindbugs.skip=true -Dcheckstyle.skip=true -Djacoco.skip=true```

File: Jenkinsfile [hunk 11]
```diff
@@ -0,0 +1,61 @@
...
+                            }
+                        }
+                    }
+                )
+            }
+        }
+        stage('Deploy') {
+            agent { label 'linux-small' }
+            steps{
+                withMaven(maven: 'M3', globalMavenSettingsConfig: 'phxOfficeGlobalSettings', mavenSettingsConfig: 'codice_settings.xml') {
```
oconnormi: Use `default-global-settings` instead :smile:

Jenkinsfile
```diff
@@ -0,0 +1,61 @@
...
+                    sh 'mvn javadoc:aggregate -Pjavadoc -DskipStatic=true -DskipTests=true'
+                    sh 'mvn deploy -DskipStatic=true -DskipTests=true'
```
oconnormi: Is the goal here to deploy just the javadoc or the whole project? And on all branches and prs or just branches matching the `Maj.min.X` pattern and master?

Jenkinsfile
```diff
@@ -0,0 +1,61 @@
...
+                                }
+                            }
+                        }
+                    }
+                )
+            }
+        }
+        stage('Deploy') {
+            agent { label 'linux-small' }
+            steps{
```
oconnormi: You'll probably need a `checkout scm` step first since this is a new node

oconnormi: Wow so now I see it but it totally didn't show up when I wrote this comment...

Jenkinsfile
```diff
@@ -0,0 +1,61 @@
...
+        stage('Parallel Build') {
+            // TODO refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
+            steps{
+                parallel(
+                    linux: {
+                        node('linux-large') {
+                            timeout(time: 3, unit: 'HOURS') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    sh 'mvn clean install -pl !test-itest-ddf'
```
clockard: Missing the `:` before test-test-ddf
Should we make these multi-threaded? `-T 1.5C`?

Jenkinsfile
mcalcote: Initially this jenkinsfile will only serve the purpose of a "Nightly" build and as such does not take into account the flow control that will be necessary (and added) as part of the follow on work. The expectation here is that the javadoc plus the artifacts will be uploaded with these two scripts.

Jenkinsfile
mcalcote: 👍

LinkMJB:
@mcalcote you may also want to consider increasing the build timeout for the jobs from 3 hours to 6 hours. Granted the legacy Jenkins projects perform multiple steps sequentially, so the concurrent nature of the job may make it so that the timeout can be shorter.

Jenkinsfile
mcalcote: I will go ahead and bump to 25 to match the jobs we will be replacing.

mcalcote:
@LinkMJB recent runs on the test jenkins environment had the windows build finishing in just over 2 hours and the linux build even faster. Since those tasks are run in parallel and the timeout isn't started until the agent is secured we probably don't want to move that up too much. What do you guys think @clockard, @oconnormi ?

oconnormi:
@mcalcote yeah that sounds reasonable to me 👍

Jenkinsfile
```diff
@@ -0,0 +1,61 @@
+pipeline {
```
lessarderic: ❓ Would it be worth adding a high-level comment on `pipeline` to explain what it's for and what it does?

mcalcote: I will work on it 👍

Jenkinsfile
mcalcote: I previously didn't need the checkout scm step when running the agent from the strictly declaritive state... That being said, I'm not sure on the behaviour after a 'scripted' parallel step is completed.

Jenkinsfile
mcalcote: When I used the `:` I saw an error: `Could not find the selected project in the reactor`, so I replaced with the directory which seemed to work. 

Would increasing the number of threads introduce any instability? The build time doesn't seem to be prohibitively long currently.

clockard:
build now

mcalcote:
Build Now

PR #1904, opened 2017-04-18
DDF-2938 Add unit test coverage to CachingFederationStrategy

codymacdonald:
#### What does this PR do?
Adds unit tests for CachingFederationStrategy

@ahoffer 
@emanns95 
@brendan-hofmann 
@mweser 
@AzGoalie 

[Test](https://github.com/orgs/codice/teams/test)

@lessarderic
@pklinef

Build and run unit tests.

Goal of the ticket was 70% code coverage.

[DDF-2938](https://codice.atlassian.net/browse/DDF-2938)

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/pom.xml
+++ b/catalog/core/catalog-core-standardframework/pom.xml
@@ -518,17 +518,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.67</minimum>
+                                            <minimum>0.74</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.57</minimum>
+                                            <minimum>0.64</minimum>
                                         </limit>
                                         <limit>
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.58</minimum>
+                                            <minimum>0.64</minimum>
                                         </limit>
                                     </limits>
                                 </rule>
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
@@ -26,13 +26,10 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.ExecutorCompletionService;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.Phaser;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.Validate;
 import org.codice.ddf.configuration.SystemInfo;
 import org.opengis.filter.sort.SortOrder;
 import org.slf4j.Logger;
@@ -118,7 +115,10 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
      */
     protected static final String UPDATE_QUERY_MODE = "update";
 
-    private static final int DEFAULT_MAX_START_INDEX = 50000;
+    /**
+     * package-private to allow for unit testing
+     */
+    static final int DEFAULT_MAX_START_INDEX = 50000;
 
     private static final Logger LOGGER = LoggerFactory.getLogger(CachingFederationStrategy.class);
 
@@ -140,11 +140,14 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
      */
     protected List<PostFederatedQueryPlugin> postQuery;
 
+    private SortedQueryMonitorFactory sortedQueryMonitorFactory =
+            new SortedQueryMonitorFactory(this);
+
     private ExecutorService queryExecutorService;
 
     private int maxStartIndex;
 
-    private CacheCommitPhaser cacheCommitPhaser = new CacheCommitPhaser();
+    private CacheCommitPhaser cacheCommitPhaser;
 
     private CacheBulkProcessor cacheBulkProcessor;
 
@@ -169,20 +172,46 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
             List<PreFederatedQueryPlugin> preQuery, List<PostFederatedQueryPlugin> postQuery,
             SolrCache cache, ExecutorService cacheExecutorService,
             ValidationQueryFactory validationQueryFactory, CacheQueryFactory cacheQueryFactory) {
+
+        Validate.notNull(queryExecutorService, "Valid queryExecutorService required.");
+        Validate.notNull(preQuery, "Valid List<PreFederatedQueryPlugin> required.");
+        Validate.noNullElements(preQuery, "preQuery cannot contain null elements.");
+        Validate.notNull(postQuery, "Valid List<PostFederatedQueryPlugin> required.");
+        Validate.noNullElements(postQuery, "postQuery cannot contain null elements.");
+        Validate.notNull(cache, "Valid SolrCache required.");
+        Validate.notNull(cacheExecutorService, "Valid cacheExecutorService required.");
+        Validate.notNull(validationQueryFactory, "Valid ValidationQueryFactory required.");
+        Validate.notNull(cacheQueryFactory, "Valid CacheQueryFactory required.");
+
         this.queryExecutorService = queryExecutorService;
         this.preQuery = preQuery;
         this.postQuery = postQuery;
         this.maxStartIndex = DEFAULT_MAX_START_INDEX;
         this.cache = cache;
         this.cacheExecutorService = cacheExecutorService;
+        cacheCommitPhaser = new CacheCommitPhaser(cache);
         cacheBulkProcessor = new CacheBulkProcessor(cache);
         this.validationQueryFactory = validationQueryFactory;
         this.cacheQueryFactory = cacheQueryFactory;
         cacheSource = new SolrCacheSource(cache);
     }
 
+    void setSortedQueryMonitorFactory(SortedQueryMonitorFactory sortedQueryMonitorFactory) {
+        this.sortedQueryMonitorFactory = sortedQueryMonitorFactory;
+    }
+
+    void setCacheCommitPhaser(CacheCommitPhaser cacheCommitPhaser) {
+        this.cacheCommitPhaser = cacheCommitPhaser;
+    }
+
+    void setCacheBulkProcessor(CacheBulkProcessor cacheBulkProcessor) {
+        this.cacheBulkProcessor = cacheBulkProcessor;
+    }
+
     @Override
     public QueryResponse federate(List<Source> sources, QueryRequest queryRequest) {
+        Validate.noNullElements(sources, "Cannot federate with null sources.");
+        Validate.notNull(queryRequest, "Cannot federate with null QueryRequest.");
         Set<String> sourceIds = new HashSet<>();
         for (Source source : sources) {
             sourceIds.add(source.getId());
@@ -291,10 +320,11 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
                     offset));
         }
 
-        queryExecutorService.submit(createMonitor(queryCompletion,
+        queryExecutorService.submit(sortedQueryMonitorFactory.createMonitor(queryCompletion,
                 futures,
                 queryResponseQueue,
-                modifiedQueryRequest));
+                modifiedQueryRequest,
+                postQuery));
 
         QueryResponse queryResponse;
         if (offset > 1 && sources.size() > 1) {
@@ -418,14 +448,8 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
         return input;
     }
 
-    private List<Metacard> getMetacards(List<Result> results) {
-        List<Metacard> metacards = new ArrayList<>(results.size());
-
-        for (Result result : results) {
-            metacards.add(result.getMetacard());
-        }
-
-        return metacards;
+    int getMaxStartIndex() {
+        return maxStartIndex;
     }
 
     /**
@@ -464,24 +488,28 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
         this.cacheRemoteIngests = cacheRemoteIngests;
     }
 
-    protected Runnable createMonitor(final CompletionService<SourceResponse> completionService,
-            final Map<Future<SourceResponse>, QueryRequest> futures,
-            final QueryResponseImpl returnResults, final QueryRequest request) {
-
-        return new SortedQueryMonitor(this,
-                completionService,
-                futures,
-                returnResults,
-                request,
-                postQuery);
-    }
-
     public void shutdown() {
         cacheCommitPhaser.shutdown();
         cacheBulkProcessor.shutdown();
     }
 
-    private static class OffsetResultHandler implements Runnable {
+    public boolean getShowErrors() {
+        return showErrors;
+    }
+
+    public void setShowErrors(boolean showErrors) {
+        this.showErrors = showErrors;
+    }
+
+    public boolean getShowWarnings() {
+        return showWarnings;
+    }
+
+    public void setShowWarnings(boolean showWarnings) {
+        this.showWarnings = showWarnings;
+    }
+
+    static class OffsetResultHandler implements Runnable {
 
         private QueryResponseImpl originalResults = null;
 
@@ -491,8 +519,8 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
 
         private int offset = 1;
 
-        private OffsetResultHandler(QueryResponseImpl originalResults,
-                QueryResponseImpl offsetResultQueue, int pageSize, int offset) {
+        OffsetResultHandler(QueryResponseImpl originalResults, QueryResponseImpl offsetResultQueue,
+                int pageSize, int offset) {
             this.originalResults = originalResults;
             this.offsetResultQueue = offsetResultQueue;
             this.pageSize = pageSize;
@@ -520,24 +548,6 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
         }
     }
 
-    /**
-     * Runnable that makes one party arrive to a phaser on each run
-     */
-    private static class PhaseAdvancer implements Runnable {
-
-        private final Phaser phaser;
-
-        public PhaseAdvancer(Phaser phaser) {
-            this.phaser = phaser;
-        }
-
-        @Override
-        public void run() {
-            phaser.arriveAndAwaitAdvance();
-        }
-
-    }
-
     private class CallableSourceResponse implements Callable<SourceResponse> {
 
         private final QueryRequest request;
@@ -552,8 +562,12 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
         @Override
         public SourceResponse call() throws Exception {
             QueryRequest queryRequest;
-            if (CACHE_QUERY_MODE.equals(request.getPropertyValue(QUERY_MODE)) || INDEX_QUERY_MODE.equals(request.getPropertyValue(QUERY_MODE))) {
-                queryRequest = new QueryRequestImpl(request.getQuery(), false, request.getSourceIds(), request.getProperties());
+            if (CACHE_QUERY_MODE.equals(request.getPropertyValue(QUERY_MODE))
+                    || INDEX_QUERY_MODE.equals(request.getPropertyValue(QUERY_MODE))) {
+                queryRequest = new QueryRequestImpl(request.getQuery(),
+                        false,
+                        request.getSourceIds(),
+                        request.getProperties());
             } else {
                 queryRequest = new QueryRequestImpl(request.getQuery(), request.getProperties());
             }
@@ -586,71 +600,16 @@ public class CachingFederationStrategy implements FederationStrategy, PostIngest
                     .map(ResultImpl::new)
                     .collect(Collectors.toList());
 
-            return new QueryResponseImpl(sourceResponse.getRequest(), clonedResults, true, sourceResponse.getHits(), sourceResponse.getProperties());
+            return new QueryResponseImpl(sourceResponse.getRequest(),
+                    clonedResults,
+                    true,
+                    sourceResponse.getHits(),
+                    sourceResponse.getProperties());
         }
     }
 
     /**
      * Phaser that forces all added metacards to commit to the cache on phase advance
      */
-    private class CacheCommitPhaser extends Phaser {
-
-        private final ScheduledExecutorService phaseScheduler =
-                Executors.newSingleThreadScheduledExecutor();
-
-        public CacheCommitPhaser() {
-            // There will always be at least one party which will be the PhaseAdvancer
-            super(1);
-
-            // PhaseAdvancer blocks waiting for next phase advance, delay 1 second between advances
-            // this is used to block queries that request to be indexed before continuing
-            // committing Solr more often than 1 second can cause performance issues and exceptions
-            phaseScheduler.scheduleWithFixedDelay(new PhaseAdvancer(this), 1, 1, TimeUnit.SECONDS);
-        }
-
-        @Override
-        protected boolean onAdvance(int phase, int registeredParties) {
-            // registeredParties should be 1 since all parties other than the PhaseAdvancer
-            // will arriveAndDeregister in the add method
-            cache.forceCommit();
-
-            return super.onAdvance(phase, registeredParties);
-        }
-
-        /**
-         * Adds results to cache and blocks for next phase advance
-         *
-         * @param results metacards to add to cache
-         */
-        public void add(List<Result> results) {
-            // block next phase
-            this.register();
-            // add results to cache
-            cache.create(getMetacards(results));
-            // unblock phase and wait for all other parties to unblock phase
-            this.awaitAdvance(this.arriveAndDeregister());
-        }
-
-        public void shutdown() {
-            this.forceTermination();
-            phaseScheduler.shutdown();
-        }
-    }
-
-    public void setShowErrors(boolean showErrors) {
-        this.showErrors = showErrors;
-    }
-
-    public boolean getShowErrors() {
-        return showErrors;
-    }
-
-    public void setShowWarnings(boolean showWarnings) {
-        this.showWarnings = showWarnings;
-    }
-
-    public boolean getShowWarnings() {
-        return showWarnings;
-    }
 
 }
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SortedQueryMonitorFactory.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SortedQueryMonitorFactory.java
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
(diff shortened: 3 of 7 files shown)


catalog/core/catalog-core-standardframework/pom.xml
```diff
@@ -311,6 +311,47 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ddf.test.itests</groupId>
+            <artifactId>test-itests-common</artifactId>
+            <version>2.11.0-SNAPSHOT</version>
```
bdeining: ${project.version} for all of these

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
```diff
@@ -183,6 +184,8 @@ public CachingFederationStrategy(ExecutorService queryExecutorService,
 
     @Override
     public QueryResponse federate(List<Source> sources, QueryRequest queryRequest) {
+        Validate.noNullElements(sources);
```
bdeining: Would be worthwhile to add the optional description string to these `Validates`

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
+
+                return new SortedQueryMonitor(this,
+                        completionService,
+                        futures,
+                        returnResults,
+                        request,
+                        postQuery) {
+                    @Override
+                    public void run() {
+                        returnResults.setHits(42);
```
bdeining: Why 42?

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
+
+        mockSource.setQueryResponse(mockResponse);
+
+        List<Source> sources = new ArrayList<>();
+        sources.add(mockSource);
+
+        QueryResponse response = strategy.federate(sources, fedQueryRequest);
+
+        WaitCondition.expect("Wait until the result appears in the response")
+                .until(() -> response.getHits(), is(42L))
```
bdeining: I recommend making the result hits a constant so if it ever changes, we only have to change it in one place

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
+                .checkEvery(100, TimeUnit.MILLISECONDS)
+                .within(1, TimeUnit.SECONDS);
+
+        assertThat(response.getHits(), is(42L));
```
bdeining: same comment about result hits constant

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
+
+        mockSource.setQueryResponse(mockResponse);
+
+        List<Source> sources = new ArrayList<>();
+        sources.add(mockSource);
+
+        QueryResponse response = strategy.federate(sources, fedQueryRequest);
+
+        WaitCondition.expect("Wait until the result appears in the response")
+                .until(() -> response.getHits(), is(42L))
```
bdeining: same comment about result hits constant

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
```
bdeining: same comment about result hits constant

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
```
bdeining: same comment about result hits constant

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
+                5,
+                mock(SortBy.class),
+                true,
+                1000);
+
+        QueryRequest fedQueryRequest = new QueryRequestImpl(mockQ, properties);
+
+        MetacardImpl mockMetacard = new MetacardImpl();
+
+        Result mockResult = new ResultImpl(mockMetacard);
```
bdeining: If you haven't already, I would recommend investigate how much of this can go in `@Before` to clean up the readability of the unit tests

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,390 @@
...
+    public void testDeleteResponse() throws Exception {
+        Map<String, Serializable> testMap = new HashMap<>();
+        testMap.put(Constants.SERVICE_TITLE, "mock response");
+        DeleteResponse response = new DeleteResponseImpl(mock(DeleteRequestImpl.class),
+                testMap,
+                new ArrayList<>());
+
+        response = strategy.process(response);
+
+        assertThat(response.getPropertyValue(Constants.SERVICE_TITLE), is("mock response"));
```
bdeining: I see `"mock response"`is used in various places; could be made a constant

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/impl/MockSource.java
```diff
@@ -109,7 +111,11 @@ public String getVersion() {
 
     @Override
     public SourceResponse query(QueryRequest query) {
-        return null;
+        return sourceResponse;
+    }
+
+    public void setQueryResponse(SourceResponse sourceResponse) {
```
bdeining: Was this added only for unit testing?

codymacdonald: Yes - originally `query()` in the `MockSource` class simply returned `null`. I needed it to be able to return a meaningful value for my unit tests. It still returns `null` as a default value, I only added the option to change it.

bdeining: It's generally frowned upon to add methods for the purposes of testing, but if there are no alternatives, we should at least make it protected

codymacdonald: I will take another look and try to find a workaround again. I remember this being the only option for `MockSource` - Perhaps it would be better to use a different class, but it seems to me that `MockSource` should be used here for mocking a source :).

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: It is just an arbitrary value that is used later for assertions. @ahoffer and I decided to override the `run()` method in `SortedQueryMonitor` because, while a `SortedQueryMonitor` instance needs to be created and ran in `CachingFederationStrategy` for full coverage, the `CachingFederationStrategyTest` unit tests are not concerned with the behavior in `SortedQueryMonitor`. Honestly, this line could be deleted, but then there would be nothing to `assert` on. It is also helpful in simply showing that a `SortedQueryMonitor` is created and `run()`.

catalog/core/catalog-core-standardframework/pom.xml
```diff
@@ -311,6 +311,47 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ddf.test.itests</groupId>
+            <artifactId>test-itests-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
ahoffer: Looks like this dependency is repeated multiple times `catalog-core-solr`

codymacdonald: Updated

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/impl/MockSource.java
AzGoalie: @bdeining This is a test class, i believe this is ok.

bdeining: Whoops, didn't realize that

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+        QueryRequest fedQueryRequest = new QueryRequestImpl(mockQuery, properties);
+
+        MockSource mockSource = getMockSource(mockResponse);
+
+        List<Source> sources = Arrays.asList(mockSource);
+
+        QueryResponse response = strategy.federate(sources, fedQueryRequest);
+
+        WaitCondition.expect("Wait until the result appears in the response")
```
AzGoalie: Should we check that the cache wasn't updated?

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,328 @@
...
```
AzGoalie: Should we check that the result came from a different source and that the cache was updated?

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,328 @@
...
```
AzGoalie: Same as above, should we check that the result came from a different source and that the cache was updated?

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,328 @@
...
+        // Multiple sources needed for OffsetResultHandler to be used
+        for (int i = 0; i < 2; i++) {
+            sources.add(getMockSource(mockResponse));
+        }
+
+        QueryResponse response = strategy.federate(sources, fedQueryRequest);
+
+        // Response is changed by OffsetResultHandler, so getHits() returns 0
+        WaitCondition.expect("Wait until the result appears in the response")
+                .until(() -> response.getHits(), is(0L))
```
AzGoalie: Wouldn't the returning result still be 1? Might be better to have the two sources return different amount of results and assert on that. Source 1 has 1 result while source 2 has 2 results.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+        QueryRequest fedQueryRequest = new QueryRequestImpl(mockQuery, properties);
+
+        MockSource mockSource = getMockSource(mockResponse);
+
+        List<Source> sources = Arrays.asList(mockSource);
+
+        QueryResponse response = strategy.federate(sources, fedQueryRequest);
+
+        WaitCondition.expect("Wait until the result appears in the response")
```
ahoffer: The value is actual set on line #127in the test class.

codymacdonald: So should I make changes to this test case or not?

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,383 @@
...
+        testMap.put(Constants.SERVICE_TITLE, MOCK_RESPONSE_TITLE);
+        DeleteResponse response = new DeleteResponseImpl(mock(DeleteRequestImpl.class),
+                testMap,
+                new ArrayList<>());
+
+        response = strategy.process(response);
+
+        assertThat(response.getPropertyValue(Constants.SERVICE_TITLE), is(MOCK_RESPONSE_TITLE));
+    }
+
```
lessarderic: ❗️We need a test for the `process(CreateResponse)` method.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,383 @@
...
+        testMap.put(Constants.SERVICE_TITLE, MOCK_RESPONSE_TITLE);
+        UpdateResponse response = new UpdateResponseImpl(mock(UpdateRequestImpl.class),
+                testMap,
+                new ArrayList<>());
+
+        response = strategy.process(response);
+
+        assertThat(response.getPropertyValue(Constants.SERVICE_TITLE), is(MOCK_RESPONSE_TITLE));
+    }
+
```
lessarderic: ❗️ For `process(UpdateResponse)`, we're missing the following paths:
* `cacheRemoteIngests` is `false` and the update request is not local (L386)
* The service title in the request is the same as `cacheSource.getId()` (L393)
* The update request contains metacards (L400)

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,383 @@
...
+    public void testUpdateResponse() throws Exception {
+        Map<String, Serializable> testMap = new HashMap<>();
+        testMap.put(Constants.SERVICE_TITLE, MOCK_RESPONSE_TITLE);
+        UpdateResponse response = new UpdateResponseImpl(mock(UpdateRequestImpl.class),
+                testMap,
+                new ArrayList<>());
+
+        response = strategy.process(response);
+
+        assertThat(response.getPropertyValue(Constants.SERVICE_TITLE), is(MOCK_RESPONSE_TITLE));
```
lessarderic: ❗️ We need to verify that `SolrCache.create()` was called with the expected metacards (L404).

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,383 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testValidCollectionWithNullSource() throws Exception {
+        List<Source> sources = new ArrayList<>();
+        sources.add(null);
+        QueryRequest fedQueryRequest = mock(QueryRequest.class);
+        when(fedQueryRequest.getQuery()).thenReturn(mockQuery);
+        strategy.federate(sources, fedQueryRequest);
+    }
+
```
lessarderic: ❗️ We need to test that `setMaxStartIndex()` works as expected when the index is <= 0.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,383 @@
...
+        QueryResponse federateResponse = strategy.federate(sources, fedQueryRequest);
+
+        WaitCondition.expect("Wait until the result appears in the response")
+                .until(() -> federateResponse.getHits(), is(HITS_RESULT))
+                .checkEvery(POLLING_INTERVAL, TimeUnit.MILLISECONDS)
+                .within(WITHIN_TIMEOUT, TimeUnit.SECONDS);
+    }
+
+    @Test
+    public void testFederateQueryCacheWithOffsetHandler() throws Exception {
```
lessarderic: ❗️ Somehow we're not getting coverage on the while loop in `OffsetResultHandler.run()`. We may need to change this test and/or add new ones to make sure we cover all possible paths.

Since the `OffsetResultHandler.run()` method contains many conditions, it may be easier to only make sure the `OffsetResultHandler.run()` is called in this test and have separate tests for the different paths and conditions. We could do this by making the `OffsetResultHandler` class package-private (and optionally moving it to its own file) and testing the `run` method directly.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
lessarderic: ❗️ We may want to rethink how `CallableSourceResponse` is written so we can cover all the possible paths and conditions, including L590-594 which aren't currently covered.

lessarderic: ❗️ Same for `CacheCommitPhaser`.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,383 @@
...
+import ddf.catalog.operation.impl.DeleteResponseImpl;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.operation.impl.QueryResponseImpl;
+import ddf.catalog.operation.impl.UpdateRequestImpl;
+import ddf.catalog.operation.impl.UpdateResponseImpl;
+import ddf.catalog.plugin.PreFederatedQueryPlugin;
+import ddf.catalog.source.Source;
+
+public class CachingFederationStrategyTest {
```
lessarderic: ❗️ General comment on all the tests: we need to make sure we have `verify` calls with the proper argument matchers for all the dependencies the `CachingFederationStrategy` is using, otherwise the unit tests may not fail when a breaking change is introduced.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SortedQueryMonitorFactory.java
```diff
@@ -0,0 +1,32 @@
...
+import ddf.catalog.plugin.PostFederatedQueryPlugin;
+
+public class SortedQueryMonitorFactory {
+    private final CachingFederationStrategy cachingFederationStrategy;
+
+    public SortedQueryMonitorFactory(CachingFederationStrategy cachingFederationStrategy) {
+        this.cachingFederationStrategy = cachingFederationStrategy;
+    }
+
+    Runnable createMonitor(final CompletionService<SourceResponse> completionService,
```
brendan-hofmann: Would be hard to write the assertions unless we add package-private getters.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: An `OffsetResultHandler` is only created if the offset in `federate()` is > 1, but the if-statement in the while loop in `OffsetResultHandler.run()` only passes if the `queryResultIndex` variable in `OffsetResultHandler.run()` is >= offset, but `queryResultIndex` is always set to 1. Maybe I'm missing something, but I don't see how that if-statement could ever pass.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: Regarding checking for `null` constructor arguments: should we add a test for each argument or one test for all arguments?

lessarderic:
@codymacdonald, let me know when the PR has been updated and is ready for another look.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: ✅

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: ✅

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: ✅

codymacdonald:
build now

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
```diff
@@ -118,7 +119,7 @@
      */
     protected static final String UPDATE_QUERY_MODE = "update";
 
-    private static final int DEFAULT_MAX_START_INDEX = 50000;
+    static final int DEFAULT_MAX_START_INDEX = 50000;
```
lessarderic: ✏️  We could add a comment here saying that this is package private for unit testing purposes.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,748 @@
...
+
+        verify(sources.get(0), atLeastOnce()).query(any(QueryRequest.class));
+        verify(sources.get(1), times(0)).query(any(QueryRequest.class));
+    }
+
+    @Test
+    public void testStartIndexGreaterThanMaxStartIndex() throws Exception {
+        // startIndex and pageSize must be > 1
+        Query mockQ = new QueryImpl(mock(NullFilterImpl.class),
+                strategy.DEFAULT_MAX_START_INDEX + 5,
```
lessarderic: ❗️ We should use the class name instead of `strategy` to access the constant.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,748 @@
...
+
+        QueryRequest fedQueryRequest = new QueryRequestImpl(mockQuery, properties);
+
+        strategy.federate(Arrays.asList(mock(Source.class)), fedQueryRequest);
+        verify(mockPlug).process(any(Source.class), any(QueryRequest.class));
+    }
+
+    @Test
+    public void testCatchStopProcessingException() throws Exception {
+        PreFederatedQueryPlugin mockPlug = mock(PreFederatedQueryPlugin.class), mockPlug2 = mock(
```
lessarderic: ❗️ We should keep the two variables on separate lines.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,748 @@
...
+                new ArrayList<>(),
+                cache,
+                cacheExecutor,
+                mock(ValidationQueryFactory.class),
+                new CacheQueryFactory(new GeotoolsFilterBuilder()));
+
+        QueryRequest fedQueryRequest = new QueryRequestImpl(mockQuery, properties);
+
+        strategy.federate(Arrays.asList(mock(Source.class)), fedQueryRequest);
+        verify(mockPlug).process(any(Source.class), any(QueryRequest.class));
```
lessarderic: ❗️ We should do like we did for the other test and have two mocks. Make sure the second one is actually called in this case.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,748 @@
...
+    @Test
+    public void testCatalogProviderSource() throws Exception {
+        CatalogProvider catalogProviderSource = mock(CatalogProvider.class);
+        when(catalogProviderSource.getId()).thenReturn(SystemInfo.getSiteName());
+
+        QueryRequest fedQueryRequest = new QueryRequestImpl(mockQuery, properties);
+
+        strategy.federate(Arrays.asList(catalogProviderSource), fedQueryRequest);
+
+        verify(catalogProviderSource, atLeastOnce()).getId();
```
lessarderic: ❗️ We should verify that `validationQueryFactory. getQueryRequestWithValidationFilter()` has been called with the proper arguments.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,748 @@
...
+    }
+
+    @Test
+    public void testStartIndexLessThanZero() throws Exception {
```
lessarderic: ✏️  We should consider splitting those up.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
codymacdonald: ✅

codymacdonald:
build now

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/cache/solr/impl/CachingFederationStrategyTest.java
```diff
@@ -0,0 +1,715 @@
...
+        verify(mockSource, times(0)).query(any(QueryRequest.class));
+
+        assertThat(federateResponse.getRequest()
+                        .getQuery(),
+                is(requestArgumentCaptor.getValue()
+                        .getQuery()));
+    }
+
+    @Test
+    public void testFederateQueryCacheNoUpdate() throws Exception {
```
pklinef: I would have expected this test and the next two to verify if the cache is or is not updated as expected based on the query mode.

codymacdonald:
build now

PR #1907, opened 2017-04-18
DDF-2956 Remove grunt from Intrigue

andrewkfiedler:
#### What does this PR do?
 - Switch to using webpack for the remaining grunt dependencies.
 - Moved fonts into their own css file to fix issues with source maps not working properly when they aren't.
 - Updated bundling to include fonts so they aren't an extra request.
 - Renamed less folder to styles.
 - Updated ApplicationSetup to interact with the react hot loader so that dev's have a better experience.  Now only css will be hot loaded, and changes to other files won't cause the dev server to refresh the page.

@andrewkfiedler
@clockard 
@rzwiefel 
@djblue 

Test out the dev server for development.  Make changes to a less file and verify it updates the page.  Make changes to a js file and verify the page doesn't update or refresh.

Verify the production build still works as intended.

Our recent switch to using webpack instead of grunt left our less hot reloading behind.  This brings back the hot reloading, along with less source maps.

https://codice.atlassian.net/browse/DDF-2956


rzwiefel:
build now

catalog/ui/catalog-ui-search/package.json
```diff
@@ -61,14 +64,15 @@
...
     "react-addons-test-utils": "15.2.1",
-    "react-hot-loader": "1.3.0",
+    "react-hot-loader": "1.3.1",
     "script-loader": "0.7.0",
     "sinon": "1.17.3",
     "squirejs": "0.2.1",
     "stack-source-map": "1.0.5",
+    "url-loader": "0.5.8",
     "webpack": "1.13.0",
     "webpack-dev-middleware": "1.6.1",
```
djblue: We can get rid of this dependency.

catalog/ui/catalog-ui-search/package.json
```diff
@@ -61,14 +64,15 @@
...
-    "webpack-dev-server": "1.14.1",
+    "webpack-dev-server": "1.16.3",
     "webpack-hot-middleware": "2.12.1",
```
djblue: We can get rid of this one too.

catalog/ui/catalog-ui-search/test.js
```diff
@@ -30,7 +30,6 @@ app.use('/css',express.static(__dirname + '/target/webapp/css'));
 app.use('/lib',express.static(__dirname + '/target/webapp/lib'));
 app.use(express.static(__dirname + '/src/main/webapp'));
 
-//if we're mocking, it is being run by grunt
```
djblue: I think we can delete the whole file. I don't think we use it anywhere.

catalog/ui/catalog-ui-search/webpack/config/dev.js
```diff
@@ -1,24 +1,45 @@
...
-                loaders: ['react-hot']
-            }
-        ]
+        loaders: [{
+            test: /\.jsx$/,
+            loaders: ['react-hot']
+        }]
     },
     plugins: [
+        new webpack.NoErrorsPlugin(),
```
djblue: You should also add `new webpack.NamedModulesPlugin()` so the console writes the module names instead of their ids.

catalog/ui/catalog-ui-search/package.json
```diff
@@ -5,11 +5,13 @@
   "version": "0.0.1",
   "license": "LGPL-3.0",
   "scripts": {
-    "startplus": "grunt startplus --webpackConfig=./webpack/config/dev.js",
```
djblue: We should remove all grunt related dependencies.

djblue: Looking through the list, we no longer require these dependencies:

- compression
- connect-livereload
- cors
- grunt
- grunt-cli
- grunt-contrib-clean
- grunt-contrib-cssmin
- grunt-contrib-jshint
- grunt-contrib-less
- grunt-contrib-watch
- grunt-express-server
- grunt-open
- grunt-webpack
- http-proxy
- load-grunt-tasks
- morgan

catalog/ui/catalog-ui-search/pom.xml
```diff
@@ -274,7 +258,16 @@
                             <goal>grunt</goal>
```
clockard: If we are removing grunt why do we still have this grunt build

djblue: I think it's so we can skip the default grunt build, otherwise the build would fail because there is no longer a `Gruntfile.js`.

catalog/ui/catalog-ui-search/package.json
andrewkfiedler: Having some build issues when I remove some of these, but not sure why.  The maven bundle plugin is failing.

PR #1910, opened 2017-04-19
DDF-2948 update spark endpoints to handle remote associations

bdeining:
#### What does this PR do?
This PR adds the ability for the front end to query writable sources (implementing CatalogStore API) to allow for remote associations.  The associations endpoint was updated to include CatalogStores in its calls to the CatalogFramework so that associations can be updated for remote sources.  It additionally adds itests to test this functionality.  Note: https://codice.atlassian.net/browse/DDF-2957 currently prevents remote associations from working, which is currently ignored in itests until that ticket is completed.
@glenhein @rzwiefel 
@troymohl 
@jlcsmith
@lessarderic 
Build / Install / Verify local associations still work as intended with a remote source configured.  Since remote associations are ignored for now, this functionality can be tested if the ContentUriAccessPlugin is turned off.
This PR relies on https://github.com/codice/ddf/pull/1929
[DDF-2957](https://codice.atlassian.net/browse/DDF-2957)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -611,15 +624,21 @@ protected UpdateResponse patchMetacards(List<MetacardChanges> metacardChanges)
...
-                    Metacard result = results.get(id)
-                            .getMetacard();
+
+                    Result result = results.get(id);
+                    if (result == null) {
+                        LOGGER.debug("Metacard with id {} not found in result set", id);
+                        continue;
+                    }
+
+                    Metacard metacard = result.getMetacard();
```
rzwiefel: So this is changing the behavior quite a bit and there's a couple options depending on what route we want to go. It's one of those cases where we have to decide if we want to allow partial successes or not.

1) If we leave it like this (so it silently skips), we need to update the method that calls this to remove the things that we didn't actually update from the response we give back. 
2) Make it so we throw an exception here and it fails the transaction completely. (and make sure we return the proper error message to the user).

Personally I would suggest throwing an exception instead of continuing because its less ambiguous. Partial updates can be very confusing to a user (especially since we don't have anything in place to let the user know which things succeeded and failed).

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -433,14 +442,18 @@ public void init() {
...
+            LOGGER.debug("Exception occurred.", ex);
             res.status(404);
             res.header(CONTENT_TYPE, APPLICATION_JSON);
             res.body(util.getJson(ImmutableMap.of("message",
                     "Could not find what you were looking for")));
         });
     }
 
+    private String getJsonWritableSources() {
+        return util.getJson(getWritableSources());
```
rzwiefel: just a nit which you can fix or ignore: i would just inline this since its only 1 line and not even a very long line.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -692,4 +711,10 @@ public InputStream openStream() throws IOException {
             return supplier.get();
         }
     }
+
+    private List<String> getWritableSources() {
```
rzwiefel: Also just a nit, can we keep the static class at the bottom and move this up next to the other private methods?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
```diff
@@ -160,6 +162,11 @@ private void addEdge(Edge edge, Map<String, Metacard> metacards,
         String id = edge.parent.get(Metacard.ID)
                 .toString();
         Metacard target = changedMetacards.getOrDefault(id, metacards.get(id));
+
+        if (target == null) {
+            return;
+        }
```
rzwiefel: Why this check only on  `addEdge`? should we add it to `removeEdge` too?  
Also if that target metacard is not there that means we can't complete part of the requested update. Like before not sure if we want to fail or allow the partial update.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -61,13 +62,17 @@
...
+import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.QueryResponse;
 import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
 
 public class EndpointUtil {
+
+    public static final Integer QUERY_DEFAULT_PAGE_SIZE = 2000000000;
```
rzwiefel: one nit: could use number separators `2_000_000_000`
Second, Is there any way we can use the constant that is already defined for the solr cloud stuff? I see that they have an undocumented system property which I'm sure would be allowable to use. The property name is private so we would have to use the string literal.
`Integer.valueOf(System.getProperty("solr.query.rows",  Integer.toString(2_000_000_000)))`

rzwiefel: Eventually we should replace large page sizes with paging thru the catalog, but that is outside of the scope of your PR.

distribution/test/itests/test-itests-common/src/main/resources/default-csw-mock-query-empty-response.xml
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<csw:GetRecordsResponse version="2.0.2" xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dct="http://purl.org/dc/terms/" xmlns:ows="http://www.opengis.net/ows" xmlns:xml="http://www.w3.org/XML/1998/namespace" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <csw:SearchStatus timestamp="2016-06-07T10:08:46.223-07:00"/>
+    <csw:SearchResults elementSet="full" nextRecord="0" numberOfRecordsMatched="0" numberOfRecordsReturned="0" recordSchema="http://www.opengis.net/cat/csw/2.0.2"></csw:SearchResults>
```
rzwiefel: just a nit: could help to format this file to use a little more vertical space instead of horizontal

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java
```diff
@@ -40,244 +57,432 @@
...
-        Map<String, String> workspace = ImmutableMap.of("title", "my workspace");
-        expect(asGuest().body(stringify(workspace)), 404).post(api());
+        Map<String, String> workspace = ImmutableMap.of(Core.TITLE, "my workspace");
+        expect(asGuest().body(stringify(workspace)), 404).post(workspaceApi());
     }
 
     @Test
     public void testGuestCanCreateWorkspacesForOthers() {
-        Map<String, String> workspace = ImmutableMap.of("title",
+        Map<String, String> workspace = ImmutableMap.of(Core.TITLE,
```
rzwiefel: I don't think @djblue meant the actual attribute title. The attributes coming in on the workspace json do not always reflect what they get transformed to for backend attributes.
So its technically not wrong, but it also just "happens" to match. May never change, but it is slightly misleading.
See WorkspaceTransformer L71 - L99 for the transforms that happen between the json and the metacard attribute names.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java
```diff
@@ -40,244 +57,432 @@
...
+        AssociationsGetResponse associationGetResponse =
+                (AssociationsGetResponse) associationGetResponses.get(0);
+        parentId = ((CharSequenceValue) associationGetResponse.getParent()
+                .get(Core.ID)).stringValue();
+        assertThat(parentId, is(metacardId1));
+        childId = ((CharSequenceValue) associationGetResponse.getChild()
+                .get(Core.ID)).stringValue();
+        assertThat(childId, is(metacardId2));
+
+        clearCatalog();
```
rzwiefel: just a nit: we could move `clearCatalog` to an `@After` method if we call it after every test.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
```diff
@@ -134,9 +135,10 @@ public void putAssociations(String id, Collection<Edge> edges)
             return;
         }
 
-        catalogFramework.update(new UpdateRequestImpl(changedMetacards.keySet()
-                .toArray(new String[0]), new ArrayList<>(changedMetacards.values())));
-
+        catalogFramework.update(new UpdateRequestImpl(new HashSet<>(writableSourceIds),
+                changedMetacards.keySet()
+                        .toArray(new String[0]),
```
glenhein: I believe it's more efficient to create the string array with expected size instead of zero.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: If you use the undocumented solr property, I would include a default just in case they change property names. Wouldn't want the code to break!

rzwiefel: The second part of the getProperty method is the default.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
bdeining: Seems like there is some disagreement between specifying zero, or specifying the expected size for `toArray()` : https://shipilev.net/blog/2016/arrays-wisdom-ancients/

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -510,6 +510,11 @@ private UpdateResponse processPostIngestPlugins(UpdateResponse updateResponse) {
 
     private UpdateResponse performRemoteUpdate(UpdateRequest updateRequest,
             UpdateResponse updateResponse) {
+        // All metacards were updated locally
+        if (updateResponse != null && updateRequest.getUpdates().size() == updateResponse.getUpdatedMetacards().size()) {
```
bdeining: `updateRequest.getUpdates()` is validated before this method : https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java but it looks like `getUpdatedMetacards` can be null at this point.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
bdeining: Moved

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
bdeining: Refactored to use the system property or fallback to 2,000,000,000

distribution/test/itests/test-itests-common/src/main/resources/default-csw-mock-query-empty-response.xml
bdeining: done; turns out there were a bunch of unneeded schemas

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java
bdeining: reverted this

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java
bdeining: done

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
bdeining: updated to check for null on `getUpdatedMetacards`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
bdeining: I agree; It seems more reasonable to fail here with an exception than continue.  The main motivation for this was to prevent an observed potential NPE in the case that `results.get(id)` is null

bdeining: Changed to throw an exception

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
bdeining: Changed to be an exception

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
```diff
@@ -83,8 +82,8 @@ public Associated(EndpointUtil util, CatalogFramework catalogFramework) {
                         .equals(metacardId))
                 .collect(Collectors.toList());
 
-        Map<String, Metacard> childMetacardMap = query(withNonrestrictedTags(forChildAssociations(
-                root)));
+        Map<String, Metacard> childMetacardMap = query(writableSourceIds, withNonrestrictedTags(
+                forChildAssociations(root)));
```
jlcsmith: root is potentially null here

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
```diff
@@ -109,7 +108,9 @@ public void putAssociations(String id, Collection<Edge> edges)
                 .distinct()
                 .collect(Collectors.toList());
 
-        Map<String, Metacard> metacards = util.getMetacards(ids, getNonrestrictedTagsFilter())
+        Map<String, Metacard> metacards = util.getMetacards(writableSourceIds,
```
rzwiefel: This one is my fault. I've used the `util` convention for the EndpointUtil class consistently throughout the intrigue backend, so we should probably keep it util unless/until we change all of them.

jlcsmith:
@bdeining  needs rebase

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/UpdateRequestImpl.java
```diff
@@ -92,6 +92,23 @@ public UpdateRequestImpl(String[] ids, List<Metacard> metacards)
     }
 
     /**
+     * Instantiates a new UpdateRequestImpl from a set of destinations,
+     * an array of ids, and a list of {@link Metacard}
+     *
+     * @param destinations the destination ids this request should be sent to
+     * @param ids          the ids associated with the {@link Metacard} list
```
bdeining: These are the metacard IDs, but this constructor was built this way so we can utilize some of the other functionality in `UpdateRequestImpl` such as `formatEntryList`.  Since this class stores the Metacards as a List<Entry>, this seemed to be the best approach.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
```diff
@@ -264,20 +276,22 @@ private Filter forChildAssociations(Metacard metacard) {
...
             throws UnsupportedQueryException, SourceUnavailableException, FederationException {
         if (filter == null) {
             return Collections.emptyMap();
         }
 
-        QueryResponse query = catalogFramework.query(new QueryRequestImpl(new QueryImpl(filter,
+        QueryRequestImpl queryRequest = new QueryRequestImpl(new QueryImpl(filter,
                 1,
-                0,
+                EndpointUtil.DEFAULT_QUERY_PAGE_SIZE,
```
bdeining: This query is called to get the associated Metacards by a Metacard ID.  I'm not sure the amount of associated Metacards would ever get to this point.  To avoid blowing up SolrCloud, it may be more reasonable to reduce this  page size to a few thousand records.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -78,10 +83,19 @@
...
-    private static final int DEFAULT_PAGE_SIZE = 10;
-
     private static final String ISO_8601_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
 
+    static {
+        Integer solrQueryRowSize;
+        try {
+            solrQueryRowSize = Integer.valueOf(System.getProperty("solr.query.rows", Integer.toString(2_000_000_000)));
+        } catch (NumberFormatException e) {
+            solrQueryRowSize = 2_000_000_000;
```
bdeining: Right now this is only being used in one place.  re : `This query is called to get the associated Metacards by a Metacard ID. I'm not sure the amount of associated Metacards would ever get to this point. To avoid blowing up SolrCloud, it may be more reasonable to reduce this page size to a few thousand records.`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -161,23 +175,24 @@ public Metacard getMetacard(String id)
...
+                attributeName,
                 ids,
                 filterBuilder.attribute(Metacard.TAGS)
                         .is()
                         .like()
                         .text(tag));
     }
 
-    public Map<String, Result> getMetacards(String attributeName, Collection<String> ids,
+    public Map<String, Result> getMetacards(List<String> writableSources, String attributeName, Collection<String> ids,
```
bdeining: added @Nullable

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -161,23 +175,24 @@ public Metacard getMetacard(String id)
...
             throws UnsupportedQueryException, SourceUnavailableException, FederationException {
-        return getMetacards(Metacard.ID, ids, tagFilter);
+        return getMetacards(writableSources, Metacard.ID, ids, tagFilter);
     }
 
     public Map<String, Result> getMetacards(String attributeName, Collection<String> ids,
             String tag)
             throws UnsupportedQueryException, SourceUnavailableException, FederationException {
-        return getMetacards(attributeName,
+        return getMetacards(null,
```
bdeining: In my opinion, I don't see any benefit in creating an empty reference when passing null suffices

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -510,6 +510,11 @@ private UpdateResponse processPostIngestPlugins(UpdateResponse updateResponse) {
 
     private UpdateResponse performRemoteUpdate(UpdateRequest updateRequest,
             UpdateResponse updateResponse) {
+        // All metacards were updated locally
```
bdeining: extracted

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -133,13 +136,16 @@
...
     public MetacardApplication(CatalogFramework catalogFramework, FilterBuilder filterBuilder,
             EndpointUtil endpointUtil, Validator validator, WorkspaceTransformer transformer,
             ExperimentalEnumerationExtractor enumExtractor,
             SubscriptionsPersistentStore subscriptions, List<MetacardType> types,
-            Associated associated) {
+            Associated associated,
+            List<CatalogStore> catalogStores) {
         this.catalogFramework = catalogFramework;
         this.filterBuilder = filterBuilder;
         this.util = endpointUtil;
```
bdeining: renamed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -680,6 +698,12 @@ private Metacard saveMetacard(Metacard metacard)
 
     }
 
+    private List<String> getWritableSources() {
+        List<String> stringList = catalogStores.stream().map(CatalogStore::getId).collect(Collectors.toList());
```
bdeining: renamed to `sourceIdList`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
bdeining: renamed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -78,10 +83,19 @@
...
     private final AttributeRegistry attributeRegistry;
 
-    private static final int DEFAULT_PAGE_SIZE = 10;
-
     private static final String ISO_8601_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
 
+    static {
+        Integer solrQueryRowSize;
+        try {
+            solrQueryRowSize = Integer.valueOf(System.getProperty("solr.query.rows", Integer.toString(2_000_000_000)));
```
bdeining: this has been refactored

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
bdeining: refactored

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -195,13 +210,27 @@ public Metacard getMetacard(String id)
...
-                SortBy.NATURAL_ORDER,
-                false,
-                TimeUnit.SECONDS.toMillis(10)), false));
+        QueryRequest queryRequest;
+
+        if (CollectionUtils.isEmpty(writableSources)) {
+            queryRequest = new QueryRequestImpl(new QueryImpl(
+                    queryFilter,
+                    1,
+                    -1,
```
bdeining: refactored to use configurationApplication instead

bdeining:
@jlcsmith @lessarderic all comments addressed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
bdeining: added a check for root being null

bdeining:
build now

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -509,8 +509,17 @@ private UpdateResponse processPostIngestPlugins(UpdateResponse updateResponse) {
         return updateResponse;
     }
 
+    private boolean metacardsUpdatedLocally(UpdateResponse updateResponse, UpdateRequest updateRequest) {
+        return updateResponse != null && updateResponse.getUpdatedMetacards() != null && updateRequest.getUpdates().size() == updateResponse.getUpdatedMetacards().size();
```
rzwiefel: formatting doesn't look quite right

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -82,12 +89,14 @@
...
             FilterBuilder filterBuilder, List<InjectableAttribute> injectableAttributes,
-            AttributeRegistry attributeRegistry) {
+            AttributeRegistry attributeRegistry,
+            ConfigurationApplication configurationApplication) {
         this.metacardTypes = metacardTypes;
         this.catalogFramework = catalogFramework;
         this.filterBuilder = filterBuilder;
         this.injectableAttributes = injectableAttributes;
         this.attributeRegistry = attributeRegistry;
+        this.querySize = configurationApplication.getResultCount();
```
rzwiefel: ✏️  / ❗️  Just a heads up, since this is instantiated in the constructor it will use the value stored in the configuration but will **not** update when it is changed. We can just create a ticket to fix it in another PR if not done in this PR.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -193,13 +203,27 @@ public Metacard getMetacard(String id)
...
+                    SortBy.NATURAL_ORDER,
+                    false,
+                    TimeUnit.SECONDS.toMillis(10)), false);
+        } else {
+            queryRequest = new QueryRequestImpl(new QueryImpl(
+                    queryFilter,
+                    1, querySize,
+                    SortBy.NATURAL_ORDER,
+                    false,
+                    TimeUnit.SECONDS.toMillis(10)), writableSources);
```
rzwiefel: ✏️ / ❗️ Timeout for a federated query we probably should move to 30 seconds (or maybe 5 minutes or whatever the default is for DDF, though I'm fine pushing that work into a different ticket as well so we can get this merged.)

rzwiefel:
@bdeining Build failed for checkstyle 
```
/jenkins/workspace/DDF-Linux-PR/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java:24:15: error: Unused import - org.ops4j.pax.exam.CoreOptions.mavenBundle.
/jenkins/workspace/DDF-Linux-PR/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java:25:15: error: Unused import - org.ops4j.pax.exam.CoreOptions.options.
/jenkins/workspace/DDF-Linux-PR/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java:50:8: error: Unused import - org.ops4j.pax.exam.Option.
```

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalogSearchUi.java
```diff
@@ -13,23 +13,41 @@
...
+import static org.codice.ddf.itests.common.AbstractIntegrationTest.DynamicUrl.INSECURE_ROOT;
 import static org.codice.ddf.itests.common.AbstractIntegrationTest.DynamicUrl.SECURE_ROOT;
+import static org.codice.ddf.itests.common.catalog.CatalogTestCommons.ingest;
+import static org.codice.ddf.itests.common.csw.CswTestCommons.CSW_TRANSACTIONAL_SOURCE_FACTORY_PID;
+import static org.codice.ddf.itests.common.csw.CswTestCommons.getCswTransactionalSourceProperties;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.core.Is.is;
-import static org.junit.Assert.assertNotNull;
+import static org.ops4j.pax.exam.CoreOptions.mavenBundle;
```
jlcsmith: ❗️ This class is failing checkstyle due to unused imports

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
bdeining: Changed to 30 seconds

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
bdeining: fixed

bdeining:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/UpdateRequestImpl.java
lessarderic: This feels like an implementation detail that shouldn't be exposed outside this class, but since we already have other constructors doing the same thing...

❗️ Should we change the `formatEntryList()` method to validate that the list of metacard IDs or URLs match with the metacards provided (`ids[i] == metacards.get(i).getId()`) to make sure every ID or URL maps to the correct metacard? Otherwise we may end up an ID or URL that maps to the wrong metacard and that seems pretty bad to me. @pklinef, @coyotesqrl, thoughts?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
```diff
@@ -262,20 +283,22 @@ private Filter forChildAssociations(Metacard metacard) {
...
             throws UnsupportedQueryException, SourceUnavailableException, FederationException {
         if (filter == null) {
             return Collections.emptyMap();
         }
 
-        QueryResponse query = catalogFramework.query(new QueryRequestImpl(new QueryImpl(filter,
+        QueryRequestImpl queryRequest = new QueryRequestImpl(new QueryImpl(filter,
                 1,
-                0,
+                querySize,
```
lessarderic: ❓ Should we consider waiting for https://github.com/codice/ddf/pull/2080 to get merged and use the new result iterator/pagination utility? That way we wouldn't need to ever worry about `querySize` getting too big or capped.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -193,13 +203,27 @@ public Metacard getMetacard(String id)
...
-                SortBy.NATURAL_ORDER,
-                false,
-                TimeUnit.SECONDS.toMillis(10)), false));
+        QueryRequest queryRequest;
+
+        if (CollectionUtils.isEmpty(writableSources)) {
+            queryRequest = new QueryRequestImpl(new QueryImpl(
+                    queryFilter,
+                    1,
+                    querySize,
```
lessarderic: ❓ For this query and the one in the `else` block, should we consider waiting for https://github.com/codice/ddf/pull/2080 to get merged and use the new result iterator/pagination utility? That way we wouldn't need to ever worry about `querySize` getting too big or capped.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -159,23 +168,24 @@ public Metacard getMetacard(String id)
...
+                attributeName,
                 ids,
                 filterBuilder.attribute(Metacard.TAGS)
                         .is()
                         .like()
                         .text(tag));
     }
 
-    public Map<String, Result> getMetacards(String attributeName, Collection<String> ids,
+    public Map<String, Result> getMetacards(@Nullable List<String> writableSources, String attributeName, Collection<String> ids,
```
lessarderic: ✏️  Sorry, I didn't realize this was a `public` method in the first round of review when I asked for the `@Nullable`. As a general rule we try to avoid `null` as a valid value in public methods to prevent unexpected NPEs. In this case, if callers don't have a list of `writableSources`, they could pass `Collections.emptyList()`, which is a singleton list that doesn't need to be instantiated.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
rzwiefel: @lessarderic If this PR hadn't been open for 2 months, I would say yes. It will likely be more effective to create a ticket and address it in a smaller, separate PR.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
bdeining: Removed `@Nullable` and changed to use `Collections.emptyList()`.  I was actually unaware that this was a singleton list that doesn't need to be instantiated. 👍

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -61,13 +64,17 @@
...
+import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.QueryResponse;
 import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
 
 public class EndpointUtil {
+
+    public int querySize;
```
bdeining: fixed

bdeining:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/UpdateRequestImpl.java
pklinef: I think the Catalog Framework should protect against that if we think that might happen.  `validateUpdateRequest` could verify that the attribute values match the update attribute key values.

https://github.com/codice/ddf/blob/08651d5cc59e559b940b4faaec6334657948c18c/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java#L349

lessarderic: Should we add this check as part of this PR or have a separate ticket?

pklinef: That looks pretty out of scope for this ticket.  I would suggest creating a ticket for that instead.

LinkMJB:
build now

LinkMJB:
build now

bdeining:
build now

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
lessarderic: @rzwiefel, @bdeining, a separate ticket works for me as long as it's tagged for the 2.11 release.

rzwiefel:
I had the same test failures as CI did above when running locally

jlcsmith:
build now

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/associations/Associated.java
bdeining: https://codice.atlassian.net/browse/DDF-3164

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
bdeining: https://codice.atlassian.net/browse/DDF-3164

bdeining:
build now

PR #1915, opened 2017-04-22
DDF 2951 Test transformation of federated results

ahoffer:
#### What does this PR do?
Enhance TestFederation itests to verify results can be transformed back into metacards. The basic flow is to use the CSW endpoint to retrieve a metacard using the three supported output schema: Open GIS, native Metacard, and GDM. The record return by the query is then passed to the appropriate input transformer. The test case verifies (implicitly) that no exception occurs and that the newly created metacard has a few of the attributes that it does. 
@clockard 
@coyotesqrl 

Run the integration tests on DDF. 
Part of team effort to increase itest coverage.
[DDF-2951](https://codice.atlassian.net/browse/DDF-2951)
#### Checklist:
- [X ] Update / Add Integration Tests

### DIFF
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
@@ -937,7 +937,7 @@ public class TestFederation extends AbstractIntegrationTest {
         ValidatableResponse response = getAndValidateCswResponse(metacardTitle, OPEN_GIS_SCHEMA_URI,
                 assertions);
 
-        // Extract the port of the response that can be transformed into a metacard
+        // Extract the part of the response that can be transformed into a metacard
         String cswRecordXml = extractRecord(response, extractXmlRegex);
 
         // Get the OPEN GIS CSW Record input transformer and create a metacard
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
@@ -82,6 +82,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
 import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpressionException;
 
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.io.FileUtils;
@@ -153,6 +154,8 @@ public class TestFederation extends AbstractIntegrationTest {
     public static final String OPEN_GIS_SCHEMA_URI = "http://www.opengis.net/cat/csw/2.0.2";
 
     public static final String METACARD_URI = "urn:catalog:metacard";
+
+    public static final String THUMB_NAIL_BASE_64_ENC_SUBSTRING = "/9j/4AAQSkZJRgABAQAAAQABAAD";
 
     protected static final Logger LOGGER = LoggerFactory.getLogger(TestFederation.class);
 
@@ -855,13 +858,12 @@ public class TestFederation extends AbstractIntegrationTest {
     public void testCswQueryForMetacardXml() throws Exception {
 
         // The first few bytes of the metacard's base-64 encoded thumbnail image
-        String thumbNailBase64EncSubstring = "/9j/4AAQSkZJRgABAQAAAQABAAD";
         String metacardTitle = "myTitle";
 
         // Define the assertions that test the query results conform to expectations
         List<Matcher<?>> assertions = getXpathMatchers(METACARD_URI, 1);
         assertions.add(hasXPath("//*[@name='thumbnail']/*/text()",
-                startsWith(thumbNailBase64EncSubstring)));
+                startsWith(THUMB_NAIL_BASE_64_ENC_SUBSTRING)));
         assertions.add(hasXPath("//*[@name='title']/*/text()", equalTo(metacardTitle)));
         assertions.add(hasXPath("//pos/text()", equalTo("30.0 10.0")));
 
@@ -870,11 +872,7 @@ public class TestFederation extends AbstractIntegrationTest {
                 METACARD_URI,
                 assertions);
 
-        XPathHelper xpHelper = new XPathHelper(response.extract()
-                .body()
-                .asString());
-        Node node = (Node) xpHelper.evaluate("//*[local-name()='metacard']", XPathConstants.NODE);
-        String metacardXml = XPathHelper.print(node);
+        String metacardXml = extractXml(response, "//*[local-name()='metacard']");
 
         // Get the metacard XML input transformer and attempt to create a metacard object
         Metacard metacard = getInputTransformer("(id=xml)").transform(IOUtils.toInputStream(
@@ -886,7 +884,7 @@ public class TestFederation extends AbstractIntegrationTest {
         assertThat("Incorrect thumbnail",
                 new String(Base64.getEncoder()
                         .encode(metacard.getThumbnail())),
-                startsWith(thumbNailBase64EncSubstring));
+                startsWith(THUMB_NAIL_BASE_64_ENC_SUBSTRING));
         assertThat("Incorrect geometry", "POINT (30 10)", equalTo(metacard.getLocation()));
     }
 
@@ -894,7 +892,6 @@ public class TestFederation extends AbstractIntegrationTest {
     public void testCswQueryForGmdXml() throws Exception {
 
         // The first few bytes of the metacard's base-64 encoded thumbnail image
-        String thumbNailBase64EncSubstring = "/9j/4AAQSkZJRgABAQAAAQABAAD";
         String metacardTitle = "myTitle";
 
         // Define the assertions that test the query results conform to expectations
@@ -907,14 +904,7 @@ public class TestFederation extends AbstractIntegrationTest {
         ValidatableResponse response = getAndValidateCswResponse(metacardTitle,
                 GMD_SCHEMA_URI,
                 assertions);
-
-        // Extract the part of the response that can be transformed into a metacard
-        XPathHelper xpHelper = new XPathHelper(response.extract()
-                .body()
-                .asString());
-        Node node = (Node) xpHelper.evaluate("//*[local-name()='MD_Metadata']",
-                XPathConstants.NODE);
-        String xml = XPathHelper.print(node);
+        String xml = extractXml(response, "//*[local-name()='MD_Metadata']");
 
         // Get the metacard XML input transformer and attempt to create a metacard object
         Metacard metacard =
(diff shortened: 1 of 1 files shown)


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
-                .post(CSW_PATH.getUrl())
-                .then()
-                .assertThat()
-                .body(hasXPath("/GetRecordsResponse/SearchResults/metacard/@id",
-                        is(metacardIds[GEOJSON_RECORD_INDEX])),
-                        hasXPath("/GetRecordsResponse/SearchResults/@numberOfRecordsReturned",
-                                is("1")),
-                        hasXPath("/GetRecordsResponse/SearchResults/@recordSchema",
-                                is("urn:catalog:metacard")));
+        // The first few bytes of the metacard's base-64 encoded thumbnail image
```
ahoffer: This PR adds tests the 3 output schemas of the CSW endpoint. I basically co-opted the `testCswQueryForMetacardXml` to test the native DDF Metacard output schema. The summary of changes to this method are:
1. Added additional assertions to verify the record returned by the CSW federated query.
2. Extract the section of the xml that represents the metacard using a regex.
3. Get the input transformer service for metacard XML. 
4. Transform the XML into a metcard object and assert a couple of properties to ensure some basic data is present. The intention was not to test EVERY attribute, just a couple.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+        String thumbNailBase64EncSubstring = "/9j/4AAQSkZJRgABAQAAAQABAAD";
```
ahoffer: Capture the first dozen or two bytes of the thumbnail. Just enough to verify it exists as intended.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+        String metacardTitle = "myTitle";
+        String extractXmlRegex = ".*?(<metacard.*?</metacard>).*";
```
ahoffer: Regex was the most straight-forward way to be to grab the XML snippet that gets fed into an input transformer. Today I realized I could use the XPathHelper and its pretty print to do the same thing , but it doesn't really buy me anything I left the test as-is.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+
+        // Define the assertions that test the query results conform to expectations
+        List<Matcher<?>> assertions = getXpathMatchers(METACARD_URI, 1);
```
ahoffer: These assertion verify the result from the CWS query. We want to make sure the query result it good before we try to transform it, or else subsequent assertions will fail. Those failures would mask the real source of the problem-- the query results.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -1852,17 +1957,10 @@ private boolean foundExpectedActivity(List<String> activities, String filename,
...
-                    .anyMatch(activity -> {
-                        if (activity.toString()
-                                .contains(messageToFind) && activity.toString()
-                                .contains(filename)) {
-                            return true;
-                        }
-                        return false;
-                    });
-
+                    .anyMatch(activity -> activity.toString()
```
ahoffer: IntelliJ 2017 decided to refactoring this thing, all on its own.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -2416,6 +2514,58 @@ public void testAsyncDownloadActionPresentUsingCometDClient() throws Exception {
                 is(expectedUrl));
     }
 
+    protected List<Matcher<?>> getXpathMatchers(String expectedOutputSchema,
```
ahoffer: What follows is a bunch of helper methods to cut down on duplicated code in the `@Test`methods.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+                startsWith(thumbNailBase64EncSubstring)));
+        assertions.add(hasXPath("//*[@name='title']/*/text()", equalTo(metacardTitle)));
+        assertions.add(hasXPath("//pos/text()", equalTo("30.0 10.0")));
+
+        // Query for CSW result and assert expected results
+        ValidatableResponse response = getAndValidateCswResponse(metacardTitle,
+                METACARD_URI,
+                assertions);
+
+        // Extract the port of the response that can be transformed into a metacard
```
emanns95: *part

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+        String metacardXml = extractRecord(response, extractXmlRegex);
+
+        // Get the metacard XML input transformer and attempt to create a metacard object
+        Metacard metacard = getInputTransformer("(id=xml)").transform(IOUtils.toInputStream(
+                metacardXml,
+                "UTF-8"));
+
+        // Assert the newly created metacard's attributes are properly populated.
+        assertThat("Incorrect metacard's title", metacard.getTitle(), equalTo(metacardTitle));
```
emanns95: Nit: I'd say "metacard title" instead

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+        assertions.add(hasXPath("//*[name()='title']/*/text()", equalTo(metacardTitle)));
+        assertions.add(hasXPath("//*[name()='westBoundLongitude']/*[local-name()='Decimal']/text()",
+                equalTo("30.0")));
+
+        // Query for CSW result and assert expected results
+        ValidatableResponse response = getAndValidateCswResponse(metacardTitle,
+                GMD_SCHEMA_URI,
+                assertions);
+
+        // Extract the port of the response that can be transformed into a metacard
```
emanns95: *part

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+        assertions.add(hasXPath("//*[local-name()='references']/text()",
+                startsWith(thumbNailBase64EncSubstring)));
+        assertions.add(hasXPath("//references/text()", startsWith(thumbNailBase64EncSubstring)));
+        assertions.add(hasXPath("//*[local-name()='LowerCorner']/text()", equalTo("30.0 10.0")));
+
+        // Query for CSW result and assert expected results
+        ValidatableResponse response = getAndValidateCswResponse(metacardTitle, OPEN_GIS_SCHEMA_URI,
+                assertions);
+
+        // Extract the port of the response that can be transformed into a metacard
```
emanns95: *part

ahoffer:
build now

ahoffer:
@emanns95. The TODO now references a DDF ticket for the issue. Comments now also include a better description.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -949,7 +949,11 @@ public void testCswQueryForOpenGisXml() throws Exception {
         assertThat("Incorrect title", metacard.getTitle(), equalTo(metacardTitle));
         assertThat("Incorrect geometry", "POINT (30.0 10.0)", equalTo(metacard.getLocation()));
 
-        // TODO: THIS ASSERTION FAILS. VERIFY WITH CHRIS IT IS WORKING AS DESIGNED.
+        // TODO: THIS ASSERTION FAILS. SEE DDF-2476 (https://codice.atlassian.net/browse/DDF-2476)
```
emanns95: Do we want to add a comment on that ticket to revisit this block? There's a lot of TODOs that get left around after they're actually resolved

ahoffer: Updated DDF-2476 with reference to remote the comment and enable the assertion.
screen shot 2017-04-24 at 8 13 50 pm

ahoffer:
build now

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
brendan-hofmann: In addition to being inconsistent with the rest of our tests, the problem with using xpath is that it increases the risk of false-positives because you are no longer asserting anything about the overall structure of the response. Since we use xpath matchers right below this, why not just be consistent for the entire test?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +849,112 @@ public void testCswQueryByTitle() throws Exception {
...
+        // Query for CSW result and assert expected results
+        ValidatableResponse response = getAndValidateCswResponse(metacardTitle,
+                METACARD_URI,
+                assertions);
+
+        // Extract the port of the response that can be transformed into a metacard
+        String metacardXml = extractRecord(response, extractXmlRegex);
+
+        // Get the metacard XML input transformer and attempt to create a metacard object
+        Metacard metacard = getInputTransformer("(id=xml)").transform(IOUtils.toInputStream(
```
brendan-hofmann: This doesn't go through the same flow as a query over csw. I think we should be creating the metacards by doing a loopback query specifying this as the outputformat, so that we verify things like caching, query plugins, and schema validation in cxf play nicely with each schema. You could also keep this step and do the above as an additional, final test for extra granularity, I guess.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
brendan-hofmann: It looks like this string could be a constant as it seems to be the same in each test.

ahoffer:
build now

ahoffer:
build now

ahoffer:
build now

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -833,23 +856,114 @@ public void testCswQueryByTitle() throws Exception {
...
+        Metacard metacard =
+                getInputTransformer("(id=gmd:MD_Metadata)").transform(IOUtils.toInputStream(xml,
+                        "UTF-8"));
+
+        // Assert the newly created metacard's attributes are properly populated.
+        assertThat("Incorrect title", metacard.getTitle(), equalTo(metacardTitle));
+        assertThat("Incorrect geometry", "POINT (30 10)", equalTo(metacard.getLocation()));
+    }
+
+    protected String extractXml(ValidatableResponse response, String xpathQuery)
```
coyotesqrl: Nit: this method should be moved below the public test methods.

ahoffer: will do.

ahoffer:
Need to do a partial rewrite based on conversation with @brendan-hofmann

PR #1926, opened 2017-04-25
DDF-2964 Fix GeoJsonInputTransformer to explicitly set the metacard location with the geojson geometry

roelens8:
#### What does this PR do?
GeoJsonInputTransformer explicitly sets the value of the metacard location field with the value of the GeoJson's geometry field.
@ryeats 
@harrison-tarr 
@kcwire
@pklinef 
@jaymcnallie 
-Run a full build clean build. 
-Ingest geojson containing geometry data and verify the metacard location field is populated.
When using the GeoJsonInputTransfomer to convert from GeoJson to a metacard, if the metacard's metacard type had multiple attribute descriptors with their types equal to "GEO", the old transformer logic would put the value of the geojson's geometry field into the last field it finds that has a type equal to "GEO" instead of immediately putting it into the metacard's location field.
[DDF-2964](https://codice.atlassian.net/browse/DDF-2964)
#### Checklist:
- [N/A] Documentation Updated
- [N/A] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/transformer/catalog-transformer-geojson-input/src/main/java/ddf/catalog/transformer/input/geojson/GeoJsonInputTransformer.java
+++ b/catalog/transformer/catalog-transformer-geojson-input/src/main/java/ddf/catalog/transformer/input/geojson/GeoJsonInputTransformer.java
@@ -31,6 +31,7 @@ import java.util.function.Function;
 
 import javax.xml.bind.DatatypeConverter;
 
+import org.apache.commons.lang3.StringUtils;
 import org.boon.json.JsonFactory;
 import org.boon.json.ObjectMapper;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


catalog/transformer/catalog-transformer-geojson-input/src/main/java/ddf/catalog/transformer/input/geojson/GeoJsonInputTransformer.java
```diff
@@ -149,22 +150,8 @@ public Metacard transform(InputStream input, String id)
...
-
-        if (geoJsonGeometry != null) {
-            if (geoAttributeName != null) {
-                metacard.setAttribute(geoAttributeName, geoJsonGeometry.toWkt());
-            } else {
-                LOGGER.info("Loss of data, could not place geometry [{}] in metacard",
-                        geoJsonGeometry.toWkt());
-            }
+        if (geoJsonGeometry != null && StringUtils.isNotEmpty(geoJsonGeometry.toWkt())) {
+            metacard.setLocation(geoJsonGeometry.toWkt());
```
harrison-tarr: What happens if the metacard shouldn't have the location field? Are there any metacard types where we don't want to populate the location field in favor of a different GEO field?

roelens8: The metacard that the input transformer returns will always have the location field. The input transformer creates a MetacardImpl which uses the base MetacardType and the base MetacardType includes the location attribute.

roelens8: I don't know of any other metacard types where location wouldn't be the best fit for the geojson geometry.

kcwire: We shouldn't inherently rely on how MetacardTypeImpl is implemented. It could change in the future to not include the Location attribute since it is an implementation.

harrison-tarr:
Successfully manually heroed and verified fix. 👍

PR #1952, opened 2017-05-02
DDF-2969 Add static analysis steps to Jenkins file

clockard:
#### What does this PR do?
Adds static analysis steps to jenkins file.
It also includes:
- changes needed to make owas/nsp scans work correctly.
- sends itest output to a log file 
- makes the base builds parallel builds
- updates the surefire plugin version
@oconnormi @mcalcote 
@lessarderic
@shaundmorris

Verify that all static analysis tools run. RIght now we are expecting owas to fail because of real dependency issues
[DDF-2969](https://codice.atlassian.net/browse/DDF-2969)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,10 @@ pipeline {
     triggers {
         cron('H H(20-23) * * *')
     }
+    environment {
+        DOCS = 'distribution/docs'
+        ITESTS= 'distribution/test/itests/test-itests-ddf'
+    }
     stages {
         stage('Setup') {
             steps{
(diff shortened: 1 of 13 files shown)


Jenkinsfile
```diff
@@ -20,28 +20,88 @@ pipeline {
...
+                        },
+                        coverity: {
+                            node('linux-medium') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    withCredentials([string(credentialsId: 'ddf-coverity-token', variable: 'COVERITY_TOKEN')]) {
+                                        withEnv(["PATH=${tool 'coverity-linux'}/bin:${env.PATH}"]) {
+                                            sh '''cov-build --dir cov-int mvn -B -DskipTests=true -Dfindbugs.skip=true -Dcheckstyle.skip=true -Djacoco.skip=true install -pl !distribution/docs --settings $MAVEN_SETTINGS
+                                              tar czvf ddf.tgz cov-int
+                                              curl --form token=$COVERITY_TOKEN --form email=cmp-security-team@connexta.com --form file=@ddf.tgz --form version="master" --form description="Description: from CI" https://scan.coverity.com/builds?project=codice%2Fddf'''
```
shaundmorris: Should updated the description here? I think it was fine but was a bit generic for coverity pushes.

pom.xml
```diff
@@ -778,7 +778,7 @@
                     <plugin>
                         <groupId>org.owasp</groupId>
                         <artifactId>dependency-check-maven</artifactId>
-                        <version>1.2.9</version>
+                        <version>1.4.5</version>
```
shaundmorris: Nice, did you look at any new configuration settings? https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html

Jenkinsfile
```diff
@@ -20,28 +20,88 @@ pipeline {
...
+                                    sh 'mvn install -q -B -Powasp -DskipTests=true -Dpmd.skip=true -Dfindbugs.skip=true -Djacoco.skip=true -Dcheckstyle.skip=true -pl !libs/owasp-diff-runner,!distribution/docs'
+                                }
+                            }
+                        },
+                        sonarqube: {
+                            node('linux-large') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    withCredentials([string(credentialsId: 'sonarqube-token', variable: 'SONAR_TOKEN')]) {
+                                        sh 'mvn -q -B -Dfindbugs.skip=true -Dcheckstyle.skip=true org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=$SONAR_TOKEN  -Dsonar.organization=codice -Dsonar.projectKey=ddf -pl !platform/solr/platform-solr-server-standalone,!distribution/docs,!distribution/test/itests/test-itests-ddf'
```
shaundmorris: If we have addressed issues with our maven phases we may not need `install` here. We needed previously because a kar was not being created when we got to the `distribution` module. `Package` may work as well. @oconnormi thoughts?

clockard: Package doesn't work right now for the reasons you mentioned. Once we resolve the phase problems next sprint we can change this.

Jenkinsfile
clockard: `DDF CI Build`?

pom.xml
clockard: I did look at them. I had added the `skipTestScope` to the configuration but somehow that was lost in my rebasing. I will add that option back in.

clockard:
build now

Jenkinsfile
```diff
@@ -20,35 +20,96 @@ pipeline {
...
             }
         }
+        stage('Static Analysis') {
+            steps {
+                parallel(
+                        owasp: {
+                            node('linux-large') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    sh 'mvn install -q -B -Powasp -DskipTests=true -Dpmd.skip=true -Dfindbugs.skip=true -Djacoco.skip=true -Dcheckstyle.skip=true -pl !libs/owasp-diff-runner,!distribution/docs'
```
lessarderic: ❓ Can't we use `-DskipStatic=true` here instead of listing every single static analysis tool we want to skip?

lessarderic: ❓ What is `owasp-diff-runner`?

Jenkinsfile
```diff
@@ -20,35 +20,96 @@ pipeline {
...
+                                }
+                            }
+                        },
+                        sonarqube: {
+                            node('linux-large') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    withCredentials([string(credentialsId: 'sonarqube-token', variable: 'SONAR_TOKEN')]) {
+                                        sh 'mvn -q -B -Dfindbugs.skip=true -Dcheckstyle.skip=true org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar -Dsonar.host.url=https://sonarqube.com -Dsonar.login=$SONAR_TOKEN  -Dsonar.organization=codice -Dsonar.projectKey=ddf -pl !platform/solr/platform-solr-server-standalone,!distribution/docs,!distribution/test/itests/test-itests-ddf'
```
lessarderic: ❓ Is there a way with pipeline scripts to define variables/constants? If so, should we consider extracting the list of modules to skip into one instead of having the list hard-coded here?

Jenkinsfile
```diff
@@ -20,35 +20,96 @@ pipeline {
...
+                        },
+                        coverity: {
+                            node('linux-medium') {
+                                checkout scm
+                                withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings') {
+                                    withCredentials([string(credentialsId: 'ddf-coverity-token', variable: 'COVERITY_TOKEN')]) {
+                                        withEnv(["PATH=${tool 'coverity-linux'}/bin:${env.PATH}"]) {
+                                            configFileProvider([configFile(fileId: 'coverity-maven-settings', replaceTokens: true, variable: 'MAVEN_SETTINGS')]) {
+                                                echo sh(returnStdout: true, script: 'env')
+                                                sh 'cov-build --dir cov-int mvn -DskipTests=true -Dfindbugs.skip=true -Dcheckstyle.skip=true -Djacoco.skip=true install -pl !distribution/docs --settings $MAVEN_SETTINGS'
```
lessarderic: ❓ `-DskipStatic=true`?

Jenkinsfile
```diff
@@ -20,35 +20,96 @@ pipeline {
...
+                                                sh 'tar czvf ddf.tgz cov-int'
+                                                sh 'curl --form token=$COVERITY_TOKEN --form email=cmp-security-team@connexta.com --form file=@ddf.tgz --form version="master" --form description="Description: DDF CI Build" https://scan.coverity.com/builds?project=codice%2Fddf'
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        nodesecurity: {
```
lessarderic: ✏️  Just to make it clear here that not referring to a Jenkins `node` here, should we call this `nodeJsSecurity` or something similar?

Jenkinsfile
```diff
@@ -20,35 +20,96 @@ pipeline {
...
+            }
+        }
         stage('Deploy') {
             agent { label 'linux-small' }
             steps{
                 withMaven(maven: 'M3', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice_settings.xml') {
                     checkout scm
                     sh 'mvn javadoc:aggregate -DskipStatic=true -DskipTests=true'
-                    sh 'mvn deploy -DskipStatic=true -DskipTests=true'
+                    retry(2) {
```
lessarderic: ❓ Extract the number of retries to a constant?

lessarderic: ❓ Is this actually going to work? If the first deploy fails in the middle, won't Nexus prevent us from uploading the artifacts that were already successfully uploaded?

Jenkinsfile
```diff
@@ -20,35 +20,96 @@ pipeline {
...
+                        sh 'mvn deploy -DskipStatic=true -DskipTests=true'
```
lessarderic: ❓ Could we use the `-T 1C` option here as well to speed up the uploads?

Jenkinsfile
clockard: Updated

Jenkinsfile
clockard: These were mostly copied from the current jenkins configuration. I think we can change to use skipStatic. I am not entirely sure what the owasp diff runner is. I think it was a tool to help find changed poms and only run owasp on those particular ones. It is not part of our executable distribution and it was causing the owasp run to have an exception. I will look into it a little more.

Jenkinsfile
clockard: changed name

Jenkinsfile
clockard: Good point. Since we can't overwrite artifacts this would fail. Removing retry.

Jenkinsfile
clockard: We have avoided the multi-threaded deploys in the past because if something goes wrong it is hard to recover from. However since this is a snapshot it shouldn't be a problem. Adding the multi threaded option.

clockard:
Build passed in our test jenkins envirionment

PR #1953, opened 2017-05-02
[WIP] DDF-2979 ContentProducerDataAccessObject is not accessing existing metacard on an update. Without the metacard attribute overrides set during the create step will be lost.

gordocanchola:
Adding and exposing filter builder to ContentComponent

Updating ContentProducer's thrown exceptions

Updating ContentProducerDataAccessObject's createContentItem.
  On ENTRY_MODIFY get the existing metacard for the file being updated/ingested. This metacard will be set in the UpdateStorageRequest instead of null.

#### What does this PR do?
  This PR changes the content component to retrieve and use the metacard associated to the file being updated.
@brendan-hofmann @dcruver 

@stustison
@clockard 
DDF-2979
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
+++ b/catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
@@ -29,6 +29,8 @@ import org.apache.camel.Message;
 import org.apache.camel.component.file.GenericFile;
 import org.apache.camel.component.file.GenericFileMessage;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang.StringUtils;
+import org.opengis.filter.Filter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -42,14 +44,21 @@ import ddf.catalog.content.operation.UpdateStorageRequest;
 import ddf.catalog.content.operation.impl.CreateStorageRequestImpl;
 import ddf.catalog.content.operation.impl.UpdateStorageRequestImpl;
 import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
+import ddf.catalog.federation.FederationException;
+import ddf.catalog.filter.FilterBuilder;
 import ddf.catalog.operation.CreateResponse;
 import ddf.catalog.operation.DeleteRequest;
 import ddf.catalog.operation.DeleteResponse;
+import ddf.catalog.operation.QueryResponse;
 import ddf.catalog.operation.Update;
 import ddf.catalog.operation.UpdateResponse;
 import ddf.catalog.operation.impl.DeleteRequestImpl;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
 import ddf.mime.MimeTypeMapper;
 import ddf.mime.MimeTypeResolutionException;
 
@@ -121,7 +130,9 @@ public class ContentProducerDataAccessObject {
 
     public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEndpoint endpoint,
             File ingestedFile, WatchEvent.Kind<Path> eventType, String mimeType,
-            Map<String, Object> headers) throws SourceUnavailableException, IngestException {
+            Map<String, Object> headers)
+            throws SourceUnavailableException, IngestException, FederationException,
+            UnsupportedQueryException {
         LOGGER.debug("Creating content item.");
 
         String key = String.valueOf(ingestedFile.getAbsolutePath()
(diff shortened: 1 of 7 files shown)


catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -148,13 +159,41 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
                 }
             }
         } else if (StandardWatchEventKinds.ENTRY_MODIFY.equals(eventType)) {
+
+            Metacard metacard = null;
+
+            if (StringUtils.isNotEmpty(id)) {
```
brendan-hofmann: If any of these checks fail, we will fall back to the previous behavior. Shouldn't we throw an exception instead?

catalog/core/catalog-core-camelcomponent/src/test/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObjectTest.java
```diff
@@ -61,6 +63,8 @@
     ContentProducerDataAccessObject contentProducerDataAccessObject =
             new ContentProducerDataAccessObject();
 
+    private FilterBuilder filterBuilder = new GeotoolsFilterBuilder();
```
brendan-hofmann: We should add a `testUpdateContentItem` that asserts that the existing metacard is looked up and used for the update, and that the attribute overrides from the create are not overwritten. We should also cover any negative cases for the conditional branches.

Likewise, we should update the existing integration test to verify that attribute overrides are preserved on update: https://github.com/codice/ddf/blob/ce0646e54dede07c57bf939a70c27e68a313a844/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java#L1514-L1514

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -148,13 +159,41 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
+
+                FilterBuilder filterBuilder = endpoint.getComponent()
+                        .getFilterBuilder();
+                if (filterBuilder != null) {
+                    Filter filter = filterBuilder.attribute(Metacard.ID)
+                            .is()
+                            .like()
```
clockard: Since this is the metacard id should this be equals instead of like?

clockard: Since this is the metacard id should this be equals instead of like?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
gordocanchola: Changed to throw ContentComponentException.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
gordocanchola: Changed to use equalTo instead of like

catalog/core/catalog-core-camelcomponent/src/test/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObjectTest.java
```diff
@@ -242,6 +266,167 @@ public void testCreateContentItem() throws Exception {
     }
 
     @Test
+    public void testUpdateContentItem() throws Exception {
+        File testFile = temporaryFolder.newFile("testUpdateContentItem.txt");
```
clockard: Do we need to assert or verify something at the end of this test?

catalog/core/catalog-core-camelcomponent/src/test/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObjectTest.java
```diff
@@ -180,6 +191,15 @@ public void testCreateContentItem() throws Exception {
         doReturn(metacardList).when(mockUpdateResponse)
                 .getUpdatedMetacards();
 
+        Result mockResult = mock(Result.class);
```
bdeining: Might be worth reassessing the code coverage numbers with this change

gordocanchola:
This PR was to change the update flow of ContentProducerDataAccessObject to get the metacard and pass that in to the ContentItemImpl used int the UpdateStorageRequestImpl.

As part of the PR comments @brendan-hofmann suggested adding an itest which revealed bigger issues with the update flow. Because the changes would require a bigger effort and design discussions, it was determined that this should be put in a color team's sprint.

This PR as it stands would fail itests builds and is not ready to be merged. It should be used as the starting point for further investigation

rzwiefel:
Being closed for now until work is resumed on this.

PR #1956, opened 2017-05-03
[WIP] DDF-2982 Added Scripting support (groovy)

rzwiefel:
#### What does this PR do? 
- Now able to run arbitrary groovy scripts by dropping them in /etc/scripts

Still there needs to be some decision on what bindings we want to pass to the script (what we make available to the scripter. (Logger, services?, etc)

@coyotesqrl @michaelmenousek @pklinef 

@stustison [Must be on local system, so hopefully no security concerns]

https://codice.atlassian.net/browse/DDF-2982
#### Checklist:
- [ ] Documentation Updated 
- [ ] Change Log Updated 
- [ ] Update / Add Unit Tests


stustison:
Need to include documentation explaining how this scripts directory works and how to harden it for a hardened system.

michaelmenousek:
I was thinking our scripting support would be more controlled, but this is an approach as well.  This one feels more like the nuclear option though ;-)

rzwiefel:
To me, this is what I would think of as the 'dev mode' scripting. Something we could use to basically accomplish anything quickly while helping support, etc. But yes for integrators and such it should probably be locked down.

michaelmenousek:
I agree with @coyotesqrl that this is a lot like what @stustison did with the Twitter source and should probably hang out elsewhere while it cooks :-)

pklinef:
I like @coyotesqrl idea to make this command based.  We can then create an audit log for when a user executes that command.  It would be easier for the admin to rerun the command instead of having to move or touch files.  We can include warnings and ask the admin if they trust the source of the script.

PR #1968, opened 2017-05-09
DDF-2997 Remove implicit globals

andrewkfiedler:
#### What does this PR do?
 - Babel will cause the app to crash at the first implicit global it encounters, so this removes all of them.
 - Changes the label on inputs to be display block to avoid issue with whitespace between the label and the validation div (this would show up on invalid enums that were multivalued).

@djblue 
@rzwiefel 
@andrewkfiedler
@pklinef 

https://codice.atlassian.net/browse/DDF-2997

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
@@ -113,7 +113,7 @@ define([
                         if (CQLUtils.isPointRadiusFilter(filter)){
                             var pointText = filter.value.value.substring(6);
                             pointText = pointText.substring(0, pointText.length - 1);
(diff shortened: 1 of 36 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
```diff
@@ -113,7 +113,7 @@ define([
                         if (CQLUtils.isPointRadiusFilter(filter)){
                             var pointText = filter.value.value.substring(6);
                             pointText = pointText.substring(0, pointText.length - 1);
-                            latLon = pointText.split(' ');
+                            var latLon = pointText.split(' ');
```
djblue: If you have babel and you want block scope, you might want to consider using `let`.

jlcsmith:
build now

PR #1973, opened 2017-05-10
DDF-3001 Update registry to insert subject information into requests.

clockard:
#### What does this PR do?
Updates the registry such that its requests include a subject. 
Also fixed issues in UpdateOperation, DeleteOperation and AbstractCswStore where the subject wasn't getting passed along with generated requests
@emmberk @mcalcote 

@coyotesqrl
@stustison
Test the following registry functions
-Publish
-Unpublish
-Publication update
-Subscription
[DDF-3001](https://codice.atlassian.net/browse/DDF-3001)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
@@ -13,7 +13,6 @@
  */
 package ddf.catalog.impl.operations;
 
-import java.io.Serializable;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -21,6 +20,7 @@ import java.util.Set;
 
 import org.apache.shiro.SecurityUtils;
 
+import ddf.catalog.operation.Operation;
 import ddf.security.SecurityConstants;
 import ddf.security.Subject;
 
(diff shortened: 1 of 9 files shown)


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
```diff
@@ -37,4 +43,30 @@ void buildPolicyMap(HashMap<String, Set<String>> policyMap,
...
+
+    /**
+     * Creates and returns a map of properties containing a security subject. If the correct subject
+     * type can not be obtained from the passed in subject or from the thread context no subject will
+     * be added to the properties.
+     * If the passed in subject is not compatible the method will fallback to trying to pull it from
+     * the thread context.
+     * @param subject The subject to be added. Can be {@code null}
+     */
+    Map<String, Serializable> getPropertiesWithSubject(Object subject) {
```
coyotesqrl: This method feels odd. It seems that while callers might want a `Map<String, Serializable>` that you're contorting yourself here to provide that. I'd think just returning the subject or null would be sufficient. I might be missing something this early in the review, however.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
```diff
@@ -37,4 +43,30 @@ void buildPolicyMap(HashMap<String, Set<String>> policyMap,
...
+        Map<String, Serializable> properties = new HashMap<>();
+        if(subject != null && subject instanceof Subject){
```
coyotesqrl: This test can be simplified to just the `instanceof` test. A `null` input for subject will return `false` for the `instanceof` test.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
```diff
@@ -37,4 +43,30 @@ void buildPolicyMap(HashMap<String, Set<String>> policyMap,
...
+            properties.put(SecurityConstants.SECURITY_SUBJECT, (Subject)subject);
+        } else {
+            try {
+                Object subjectFromContxt = SecurityUtils.getSubject();
```
coyotesqrl: Typo if you care.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
clockard: I agree. I wrote this method a few different ways but ended with this to minimize duplicating code. One of the ways was to just return the subject for a given operation. I will go back to that way of doing it.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
clockard: Good call

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsSecuritySupport.java
clockard: Oops. Fixed

clockard:
build now

stustison:
nit: You could probably have used `Collections.singletonMap(K, V)` for most of this to avoid the map creation.

clockard:
Manually tested and verified.

PR #1985, opened 2017-05-16
DDF-3012 Guess mimeType of ingesting files without extensions

emanns95:
#### What does this PR do?
Adds a check to MimeTypeMapperImpl that uses Tika to guess the mimeType of files that do not have file extensions.

@brendan-hofmann 
@ahoffer 
@mweser 
@AzGoalie 

@clockard 
@stustison

Run full tests.
Attempt to ingest an xml file without a file extension. Query it and observe the metacardType is ddf.metacard instead of fallback.common.

Before this change, xml files without a file extension would go into the TikaInputTransformer and get a metacardType of fallback.common.

[DDF-3012](https://codice.atlassian.net/browse/DDF-3012)

#### Checklist:
- [x] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests


distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1621,12 +1624,53 @@ private ValidatableResponse assertIngestedDirectoryMonitor(String query, int num
             } catch (InterruptedException e) {
             }
         } while (TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startTime)
-                < TimeUnit.MINUTES.toMillis(1));
+                < TimeUnit.MINUTES.toMillis(3));
```
emanns95: Guessing the type of an extensionless file makes the ingest take a little longer than others. Adjusted the timeout here to reuse the method for the new test with a little bit safer buffer.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -1621,12 +1624,53 @@ private ValidatableResponse assertIngestedDirectoryMonitor(String query, int num
...
+        Configuration managedService = getServiceManager().createManagedService(
+                "org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor",
+                cdmProperties);
+
+        ValidatableResponse response = assertIngestedDirectoryMonitor("SysAdmin", 1);
+        assertThat(response.extract()
+                .response()
+                .body()
+                .prettyPrint()
+                .contains("ddf.metacard"), is(true));
```
emanns95: I actually meant to replace this with a better check but forgot. I'll get it in

platform/platform-app/src/main/resources/features.xml
```diff
@@ -837,6 +837,8 @@
         <feature>mime-core-api</feature>
         <bundle>mvn:ddf.mime.core/mime-core-impl/${project.version}</bundle>
         <bundle>mvn:ddf.mime.core/mime-core-configurableresolver/${project.version}</bundle>
+        <bundle>mvn:org.apache.tika/tika-core/${tika.version}</bundle>
+        <bundle>mvn:org.codice.thirdparty/tika-bundle/1.13.0_1</bundle>
```
emanns95: I'll change this to use the property instead of a hardcoded version

platform/mime/core/platform-mime-core-impl/pom.xml
```diff
@@ -75,6 +77,15 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.tika</groupId>
+            <artifactId>tika-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.codice.thirdparty</groupId>
+            <artifactId>tika-bundle</artifactId>
```
stustison: You generally don't use a bundle as a dependency, you use a jar. The one that you actually need is most likely tika-parsers. The bundles only belong in the features files.

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +181,25 @@ public String guessMimeType(InputStream is, String fileExtension)
         // from being used when a CustomMimeTypeResolver may be more appropriate.
         List<MimeTypeResolver> sortedResolvers = sortResolvers(mimeTypeResolvers);
 
+        if (StringUtils.isEmpty(fileExtension)) {
+            try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+                IOUtils.copy(is, baos);
```
stustison: You can't do this. This will blow up the heap if someone is attempting to upload a very large file. You need to use a TemporaryFileBackedOutputStream which can fail over to the file system in the event that the InputStream has too many bytes in it.

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +181,25 @@ public String guessMimeType(InputStream is, String fileExtension)
...
+                byte[] bytes = baos.toByteArray();
+                is = new ByteArrayInputStream(bytes);
+
+                Parser parser = new AutoDetectParser();
+                ToXMLContentHandler xmlContentHandler = new ToXMLContentHandler();
+                Metadata metadata = new Metadata();
+                parser.parse(is, xmlContentHandler, metadata, new ParseContext());
+                is.reset();
```
stustison: When you switch this over to use the file backed stream, you can simply ask it for a new stream as opposed to resetting the existing one.

platform/platform-app/pom.xml
```diff
@@ -317,6 +317,11 @@
             <artifactId>platform-email-impl</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.codice.thirdparty</groupId>
+            <artifactId>tika-bundle</artifactId>
```
stustison: Use the jar.

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
emanns95: Would that look like `temporaryFileBackedOutputStream.asByteSource().openStream()`?

platform/mime/core/platform-mime-core-impl/pom.xml
emanns95: All the other usages in the project also use it as a dependency in this way. Is that tech debt?

stustison: Might be, we shouldn't be using a bundle in this way when jars are available.

emanns95: I'll make a ticket for it once i get this working with the jar

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +182,26 @@ public String guessMimeType(InputStream is, String fileExtension)
...
+            try {
+                TemporaryFileBackedOutputStream tfbos = new TemporaryFileBackedOutputStream();
+                IOUtils.copy(is, tfbos);
+                is = tfbos.asByteSource().openStream();
+
+                Parser parser = new AutoDetectParser();
+                ToXMLContentHandler xmlContentHandler = new ToXMLContentHandler();
+                Metadata metadata = new Metadata();
+                parser.parse(is, xmlContentHandler, metadata, new ParseContext());
+                is.close();
```
stustison: If something errors, this stream is going to be left unclosed. Need to probably open this in a try-with-resources.

mweser:
Manually verified functionality

platform/mime/core/platform-mime-core-impl/src/test/java/ddf/mime/mapper/MimeTypeMapperTest.java
```diff
@@ -35,68 +38,70 @@
...
     private static final String NO_NAMESPACES_XML_FILE = "src/test/resources/no_namespaces.xml";
 
+    private static final String NITF_FILE = "src/test/resources/sampleNitf.nitf";
+
+    private static TikaMimeTypeResolver tikaMimeTypeResolver;
+
+    static {
+        tikaMimeTypeResolver = new TikaMimeTypeResolver();
+        tikaMimeTypeResolver.setPriority(-1);
+    }
```
mweser: Why is this static?

emanns95: It's in a static block so I could set the value of the static constant

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +180,28 @@ public String guessMimeType(InputStream is, String fileExtension)
...
+                    Parser parser = new AutoDetectParser();
+                    ToXMLContentHandler xmlContentHandler = new ToXMLContentHandler();
+                    Metadata metadata = new Metadata();
+                    parser.parse(inputStream, xmlContentHandler, metadata, new ParseContext());
+
+                    String metacardContentType = metadata.get(Metadata.CONTENT_TYPE);
+                    fileExtension = getFileExtensionForMimeType(metacardContentType).replace(".",
+                            "");
+                }
+                is = tfbos.asByteSource()
```
stustison: I think this should be in a `finally`. That way if anything inside this `try` errors out, then we'll still have an input stream to use for the rest of the method.

platform/platform-app/src/main/resources/features.xml
```diff
@@ -837,6 +837,8 @@
         <feature>mime-core-api</feature>
         <bundle>mvn:ddf.mime.core/mime-core-impl/${project.version}</bundle>
         <bundle>mvn:ddf.mime.core/mime-core-configurableresolver/${project.version}</bundle>
+        <bundle>mvn:org.apache.tika/tika-core/${tika.version}</bundle>
+        <bundle>mvn:org.apache.tika/tika-parsers/${tika.version}</bundle>
```
stustison: We are already including our own tika-bundle in the catalog features file, instead we can just include it here. I don't think we want multiple things exporting tika packages.

platform/platform-app/src/main/resources/features.xml
```diff
@@ -835,6 +835,7 @@
     <feature name="mime-core" install="manual" version="${project.version}"
              description="DDF MIME Core">
         <feature>mime-core-api</feature>
+        <feature>catalog-core-api</feature>
```
emanns95: @stustison Is something like this moreso what we're looking for?

stustison: No you can't do that, that'll create a circular dependency. I was meaning that we should just use the same tika-bundle that catalog is trying to use. Karaf will only pull it in once so it will essentially ignore the line in catalog, if the mime-core feature was already installed.

emanns95:
build now

emanns95:
build now

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +180,29 @@ public String guessMimeType(InputStream is, String fileExtension)
         // from being used when a CustomMimeTypeResolver may be more appropriate.
         List<MimeTypeResolver> sortedResolvers = sortResolvers(mimeTypeResolvers);
 
+        if (StringUtils.isEmpty(fileExtension)) {
+            try (TemporaryFileBackedOutputStream tfbos = new TemporaryFileBackedOutputStream()) {
+                IOUtils.copy(is, tfbos);
+                try (InputStream inputStream = tfbos.asByteSource()
+                        .openStream()) {
+                    Parser parser = new AutoDetectParser();
```
brendan-hofmann: Do we need integrator documentation for this hook?

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +180,29 @@ public String guessMimeType(InputStream is, String fileExtension)
...
+                    ToXMLContentHandler xmlContentHandler = new ToXMLContentHandler();
+                    Metadata metadata = new Metadata();
+                    parser.parse(inputStream, xmlContentHandler, metadata, new ParseContext());
```
brendan-hofmann: Will this parse the entire inputstream or stop once it has a guess? We want to break early or we'll introduce a lot of unnecessary overhead.

emanns95: I can't find that documented anywhere. My gut tells me it'll go through the entire inputstream

emanns95: I just went into the debugger and I see the inputstream position is at the end after this parse method

emanns95: Though I don't know why I'm fully parsing it here instead of using the Tika detector that the parser uses. Changing it now

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
clockard: The sleep time for this loop seems too small. Why not just do it every second?

emanns95: I don't see a reason not to. I'll change it

emanns95:
build now

emanns95:
build now

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +179,26 @@ public String guessMimeType(InputStream is, String fileExtension)
         // from being used when a CustomMimeTypeResolver may be more appropriate.
         List<MimeTypeResolver> sortedResolvers = sortResolvers(mimeTypeResolvers);
 
+        if (StringUtils.isEmpty(fileExtension)) {
+            try (TemporaryFileBackedOutputStream tfbos = new TemporaryFileBackedOutputStream()) {
+                IOUtils.copy(is, tfbos);
+                try (InputStream inputStream = tfbos.asByteSource()
+                        .openStream()) {
+                    Detector detector = new DefaultDetector();
```
emanns95: @brendan-hofmann using the Default Detector here collects all the composite detectors, which are documented to not read through the entire stream https://tika.apache.org/1.14/api/

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -66,8 +71,7 @@ public int compare(MimeTypeResolver o1, MimeTypeResolver o2) {
     /**
      * Constructs the MimeTypeMapper with a list of {@link MimeTypeResolver}s.
      *
-     * @param mimeTypeResolvers
-     *            the {@link List} of {@link MimeTypeResolver}s
+     * @param mimeTypeResolvers the {@link List} of {@link MimeTypeResolver}s
```
ahoffer: There is a TODO comment left in some code in another place in the file. It's not in anything you touched, but it would be a bonus if we could clean it up:

` // TODO: This is to force the TikaMimeTypeResolver to be called`

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +179,26 @@ public String guessMimeType(InputStream is, String fileExtension)
...
+                } finally {
+                    is = tfbos.asByteSource()
+                            .openStream();
+                }
+
+            } catch (Exception e) {
+                LOGGER.debug("Failed to guess mimeType for file without extension.");
+            }
+        }
+
```
ahoffer: A little further down (again in code you didn't touch), the method `getRootNamespace` is invoked. This class has it's own implementation of the getRootNamespace method, which build the entire DOM. The class ` org.codice.ddf.platform.util.XMLUtils` has a method named `getRootNamespace` that uses the SAX parser and only looks at the first element to get the rootnamespace. Recommend you use that implementation, assuming it does the right thing.

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
```diff
@@ -175,6 +179,26 @@ public String guessMimeType(InputStream is, String fileExtension)
         // from being used when a CustomMimeTypeResolver may be more appropriate.
         List<MimeTypeResolver> sortedResolvers = sortResolvers(mimeTypeResolvers);
 
+        if (StringUtils.isEmpty(fileExtension)) {
+            try (TemporaryFileBackedOutputStream tfbos = new TemporaryFileBackedOutputStream()) {
```
ahoffer: Why a new filestream? Couldn't we read from the InputStream and then reset it when we are done with it? Like this?
```
if(StringUtils.isEmpty(fileExtension))) {
           try {
                Detector detector = new DefaultDetector();
                MediaType mediaType = detector.detect(is, new Metadata());
                fileExtension = getFileExtensionForMimeType(mediaType.toString()).replace(".", "");
            } catch (Exception e) {
                LOGGER.debug("Failed to guess mimeType for file without extension.");
            } finally {
                try {
                    is.reset();
                } catch (IOException e) {
                    LOGGER.debug("Failed to reset input stream.");
                }
            }
}
```

platform/mime/core/platform-mime-core-impl/src/test/java/ddf/mime/mapper/MimeTypeMapperTest.java
```diff
@@ -123,4 +132,48 @@ public void testGuessMimeTypeForXmlIngest() throws Exception {
...
+                "urn:catalog:metacard"));
+        resolversWithXmlPriority.add(new MockMimeTypeResolver("CswResolver",
+                10,
+                new String[] {"xml=text/xml;id=csw"},
+                "http://www.opengis.net/cat/csw/2.0.2"));
+
+        mapper = new MimeTypeMapperImpl(resolversWithXmlPriority);
+        mimeType = mapper.getMimeTypeForFileExtension("xml");
+        LOGGER.debug("mimeType = {}", mimeType);
+        assertEquals("text/xml", mimeType);
```
ahoffer: This test adds two new resolvers with different priorities, starting on line 163. Line 160 might be a good place to end the test method and start a new one. 

Also, I can't really follow what lines 163 to the end are testing. Why does one resolver use "xml=text/xml" as the schema and the other resolver add ";id=csw" to the end of the schema? Using dummy values like
`new String[] {"xml=HIGH_PRIORITY"}`
and `new String[] {"xml=LOW_PRIORITY"}`
would be good.

Is the schema value (4th constructor parameter of the resolver) being used ? It  doesn't look like it. If it is not used, let's pass `null` or an empty string for the schema. Having real values made me think they were being used by the test somehow.

platform/pom.xml
```diff
@@ -270,6 +271,11 @@
                 <version>${tika.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.codice.thirdparty</groupId>
+                <artifactId>tika-bundle</artifactId>
+                <version>${tika.thirdparty.bundle.version}</version>
+            </dependency>
```
emanns95: We went back and forth on this, but that's how it's being done throughout catalog and I couldn't get it to work any other way

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
emanns95: The InputStream interface doesn't support `.reset()`

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
emanns95: I'll take a look

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
ahoffer: Really? The interface looks like it does. Maybe the implementation doesn't?
https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html

emanns95: Well the `reset()` method there says 
> If the method markSupported returns false, then: The call to reset may throw an IOException.

And then below for `markSupported`
>The markSupported method of InputStream returns false.

platform/mime/core/platform-mime-core-impl/src/test/java/ddf/mime/mapper/MimeTypeMapperTest.java
emanns95: The second part isn't really necessary. Was just demonstrating that it doesn't bound it to the text/xml resolver. I'm going to remove that part though

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
emanns95: I don't really see how this is a TODO. It's more of an explanation...I don't think I have a way of asking the original author so I think I'm just going to remove the TODO tag...

platform/mime/core/platform-mime-core-impl/src/main/java/ddf/mime/mapper/MimeTypeMapperImpl.java
emanns95: Done

emanns95:
build now

PR #1987, opened 2017-05-16
DDF-3009 Added service to generate Metacard UUIDs

bdeining:
#### What does this PR do?
This PR creates a configurable service to create Metacard UUIDs and extracts the use of UUID.randomUUID from areas in DDF.
@rzwiefel @adimka 
@brianfelix @troymohl 
@jlcsmith
@millerw8
- Build 
- Install 
- Ingest some data and verify IDs are still set properly
- Verify Archive / Restore / Caching has not regressed
- Configure Metacard UUID Generator (remove hyphens)
- Ingest some data and verify IDs are set properly
- Verify Archive / Restore / Caching has not regressed
- Configure a Content Directory Monitor and verify ingest works properly
[DDF-3009](https://codice.atlassian.net/browse/DDF-3009)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [X] Update / Add Unit Tests

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java
@@ -17,7 +17,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.UUID;
 
 import javax.activation.MimeType;
 import javax.activation.MimeTypeParseException;
@@ -52,22 +51,6 @@ public class ContentItemImpl implements ContentItem {
     private Metacard metacard;
 
     protected String qualifier;
-
-    /**
-     * An incoming content item whose ID will initially be
-     * <code>null</code> because the {@link ddf.catalog.CatalogFramework} will assign its GUID.
-     *
-     * @param byteSource      the {@link ContentItem}'s input stream containing its actual data
-     * @param mimeTypeRawData the {@link ContentItem}'s mime type
-     * @param filename        the {@link ContentItem}'s file name - can be null
-     * @param metacard        the {@link ContentItem}'s associated metacard
-     */
-    public ContentItemImpl(ByteSource byteSource, String mimeTypeRawData, String filename,
-            Metacard metacard) {
-        this(UUID.randomUUID()
-                .toString()
-                .replaceAll("-", ""), byteSource, mimeTypeRawData, filename, 0, metacard);
-    }
 
     /**
      * An incoming content item where the item's GUID should be known.
--- a/catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducer.java
+++ b/catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducer.java
@@ -23,6 +23,7 @@ import org.apache.camel.ExchangePattern;
 import org.apache.camel.Message;
 import org.apache.camel.impl.DefaultProducer;
 import org.apache.commons.lang.StringUtils;
+import org.codice.ddf.platform.util.uuidgenerator.UuidGenerator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,8 +45,7 @@ public class ContentProducer extends DefaultProducer {
 
     private ContentEndpoint endpoint;
 
-    ContentProducerDataAccessObject contentProducerDataAccessObject =
-            new ContentProducerDataAccessObject();
+    ContentProducerDataAccessObject contentProducerDataAccessObject;
 
     /**
      * Constructs the {@link org.apache.camel.Producer} for the custom Camel ContentComponent. This producer would
@@ -56,7 +56,12 @@ public class ContentProducer extends DefaultProducer {
     public ContentProducer(ContentEndpoint endpoint) {
         super(endpoint);
         this.endpoint = endpoint;
-
+        ContentComponent contentComponent = endpoint.getComponent();
+        UuidGenerator uuidGenerator = null;
+        if (contentComponent != null) {
+            uuidGenerator = contentComponent.getUuidGenerator();
+        }
+        this.contentProducerDataAccessObject = new ContentProducerDataAccessObject(uuidGenerator);
         LOGGER.trace("INSIDE: ContentProducer constructor");
     }
 
--- a/catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/impl/DeletedMetacardImpl.java
+++ b/catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/impl/DeletedMetacardImpl.java
@@ -16,12 +16,8 @@ package ddf.catalog.core.versioning.impl;
 import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
-import java.util.UUID;
 
 import javax.annotation.Nullable;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.ImmutableSet;
 
@@ -40,8 +36,6 @@ import ddf.catalog.data.impl.MetacardTypeImpl;
  * Represents a currently soft deleted metacard.
  */
 public class DeletedMetacardImpl extends MetacardImpl implements DeletedMetacard {
-    private static final Logger LOGGER = LoggerFactory.getLogger(DeletedMetacardImpl.class);
-
     private static final MetacardType METACARD_TYPE;
 
     private static final Set<AttributeDescriptor> DESCRIPTORS =
@@ -54,7 +48,6 @@ public class DeletedMetacardImpl extends MetacardImpl implements DeletedMetacard
                 false,
                 false,
                 BasicTypes.STRING_TYPE));
-
         DESCRIPTORS.add(new AttributeDescriptorImpl(DELETION_OF_ID,
                 true,
                 true,
@@ -70,20 +63,18 @@ public class DeletedMetacardImpl extends MetacardImpl implements DeletedMetacard
         METACARD_TYPE = new MetacardTypeImpl(PREFIX, DESCRIPTORS);
     }
 
-    public DeletedMetacardImpl(String deletionOfId, String deletedBy, String lastVersionId,
-            Metacard metacardBeingDeleted) {
+    public DeletedMetacardImpl(String id, String deletionOfId, String deletedBy,
+            String lastVersionId, Metacard metacardBeingDeleted) {
         super(metacardBeingDeleted,
                 new MetacardTypeImpl(PREFIX,
-                        getDeletedMetacardType(),
-                        metacardBeingDeleted.getMetacardType()
-                                .getAttributeDescriptors()));
+                    getDeletedMetacardType(),
+                    metacardBeingDeleted.getMetacardType()
+                        .getAttributeDescriptors()));
         this.setDeletionOfId(deletionOfId);
         this.setDeletedBy(deletedBy);
         this.setLastVersionId(lastVersionId);
         this.setTags(ImmutableSet.of(DELETED_TAG));
-        this.setId(UUID.randomUUID()
-                .toString()
-                .replace("-", ""));
+        this.setId(id);
     }
 
     public static boolean isNotDeleted(@Nullable Metacard metacard) {
--- a/platform/platform-app/src/main/resources/features.xml
+++ b/platform/platform-app/src/main/resources/features.xml
@@ -760,6 +760,17 @@
         <bundle>mvn:ddf.platform.migration/platform-migration/${project.version}</bundle>
     </feature>
 
+    <feature name="util-uuidgenerator-api" install="manual" version="${project.version}"
+             description="UUID Generator API">
+        <feature prerequisite="true">platform-dependencies</feature>
+        <bundle>mvn:ddf.platform.util/util-uuidgenerator-api/${project.version}</bundle>
+    </feature>
+
+    <feature name="util-uuidgenerator" install="manual" version="${project.version}" description="A UUID generator service">
+        <feature>util-uuidgenerator-api</feature>
+        <bundle>mvn:ddf.platform.util/util-uuidgenerator-impl/${project.version}</bundle>
+    </feature>
+
     <feature name="platform-scheduler" install="manual" version="${project.version}"
              description="Schedules tasks">
         <feature prerequisite="true">commons</feature>
@@ -1046,6 +1057,7 @@
         <feature>landing-page</feature>
         <feature>platform-filter-response</feature>
         <feature>platform-email</feature>
+        <feature>util-uuidgenerator</feature>
     </feature>
 
     <feature name="clientinfo-filter" install="auto" version="${project.version}"
--- a/platform/util/platform-util-uuidgenerator/util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/UuidGeneratorImpl.java
+++ b/platform/util/platform-util-uuidgenerator/util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/UuidGeneratorImpl.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.platform.util.uuidgenerator.impl;
+
+import java.util.UUID;
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.StringUtils;
+import org.codice.ddf.platform.util.uuidgenerator.UuidGenerator;
+
+public class UuidGeneratorImpl implements UuidGenerator {
+
+    private Pattern hexPattern = Pattern.compile("^[0-9A-Fa-f]+$");
+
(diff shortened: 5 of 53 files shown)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java
```diff
@@ -54,22 +54,6 @@
...
-     * @param byteSource      the {@link ContentItem}'s input stream containing its actual data
-     * @param mimeTypeRawData the {@link ContentItem}'s mime type
-     * @param filename        the {@link ContentItem}'s file name - can be null
-     * @param metacard        the {@link ContentItem}'s associated metacard
-     */
-    public ContentItemImpl(ByteSource byteSource, String mimeTypeRawData, String filename,
-            Metacard metacard) {
-        this(UUID.randomUUID()
-                .toString()
-                .replaceAll("-", ""), byteSource, mimeTypeRawData, filename, 0, metacard);
```
bdeining: This was removed to prevent any circumvention of the metacard uuid generator

platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+ * The {@code MetacardUuidGenerator} generates {@link java.util.UUID}s for Metacards
+ */
+public interface MetacardUuidGenerator {
+
+    /**
+     * Returns a generated uuid for a Metacard
+     *
+     * @return
+     */
+    String getMetacardUuid();
```
millerw8: Should this be "createMetacardUuid" or "generateMetacardUuid" since you aren't just retrieving an existing value?  Should we limit it to Metacards or can this just be a service for generating and validating UUIDs in general?

platform/util/pom.xml
```diff
@@ -26,5 +26,6 @@
     <modules>
         <module>platform-util</module>
         <module>platform-util-unavailableurls</module>
+        <module>uuidgenerator</module>
```
troymohl: nit: should this follow the `platform-util` naming that the other modules are using?

platform/platform-app/src/main/resources/features.xml
```diff
@@ -760,6 +760,17 @@
         <bundle>mvn:ddf.platform.migration/platform-migration/${project.version}</bundle>
     </feature>
 
+    <feature name="platform-uuidgenerator-api" install="manual" version="${project.version}"
```
troymohl: Should feature naming match the module naming and include `util` in the name (`platform-util`)?

platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/MetacardUuidGeneratorImpl.java
```diff
@@ -0,0 +1,53 @@
...
+        if (removeHyphens) {
+            uuid = uuid.replaceAll("-", "");
+        }
+        return uuid;
+    }
+
+    @Override
+    public boolean isCorrectUuidFormat(String uuid) {
+        int length = removeHyphens ? 32 : 36;
+        Pattern pattern = removeHyphens ? hexPattern : hexPatternWithHypens;
```
millerw8: Maybe for patterns with hyphens we should just confirm UUID.fromString(uuid) doesn't throw an exception.  Using exceptions is generally not the best for logic, but it does confirm that there are 4 hyphens, and they are in the right place.  

Or alternatively, this pattern should be more precise and work: ^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$, and also eliminates the length requirement, although I am guessing that was a performance optimization over ^[0-9a-fA-F]{32}$

millerw8: Actually, looking at java.util.UUID.java, we definitely want to go with a regex "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$" over fromString because fromString seems to violate its documentation, it accepts 1-2-3-4-5 to make 00000001-0002-0003-0004-000000000005, even though it claims it throws an exception when not matching that format (as defined in BNF in the toString documentation)

Also, UUIDs technically encode things like its variant in the UUID, but nothing enforces that and I don't think we need to either.

platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ *
+ * The {@code MetacardUuidGenerator} generates {@link java.util.UUID}s for Metacards
+ */
+public interface MetacardUuidGenerator {
```
pklinef: Is there any reason this needs to be Metacard specific?  Also, there is not a concept of metacard at the platform level.

bdeining: No, i'll make it generic

platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/MetacardUuidGeneratorImpl.java
rzwiefel: Just as a head up we have this here, 
https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemValidator.java#L29-L42

rzwiefel: I think originally I chose regex because i didn't want to use exceptions as flow control (the fromString method), but good to know fromString isn't even accurate xD

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java
rzwiefel: Are we allowed to remove the ctor completely? Or is this a case where we need to deprecate and remove later?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducer.java
```diff
@@ -56,7 +55,7 @@
     public ContentProducer(ContentEndpoint endpoint) {
         super(endpoint);
         this.endpoint = endpoint;
-
+        this.contentProducerDataAccessObject = new ContentProducerDataAccessObject(endpoint.getComponent().getMetacardUuidGenerator());
```
rzwiefel: Is this formatted?

platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java
rzwiefel: Could we take advantage of some of our built in java interfaces and make this interface `extend Supplier<String>` that way we can use it as a drop in supplier? (And then the overriden `get` method could just call `getMetacardUuid`)

platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    String getMetacardUuid();
+
+    /**
+     * Returns true if the UUID format is correct
+     *
+     * @param uuid - the given UUID
+     * @return
+     */
+    boolean isCorrectUuidFormat(String uuid);
```
rzwiefel: `validateUuid`? `validateId`? `validateIdentifier` `isValid`?

platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/MetacardUuidGeneratorImpl.java
rzwiefel: Also shouldn't the validator not be stateful? Meaning, make it so that both with and without hyphens are valid? (EG, someone changes the removeHyphens setting or someone does a legacy upgrade etc etc)

platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/test/java/org/codice/ddf/platform/util/uuidgenerator/impl/TestMetacardUuidGeneratorImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestMetacardUuidGeneratorImpl {
+
+    private MetacardUuidGeneratorImpl metacardUuidGenerator;
+
+    private static final List<String> INVALID_ID_LIST_NO_HYPHENS = Arrays.asList(null, "", " ", "1234", "fc3f1798fe17_66fbbcc39dcfcb232e6", "fc3f1798fe17166fbbcc39dcfcb2232e6");
+
+    private static final List<String> INVALID_ID_LIST_HYPHENS = Arrays.asList("fc3f1798fe17166fbbcc39dcfcb2232e6----", "fc3f1798fe1766fbbcc39dcfcb232e6---1", "fc3f1798fe17166fbbcc39dcfcb2232e61234");
```
rzwiefel: Formatting?

platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java
rzwiefel: Even for the validate we could make it also `implements Predicate<String>` and have the test method override `isCorrectUuidFormat`

catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/impl/DeletedMetacardImpl.java
```diff
@@ -70,20 +65,17 @@
...
-                new MetacardTypeImpl(PREFIX,
-                        getDeletedMetacardType(),
-                        metacardBeingDeleted.getMetacardType()
-                                .getAttributeDescriptors()));
+    public DeletedMetacardImpl(String id, String deletionOfId, String deletedBy,
+            String lastVersionId, Metacard metacardBeingDeleted) {
+        super(metacardBeingDeleted, new MetacardTypeImpl(PREFIX,
+                getDeletedMetacardType(),
+                metacardBeingDeleted.getMetacardType()
+                        .getAttributeDescriptors()));
```
rzwiefel: Hmmm... My IntelliJ formats it like it was previously. I just pulled ddf-support and re-imported formatters and it looks like it does it the same still. Your formatter might be out of date.
```java
public DeletedMetacardImpl(String id, String deletionOfId, String deletedBy,
            String lastVersionId, Metacard metacardBeingDeleted) {
        super(metacardBeingDeleted,
                new MetacardTypeImpl(PREFIX,
                        getDeletedMetacardType(),
                        metacardBeingDeleted.getMetacardType()
                                .getAttributeDescriptors()));
        this.setDeletionOfId(deletionOfId);
        this.setDeletedBy(deletedBy);
        this.setLastVersionId(lastVersionId);
        this.setTags(ImmutableSet.of(DELETED_TAG));
        this.setId(id);
    }```

File: catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java (hunk 1 above)
Comment (bdeining) [inline catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java]: I'm not sure; but i'd point out that this is on the impl, and not an API change.

File: platform/util/pom.xml (hunk 3 above)
Comment (bdeining) [inline platform/util/pom.xml]: Renamed packages

File: platform/platform-app/src/main/resources/features.xml (hunk 4 above)
Comment (bdeining) [inline platform/platform-app/src/main/resources/features.xml]: Renamed packages

File: platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java (hunk 2 above)
Comment (bdeining) [inline platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java]: Renamed to "generateMetacardUuid"

File: platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/MetacardUuidGeneratorImpl.java (hunk 5 above)
Comment (bdeining) [inline platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/MetacardUuidGeneratorImpl.java]: Good points; updated to use @millerw8's regex suggestion and made the validation stateless

File: catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducer.java (hunk 7 above)
Comment (bdeining) [inline catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducer.java]: Refactored; should be formatted now

File: platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java (hunk 8 above)
Comment (bdeining) [inline platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java]: Renamed to `validateUuid`

File: platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/test/java/org/codice/ddf/platform/util/uuidgenerator/impl/TestMetacardUuidGeneratorImpl.java (hunk 9 above)
Comment (bdeining) [inline platform/util/uuidgenerator/platform-util-uuidgenerator-impl/src/test/java/org/codice/ddf/platform/util/uuidgenerator/impl/TestMetacardUuidGeneratorImpl.java]: Formatted

File: catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/impl/DeletedMetacardImpl.java (hunk 10 above)
Comment (bdeining) [inline catalog/core/catalog-core-versioning/versioning-common/src/main/java/ddf/catalog/core/versioning/impl/DeletedMetacardImpl.java]: Fixed

File: platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java (hunk 6 above)
Comment (bdeining) [inline platform/util/uuidgenerator/platform-util-uuidgenerator-api/src/main/java/org/codice/ddf/platform/util/uuidgenerator/MetacardUuidGenerator.java]: Refactored to be a generic uuid generator; decided to keep it simple for now with regards to `supplier` and `predicate`

**Comment by bdeining** [comment]
build now

File: platform/util/platform-util-uuidgenerator/util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/UuidGeneratorImpl.java [hunk 11]
```diff
@@ -0,0 +1,57 @@
...
+        return uuid;
+    }
+
+    @Override
+    public boolean validateUuid(String uuid) {
+        if (StringUtils.isEmpty(uuid)) {
+            return false;
+        }
+
+        if (uuid.length() == 32) {
```
millerw8: I think we still need to check value of `this.removeHypens` so we only validate based on the current mode we are in.

bdeining:
build now

brianfelix:
Manual tested/verified. LGTM 👍

platform/util/platform-util-uuidgenerator/util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/UuidGeneratorImpl.java
rzwiefel: I can just see a scenario where we either have a discrepancy in the setting being changed or something to do with other systems interacting with us. 

It also seems odd to provide the admins an option that they can only set once (that must be before any metacards are ever created) and there is no way to go back without something custom to rewrite solr ids to be a different format. Also this prevents us from making any automatically generated metacards at system startup. 

As little as I like adding more configuration options... maybe this is a case to add a "strict validation" option thats on by default so it validates as @millerw8 is saying, but an option to turn it off and make both valid?

millerw8: This actually appears to only be called one time in non-tests, and that is during generation in a create.  Which means the scenario we are allowing is the user to generate it by the opposite format, and then check it, and get it through.

This function doesn't actually restrict anything, it just a test for its format.  The groomer update doesn't validate or change the id.

bdeining:
build now

bdeining:
build now

millerw8:
build now

PR #2006, opened 2017-05-23
DDF-3027 Update wkt validator to correctly handle MultiPoint

andrewkfiedler:
#### What does this PR do?
Updates the wkt validator to correctly handle multipoint.

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?delete the rest).
@andrewkfiedler
@millerw8
@bdeining 
@adimka 

Enter a location of `GEOMETRYCOLLECTION (MULTIPOINT ((-113 32), (-112 33)), MULTIPOLYGON (((10 10, 15 10, 15 15, 10 15, 10 10)), ((20 20, 25 20, 25 25, 20 25, 20 20))), POLYGON ((-20 5, -15 5, -15 -5, -20 -5, -20 5)))` and verify that no issues are found.

https://codice.atlassian.net/browse/DDF-3027

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/input/geometry/input-geometry.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/input/geometry/input-geometry.view.js
@@ -22,8 +22,11 @@ var _ = require('underscore');
 function convertUserValueToWKT(val){
     val = val.split(' (').join('(').split(', ').join(',');
     val = val.split('MULTIPOINT').map(function(value, index){
-        if (index % 2 === 1 && value.indexOf('((') !== -1) {
-            return value.split('((').join('(').split('),(').join(',').split('))').join(')');
+        if (value.indexOf('((') === 0) {
+            var endOfMultiPoint = value.indexOf('))') + 2;
+            var multipointStr = value.substring(0, endOfMultiPoint);
+            multipointStr = multipointStr.split('((').join('(').split('),(').join(',').split('))').join(')')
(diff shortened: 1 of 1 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/component/input/geometry/input-geometry.view.js
```diff
@@ -22,8 +22,11 @@ var _ = require('underscore');
 function convertUserValueToWKT(val){
     val = val.split(' (').join('(').split(', ').join(',');
     val = val.split('MULTIPOINT').map(function(value, index){
-        if (index % 2 === 1 && value.indexOf('((') !== -1) {
-            return value.split('((').join('(').split('),(').join(',').split('))').join(')');
+        if (index % 2 === 1 && value.indexOf('((') === 0) {
```
millerw8: The problem still occurs if they choose to have two+ multipoints.  I believe that can be addressed by changing this line to  
```
if (index % 2 !== 0 && value.indexOf('((') === 0) {
```

andrewkfiedler: Yeah, that should work.  I'll update it to just check `value.indexOf('((') === 0`.

millerw8: yep, considered that as well.  The only difference is it would modify something string that led off with a `((` which should never happen.

millerw8:
build now

PR #2022, opened 2017-05-31

mweser:
#### What does this PR do?
Provides a generic utility class to page through catalog framework query results

@rzwiefel @ahoffer @AzGoalie @brendan-hofmann 

@coyotesqrl
@lessarderic

1. Build and run DDF instance
2. Ingest large number of files
3. Attempt export command such as `catalog:export -delete=true --cql "\"id\" like '*'"`
4. Verify that zip file contains all files sought

Looking to make class designed around ExportCommand to have more generic functionality to be used as a util class for other uses, as well as to provide paging of queries to avoid memory issues.

[](https://codice.atlassian.net/browse/DDF-3034)

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
@@ -11,7 +11,7 @@
  * is distributed along with this program and can be found at
  * <http://www.gnu.org/licenses/lgpl.html>.
  */
-package org.codice.ddf.commands.util;
+package ddf.catalog.util.impl;
 
 import java.io.Serializable;
 import java.util.HashMap;
@@ -21,55 +21,42 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Spliterator;
 import java.util.Spliterators;
-import java.util.function.Function;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+
+import org.apache.commons.lang3.Validate;
 
 import ddf.catalog.CatalogFramework;
 import ddf.catalog.data.Result;
 import ddf.catalog.federation.FederationException;
-import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.SourceResponse;
+import ddf.catalog.operation.impl.QueryRequestImpl;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
 
 /**
  * Effectively a cursor over the results of a filter that automatically pages through all results
  * <br/>
- * Throws a {@link CatalogCommandRuntimeException} if anything goes wrong during iteration or
+ * Throws a {@link CatalogRuntimeException} if anything goes wrong during iteration or
  * querying
  */
 public class QueryResultIterable implements Iterable<Result> {
     private final CatalogFramework catalog;
 
-    private final Function<Integer, QueryRequest> filter;
-
-    private final int pageSize;
+    private final QueryRequestImpl queryRequest;
 
     /**
-     * For paging through a single filter with a default pageSize of 64
+     * For paging through a single filter with a default PAGE_SIZE of 64
      *
-     * @param catalog catalog to query
-     * @param filter  A dynamic supplier of a filter that takes the current index such that
-     *                the caller can control iteration based on their own logic
+     * @param catalog      catalog to query
+     * @param queryRequest A filter to query by
      */
-    public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequest> filter) {
-        this(catalog, filter, 64);
-    }
+    public QueryResultIterable(CatalogFramework catalog, QueryRequestImpl queryRequest) {
+        Validate.notNull(catalog, "Catalog is null");
+        Validate.notNull(queryRequest, "queryRequest is null");
 
-    /**
-     * For paging through a single filter.
-     *
-     * @param catalog  catalog to query
-     * @param filter   A dynamic supplier of a filter that takes the current index such that
-     *                 the caller can control iteration based on their own logic
-     * @param pageSize How many results should each page hold
-     */
-    public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequest> filter,
-            int pageSize) {
         this.catalog = catalog;
-        this.filter = filter;
-        this.pageSize = pageSize;
+        this.queryRequest = queryRequest;
     }
 
     @Override
@@ -77,6 +64,7 @@ public class QueryResultIterable implements Iterable<Result> {
         return new ResultQueryIterator();
     }
 
+    // TODO: 6/7/17 Use Spliterator to implement paging design
     @Override
     public Spliterator<Result> spliterator() {
         int characteristics = Spliterator.DISTINCT;
@@ -87,8 +75,11 @@ public class QueryResultIterable implements Iterable<Result> {
         return StreamSupport.stream(this.spliterator(), false);
     }
 
+    // TODO: 6/5/17 Cover with unit test(s) through public API
     class ResultQueryIterator implements Iterator<Result> {
         private int pageIndex = 1;
+
+        private int missingResults = 0;
 
         private boolean finished = false;
 
@@ -96,11 +87,7 @@ public class QueryResultIterable implements Iterable<Result> {
 
         private Iterator<Result> results = null;
 
-        ResultQueryIterator() {
-            if (pageSize <= 0) {
-                this.finished = true;
-            }
-        }
+        private static final int PAGE_SIZE = 64;
 
         @Override
         public boolean hasNext() {
@@ -112,8 +99,9 @@ public class QueryResultIterable implements Iterable<Result> {
                 return false;
             }
 
-            pageIndex += pageSize;
-            queryNext(pageIndex);
+            pageIndex += (PAGE_SIZE - missingResults);
+            missingResults = 0;
+            queryNext();
             return hasNext();
         }
 
@@ -126,29 +114,34 @@ public class QueryResultIterable implements Iterable<Result> {
             return results.next();
         }
 
-        private void queryNext(int index) {
+        private void queryNext() {
             try {
                 Map<String, Serializable> props = new HashMap<>();
                 // Avoid caching all results while dumping with native query mode
                 props.put("mode", "native");
-                response = catalog.query(filter.apply(index));
+                response = catalog.query(queryRequest);
+                missingResults = PAGE_SIZE - response.getResults()
+                        .size();
+
             } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
-                throw new CatalogCommandRuntimeException(e);
+                throw new CatalogRuntimeException(e);
             }
             List<Result> queryResults = response.getResults();
             this.results = queryResults.iterator();
 
             int size = queryResults.size();
-            if (size == 0 || size < pageSize) {
+            if (size == 0 || size < PAGE_SIZE) {
+                // TODO: 6/5/17 Address edge cases of this size comparison
+
                 finished = true;
             }
         }
 
-        private void ensureInitialized() throws CatalogCommandRuntimeException {
+        private void ensureInitialized() {
             if (response != null || results != null) {
                 return;
             }
-            queryNext(pageIndex);
+            queryNext();
         }
     }
 }
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
@@ -70,9 +70,9 @@ public class QueryResultIterable implements Iterable<Result> {
 //        return Spliterators.spliteratorUnknownSize(this.iterator(), characteristics);
 
         // TODO: 6/9/17 Add bean to avoid 'new' keyword
-        BatchedSpliterator batchedSpliterator = new BatchedSpliterator();
+        QueryResultSpliterator queryResultSpliterator = new QueryResultSpliterator();
 
-        return batchedSpliterator;
+        return queryResultSpliterator;
     }
 
     public Spliterator<Result> spliterator(int size) {
@@ -80,9 +80,9 @@ public class QueryResultIterable implements Iterable<Result> {
 //        return Spliterators.spliteratorUnknownSize(this.iterator(), characteristics);
 
         // TODO: 6/9/17 Add bean to avoid 'new' keyword
-        BatchedSpliterator batchedSpliterator = new BatchedSpliterator();
+        QueryResultSpliterator queryResultSpliterator = new QueryResultSpliterator();
 
-        return batchedSpliterator;
+        return queryResultSpliterator;
     }
 
     public Stream<Result> stream() {
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
@@ -63,7 +63,7 @@ public class QueryResultIterable implements Iterable<Result> {
         return new ResultQueryIterator();
     }
 
-    // TODO: 6/7/17 Use Spliterator to implement paging design
+    // TODO: 6/9/17 Determine where this is called and if already implicitly called
     @Override
     public Spliterator<Result> spliterator() {
 //        int characteristics = Spliterator.DISTINCT;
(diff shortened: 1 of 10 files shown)


mweser:
Export command successfully runs in running instance and successfully exports using paging.

mweser:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -131,9 +136,17 @@ private void queryNext(int index) {
                 Map<String, Serializable> props = new HashMap<>();
                 // Avoid caching all results while dumping with native query mode
                 props.put("mode", "native");
-                response = catalog.query(filter.apply(index));
+                response = catalog.query(new QueryRequestImpl(new QueryImpl(filter,
+                        pageIndex,
+                        pageSize,
+                        SortBy.NATURAL_ORDER,
```
lessarderic: ❗️ We should provide the ability to set the sort order.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -131,9 +136,17 @@ private void queryNext(int index) {
...
+                        false,
+                        5000l/*timeout? config? passed in?*/)));
```
lessarderic: ❓ Should we add a constant for this, but still allow clients to override the default value in the constructor?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -131,9 +136,17 @@ private void queryNext(int index) {
...
+                missingResults = pageSize - response.getResults()
+                        .size();
+
             } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
-                throw new CatalogCommandRuntimeException(e);
+                throw new CatalogRuntimeException(e);
             }
             List<Result> queryResults = response.getResults();
             this.results = queryResults.iterator();
```
lessarderic: ❗️ We can't just check if `size < pageSize` since it's possible that `size` is smaller because that's all the catalog framework will return, and not because there's nothing else.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -131,9 +136,17 @@ private void queryNext(int index) {
...
+                        pageSize,
+                        SortBy.NATURAL_ORDER,
+                        false,
+                        5000l/*timeout? config? passed in?*/)));
+                missingResults = pageSize - response.getResults()
+                        .size();
+
             } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
-                throw new CatalogCommandRuntimeException(e);
+                throw new CatalogRuntimeException(e);
```
lessarderic: ❓ Should we hide all those exceptions? If we're hoping to have most of the code use this class instead of `query` method, we may not want to change the error reporting too much. Thoughts?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -144,7 +157,7 @@ private void queryNext(int index) {
             }
         }
 
-        private void ensureInitialized() throws CatalogCommandRuntimeException {
+        private void ensureInitialized() throws CatalogRuntimeException {
```
brendan-hofmann: This is cool! :D

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -53,7 +56,7 @@
      * @param filter  A dynamic supplier of a filter that takes the current index such that
```
mweser: Which part of comments need changing? It appears that there weren't any param comments for the removed Function parameter.

rzwiefel: It's not a dynamic supplier of filter that takes an index, just the filter now.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: 🎉

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -21,28 +21,31 @@
...
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
 
 /**
  * Effectively a cursor over the results of a filter that automatically pages through all results
  * <br/>
- * Throws a {@link CatalogCommandRuntimeException} if anything goes wrong during iteration or
+ * Throws a {@link CatalogRuntimeException} if anything goes wrong during iteration or
  * querying
  */
```
rzwiefel: I don't believe many iterators are threadsafe , but it wouldn't hurt to put it explicitly. We could also mention this (since it is an iterator) is a stateful, mutable object.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -65,8 +68,7 @@ public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequ
      *                 the caller can control iteration based on their own logic
      * @param pageSize How many results should each page hold
      */
-    public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequest> filter,
-            int pageSize) {
+    public QueryResultIterable(CatalogFramework catalog, Filter filter, int pageSize) {
         this.catalog = catalog;
```
mweser: Should we allow for pageSize == -1 as valid?

mweser: And in case of invalid parameters, is throwing an exception acceptable?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -65,8 +68,7 @@ public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequ
      *                 the caller can control iteration based on their own logic
      * @param pageSize How many results should each page hold
      */
-    public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequest> filter,
-            int pageSize) {
+    public QueryResultIterable(CatalogFramework catalog, Filter filter, int pageSize) {
```
rzwiefel: I like the idea of being able to specify a guaranteed result-size (it would be good to not over-query for too many results if they only want very something that is 1 result over a new query boundry).
Though, Why do you think it would be better split into two separate classes instead of just an overloaded ctor for this that either allows them to specify max and then the other can hide the page-size/result-size all together.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -131,9 +136,17 @@ private void queryNext(int index) {
                 Map<String, Serializable> props = new HashMap<>();
```
mweser: Found it belongs in QueryRequestImpl but we were using an overloaded method which didn't require it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
mweser: Would a setter method be appropriate for this? I have a class field set up.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
mweser: Would a setter method be appropriate for this? I have a class field set up.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: We can try to combine them, but if we want an API that can return an `Iterator`/`Stream` of individual `Result` objects, and one that can return batches of `Result` objects, I'm not sure how clear the API will be. The class could always return `Iterator<List<Result>>` but that feels awkward when the client doesn't need batching.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: We may want to get it as a constructor argument to prevent clients from changing it while the instance is being used and get undefined behaviors.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
rzwiefel: Ahhhhh right understood. It's a bummer we can't implement that batching of results somehow under the covers of the actual `getResults` method, which is what i would have expected it do in the first place.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: @mweser, like we discussed, I don't think the page size makes sense here. It only makes sense if we have a separate class returning batched results, in which case the argument should probably be called `batchSize` instead and have to be `>= 1`.

For this class, we probably should just get rid of the `pageSize` argument altogether.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: Constant for the default value and a setter to change it sounds like a good way to go here.

mweser:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
mweser: And by "mention", do we just mean a comment in the existing javadoc?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
mweser: @coyotesqrl @rzwiefel @brendan-hofmann ?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -90,6 +92,8 @@ public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequ
     class ResultQueryIterator implements Iterator<Result> {
         private int pageIndex = 1;
 
+        private int missingResults = 0;
+
         private boolean finished = false;
 
         private SourceResponse response = null;
```
mweser: @rzwiefel What does this check even do? It's not quite clear why it's included.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -90,6 +92,8 @@ public QueryResultIterable(CatalogFramework catalog, Function<Integer, QueryRequ
     class ResultQueryIterator implements Iterator<Result> {
```
mweser: Doesn't quite seem like we can make it static as it relies on pageSize and other state fields. Thoughts on how to fix this or refactor to make it form a static class?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: Yep.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: Agree, we can't make it `static` as it is. We should be able to determine what this `Iterator` really needs and pass that in to its constructor instead of sharing the member variables between the two classes.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: Actually, now that I think of it, the `QueryResultIterable` _is_ thread safe, i.e., it is currently immutable and a client can use it to create as many `Iterator`s or `Stream`s as it wants and it will keep working. The `ResultQueryIterator` is actually the one that isn't thread safe.

Sooo... maybe we should remove the _Not thread safe_ comment for now. Sorry about that 😕

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
rzwiefel: I would rather throw the exception and simply hide it explicitly where we use this to replace things than hide them in here.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
rzwiefel: @mweser Which check where?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -131,9 +136,17 @@ private void queryNext(int index) {
                 Map<String, Serializable> props = new HashMap<>();
                 // Avoid caching all results while dumping with native query mode
                 props.put("mode", "native");
-                response = catalog.query(filter.apply(index));
+                response = catalog.query(new QueryRequestImpl(new QueryImpl(filter,
+                        pageIndex,
+                        pageSize,
```
rzwiefel: @lessarderic Couldn't using -1 affect compatibility with legacy systems? Would it be safer to just query with a concrete page size? (as apposed to assuming the implementation will set it to some cap like ours now does)

lessarderic: Good question. `-1` used to mean get everything and was actually caped by some sources and endpoints when necessary. So it could technically cause some OOM errors and such when interfacing with endpoints that do not limit that number (e.g., CSW), but that's already a problem today.

Having said that, we could use a constant number here to prevent potential problems. The flip side though is that if the system property used to cap the number of metacards is set higher than this number, say on a system that stores small metacards, then this code will not take full advantage of that maximum page size.

I guess as long as we make this number high enough and don't expose it to the clients of this class, I'm ok either way.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: So @rzwiefel, your vote is no catch and rethrow here? Just let the three exceptions bubble up?

rzwiefel: Certainly. At least for the errors that pop up from the catalog framework. Would much rather not hide those unless anyone has strong feelings for hiding them. Especially if we will already have to modify code to replace instances with this, we should also make sure we properly deal with exceptions.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
rzwiefel: I would prefer to try and use an abitrary number and have maybe be not quite as efficient than to use -1 if the risk of breaking things is even a remote possibility.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -20,87 +20,77 @@
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Spliterator;
-import java.util.Spliterators;
-import java.util.function.Function;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import org.apache.commons.lang3.Validate;
```
brendan-hofmann: `Validate` is in commons-lang 2.6, we don't need to embed lang3 here.

PR #2029, opened 2017-06-05
DDF-3050 Added component tests for directory monitor

AzGoalie:
#### What does this PR do?
Adds component tests for the Content Directory Monitor to test the interactions between the Content Directory Monitor, Camel Context, Camel Component and Catalog Framework.
@brendan-hofmann @jhunzic @ahoffer 
@lessarderic [Test](https://github.com/orgs/codice/teams/test)
@coyotesqrl
@pklinef
https://codice.atlassian.net/projects/DDF/issues/DDF-3050
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests


catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -110,6 +117,159 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.ops4j.pax.exam</groupId>
+            <artifactId>pax-exam</artifactId>
+            <version>${pax.exam.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.ops4j.pax.url</groupId>
+            <artifactId>pax-url-aether</artifactId>
```
coyotesqrl: I can't tell from here...are you getting the version from a parent pom or is this dependency just missing its version?

catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -110,6 +117,159 @@
...
+        <dependency>
+            <groupId>ddf.distribution</groupId>
+            <artifactId>ddf-branding-plugin</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ddf.mime.core</groupId>
+            <artifactId>mime-core-impl</artifactId>
+            <scope>test</scope>
```
coyotesqrl: Same: inheriting the version or missed?

catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -110,6 +117,159 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.tika</groupId>
+            <artifactId>tika-core</artifactId>
+            <scope>test</scope>
```
coyotesqrl: Same: inheriting the version or missed?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorComponentTest.java
```diff
@@ -0,0 +1,308 @@
...
+    public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+    private String directoryPath;
+
+    private CatalogFramework catalogFramework;
+
+    @Configuration
+    public Option[] config() throws IOException {
+
+        return options(testDependencies(),
```
coyotesqrl: Does it make sense to extract out an abstract base class (to live somewhere I can't currently think of) as we have an abstract base for our itests? I would think we'll see a lot of commonality between different component tests and it's probably not too early to make this optimization to avoid refactoring later.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorComponentTest.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private Option initContentDirectoryMonitorConfig() throws IOException {
+        return editConfigurationFilePut(
+                "etc/org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor-test.cfg",
+                "monitoredDirectoryPath",
+                "");
+    }
+
+    private Option contentDirectoryMonitorDependencies() {
```
coyotesqrl: Even in the case of this, very specific method, with a base class you could yank all the bundle starting knowledge out of the individual component test classes and pass in a list of tuples of group/artifact.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/features/CamelFeatures.java
```diff
@@ -0,0 +1,135 @@
...
+import static org.ops4j.pax.exam.CoreOptions.maven;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.features;
+
+import java.util.Arrays;
+
+import org.ops4j.pax.exam.Option;
+
+public class CamelFeatures {
+    public enum CamelFeature {
+        XML_SPECS_API("xml-specs-api"), CAMEL("camel"), CAMEL_CORE("camel-core"), CAMEL_CATALOG(
```
coyotesqrl: It doesn't appear this enum's values are used anywhere and that the only use of the enum at all is in the `camelFeatures` method where you iterate over it. Could probably just use a `final static String[]` or a comma-delimited String in a resource bundle to simplify.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/features/CatalogAppFeatures.java
```diff
@@ -0,0 +1,92 @@
...
+
+import static org.ops4j.pax.exam.CoreOptions.maven;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.features;
+
+import java.util.Arrays;
+
+import org.ops4j.pax.exam.Option;
+
+public class CatalogAppFeatures {
+    public enum CatalogFeature {
```
coyotesqrl: Same comment about a static array or resource bundle.

As it appears this model is going to be duplicated a lot here, a single resource bundle with several properties of comma-delimited strings would probably be the simplest solution.

If these are things that might be common across many component tests, then I would suggest pulling them out to a resource to be loaded by that abstract class I was mentioning might be worthwhile.

You could maintain a set of consts in the abstract class that the inheritors could reference; those consts would then be lookup keys into the resource bundle.

catalog/core/catalog-core-directorymonitor/pom.xml
AzGoalie: This one was missing, good catch

catalog/core/catalog-core-directorymonitor/pom.xml
AzGoalie: It's being inherited.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/features/CamelFeatures.java
AzGoalie: The enums were generated out of the camel feature file. I thought it would be nice to have all features of camel available if we needed them in other component tests. Reasoning for an enum over a string array was for compile time checks to make sure we were not trying to start feature that wasn't available in the camel feature file.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/features/CatalogAppFeatures.java
AzGoalie: Same comment as above, but I could try refactoring out the core logic.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorComponentTest.java
AzGoalie: Originally i tried to have an abstract base class but it was failing to configure correctly. I don't remember why, but i can look into it again.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorComponentTest.java
AzGoalie: Agreed.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorComponentTest.java
```diff
@@ -0,0 +1,272 @@
...
+import ddf.catalog.content.data.ContentItem;
+import ddf.catalog.content.operation.CreateStorageRequest;
+import ddf.catalog.operation.CreateResponse;
+import ddf.catalog.source.IngestException;
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.security.service.SecurityManager;
+
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerClass.class)
+public class ContentDirectoryMonitorComponentTest extends AbstractComponentTest {
```
lessarderic: ❓ Should we use failsafe instead of surefire to run component tests? That may be a more logical phase to run them in. If we do, we should also use `IT` as the test class suffix instead of `Test`.

AzGoalie:
I'm not sure how to use the CXF local transport. I'm not even sure what requires CXF to begin with.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,272 @@
...
+
+                new BundleInfo("com.google.guava", "guava"),
+                new BundleInfo("org.apache.tika", "tika-core"),
+
+                new BundleInfo("org.apache.shiro", "shiro-core"),
+                new BundleInfo("javax.servlet", "javax.servlet-api"),
+                new BundleInfo("javax.validation", "validation-api"),
+                new BundleInfo("org.bouncycastle", "bcprov-jdk15on"),
+
+                new BundleInfo("ddf.platform.util", "util-uuidgenerator-api"),
```
lessarderic: ❓ Should we try to clean up the UUID generator feature definition so we don't have to import the individual bundles?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,272 @@
...
+                new BundleInfo("ddf.platform.util", "util-uuidgenerator-impl"),
+                new BundleInfo("ddf.catalog.core", "catalog-core-camelcontext"),
+                new BundleInfo("ddf.platform.api", "platform-api"),
```
lessarderic: ❓ Any reason we can't just pull in the `platform-api` feature from the feature file?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,272 @@
...
+                new BundleInfo("org.codice.thirdparty", "gt-opengis"),
+                new BundleInfo("ddf.catalog.core", "catalog-core-api"),
+                new BundleInfo("ddf.security.core", "security-core-api"),
```
lessarderic: ❓ Do we absolutely need to include the security framework in these tests? If this component doesn't do anything special security-wise, we should be able to safely assume security works and not have to pull it in.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,272 @@
...
+
+    @Rule
+    public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+    private String directoryPath;
+
+    private CatalogFramework catalogFramework;
+
+    @Override
+    protected Option setupDistribution() {
```
lessarderic: ✏️  We should keep all the `public` methods at the top of the class instead of having the `protected` and `private` ones. Makes it easier to see what the tests are without having to scroll way down.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KarafConfigurator.java
```diff
@@ -0,0 +1,61 @@
...
+                        .useDeployFolder(false),
+                configureConsole().ignoreLocalConsole(),
+                logLevel().logLevel(LogLevelOption.LogLevel.WARN),
+                setSystemProperties(),
+                when(Boolean.getBoolean("isDebugEnabled")).useOptions(debugConfiguration()),
+                karafStandardFeatures(STANDARD));
+    }
+
+    private static Option setSystemProperties() {
+        return composite(systemProperty("ddf.home").value("${karaf.home}"),
```
lessarderic: ❓ Is it better to change the configuration file instead:

```
editConfigurationFilePut("etc/system.properties", "ddf.home", "${karaf.home}"),
editConfigurationFilePut("etc/system.properties", "org.codice.ddf.system.hostname", "localhost")
```

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KeystoreTruststoreConfigurator.java
```diff
@@ -0,0 +1,92 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+
+import org.apache.commons.io.IOUtils;
+import org.ops4j.pax.exam.Option;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KeystoreTruststoreConfigurator {
```
lessarderic: ❓ See my question about security above. If we can ignore security, we should be able to get rid of this class and simplify the component test configuration and maintenance.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/features/CamelFeatures.java
lessarderic: Not sure I'm a fan of those big enum files either. I'm thinking that most of the time, components will only need to pull in one or two specific dependencies from existing third-party feature files and using simple feature names should be good enough. Maintaining those enums in that case seems somewhat overkill imho.

Since this code is local to this component test, i'd be ok with leaving it for now until I'm done with my own component test. Then we can compare and start re-thinking the approach and move things to a common component test module that everyone can use.

Thoughts?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,272 @@
...
+            throws IOException, SourceUnavailableException, IngestException {
+        updateContentDirectoryMonitor(directoryPath, ContentDirectoryMonitor.DELETE);
+
+        File file = createTestFile(directoryPath);
+        File directory = Paths.get(directoryPath)
+                .toFile();
+
+        await("file deleted").atMost(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS)
+                .until(() -> !file.exists());
+
```
coyotesqrl: ✏️ Probably be good to assert that the file is also not in the movedPath here.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
AzGoalie: The directory monitor imports a bunch of security packages and they must be available for the bundle to start. We could make a mock bundle that exports all the packages that the content directory monitor requires.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KarafConfigurator.java
AzGoalie: Done

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
AzGoalie: I did that with asserting that the directory is empty with the assert below this.

AzGoalie:
build now

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KeystoreTruststoreConfigurator.java
AzGoalie: I don't think there's any way to remove this class, we embed security commons in the content directory monitor bundle which expects to be able to find these keystore/truestore files.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/AbstractComponentTest.java
```diff
@@ -0,0 +1,74 @@
...
+        return mavenBundle(bundle.groupId, bundle.artifactId).versionAsInProject()
+                .start();
+    }
+
+    protected <T> void registerService(T service, Class<T> registerClass) {
+        ServiceRegistration<T> registration = bundleContext.registerService(registerClass,
+                service,
+                null);
+        serviceRegistrations.add(registration);
+        await("service registration").atMost(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS)
```
lessarderic: ❓ As far as I could find service registration is a synchronous call so we shouldn't need to wait here. Did you find otherwise or did you run into any issues?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KeystoreTruststoreConfigurator.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public static Option createKeystoreAndTruststore(InputStream keystore, InputStream truststore) {
+        String keystorePath = copyKeystoreAndGetPath(keystore);
+        String truststorePath = copyTruststoreAndGetPath(truststore);
+
+        return setupSystemProperties(keystorePath, truststorePath);
+    }
+
+    private static Option setupSystemProperties(String keystorePath, String truststorePath) {
+        return composite(systemProperty(KEYSTORE_TYPE).value("jks"),
```
lessarderic: ❓ Should we use `editConfigurationFilePut()` like we did for the other system properties? It makes it easier to troubleshoot problems when saving off the exam folder as the values get written out to a file instead of being stored only in memory.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KeystoreTruststoreConfigurator.java
```diff
@@ -0,0 +1,92 @@
...
+            return createTempFile("truststore", truststore);
+        } catch (IOException e) {
+            LOG.error("Failed to copy and create truststore.jks", e);
+        }
+
+        return "";
+    }
+
+    private static String createTempFile(String filename, InputStream contents) throws IOException {
+        File tempFile = Files.createTempFile(filename, ".jks")
```
lessarderic: ❓ Instead of putting those files in a temporary directory, should we use `KarafDistributionOption .replaceConfigurationFile()` and put them under `etc/keystores`? That way they would still be available for troubleshooting if the exam folder is kept and they would automatically get deleted if not.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,264 @@
...
+        assertThat(directory.list().length, is(0));
+    }
+
+    @Override
```
lessarderic: ❓ Not sure how much coverage we want to get out of those tests, but shouldn't be testing these use cases:
* Bad files and bad file extensions options?
* Creating a file in a sub-directory (recursive watch)?
* Test create, rename and delete for all three configurations, i.e., in-place, move and delete?
* Files don't get processed anymore after feature has been stopped?
* Files already in the directory to monitor are actually picked up on startup and after a restart?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -0,0 +1,264 @@
...
+        File file = createTestFile(directoryPath);
+
+        ArgumentCaptor<CreateStorageRequest> createStorageRequest = ArgumentCaptor.forClass(
+                CreateStorageRequest.class);
+        verify(catalogFramework).create(createStorageRequest.capture());
+        ContentItem result = createStorageRequest.getValue()
+                .getContentItems()
+                .get(0);
+
+        assertThat(result.getFilename(), is(file.getName()));
```
lessarderic: ❗️ Since our goal here is to verify that the directory monitor always creates the proper content item, we should verify _all_ the `ContentItem` fields, not just the file name.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KeystoreTruststoreConfigurator.java
AzGoalie: Done

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/configurators/KeystoreTruststoreConfigurator.java
AzGoalie: Done

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
AzGoalie: This was to get a start on creating component tests so we could test the web dav component being added. We can add more tests at a later time.

AzGoalie:
build now

shaundmorris:
Merging, last build was successful

PR #2039, opened 2017-06-07
DDF-3060 changed to extend jetty's impl instead of the abstract class to avoid a hack in pax web

stustison:
… to avoid a hack in pax web

@jlcsmith @beyelerb @millerw8 
@coyotesqrl
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/platform/platform-paxweb-jettyconfig/src/main/java/org/codice/ddf/security/session/AttributeSharingHashSessionIdManager.java
+++ b/platform/platform-paxweb-jettyconfig/src/main/java/org/codice/ddf/security/session/AttributeSharingHashSessionIdManager.java
@@ -47,22 +47,23 @@ import javax.servlet.http.HttpSession;
 
 import org.codice.ddf.platform.util.RandomNumberGenerator;
 import org.eclipse.jetty.server.session.AbstractSession;
-import org.eclipse.jetty.server.session.AbstractSessionIdManager;
+import org.eclipse.jetty.server.session.HashSessionIdManager;
 
 /**
  * Custom implementation of the {@link org.eclipse.jetty.server.SessionIdManager} that shares session
  * data between sessions in a cluster.
  */
-public class HashSessionIdManager extends AbstractSessionIdManager {
+public class AttributeSharingHashSessionIdManager extends HashSessionIdManager {
+    //changed this to extend Jetty's impl instead of the abstract class because of a hack in pax web
 
     private final Map<String, Set<WeakReference<HttpSession>>> sessions =
             new ConcurrentHashMap<>();
 
-    public HashSessionIdManager() {
+    public AttributeSharingHashSessionIdManager() {
         super(RandomNumberGenerator.create());
     }
 
-    public HashSessionIdManager(Random random) {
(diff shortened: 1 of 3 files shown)


stustison:
build now

stustison:
build now

stustison:
build now

platform/platform-paxweb-jettyconfig/src/main/java/org/codice/ddf/security/session/HashSessionIdManager.java
```diff
@@ -47,13 +47,14 @@
...
 import org.eclipse.jetty.server.session.AbstractSession;
-import org.eclipse.jetty.server.session.AbstractSessionIdManager;
 
 /**
  * Custom implementation of the {@link org.eclipse.jetty.server.SessionIdManager} that shares session
  * data between sessions in a cluster.
  */
-public class HashSessionIdManager extends AbstractSessionIdManager {
+@edu.umd.cs.findbugs.annotations.SuppressWarnings(value = "NM_SAME_SIMPLE_NAME_AS_SUPERCLASS", justification = "not changing")
+public class HashSessionIdManager extends org.eclipse.jetty.server.session.HashSessionIdManager {
```
coyotesqrl: Sonarqube doesn't like it either. :/

Is a class rename out of the question? If not, we'll need to mark this as a non-issue for Sonarqube as well.

platform/platform-paxweb-jettyconfig/src/main/java/org/codice/ddf/security/session/HashSessionIdManager.java
```diff
@@ -47,13 +47,14 @@
...
+    //changed this to extend Jetty's impl instead of the abstract class because of a hack in pax web
```
millerw8: I'm a little confused, this seems to be going from a jetty abstract class to a jetty class. Where does the pax web hack come in?  Also, What's the difference between this class and the jetty version of HashSessionIdManager.  It seems we override all of its functions anyway, so the fact that we are using it over AbstractSessionIdManager doesn't change anything.  And that class seems to be the same as this class except this uses a concurrentHashMap and that class uses synchronized everywhere.

stustison: Pax Web is expecting an instance of the Jetty implementation, which ours was not. That ended up skipping a bunch of code that we actually need. There are also additional changes to our id manager that aren't in the Jetty Impl. This is what we use to hook in SSO.

platform/platform-paxweb-jettyconfig/src/main/java/org/codice/ddf/security/session/HashSessionIdManager.java
stustison: I really had no desire to change the class name.

stustison:
I just changed the name because I was tired of fighting all the static analysis tools.

stustison:
build now

PR #2040, opened 2017-06-07
DDF-3061 synchronized the session renewal and added a cache to return the last returned token

stustison:
@jlcsmith @beyelerb @millerw8 
@coyotesqrl
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
+++ b/platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import javax.xml.stream.XMLStreamException;
 
@@ -56,6 +57,8 @@ import org.w3c.dom.ls.DOMImplementationLS;
 import org.w3c.dom.ls.LSSerializer;
 
 import com.google.common.base.Splitter;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 
 import ddf.security.PropertiesLoader;
 import ddf.security.assertion.SecurityAssertion;
@@ -109,6 +112,10 @@ public abstract class AbstractStsRealm extends AuthenticatingRealm
     private String keySize = null;
 
     private Boolean useKey = null;
+
+    private Cache<Element, SecurityToken> cache = CacheBuilder.newBuilder()
+            .expireAfterAccess(1, TimeUnit.MINUTES)
+            .build();
 
     public AbstractStsRealm() {
         this.bus = getBus();
@@ -237,32 +244,39 @@ public abstract class AbstractStsRealm extends AuthenticatingRealm
      * @param securityToken The token being renewed.
      * @return security token (SAML assertion)
      */
-    protected SecurityToken renewSecurityToken(SecurityToken securityToken) {
-        SecurityToken token = null;
+    protected SecurityToken renewSecurityToken(final SecurityToken securityToken) {
         String stsAddress = getAddress();
 
         try {
             LOGGER.debug("Renewing security token from STS at: {}.", stsAddress);
 
             if (securityToken != null) {
-                LOGGER.debug("Telling the STS to renew a security token on behalf of the auth token");
-                STSClient stsClient = configureStsClient();
+                synchronized (securityToken.getToken()) {
+                    return cache.get(securityToken.getToken(), () -> {
+                        LOGGER.debug(
+                                "Telling the STS to renew a security token on behalf of the auth token");
+                        STSClient stsClient = configureStsClient();
 
-                stsClient.setWsdlLocation(stsAddress);
-                stsClient.setTokenType(getAssertionType());
-                stsClient.setKeyType(getKeyType());
-                stsClient.setKeySize(Integer.parseInt(getKeySize()));
-                stsClient.setAllowRenewing(true);
-                token = stsClient.renewSecurityToken(securityToken);
-                LOGGER.debug("Finished renewing security token.");
+                        stsClient.setWsdlLocation(stsAddress);
+                        stsClient.setTokenType(getAssertionType());
+                        stsClient.setKeyType(getKeyType());
+                        stsClient.setKeySize(Integer.parseInt(getKeySize()));
+                        stsClient.setAllowRenewing(true);
+                        SecurityToken token = stsClient.renewSecurityToken(securityToken);
+                        cache.put(securityToken.getToken(), token);
+                        LOGGER.debug("Finished renewing security token.");
+
+                        return token;
+                    });
+                }
+            } else {
+                return null;
             }
         } catch (Exception e) {
             String msg = "Error renewing the security token from STS at: " + stsAddress + ".";
             LOGGER.debug(msg, e);
             throw new AuthenticationException(msg, e);
         }
-
-        return token;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stustison:
build now

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
```diff
@@ -110,6 +113,10 @@
 
     private Boolean useKey = null;
 
+    private Cache<Element, SecurityToken> cache = CacheBuilder.newBuilder()
+            .expireAfterWrite(1, TimeUnit.MINUTES)
```
rzwiefel: ❓ Will no session ever have a life of shorter than 1 minute (such that there might be a bug with caching them for up to a minute after creation) ?

stustison: You cannot have a session with a life shorter than 1 minute. The trigger for us to renew a session is at the 1 minute mark.

stustison:
build now

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
rzwiefel: Should this cache be slightly less than one minute to help prevent any possible race conditions then if one *were* to set it at a minute? (I understand here this is just a possibility, not the norm)

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
```diff
@@ -237,24 +244,32 @@ protected SecurityToken requestSecurityToken(Object authToken) {
...
-
-                stsClient.setWsdlLocation(stsAddress);
-                stsClient.setTokenType(getAssertionType());
-                stsClient.setKeyType(getKeyType());
-                stsClient.setKeySize(Integer.parseInt(getKeySize()));
-                stsClient.setAllowRenewing(true);
-                token = stsClient.renewSecurityToken(securityToken);
-                LOGGER.debug("Finished renewing security token.");
+                synchronized (securityToken.getToken()) {
+                    SecurityToken cachedToken = cache.getIfPresent(securityToken.getToken());
```
coyotesqrl: I think you should change the cache builder to `expireAfterAccess` instead of write. Otherwise, you could get the token here at 0:59 and return it but it would be evicted just after returning it (when the lock is released). Unless the semantics you want are for it potentially to be evicted if the timing worked out that way.

_Note: I'm not sure that Guava's cache tries to lock on the key; it's possible it would evict it before you returned. But either case seems wrong to me._

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
```diff
@@ -684,10 +699,9 @@ public void setUseKey(Boolean useKey) {
         @Override
         public boolean doCredentialsMatch(AuthenticationToken token, AuthenticationInfo info) {
             if (token instanceof SAMLAuthenticationToken) {
-                SecurityToken oldToken = (SecurityToken) token.getCredentials();
-                SecurityToken newToken = (SecurityToken) info.getCredentials();
-                return oldToken.getId()
-                        .equals(newToken.getId());
+                Object oldToken = token.getCredentials();
```
coyotesqrl: ❓ Does it matter in this case if `info` is not a `SAMLAuthenticationToken`?

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
rzwiefel: But `expireAfterAccess` means if it was continually accessed (intervals less than the expiry time) it would continue to be returned after actual assertion expires even.

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
```diff
@@ -237,24 +244,32 @@ protected SecurityToken requestSecurityToken(Object authToken) {
...
-                STSClient stsClient = configureStsClient();
-
-                stsClient.setWsdlLocation(stsAddress);
-                stsClient.setTokenType(getAssertionType());
-                stsClient.setKeyType(getKeyType());
-                stsClient.setKeySize(Integer.parseInt(getKeySize()));
-                stsClient.setAllowRenewing(true);
-                token = stsClient.renewSecurityToken(securityToken);
-                LOGGER.debug("Finished renewing security token.");
+                synchronized (securityToken.getToken()) {
```
lessarderic: ❓ Is it safe to synchronize on the object returned by `getToken`? Are we always guaranteed that it will always the same instance?

stustison: We are, but that knowledge is an implementation detail. Unfortunately, if I synchronize on anything else, I'll single thread all renews from all users, which is not at all what we want.

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
stustison: Nope, that is fine.

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
stustison: I changed to `expireAfterAccess`. Ideally this could only occur if the system was under a huge amount of strain. You'd need X number of requests to all come in at the same time, the first one gets through and then the remaining X-1 requests come in staggered by some large amount of time. Any other requests coming in to the system would already get the new token that the first request returned and wouldn't hit this logic anyways. This way it provides for about a minute between each of the requests before finally tossing it out of the cache.

stustison: I think the most common case is that all X of the requests will hit this synchronized block at roughly the same time and 1 will win.

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
```diff
@@ -237,24 +244,32 @@ protected SecurityToken requestSecurityToken(Object authToken) {
      * @param securityToken The token being renewed.
      * @return security token (SAML assertion)
      */
-    protected SecurityToken renewSecurityToken(SecurityToken securityToken) {
+    protected SecurityToken renewSecurityToken(final SecurityToken securityToken) {
```
coyotesqrl: A little cleaner version of this method is here: https://gist.github.com/coyotesqrl/f4ffa0c585fabd6500241c91ef2dc54b

Either works, but this takes advantage of the cache's functionality a bit more.

platform/security/sts/security-sts-realm/src/main/java/ddf/security/realm/sts/AbstractStsRealm.java
```diff
@@ -684,10 +699,9 @@ public void setUseKey(Boolean useKey) {
         @Override
         public boolean doCredentialsMatch(AuthenticationToken token, AuthenticationInfo info) {
             if (token instanceof SAMLAuthenticationToken) {
-                SecurityToken oldToken = (SecurityToken) token.getCredentials();
-                SecurityToken newToken = (SecurityToken) info.getCredentials();
-                return oldToken.getId()
-                        .equals(newToken.getId());
+                Object oldToken = token.getCredentials();
+                Object newToken = info.getCredentials();
+                return oldToken.equals(newToken);
```
millerw8: So the thought here is we can just use Object.equals because we would have pulled oldToken from the cache?

stustison: Yeah, it really is the same object for the renewal case, which is what this section deals with. The renewal itself was constantly throwing an exception due to the fact that multiple requests got there at the same exact time. This particular code was throwing a null pointer exception once I fixed the concurrent requests.

stustison:
build now

PR #2042, opened 2017-06-07
DDF-3062 Adds null check for application name regex

vinamartin:
#### What does this PR do?
Adds in a null check for the application name regex in the case where an application neither has a version or a version in its name. The `.match` method returns null when the regex isn't matched. 
This isn't currently an issue with any of our apps, but it could become an issue in the future. 

@djblue @andrewkfiedler 

@clockard
@lessarderic

Run through the installer and make sure everything installs properly

This is related to work being done for the Karaf 4.1.1 upgrade.

### DIFF
--- a/platform/admin/ui/src/main/webapp/js/models/Applications.js
+++ b/platform/admin/ui/src/main/webapp/js/models/Applications.js
@@ -132,7 +132,7 @@ define([
             changeObj.displayName = this.get('name');
             if (this.get('version') === '0.0.0') {
                 var matches = this.get('name').match(versionRegex);
-                if (matches.length === 3) {
+                if (matches !== null && matches.length === 3) {
                     changeObj.displayName = matches[1];
                     changeObj.version = matches[2];
                 }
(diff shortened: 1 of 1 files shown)


vinamartin:
build now

platform/admin/ui/src/main/webapp/js/models/Applications.js
```diff
@@ -132,7 +132,7 @@ define([
             changeObj.displayName = this.get('name');
             if (this.get('version') === '0.0.0') {
                 var matches = this.get('name').match(versionRegex);
-                if (matches.length === 3) {
+                if (matches !== null && matches.length === 3) {
```
djblue: I think you can also remove the length check. If the match isn't null, then all the groups in the regex were found, so it will always be of length three.

vinamartin: Ah yeah good point

vinamartin:
build now

PR #2051, opened 2017-06-10
DDF-3070 Allow the ability to logout if currently guest.

pklinef:
#### What does this PR do?
- Fixed bug that would not allow logging out if guest.
- Fixed IdP logout action provider to use the user’s display name
  instead of their full name id.
- Fixed logout confirmation to URI decode the response message.

@brjeter 
@harrison-tarr 

@lessarderic
@stustison

Login as guest.  Go to `https://localhost:8993/logout` and verify that you can logout as guest and log in as a different user.

[DDF-3070](https://codice.atlassian.net/browse/DDF-3070)

idp guest logout
before
after

#### Checklist:
- [ n/a ] Change Log Updated
- [ n/a ] Update / Add Unit Tests

### DIFF
--- a/platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpLogoutActionProvider.java
+++ b/platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpLogoutActionProvider.java
@@ -56,7 +56,9 @@ public class IdpLogoutActionProvider implements ActionProvider {
             try {
 
                 @SuppressWarnings("unchecked")
-                String nameId = SubjectUtils.getName((Subject) ((Map) realmSubjectMap).get("idp"));
+                String nameId = SubjectUtils.getName((Subject) ((Map) realmSubjectMap).get("idp"),
+                        "You",
+                        true);
 
                 String nameIdTimestamp = nameId + "\n" + System.currentTimeMillis();
                 nameIdTimestamp = encryptionService.encrypt(nameIdTimestamp);
--- a/platform/security/platform-security-core-api/src/test/java/ddf/security/SubjectUtilsTest.java
+++ b/platform/security/platform-security-core-api/src/test/java/ddf/security/SubjectUtilsTest.java
@@ -32,12 +32,14 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedSet;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import javax.security.auth.kerberos.KerberosPrincipal;
 import javax.security.auth.x500.X500Principal;
 
 import org.apache.shiro.mgt.DefaultSecurityManager;
@@ -58,6 +60,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 
 import ddf.security.assertion.SecurityAssertion;
+import ddf.security.principal.GuestPrincipal;
 
 /**
  * Tests out the SubjectUtils class
@@ -67,6 +70,8 @@ public class SubjectUtilsTest {
     private static final String TEST_NAME = "test123";
 
     private static final String DEFAULT_NAME = "default";
+
+    public static final String X500_DN = "CN=Foo,OU=Engineering,OU=Dev,O=DDF,ST=AZ,C=US";
 
     private Principal principal;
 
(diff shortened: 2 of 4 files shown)


platform/security/servlet/security-servlet-logout/src/main/java/org/codice/ddf/security/servlet/logout/LogoutService.java
```diff
@@ -80,17 +80,14 @@ public Response getActionProviders(@Context HttpServletRequest request)
                             "",
                             true);
 
-                    if (displayName != null
```
brjeter: Did we still want the null check here? In which scenarios would the displayName be null?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpLogoutActionProvider.java
```diff
@@ -56,7 +56,9 @@
             try {
 
                 @SuppressWarnings("unchecked")
-                String nameId = SubjectUtils.getName((Subject) ((Map) realmSubjectMap).get("idp"));
+                String nameId = SubjectUtils.getName((Subject) ((Map) realmSubjectMap).get("idp"),
```
lessarderic: ❗️ We should add a unit test to cover the case where the name couldn't be found.

platform/security/servlet/security-servlet-logout/src/main/java/org/codice/ddf/security/servlet/logout/LogoutService.java
```diff
@@ -80,17 +80,14 @@ public Response getActionProviders(@Context HttpServletRequest request)
...
-                        actionProperties.put("title", action.getTitle());
-                        actionProperties.put("realm", realm);
-                        actionProperties.put("auth", displayName);
-                        actionProperties.put("description", action.getDescription());
-                        actionProperties.put("url",
-                                action.getUrl()
-                                        .toString());
-                        realmToPropMaps.add(actionProperties);
-                    }
+                    actionProperties.put("title", action.getTitle());
```
lessarderic: ❗️ Could we add some unit tests to cover the scenarios where `displayName` is empty and equal to `GUEST_DISPLAY_NAME`?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpLogoutActionProvider.java
```diff
@@ -56,7 +56,9 @@
...
+                        "You",
```
lessarderic: ❓ Is this `You` string displayed as-is in the UI somehow? If so, is it safe to assume this will get displayed ok in the sentence used by the UI?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpLogoutActionProvider.java
stustison: SubjectUtils already has unit tests to cover when the name can't be found and it falls back on the default value, which is "You" in this case.

platform/security/servlet/security-servlet-logout/src/main/java/org/codice/ddf/security/servlet/logout/LogoutService.java
stustison: I think that is already covered by the SubjectUtils unit tests.

platform/security/servlet/security-servlet-logout/src/main/java/org/codice/ddf/security/servlet/logout/LogoutService.java
pklinef: The second parameter of `SubjectUtils.getName` provides a default value.  Display name should never be null at this point.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpLogoutActionProvider.java
pklinef: This parameter is the default value for the display name.  In case the current subject does not have a display or name ID then You will be used.  The message gets displayed as "You logged out successfully.".

stustison:
build now

platform/security/servlet/security-servlet-logout/src/main/java/org/codice/ddf/security/servlet/logout/LogoutService.java
stustison: I think that we should test that `realmToPropMaps` always has values added to it no matter what the value of the `displayName` is.

lessarderic:
build now

platform/security/platform-security-core-api/src/test/java/ddf/security/SubjectUtilsTest.java
```diff
@@ -112,6 +120,62 @@ public void testGetDefaultName() {
     }
 
     @Test
+    public void testGuestDisplayName() {
+        Subject subject = getSubjectWithPrincipal(new GuestPrincipal("127.0.0.1"));
+        assertEquals(SubjectUtils.GUEST_DISPLAY_NAME, SubjectUtils.getName(subject, null, true));
```
lessarderic: ✏️  We could use `assertThat(..., is(...))` instead of `assertEquals()`.

pklinef:
build now

PR #2052, opened 2017-06-10
DDF-3069 Geospatial search by named location (keyword)

albany551:
#### What does this PR do?
- Adds a "Keyword" button to the search UI that queries a service for geographic place names. Upon choosing a name from the list, the corresponding shape (polygon, bounding box, etc.) will be loaded into the form.
- Adds the interface for the keyword search service, and provides a stub implementation that always returns one each of a polygon, bounding box, and point/radius location. The real implementation will be in a forthcoming PR.

 Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@andrewkfiedler 
@rzwiefel 

@andrewkfiedler
@millerw8
1. Open search UI and edit an existing search
2. Choose "Somewhere Specific" for the "Located" filed and select the "Keyword" tab.
3. Type 2 or more characters to get the 3 stubbed test suggestions.
4. For each suggestion:
   1. Select the suggestion and ensure that its shape becomes visible on the map and the appropriate edit fields are shown (polygon field for a polygon, north/south/east/west for a bounding box, etc.).
   2. Save the search, then edit it again. Ensure the the shape has been properly saved. (Note that bounding boxes get converted to polygons after saving).

https://codice.atlassian.net/browse/DDF-3069
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -13,7 +13,11 @@
  */
 package org.codice.ddf.catalog.ui.query;
 
-import static spark.Spark.*;
+import static spark.Spark.after;
+import static spark.Spark.before;
+import static spark.Spark.exception;
+import static spark.Spark.get;
+import static spark.Spark.post;
 import static spark.route.RouteOverview.enableRouteOverview;
 
 import java.util.List;
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -13,11 +13,10 @@
  */
 package org.codice.ddf.catalog.ui.query;
 
-import static spark.Spark.after;
-import static spark.Spark.exception;
-import static spark.Spark.post;
+import static spark.Spark.*;
 import static spark.route.RouteOverview.enableRouteOverview;
 
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import org.boon.json.JsonParserFactory;
@@ -26,6 +25,7 @@ import org.boon.json.ObjectMapper;
 import org.boon.json.implementation.ObjectMapperImpl;
 import org.codice.ddf.catalog.ui.query.cql.CqlQueryResponse;
 import org.codice.ddf.catalog.ui.query.cql.CqlRequest;
+import org.codice.ddf.catalog.ui.query.geofeature.FeatureService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -56,6 +56,8 @@ public class QueryApplication implements SparkApplication {
 
     private ActionRegistry actionRegistry;
 
+    private FeatureService featureService;
+
     private ObjectMapper mapper = new ObjectMapperImpl(new JsonParserFactory().usePropertyOnly(),
             new JsonSerializerFactory().includeEmpty()
                     .includeNulls()
@@ -64,6 +66,9 @@ public class QueryApplication implements SparkApplication {
 
     @Override
     public void init() {
+        before((req, res) -> {
+            res.type(APPLICATION_JSON);
+        });
 
         post("/cql", APPLICATION_JSON, (req, res) -> {
             CqlRequest cqlRequest = mapper.readValue(req.body(), CqlRequest.class);
@@ -74,8 +79,18 @@ public class QueryApplication implements SparkApplication {
         });
 
         after("/cql", (req, res) -> {
-            res.type(APPLICATION_JSON);
             res.header("Content-Encoding", "gzip");
+        });
+
+        get("/geofeature/suggestions", (req, res) -> {
+            String query = req.queryParams("q");
+            List<String> results = this.featureService.getSuggestedFeatureNames(query, 10);
+            return mapper.toJson(results);
+        });
+
+        get("/geofeature", (req, res) -> {
+            String name = req.queryParams("name");
+            return mapper.toJson(this.featureService.getFeatureByName(name));
         });
 
         exception(UnsupportedQueryException.class, (e, request, response) -> {
@@ -127,5 +142,9 @@ public class QueryApplication implements SparkApplication {
 
     public void setActionRegistry(ActionRegistry actionRegistry) {
         this.actionRegistry = actionRegistry;
+    }
+
+    public void setFeatureService(FeatureService featureService) {
+        this.featureService = featureService;
     }
 }
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -29,6 +29,7 @@ import org.boon.json.ObjectMapper;
 import org.boon.json.implementation.ObjectMapperImpl;
 import org.codice.ddf.catalog.ui.query.cql.CqlQueryResponse;
 import org.codice.ddf.catalog.ui.query.cql.CqlRequest;
+import org.codice.ddf.catalog.ui.query.geofeature.Feature;
 import org.codice.ddf.catalog.ui.query.geofeature.FeatureService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -94,7 +95,8 @@ public class QueryApplication implements SparkApplication {
 
         get("/geofeature", (req, res) -> {
             String name = req.queryParams("name");
-            return mapper.toJson(this.featureService.getFeatureByName(name));
+            Feature feature = this.featureService.getFeatureByName(name);
+            return mapper.toJson(feature.getJsonObject());
         });
 
         exception(UnsupportedQueryException.class, (e, request, response) -> {
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
@@ -1,0 +1,44 @@
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+public class BoundingBoxFeature extends Feature {
+    public BoundingBoxFeature() {
+        this.type = "bbox";
+    }
+
+    public double getNorth() {
+        return north;
+    }
+
+    public void setNorth(double north) {
+        this.north = north;
+    }
+
+    public double getSouth() {
+        return south;
+    }
+
+    public void setSouth(double south) {
+        this.south = south;
+    }
+
+    public double getEast() {
+        return east;
+    }
+
+    public void setEast(double east) {
+        this.east = east;
+    }
+
+    public double getWest() {
+        return west;
+    }
+
+    public void setWest(double west) {
+        this.west = west;
+    }
+
+    private double north;
+    private double south;
+    private double east;
+    private double west;
+}
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
@@ -13,6 +13,12 @@
  */
 package org.codice.ddf.catalog.ui.query.geofeature;
 
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A feature with bounding-box geometry specified by north, south, east, and west extents.
+ */
 public class BoundingBoxFeature extends Feature {
     private double north;
 
@@ -22,40 +28,34 @@ public class BoundingBoxFeature extends Feature {
 
     private double west;
 
-    public BoundingBoxFeature() {
-        this.type = "bbox";
-    }
-
-    public double getNorth() {
-        return north;
-    }
-
     public void setNorth(double north) {
         this.north = north;
-    }
-
-    public double getSouth() {
-        return south;
     }
 
     public void setSouth(double south) {
         this.south = south;
     }
 
-    public double getEast() {
-        return east;
-    }
-
     public void setEast(double east) {
         this.east = east;
-    }
-
-    public double getWest() {
-        return west;
     }
 
     public void setWest(double west) {
         this.west = west;
     }
 
+    @Override
+    public String getType() {
+        return "bbox";
+    }
+
+    @Override
+    public Map<String, Object> getGeometryJsonObject() {
+        Map<String, Object> result = new HashMap<>();
+        result.put("north", this.north);
+        result.put("south", this.south);
+        result.put("east", this.east);
+        result.put("west", this.west);
+        return result;
+    }
 }
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
@@ -1,6 +1,27 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
 package org.codice.ddf.catalog.ui.query.geofeature;
 
 public class BoundingBoxFeature extends Feature {
+    private double north;
+
+    private double south;
+
+    private double east;
+
+    private double west;
+
     public BoundingBoxFeature() {
         this.type = "bbox";
     }
@@ -37,8 +58,4 @@ public class BoundingBoxFeature extends Feature {
         this.west = west;
     }
 
-    private double north;
-    private double south;
-    private double east;
-    private double west;
 }
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Coordinate.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Coordinate.java
@@ -1,6 +1,24 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
 package org.codice.ddf.catalog.ui.query.geofeature;
 
 public class Coordinate {
+
+    private double latitude;
+
+    private double longitude;
+
     public double getLatitude() {
         return latitude;
     }
@@ -22,6 +40,4 @@ public class Coordinate {
         this.longitude = longitude;
     }
 
-    private double latitude;
-    private double longitude;
 }
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Coordinate.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Coordinate.java
@@ -13,11 +13,22 @@
  */
 package org.codice.ddf.catalog.ui.query.geofeature;
 
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Simple container for a latitude/longitude pair
+ */
 public class Coordinate {
 
     private double latitude;
 
     private double longitude;
+
+    public Coordinate(double latitude, double longitude) {
+        this.latitude = latitude;
+        this.longitude = longitude;
+    }
 
     public double getLatitude() {
         return latitude;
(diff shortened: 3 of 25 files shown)


rzwiefel:
Couple code formatting things seem to be off, We should double check you have the intelliJ formatter installed. 

On note irrelevant to your PR,  Can't seem to find our "Code Style" guidelines anywhere under here (https://codice.atlassian.net/wiki/display/DDF/Guidelines%2C+Standards+and+Best+Practices), Not sure if i've missed them or we don't have a page for them yet. It might be good to have a page (or comment under code style) about how to install the formatter as well.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -13,11 +13,10 @@
  */
 package org.codice.ddf.catalog.ui.query;
 
-import static spark.Spark.after;
-import static spark.Spark.exception;
-import static spark.Spark.post;
+import static spark.Spark.*;
```
bdeining: We should import the packages directly instead of using `.*`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public double getWest() {
+        return west;
+    }
+
+    public void setWest(double west) {
+        this.west = west;
+    }
+
+    private double north;
```
bdeining: nit: we've generally put the globals at the top of the class

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Coordinate.java
```diff
@@ -0,0 +1,27 @@
...
+    public void setLongitude(double longitude) {
+        this.longitude = longitude;
+    }
+
+    public Coordinate(double latitude, double longitude) {
+        this.latitude = latitude;
+        this.longitude = longitude;
+    }
+
+    private double latitude;
```
bdeining: same nit here about placement of globals

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Feature.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public String getType() {
+        return type;
+    }
+
+    private String name;
+    // this shouldn't be an instance property (it's constant for the class),
+    // but I haven't seen the right way yet to customize boon's serialization
+    // to get anything but properties...
+    protected String type;
```
bdeining: same nit here about placement of globals

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/FeatureService.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ */
+
+public interface FeatureService {
+    List<String> getSuggestedFeatureNames(String query, int maxResults);
```
bdeining: While it may be obvious what these methods are doing, it would be nice to have javadoc on this interface

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PointRadiusFeature.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public double getRadius() {
+        return radius;
+    }
+
+    public void setRadius(double radius) {
+        this.radius = radius;
+    }
+
+    private Coordinate center;
```
bdeining: same nit here about placement of globals

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PolygonFeature.java
```diff
@@ -0,0 +1,17 @@
...
+    public PolygonFeature() {
+        this.type = "polygon";
+        this.coordinates = new ArrayList<>();
+    }
+
+    public List<Coordinate> getCoordinates() {
+        return coordinates;
+    }
+
+    private List<Coordinate> coordinates;
```
bdeining: same nit here about placement of globals

catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -327,4 +328,8 @@
     <service ref="workspaceQueryService"
              interface="org.codice.ddf.catalog.ui.query.monitor.api.WorkspaceQueryService"/>
 
+    <bean id="featureService" class="org.codice.ddf.catalog.ui.query.geofeature.StubFeatureService"></bean>
```
bdeining: nit : ` <bean id="featureService" class="org.codice.ddf.catalog.ui.query.geofeature.StubFeatureService"/>`.  Can remove the closing bean element

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
albany551: I think intelliJ snuck this in on me, but will be fixing the formatting issues presently now that I've imported the formatted .jar. :+1:

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/BoundingBoxFeature.java
```diff
@@ -0,0 +1,61 @@
...
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+public class BoundingBoxFeature extends Feature {
```
glenhein: Please add some javadoc to the class.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Coordinate.java
```diff
@@ -0,0 +1,43 @@
...
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+public class Coordinate {
```
glenhein: Please add some javadoc to the class.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Feature.java
```diff
@@ -0,0 +1,36 @@
...
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+public abstract class Feature {
+    private String name;
+
+    // this shouldn't be an instance property (it's constant for the class),
```
glenhein: This should probably be formatted as javadoc.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PointRadiusFeature.java
```diff
@@ -0,0 +1,40 @@
...
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+public class PointRadiusFeature extends Feature {
```
glenhein: Please add some javadoc for this class.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PointRadiusFeature.java
```diff
@@ -0,0 +1,40 @@
...
+    private Coordinate center;
+
+    private double radius;
+
+    public PointRadiusFeature() {
+        this.type = "point-radius";
```
glenhein: would it be cleaner to call a protected ctor in the base class instead of setting the member directly?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PolygonFeature.java
```diff
@@ -0,0 +1,30 @@
...
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PolygonFeature extends Feature {
+    private List<Coordinate> coordinates;
+
+    public PolygonFeature() {
+        this.type = "polygon";
```
glenhein: would it be cleaner to call a protected ctor in the base class instead of setting the member directly?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PolygonFeature.java
```diff
@@ -0,0 +1,30 @@
...
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PolygonFeature extends Feature {
```
glenhein: Please add some javadoc to this class.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/StubFeatureService.java
```diff
@@ -0,0 +1,81 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class StubFeatureService implements FeatureService {
```
glenhein: Please add some javadoc to this class.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/StubFeatureService.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override
+    public List<String> getSuggestedFeatureNames(String query, int maxResults) {
+        List<String> results = new ArrayList<>();
+        results.add("Test City - Phoenix");
```
glenhein: This looks like a normal class, but it has test data??

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/PolygonFeature.java
albany551: That sounds better for this implementation, yeah - `type` is only a property at all because the JSON serializer being used doesn't seem to support non-properties. 
I'm thinking to avoid it entirely I'll just give the base Feature a getDataObject() method and let subclasses specify their data to be serialized as a return value.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/StubFeatureService.java
albany551: My plan was for this to go away in the next PR, or more likely get moved into a test in some form.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/autocomplete/input-autocomplete.hbs
```diff
@@ -0,0 +1,16 @@
...
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+ --}}
+<select name="{{property.id}}">
```
andrewkfiedler: We should be able to remove the `name` attribute and the `<option>` block inside here.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/autocomplete/input-autocomplete.less
```diff
@@ -0,0 +1,35 @@
+@{customElementNamespace}input.is-autocomplete {
+    .select2-container--default .select2-selection--single {
+        background: @darken-inherited-background-color;
+        color: @button-text-color;
```
andrewkfiedler: We should use `@text-color` here.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/autocomplete/input-autocomplete.less
```diff
@@ -0,0 +1,35 @@
...
+
+}
+
+// can't namespace because gets appended to doc
+.select2-dropdown {
+    z-index: @zIndexDropdown;
+    background-color: @dropdown-background-color;
+    border-radius: @border-radius;
+    border-color: @heavily-lighten-inherited-background-color;
+    color: @button-text-color;
```
andrewkfiedler: We should use `@text-color` here.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/autocomplete/input-autocomplete.less
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.select2-container--default .select2-results__option--highlighted[aria-selected] {
+    background: @primary-color;
```
andrewkfiedler: We can use `@lighten-inherited-background-color` here to bring it in line with the other dropdowns.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/autocomplete/input-autocomplete.less
```diff
@@ -0,0 +1,35 @@
...
+            top: 0px;
+            height: 100%;
+            width: @minimumButtonSize;
+        }        
+    }
+
+
+}
+
+// can't namespace because gets appended to doc
```
andrewkfiedler: I think it'd be worth adding:
```
.select2-results__option {
    height: @minimumButtonSize;
    line-height: @minimumButtonSize;
    overflow: hidden;
    text-overflow: ellipsis;
    padding: 0px @minimumSpacing;
    white-space: nowrap;
}
```
This will bring it more in line with the other dropdowns and handle text overflow.

troymohl:
Hero build passed. Test locations were returned and geospatial bounds were correctly saved in the search.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/autocomplete/input-autocomplete.view.js
```diff
@@ -0,0 +1,91 @@
...
+                if (result.loading) { return result.text; }
+                return getLabelForResult(result);
+            },
+            templateSelection(result) {
+                if (!result.id) { return result.text; /* nothing selected */ }
+                return getLabelForSelection(result);
+            }            
+        });
+    },
+    getCurrentValue(){
```
albany551: I have no qualms about destroying select2 :-)

catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
```diff
@@ -426,10 +433,77 @@ define([
             this.$('#utmZone').multiselect(singleselectOptions);
             this.$('#utmHemisphere').multiselect(singleselectOptions);
 
+            var keywordProperty = new Property({
```
andrewkfiedler: It'd be worth putting this block into it's own method `showKeywordPropertyView` so the logic is isolated, and so we can reshow the keyword property on it's own if need be.  Not critical or anything though.

catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
```diff
@@ -426,10 +433,77 @@ define([
...
+                    var lat = data.center.latitude;
+                    var lon = data.center.longitude;
+                    var radius = data.radius;
+                    attrsToSet = { lat, lon, radius, bbox: undefined };
+                    eventToDrawShape = 'search:circledisplay';                    
+                    break;
+                }
+            }
+
+            if (attrsToSet && eventToDrawShape) {
```
andrewkfiedler: Do we want to consider putting an else here for malformed geometries or geometries we can't currently handle?  Then we could do something to let the user know.

catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
```diff
@@ -426,10 +433,77 @@ define([
...
+                    });
+                    attrsToSet = { polygon, bbox: undefined };
+                    eventToDrawShape = 'search:polydisplay';
+                    break;
+                }
+                case "point-radius": {
+                    var lat = data.center.latitude;
+                    var lon = data.center.longitude;
+                    var radius = data.radius;
+                    attrsToSet = { lat, lon, radius, bbox: undefined };
```
millerw8: why is bbox defined?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Feature.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.codice.ddf.catalog.ui.query.geofeature;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Base class for results returned from the FeatureService.
+ */
+public abstract class Feature {
```
millerw8: Any reason we need to do something like this as opposed to using existing representations?

catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
albany551: For whatever reason, if I didn't effectively unset bbox here and and bbox had been previously defined, the ui would remain in bounding box mode. Possibly a data-binding issue.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Feature.java
albany551: Was planning to ask someone about this at some point, possibly for next or future PR. I saw that there's lots of existing code dealing with shapes but didn't manage to find any structure I could return that can be all of three supported feature types.

I started down the road of trying to use only opengis interfaces (with the JTS implementations?) but it was looking verbose and I figured it could wait. I'm assuming that's the correct path, though, as we already use those elsewhere?

catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
millerw8: I could see how a defined bounding box would be an issue. But I struggle to see a scenario where bounding box would be defined.  You start with a null variable, hit 1 of 3 disjoint pieces of code (data.type can't be a bbox and a polygon right? and you don't loop, and all cases have a break) and set the value once.

A couple of styling things, unusual to see curly braces on a case statement.  Not sure if we have styling guidelines about that.

Also, could remove the if/else below and make the Announcement (with a `return;`) the default of the switch statement.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/geofeature/Feature.java
millerw8: So, it may be more trouble than it is worth, not sure.  The challenge is always the circle.  Ultimately, you are creating these so some service can serve them out right?  And you are serving them out by doing a toJson of these objects.  So the real goal is to serialize out geometries to the system.

We tend to work with three standard for these geometries: geoJson (json), gml (xml), and wkt (human readable).

gml has the best representation of a circle with a few different options, the most obvious being CircleByCenterPoint.  geojson doesn't have a circle but it has a Feature which includes a geometry (which could be a point) and properties (which could include a radius).  There is an old proposal to add circles and ellipses, but it seems to be going no where.  WKT doesn't really have anything, and often uses polygon approximations.

JTS seems to follow closely to what WKT can support, and I don't really see much circle support there.  Again, it relies on polygon approximations.

So, the answer might well be "we do need to do something like this"...

catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.view.js
albany551: There's more code running than what is seen in this block, though, because events get triggered and handlers run (which then auto-magically fill in fields in the DOM). 
I mispoke/forgot the actual symptom - searching for a bounding box feature first and then a different feature type afterwards (could happen once I add country polygons) results in an invalid polygon being saved (all NaNs). I believe it has to do with existing code that converts any bounding box (include one you draw yourself) into a polygon on save.

millerw8:
build now

millerw8:
build now

albany551:
build now

albany551:
build now

PR #2055, opened 2017-06-12
DDF-3074 Add WebDAV support to in-place monitoring

brendan-hofmann:
#### What does this PR do?
Adds WebDAV support to the in-place directory monitor. Unit tests and documentation are in progress. I would also like to add component tests once #2029 is merged.
@emanns95  @jhunzik @AzGoalie @ahoffer @mweser 
@coyotesqrl 
@shaundmorris 
The easiest way to test this is with tomcat. Download tomcat and follow the instructions for the web.xml: https://wiki.apache.org/tomcat/Tomcat/WebDav
[DDF-3074](https://codice.atlassian.net/browse/DDF-3074)


catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -131,22 +131,31 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
-        String id = fileIdMap.loadAllKeys()
-                .contains(key) ? String.valueOf(fileIdMap.loadFromPersistence(key)) : null;
-
+        String refKey = (String) headers.get(Constants.STORE_REFERENCE_KEY);
+        String safeKey = null;
+        String id = null;
+        String name = ingestedFile.getName();
+
+        if (refKey != null) {
+            name = FilenameUtils.getName(refKey);
```
ahoffer: `name` is assigned in line 137? Why re-assign it in line 140 ?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -131,22 +131,31 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
         if (StandardWatchEventKinds.ENTRY_CREATE.equals(eventType)) {
             CreateStorageRequest createRequest =
                     new CreateStorageRequestImpl(Collections.singletonList(new ContentItemImpl(
                             uuidGenerator.generateUuid(),
                             Files.asByteSource(ingestedFile),
                             mimeType,
-                            ingestedFile.getName(),
+                            name,
                             0L,
                             null)), null);
```
ahoffer: The `processHeaders` sets properties on the RequestImpl objects. Change line 156 from:

`null)), null);`

to:

` null)), Maps.transformValues(headers, (v) -> (Serializable) v));`

The properties are set. Let Guava do the work of changing `Map<String, Object>` to `Map<String, Serializable>`. Then you can remove the `processHeaders` method.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -131,22 +131,31 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
-                .hashCode());
-        String id = fileIdMap.loadAllKeys()
-                .contains(key) ? String.valueOf(fileIdMap.loadFromPersistence(key)) : null;
-
+        String refKey = (String) headers.get(Constants.STORE_REFERENCE_KEY);
+        String safeKey = null;
+        String id = null;
+        String name = ingestedFile.getName();
+
+        if (refKey != null) {
```
ahoffer: Under what circumstances would the method be called with a with a `null` refKey? What kind of behavior would it exhibit?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -155,7 +164,9 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
                 List<Metacard> createdMetacards = createResponse.getCreatedMetacards();
 
                 for (Metacard metacard : createdMetacards) {
-                    fileIdMap.store(key, metacard.getId());
+                    if (safeKey != null) {
```
shaundmorris: ❓  should `if (safeKey != null)` wrap the entire for loop? if `safeKey` is null then it looks like it just logs the debug statement for each metacard.

brendan-hofmann:
❗️ @ahoffer pointed out that the files need to be URI-encoded before retrieving them from the server. Otherwise spaces in the paths (among other characters) won't work.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileEndpoint.java
```diff
@@ -192,4 +112,92 @@ public void setFile(File file) {
...
+        public boolean renameFile(String from, String to)
+                throws GenericFileOperationFailedException {
+            File srcFile = new File(from);
+            File destFile = new File(to);
+            try {
+                FileUtils.copyFile(srcFile, destFile);
+            } catch (IOException e) {
+                return false;
+            }
+            return !srcFile.exists() && destFile.exists();
```
emanns95: Will this ever be true if we're copying the file's contents to a new file instead of moving it? I think the srcFile will still exist

catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -125,11 +138,16 @@
                             platform-util,
                             platform-util-unavailableurls,
                             failsafe,
-                            hazelcast
+                            hazelcast,
+                            sardine,
+                            httpclient,
```
AzGoalie: looks like `httpclient` and `httpcore` are osgi bundles, any reason why we're embedding instead of using their bundles?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,306 @@
...
+    /**
+     * Compare two file lists for files which have been created, modified or deleted.
+     *
+     * @param parent   The parent entry
+     * @param previous The original list of files
+     * @param incoming The current list of files
+     */
+    private void checkAndNotify(final DavEntry parent, final DavEntry[] previous,
+            final DavResource[] incoming) {
+        int c = 0;
```
AzGoalie: Should we use a more descriptive name for this variable?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,306 @@
...
+                c++;
+
+            } else {
+                // entry no longer exists, so delete it
+                checkAndNotify(entry, entry.getChildren(), EMPTY_RESOURCES);
+                doDelete(entry);
+            }
+        }
+        // create anything remaining at the end
+        for (; c < incoming.length; c++) {
```
AzGoalie: Could use a while loop instead, but personal preference.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -131,22 +131,31 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
+        String id = null;
+        String name = ingestedFile.getName();
+
+        if (refKey != null) {
+            name = FilenameUtils.getName(refKey);
+            // guards against impermissible filesystem characters
+            safeKey = String.valueOf(refKey.hashCode());
+            if (fileIdMap.loadAllKeys()
+                    .contains(safeKey)) {
+                id = String.valueOf(fileIdMap.loadFromPersistence(safeKey));
```
brendan-hofmann: Using the reference's hash as the key guards against this. I can add a comment to that effect.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -192,15 +203,16 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
                 List<Metacard> deletedMetacards = deleteResponse.getDeletedMetacards();
 
                 for (Metacard delete : deletedMetacards) {
```
brendan-hofmann: It is used in the logging statement.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: If the file is being ingested into content the store, the reference will be null. References are for files that live outside of the content store. I can add an explanatory comment.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: That was deliberate to preserve the existing behavior. If we dislike the logging loops I can refactor them away?

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -411,58 +414,80 @@ private ContentItem readContent(URI uri) throws StorageException {
...
-        String extension = FilenameUtils.getExtension(file.getFileName()
-                .toString());
+        String filename = file.getFileName()
+                .toString();
+        String extension = FilenameUtils.getExtension(filename);
+        URI reference = null;
         if (REF_EXT.equals(extension)) {
-            extension = FilenameUtils.getExtension(FilenameUtils.removeExtension(file.getFileName()
-                    .toString()));
+            extension = FilenameUtils.getExtension(FilenameUtils.removeExtension(filename));
```
brendan-hofmann: Yes, in the case of external references. The real extension, and then a special reference extension is appended to that so we know how to retrieve the real file contents later. I can add a comment.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: Didn't know about that one, will do. 👍

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileEndpoint.java
brendan-hofmann: This endpoint doesn't know about that, so the implementation of this interface is still technically valid. It's not currently used but I could see it being used in the future.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: There's already a comment on the instantiation of `safeKey`. 😃

catalog/core/catalog-core-directorymonitor/pom.xml
brendan-hofmann: It looks like we do this in a lot of places. I tried without the embeds initially and their packages were not available. It looks like technical debt to make them available in the container and remove all the embeds.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
brendan-hofmann: Personally I like the convention of single character names for indexes that are incremented in a loop.  Given the number of times that index is referenced, I don't think renaming it would make the code more readable.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
brendan-hofmann: I think I like the for loop better, keeps the step and condition in the same place.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
shaundmorris: I'm okay either way on it. I just thought it looked a bit odd. If you think there is value in logging every content item created with its content ID I would leave it. Otherwise I would refactor away because it could be a lot of noise and not as valuable to an admin looking at the log.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,306 @@
...
+import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import com.github.sardine.DavResource;
+import com.github.sardine.Sardine;
+
+public class DavAlterationObserver implements Serializable {
```
shaundmorris: ❓  should we add JavaDoc at the class level?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: It is only logged at debug so I am fine with keeping it.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
brendan-hofmann: Can do.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
```diff
@@ -0,0 +1,317 @@
...
+                0);
+        setLength(isExists() && !isDirectory() ? davResource.getContentLength() : 0);
+        seteTag(isExists() ? davResource.getEtag() : "0");
+
+        if (!isDirectory() && (file != null || origLastModified != getLastModified())) {
+            FileUtils.deleteQuietly(file);
+        }
+
+        // Return if there are changes
+        return isExists() != origExists || getLastModified() != origLastModified
```
coyotesqrl: ✏️ Might be good to enforce more readable formatting here with EOL comments. Something like...
```java
return isExists() != origExists \\
        || getLastModified() != origLastModified \\
        || ...
```
That should make the formatter leave it alone.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
```diff
@@ -0,0 +1,317 @@
...
+    /**
+     * Return a local cache of the file being monitored.
+     * This file is invalidated if necessary when refresh() is called.
+     *
+     * @param sardine
+     * @return the file being monitored
+     */
+    public File getFile(Sardine sardine) throws IOException {
+        if (file == null || !file.exists()) {
+            Path dav = Files.createTempDirectory("dav");
```
coyotesqrl: ❓ You're never deleting the temp directory?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
```diff
@@ -0,0 +1,317 @@
...
+            File dest = new File(dav.toString(),
```
coyotesqrl: ✏️ Could call `dav.toFile()` here instead.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
```diff
@@ -0,0 +1,317 @@
...
+                    URLDecoder.decode(FilenameUtils.getName(getLocation())));
+            IOUtils.copy(sardine.get(getLocation()), new FileOutputStream(dest));
```
coyotesqrl: ❗️ Should close the output stream.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.camel.component.file.GenericFileOperations;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.github.sardine.Sardine;
+import com.github.sardine.SardineFactory;
+
+public class DurableWebDavFileConsumer extends AbstractDurableFileConsumer {
+    private static final Logger LOGGER =
+            LoggerFactory.getLogger(DurableFileSystemFileConsumer.class);
```
coyotesqrl: ❗️ Wrong class for logger.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/DavEntryTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import com.github.sardine.DavResource;
+import com.github.sardine.Sardine;
+
+@RunWith(JUnit4.class)
+public class DavEntryTest {
```
coyotesqrl: ✏️ Might consider adding serialization tests.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserverTest.java
```diff
@@ -0,0 +1,230 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import com.github.sardine.DavResource;
+import com.github.sardine.Sardine;
+
+@RunWith(JUnit4.class)
+public class DavAlterationObserverTest {
```
coyotesqrl: ✏️ Might want to add serialization tests.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
stustison: Should probably keep track of the items created and log after the loop.

shaundmorris: yeah @stustison that sounds better to me. Seems like the code should be intentional on whether it is logging for debugging purposes or looping through the created metacards for another reason.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,310 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Final processing.
+     *
+     * @throws Exception if an error occurs
+     */
+    public void destroy() throws Exception {
```
shaundmorris: ❓ why do you have this method if does not do anything? I see it is being called by DurableWebDavFileConsumer, but since it is calling this implementation directly should something actually happen or does DurableWebDavFileConsumer not really need to call this method?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
```diff
@@ -0,0 +1,134 @@
...
+    private Sardine sardine = SardineFactory.begin();
+
+    DurableWebDavFileConsumer(GenericFileEndpoint<EventfulFileWrapper> endpoint,
+            Processor processor, GenericFileOperations<EventfulFileWrapper> operations) {
+        super(endpoint, processor, operations);
+    }
+
+    @Override
+    protected boolean pollDirectory(String fileName, List list, int depth) {
+        String remaining = ((DurableFileComponent) endpoint.getComponent()).remaining;
```
shaundmorris: ❗️ I would recommend doing some more null checking in here. A DefaultEndpoint can be constructed with a null `component` and a DurableFileComponent can be constructed with a null `remaining`. I may be missing Javadoc that says those won't be null. Seems like `remaining` can only get populated if `DurableFileComponent.buildFileEndpoint()` is called.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileComponent.java
```diff
@@ -32,18 +36,28 @@
                     + " Use the fileName option to set the dynamic expression.");
         }
 
-        File file = new File(remaining);
-
-        DurableFileEndpoint result = new DurableFileEndpoint(uri, this);
-        result.setFile(file);
+        this.remaining = remaining;
+        String isDav = String.valueOf(parameters.get("isDav"));
```
shaundmorris: ❗️ this local variable is hiding the global `isDav` Boolean. Recommend renaming.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,310 @@
...
+     * the empty or the file is not a directory
+     */
+    private DavResource[] listFiles(final String file) {
+        DavResource[] children = null;
+        try {
+            List<DavResource> list = sardine.list(file);
+            // the returned list includes the parent
+            if (list.size() > 1 && list.get(0)
+                    .isDirectory()) {
+                children = list.subList(1, list.size())
```
brendan-hofmann: Almost. `sort` returns void so it can't be chained like that. Definitely don't need the extra conversions though. 😃

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,310 @@
...
+     */
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "[file='" + getDirectory() + '\'' + ", listeners="
+                + listeners.size() + "]";
+    }
+
+    private void readObject(java.io.ObjectInputStream in)
+            throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
```
brendan-hofmann: I don't think that's necessary here. The serialization is stored in `DDF_HOME/data` so an attacker would have to access to the filesystem, at which point there's much worse things they could be doing. We've always treated the installation directory as secure.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
brendan-hofmann: I think comment slashes lean in the other direction. 😃 Will do though.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
brendan-hofmann: Good catch, added appropriate null checks to both classes to guarantee this.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/EntryAlterationListener.java
```diff
@@ -0,0 +1,28 @@
...
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.content.monitor;
+
+public interface EntryAlterationListener {
```
shaundmorris: ❓ should we add any Javadoc for this interface?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    public void shutdown() throws Exception {
+        super.shutdown();
+        if (observer != null) {
+            observer.destroy();
+        }
+    }
+
+    private void createExchangeHelper(DavEntry entry, WatchEvent.Kind<Path> fileEvent) {
```
shaundmorris: ❓ since this is a private method and only called in the `EnteryAlterationListnerImpl` class, should it just be moved there or is there a bigger purpose to this method ?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,310 @@
...
+     * @param previous The original list of files
+     * @param incoming The current list of files
+     */
+    private void checkAndNotify(final DavEntry parent, final DavEntry[] previous,
+            final DavResource[] incoming) {
+        int c = 0;
+        final DavEntry[] current =
+                incoming.length > 0 ? new DavEntry[incoming.length] : DavEntry.getEmptyEntries();
+        // this loop relies on lexicographical sorting, so compareTo < 0 will never be hit
+        for (final DavEntry entry : previous) {
```
brendan-hofmann: Added an explanation with examples to hopefully demonstrate the logic better. Also added a unit test to prove the first question works.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
brendan-hofmann: Added an on-completion hook to remove the temporary cache after it is successfully processed.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/EntryAlterationListener.java
brendan-hofmann: I think the code is sufficiently self-explanatory that there's nothing to add. What did you have in mind?

brendan-hofmann:
build now

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/EntryAlterationListener.java
shaundmorris: Wasn't sure if it was going to be used beyond its current implementation. If there were going to be more implementations was wondering if we needed more guidance on what has to be implemented (e.g. noop). Just a question, I don't think anything specific needs to be added.

brendan-hofmann: Ok. Since it's a listener, nothing is *required* anyways. It is up to the implementer which, if any, events to react to.

brendan-hofmann:
build now

brendan-hofmann:
build now

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -203,27 +218,18 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
-        }
-        if (headers.containsKey(Constants.STORE_REFERENCE_KEY)) {
-            storageRequest.getProperties()
-                    .put(Constants.STORE_REFERENCE_KEY, ingestedFile.getAbsolutePath());
-        }
+    protected HashMap<String, Serializable> getProperties(Map<String, Object> headers) {
+        return Maps.newHashMap(Maps.transformValues(headers, (v) -> (Serializable) v));
     }
 
     private void waitForAvailableSource(CatalogFramework catalogFramework)
```
brendan-hofmann: SonarQube doesn't understand the lambda usage.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -139,22 +140,35 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
+        if (refKey != null) {
+            // guards against impermissible filesystem characters
+            safeKey = DigestUtils.sha1Hex(refKey);
+            if (fileIdMap.loadAllKeys()
+                    .contains(safeKey)) {
+                id = String.valueOf(fileIdMap.loadFromPersistence(safeKey));
+            } else if (!StandardWatchEventKinds.ENTRY_CREATE.equals(eventType)) {
+                LOGGER.warn("Unable to look up id for {}, not performing {}",
+                        refKey,
+                        eventType.name());
```
brendan-hofmann: This is just a getter.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/AbstractDurableFileConsumer.java
```diff
@@ -0,0 +1,112 @@
...
+        @Override
+        public void onComplete(Exchange exchange) {
+            // no-op
+        }
+
+        @Override
+        public void onFailure(Exchange exchange) {
+            INGEST_LOGGER.error("Delivery failed for {} event on {}",
+                    file,
+                    fileEvent.name(),
```
brendan-hofmann: Just a getter.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
```diff
@@ -0,0 +1,324 @@
...
+            URI uri = new URI(url.getProtocol(),
+                    url.getUserInfo(),
+                    url.getHost(),
+                    url.getPort(),
+                    url.getPath(),
+                    url.getQuery(),
+                    url.getRef());
+            location = uri.toASCIIString();
+        } catch (MalformedURLException | URISyntaxException e) {
+            throw new RuntimeException(e);
```
brendan-hofmann: No.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
shaundmorris: yeah, we will have to check out why this is still an issue. It has popped up on a few reviews. https://jira.sonarsource.com/browse/SONARJAVA-583 indicates it should be fixed.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
```diff
@@ -0,0 +1,151 @@
...
+            remaining = ((DurableFileComponent) component).remaining;
+            if (remaining != null) {
+                initialize(remaining);
+            }
+        }
+        if (observer != null) {
+            observer.addListener(listener);
+            observer.checkAndNotify(sardine);
+            observer.removeListener(listener);
+            fileSystemPersistenceProvider.store(String.valueOf(remaining.hashCode()), observer);
```
brendan-hofmann: Outdated, guarded in parent class.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -166,33 +180,34 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
             if (createResponse != null) {
                 List<Metacard> createdMetacards = createResponse.getCreatedMetacards();
 
-                for (Metacard metacard : createdMetacards) {
-                    fileIdMap.store(key, metacard.getId());
-                    LOGGER.debug("content item created with id = {}", metacard.getId());
+                if (safeKey != null) {
+                    fileIdMap.store(safeKey,
+                            createdMetacards.get(0)
```
coyotesqrl: ❓ Why just the first metacard in the list? Should we protect against null or empty `createdMetacards`?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -166,33 +180,34 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
+                                    .getId());
                 }
+                LOGGER.debug("content item created with id = {}",
+                        createdMetacards.stream()
+                                .map(Metacard::getId));
```
coyotesqrl: ❗️ This logger statement is wrong. It will `toString()` the stream returned from the `map` method. To match above, you want to get `createdMetacards.get(0).getId()`. Conversely, 
```java
createMetacards.stream()
    .map(Metacard::getId)
    .findFirst()
    .orElse("XXX")
```

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -166,33 +180,34 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
             if (updateResponse != null) {
                 List<Update> updatedMetacards = updateResponse.getUpdatedMetacards();
 
-                for (Update update : updatedMetacards) {
-                    LOGGER.debug("content item updated with id = {}",
-                            update.getNewMetacard()
-                                    .getId());
-                }
+                LOGGER.debug("content item updated with id = {}",
+                        updatedMetacards.stream()
```
coyotesqrl: ❗️ Same. This will `toString()` your stream.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -203,27 +218,18 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
             if (deleteResponse != null) {
                 List<Metacard> deletedMetacards = deleteResponse.getDeletedMetacards();
 
-                for (Metacard delete : deletedMetacards) {
-                    fileIdMap.delete(ingestedFile.getAbsolutePath());
-                    LOGGER.debug("content item deleted with id = {}", delete.getId());
+                if (safeKey != null) {
+                    fileIdMap.delete(safeKey);
                 }
+                LOGGER.debug("content item deleted with id = {}",
```
coyotesqrl: ❗️ Same problem with `toString`

catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -271,6 +284,30 @@
...
+        <dependency>
+            <groupId>ddf.test.thirdparty</groupId>
+            <artifactId>restito</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.xalan</artifactId>
+            <version>2.7.2_3</version>
```
coyotesqrl: ✏️ We do have a `cxf.xalan.bundle.version` property in the root pom.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,318 @@
...
+ */
+public class DavAlterationObserver implements Serializable {
+
+    private static final long serialVersionUID = 42L;
+
+    private static final DavResource[] EMPTY_RESOURCES = new DavResource[] {};
+
+    private transient LinkedHashSet<EntryAlterationListener> listeners = new LinkedHashSet<>();
+
+    private final DavEntry rootEntry;
```
coyotesqrl: ✏️ Nit. You might want to move the `rootEntry` down so the non-serialized and serialized fields in the class are all grouped together.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,318 @@
...
+    public void initialize(Sardine sardine) {
+        this.sardine = sardine;
+        try {
+            DavResource davResource = sardine.list(rootEntry.getLocation())
+                    .get(0);
+            rootEntry.refresh(davResource);
+            final DavEntry[] children = doListFiles(rootEntry.getLocation(), rootEntry);
+            rootEntry.setChildren(children);
+        } catch (IOException e) {
+            // probably means root location was inaccessible
```
coyotesqrl: ❓ Is this worth a log?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
```diff
@@ -0,0 +1,318 @@
...
+            // would remove all metacards if connectivity is lost, but tough to distinguish
+            // from the hierarchy not existing remotely
+            checkAndNotify(rootEntry, rootEntry.getChildren(), EMPTY_RESOURCES);
+        } else {
+            // Didn't exist and still doesn't
+        }
+    }
+
+    /**
+     * Compare two file lists for files which have been created, modified or deleted.
```
coyotesqrl: This javadoc makes my soul happy.

brendan-hofmann:
build now

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileEndpoint.java
```diff
@@ -192,4 +112,92 @@ public void setFile(File file) {
...
+            return new File(name).exists();
+        }
+
+        @Override
+        public boolean renameFile(String from, String to)
+                throws GenericFileOperationFailedException {
+            File srcFile = new File(from);
+            File destFile = new File(to);
+            try {
+                FileUtils.copyFile(srcFile, destFile);
```
coyotesqrl: ❓ If this is a rename operation, shouldn't we call `moveFile()` here or delete the source file on completion? Or, should this method be renamed to `copyFile()`?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileSystemFileConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    protected void initialize(String fileName) {
+        if (fileSystemPersistenceProvider == null) {
+            fileSystemPersistenceProvider =
+                    new FileSystemPersistenceProvider(getClass().getSimpleName());
+        }
+        if (observer == null && fileName != null) {
+            if (fileSystemPersistenceProvider.loadAllKeys()
+                    .contains(String.valueOf(fileName.hashCode()))) {
```
coyotesqrl: ❓ Shouldn't this also be the SHA-1 of the filename?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileSystemFileConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+                observer =
+                        (FileAlterationObserver) fileSystemPersistenceProvider.loadFromPersistence(
```
coyotesqrl: ❓ Same question re: SHA-1.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
rzwiefel: This is glorious indeed.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: There's a one to one relationship between the reference key and the file, so there should always be exactly one. Batching is only possible for content items that are stored in the content store.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: Good catch, not what I intended. 😅

catalog/core/catalog-core-directorymonitor/pom.xml
brendan-hofmann: It was in platform, moved it up to root.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileEndpoint.java
brendan-hofmann: Can't rename it because it is inherited, but did fix the  copy to move. 👍

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavAlterationObserver.java
brendan-hofmann: Yep, logged. 👍

brendan-hofmann:
build now

LinkMJB:
build now

LinkMJB:
build now

brendan-hofmann:
build now

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -203,27 +217,25 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
-            storageRequest.getProperties()
-                    .put(Constants.STORE_REFERENCE_KEY, ingestedFile.getAbsolutePath());
-        }
+    protected void logIds(List<Metacard> metacards, String action) {
+        LOGGER.debug("content item(s) {} with id = {}",
+                action,
+                metacards.stream()
+                        .map(Metacard::getId)
+                        .filter(Objects::nonNull)
+                        .reduce((s, s2) -> String.format("%s, %s", s, s2)));
```
coyotesqrl: ❗️ This will log the `toString()` of the optional returned from `reduce`, giving you outputs like
`content item(s) updated with id = Optional[x, y, z]`
and potentially
`content item(s) updated with id = Optional.empty`.

There are a couple of ways around that, but the easiest is to swap out the reducer with a call to 
`.collect(Collectors.joining(", "));`

That will get you the output you want and handle the empty collection of ids.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
```diff
@@ -203,27 +217,25 @@ public void createContentItem(FileSystemPersistenceProvider fileIdMap, ContentEn
...
-            storageRequest.getProperties()
-                    .put(Constants.ATTRIBUTE_OVERRIDES_KEY,
-                            (Serializable) attributeOverrideHeaders);
-        }
-        if (headers.containsKey(Constants.STORE_REFERENCE_KEY)) {
-            storageRequest.getProperties()
-                    .put(Constants.STORE_REFERENCE_KEY, ingestedFile.getAbsolutePath());
-        }
+    protected void logIds(List<Metacard> metacards, String action) {
+        LOGGER.debug("content item(s) {} with id = {}",
```
coyotesqrl: ❗️ Assuming these can be long lists, this processing could be pretty expensive for naught in the case where the logging level isn't set high enough. Best to wrap the debug call in a conditional `ifDebugEnabled()` check.

catalog/core/catalog-core-directorymonitor/src/test/resources/exam.properties
```diff
@@ -1 +1 @@
-pax.exam.karaf.version=${karaf.version}
+pax.exam.karaf.version=4.0.7
```
coyotesqrl: ❓ Is this the version you want here?

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: Fixed.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: Good catch! Fixed. 👍

catalog/core/catalog-core-directorymonitor/src/test/resources/exam.properties
brendan-hofmann: It was a few weeks ago. 😄 Fixed.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
```diff
@@ -0,0 +1,139 @@
...
+            fileSystemPersistenceProvider.store(sha1, observer);
+            return true;
+        } else {
+            return isMatched(null, null, null);
+        }
+    }
+
+    @Override
+    protected void initialize(String fileName, String sha1) {
+        if (fileSystemPersistenceProvider == null) {
```
clockard: This method looks like it an exact copy of the one from DurableFileSystemFileConsumer with the exception of the cast. Could this be somehow moved to the abstract base class?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -123,6 +159,27 @@ public void testInPlaceMonitoring()
     }
 
     @Test
+    public void testInPlaceDavMonitoring()
+            throws IOException, InterruptedException, SourceUnavailableException, IngestException,
```
clockard: Could just throw Exception here instead of listing them all out.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
```diff
@@ -123,6 +159,27 @@ public void testInPlaceMonitoring()
...
+            MimeTypeParseException, StorageException, URISyntaxException {
+        updateContentDirectoryMonitor("http://localhost:" + stubServer.getPort() + "/webdavtest",
```
clockard: Is the stubServer port a fixed port or is it the first available port? Just want to make sure we don't run into port conflicts.

catalog/core/catalog-core-directorymonitor/src/test/resources/exam.properties
```diff
@@ -1 +1 @@
-pax.exam.karaf.version=${karaf.version}
+pax.exam.karaf.version=4.1.1
```
clockard: Does property replacement not work here? We should use the variable if at all possible.

vim.exe.stackdump
```diff
@@ -0,0 +1,19 @@
+Stack trace:
```
clockard: Something tells me this should be removed :)

platform/platform-app/pom.xml
```diff
@@ -455,7 +458,6 @@
         <cxf.xmlresolver.bundle.version>1.2_5</cxf.xmlresolver.bundle.version>
         <cxf.xerces.bundle.version>2.11.0_1</cxf.xerces.bundle.version>
         <cxf.bcel.bundle.version>5.2_4</cxf.bcel.bundle.version>
-        <cxf.xalan.bundle.version>2.7.2_3</cxf.xalan.bundle.version>
```
clockard: I realize you added this to the root pom but I don't know that we want to remove it from here since this is copied from cxf. I think by keeping them all together here they were trying to make it easier to upgrade next time.

catalog/core/catalog-core-camelcomponent/src/main/java/ddf/camel/component/catalog/content/ContentProducerDataAccessObject.java
brendan-hofmann: I was trying to be consistent due to the extra logic for the update case, but now that I think about it that will only ever have a size of one, so I think I will move the debug check into the helper method.

catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -117,6 +117,19 @@
             <groupId>com.hazelcast</groupId>
             <artifactId>hazelcast</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.lookfirst</groupId>
+            <artifactId>sardine</artifactId>
+            <version>5.7</version>
```
brendan-hofmann: Do you have a preference as to what pom level the version property lives in? I had left it local since it was only used in one place, so my first inclination would be to put the property in this pom. That had seemed a little redundant, but I can see the reasoning from a consistency standpoint.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DavEntry.java
```diff
@@ -0,0 +1,324 @@
...
+                location = location.replaceFirst("/", "");
+            }
+            if (!parentLocation.endsWith("/") && !location.startsWith("/")) {
+                location = "/" + location;
+            }
+            location = parentLocation + location;
+        }
+        try {
+            URL url = new URL(location);
+            URI uri = new URI(url.getProtocol(),
```
brendan-hofmann: Yes, the internal logic has different behaviors for handling character encoding/decoding. I will try to add an explanatory comment.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableWebDavFileConsumer.java
brendan-hofmann: I could not find an elegant way to abstract this logic given all the different types involved.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
brendan-hofmann: I dislike that pattern, especially since the IDE manages them for you, but I remain open to having my mind changed.

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
brendan-hofmann: It is dynamically chosen from available ports.

catalog/core/catalog-core-directorymonitor/src/test/resources/exam.properties
brendan-hofmann: It did not work for me, but I will try it again.

vim.exe.stackdump
brendan-hofmann: Woops. 😅

platform/platform-app/pom.xml
brendan-hofmann: It was pulled up to re-use in catalog. Should I duplicate it instead?

catalog/core/catalog-core-directorymonitor/src/test/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitorIT.java
clockard: I prefer the ease/brevity of the Exception for tests but there isn't anything wrong with listing them out 👍

catalog/core/catalog-core-directorymonitor/pom.xml
clockard: I would put it in the root pom but the local pom also works.

platform/platform-app/pom.xml
brendan-hofmann: I fixed this per @emmberk . Pattern is to declare in the 'must match' section of the root pom, then redeclare using the root declaration in platform-app so that block matches as closely as possible to the CXF block it is copied from.

pom.xml
```diff
@@ -136,6 +136,7 @@
         <dom4j.bundle.version>1.6.1_5</dom4j.bundle.version>
         <jdom.bundle.version>1.1_4</jdom.bundle.version>
         <geronimo-jms_2_spec.version>1.0-alpha-2</geronimo-jms_2_spec.version>
+        <cxf.xalan.bundle.version>2.7.2_3</cxf.xalan.bundle.version>
```
emmberk: :+1:

platform/platform-app/pom.xml
emmberk: Thanks!:+1:

platform/platform-app/pom.xml
```diff
@@ -455,7 +458,7 @@
...
+        <cxf.xalan.bundle.version>${cxf.xalan.bundle.version}</cxf.xalan.bundle.version>
```
emmberk: :+1:

brendan-hofmann:
build now

brendan-hofmann:
build now

brendan-hofmann:
build now

PR #2080, opened 2017-06-21

mweser:
#### What does this PR do?
Provides a generic utility class to page through catalog framework query results

@rzwiefel @ahoffer @emanns95  @vinamartin  

@coyotesqrl
@lessarderic

1. Build and run DDF instance
2. Ingest large number of files
3. Attempt export command such as `catalog:export --delete=true --cql "\"id\" like '*'"`
4. Verify that zip file contains all files sought

Looking to make class designed around ExportCommand to have more generic functionality to be used as a util class for other uses, as well as to provide paging of queries to avoid memory issues.

[](https://codice.atlassian.net/browse/DDF-3034)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,130 @@
...
+import ddf.catalog.data.Result;
+import ddf.catalog.federation.FederationException;
+import ddf.catalog.operation.SourceResponse;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
+
+public class QueryResultPaginator {
+
+    private final CatalogFramework catalogFramework; // needs to be passed in from on-high
```
vinamartin: ✏️ This comment is probably unnecessary.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,130 @@
...
+
+        this.catalogFramework = catalogFramework;
+        this.queryRequest = queryRequest;
+        this.indexOffset = indexOffset;
+        this.endIndex = endIndex;
+        this.maxResults = maxResults;
+        this.ceilingIndex = ceilingIndex;
+
+        if (indexOffset < 0 || indexOffset > ceilingIndex || indexOffset > endIndex) {
+            throw new Exception();
```
vinamartin: ✏️  Could we add an exception message here explaining what went wrong?

catalog/opensearch/catalog-opensearch-endpoint/src/main/resources/ddf-os.xml
```diff
@@ -39,9 +39,9 @@
     <!--
       (q) searchTerms - Space delimited list of search terms.
 
-      (mr) maxResults - Maximum # of results to return.
+      (ceilingIndexults - Maximum # of results to return.
```
vinamartin: ❓  What is a ceilingIndexults?

mweser: Not even sure how that file got into my PR, but I'll check it out.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Will do.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: True, will remove

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
```diff
@@ -0,0 +1,230 @@
...
+                startIndex,
+                endIndex,
+                maxResults,
+                ceilingIndex);
+
+        resultList = queryResultPaginator.nextPage();
+        assertThat("Stops at ceiling index", resultList.size() == 400);
+    }
+
+//    @Test
```
mweser: Sounds good.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,121 @@
...
+            queryNext();
+
+            if (results.hasNext()) {
+                return true;
+            }
+            // TODO: 6/20/17 Not sure what to do with this clause
+            //            if (isDone) {
+            //                return false;
+            //            }
+            return hasNext();
```
AzGoalie: Wouldn't this cause an infinite loop when `results.hashNext()` returns false? I think in both cases we should return `results.hasNext()`

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+    private int ceilingIndex;
+
+    private int endIndex;
+
+    private int maxResults;
+
+    private int indexOffset;
+
+    public QueryResultPaginator(CatalogFramework catalogFramework, QueryRequestImpl queryRequest) {
+        this.catalogFramework = catalogFramework;
```
AzGoalie: Do we not need to validate these arguments like we did in `QueryResultIterable`?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private List<Result> queryCatalogFramework() {
+        try {
+            Map<String, Serializable> props = new HashMap<>();
+            // Avoid caching all iteratorResults while dumping with native query mode
+            props.put("mode", "native");
+            response = catalogFramework.query(queryRequest);
+        } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            throw new CatalogRuntimeException(e);
```
AzGoalie: Should we add an exception message here too?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        List<Result> resultList = response.getResults();
+        return resultList;
+    }
+
+    private List<Result> trimListToMaxResults(List<Result> list, int maxResults, int ceilingIndex,
+            int originalIndexOffset) {
+
+        // TODO: 6/15/17 Change to feedback on number of results in last query
+        if (list.size() > maxResults) {
```
AzGoalie: What is the performance impact of dropping the unpaged results? Looks like we make a full query every time we want a new page and drop everything that isn't in the range.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+                    originalIndexOffset);
+
+            indexOffset = originalIndexOffset + resultList.size();
+        }
+        return resultList;
+    }
+
+    private List<Result> queryCatalogFramework() {
+        try {
+            Map<String, Serializable> props = new HashMap<>();
```
lessarderic: ❗️ This `Map` is not used and should be removed. Native mode should be set on the `QueryRequest` by the caller, not by this class, if needed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class QueryResultPaginator {
+
+    private final CatalogFramework catalogFramework;
+
+    private final QueryRequestImpl queryRequest;
+
+    private SourceResponse response = null;
+
+    private int ceilingIndex;
```
lessarderic: ❓ Assuming `ceilingIndex` represents the maximum total number or results to return, I wonder if we really need it. Since the client can provide start and end indices, they can limit the total number of results they can get, i.e., `ceilingIndex` = end index - start index.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return resultList;
+    }
+
+    private List<Result> queryCatalogFramework() {
+        try {
+            Map<String, Serializable> props = new HashMap<>();
+            // Avoid caching all iteratorResults while dumping with native query mode
+            props.put("mode", "native");
+            response = catalogFramework.query(queryRequest);
```
lessarderic: ❗️ We need to update the query's start index before calling the catalog otherwise we will always get the same results over and over again. We also need to make sure our unit tests catch this.

catalog/spatial/geocoding/spatial-geocoding-plugin/pom.xml
```diff
@@ -65,7 +72,7 @@
                             ddf.catalog.util.impl
                         </Private-Package>
                         <Embed-Dependency>
-                            commons-lang3
+                            commons-lang3, catalog-core-api-impl, platform-util
```
lessarderic: ❓ What's forcing us to embed `platform-util`?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/CatalogRuntimeException.java
```diff
@@ -11,22 +11,22 @@
  * is distributed along with this program and can be found at
  * <http://www.gnu.org/licenses/lgpl.html>.
  */
-package org.codice.ddf.commands.util;
+package ddf.catalog.util.impl;
 
-public class CatalogCommandRuntimeException extends RuntimeException {
-    public CatalogCommandRuntimeException() {
+public class CatalogRuntimeException extends RuntimeException {
+    public CatalogRuntimeException() {
```
rzwiefel: But what if one is merely passing an exception and no message is needed? For example the one case where this is used (https://github.com/codice/ddf/pull/2080#discussion_r123301891) There really is no message to give, the real message of failure is contained in the exception that is being wrapped.

catalog/spatial/geocoding/spatial-geocoding-plugin/pom.xml
rzwiefel: `catalog-core-api-impl` has a dependency on `platform-util` Its for a single static method that is ~4 lines long. 
(Aware this is a link that some may not be able to access) https://connexta.slack.com/archives/C0B9XC1F0/p1491413170475395?thread_ts=1491408676.160564&cid=C0B9XC1F0

mweser: These were some odd dependencies that were breaking the build. Travis knows more on why they are needed.

mweser:
@lessarderic @coyotesqrl Brought in Aaron's commits and going to build on top of this (implementing the iterator class as exists in my branch).

mweser:
build now

lessarderic:
@mweser, please address any comments not addressed on PR https://github.com/codice/ddf/pull/2097 as part of this PR.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+     * If there are pages beyond the current page, returns true.
+     *
+     * @return True if there are results beyond the current page
+     */
+    /**
+     * @return true if there are more results for the client
+     */
+    public boolean hasNext() {
+
+        if (notEmpty()) {
```
rzwiefel: ✏️ 
```java
         if (notEmpty()) {
             return true;
         }

         if (isExhausted) {
             return false;
         }

         //Buffer is empty, but query might not be exhausted.
         fetchNext();
         return notEmpty();
```

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+    public List<Result> next() throws NoSuchElementException {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+
+        while (!isExhausted && getBufferSize() < pageSize) {
+            fetchNext();
+        }
+
+        List<Result> queriedResultsBufferCopy = new ArrayList<>(pageSize);
```
rzwiefel: ❓ Should we instantiate this ArrayList with a capacity equal to `queriedResultBuffer.size()` to account for the case when there are very few results incoming and prevent over provisioning?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,129 @@
...
+
+import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Result;
+import ddf.catalog.federation.FederationException;
+import ddf.catalog.operation.SourceResponse;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
+
+public class QueryResultPaginator {
```
mweser: I have this commented above the constructor. Would it be better to move those comments to the class declaration?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Committed simplified logic.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: I would think so, as drainTo should increase the list size dynamically if need be?

catalog/opensearch/catalog-opensearch-endpoint/src/main/resources/ddf-os.xml
```diff
@@ -39,9 +39,9 @@
     <!--
       (q) searchTerms - Space delimited list of search terms.
 
-      (mr) maxResults - Maximum # of results to return.
+      (pageSizeults - Maximum # of results to return.
```
emanns95: ❓ is `pageSizeults` intentional?

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
```diff
@@ -0,0 +1,395 @@
...
+        for (int i = 0; i < pageSize; i++) {
+            ResultImpl newResult = new ResultImpl();
+            newResult.setDistanceInMeters((double) i + indexOffset);
+            resultList.add(newResult);
+        }
+        return resultList;
+    }
+
+    /*
+     private QueryResultPaginator queryResultPaginator;
```
emanns95: ❓ Should this be removed?

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
```diff
@@ -0,0 +1,395 @@
...
+        List<Result> pageTwo = paginator.next();
+        assertThat(pageTwo, hasSize(5));
+        assertThat(paginator.hasNext(), equalTo(false));
+        assertThat(paginator.getBufferSize(), equalTo(0));
+        assertThat(paginator.getIndexOffset(), equalTo(20));
+
+        try {
+            paginator.next();
+        } catch (NoSuchElementException e) {
+            //do nothing
```
emanns95: If we don't expect this exception to actually happen, then we should have the test method throw it. If we are expecting it, we should make sure that we're catching it and fail if it isn't thrown

rzwiefel:
@mweser @lessarderic I'm still not seeing any benefit to the second, broken out class `QueryResultPaginator`.  It seems like we could achieve that by rolling the logic back into the `QueryResultIterator` as it originally was. The `QueryResultPaginator` class implements the Iterator interface in action, but not in signature. The only difference between the two classes is the ability to return a single List of items, which could simply be implemented as a utility method on the original class like such:
```java
    public Iterator<List<Result>> getBatchedResults() {
        return new BatchedResultQueryIterator(new ResultQueryIterator());
    }
    
    class BatchedResultQueryIterator implements Iterator<List<Result>> {

        ResultQueryIterator iterator;
        
        BatchedResultQueryIterator(ResultQueryIterator iterator) {
            this.iterator = iterator;
        }

        @Override
        public boolean hasNext() {
            iterator.hasNext();
        }

        @Override
        public List<Result> next() {
            StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator,
                    Spliterator.DISTINCT | Spliterator.ORDERED), false)
                    .limit(query.getPageSize())
        }
    }
```

There could even be a parameter added if we want to allow modifying chunksize, etc. Does this seem reasonable?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.stream.StreamSupport;
+
+import org.apache.commons.lang3.Validate;
+
+import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Result;
+import ddf.catalog.operation.impl.QueryImpl;
+
+/**
+ * Effectively a cursor over the iteratorResults of a filter that automatically pages through all iteratorResults
```
rzwiefel: Should be marked as experimental

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
rzwiefel: Should be marked as Experimental

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+     * Assume the current index is pointed to the first result
+     * that has not been retrieved.
+     * Update the current index after the results are fetched.
+     * Add the fetched results to the queriedResultsBuffer.
+     * If no results were returned, assume there are no more results possible.
+     */
+    private void fetchNext() {
+
+        queryCopy.setStartIndex(indexOffset);
+        List<Result> resultList = queryCatalogFramework();
```
rzwiefel: ❓ Is there any reason not to just use the list of results returned by the framework as the backing data structure for each chunk instead of dumping them into an intermediate BlockingQueue and then pulling them back out and copying them back into another list?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+        queriedResultsBuffer = new LinkedBlockingDeque<>();
+    }
+
+    // TODO: 6/27/17 Move this method over to QueryResultIterable inner class
+    /**
+     * If there are pages beyond the current page, returns true.
+     *
+     * @return True if there are results beyond the current page
+     */
+    /**
```
coyotesqrl: ❗️ Remove this extra javadoc.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+
+        Validate.inclusiveBetween(1,
+                Integer.MAX_VALUE,
+                indexOffset,
+                "Query request start index must be greater than zero");
+
+        queriedResultsBuffer = new LinkedBlockingDeque<>();
+    }
+
+    // TODO: 6/27/17 Move this method over to QueryResultIterable inner class
```
coyotesqrl: ❗️ This todo is defunct now, yes?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+    private final QueryImpl queryCopy;
+
+    private int indexOffset;
+
+    private int pageSize;
+
+    //Set to true when a query returns zero results.
+    private boolean isExhausted;
+
+    private BlockingQueue<Result> queriedResultsBuffer;
```
coyotesqrl: ✏️ Queue might be overkill (though it's not a terrible choice) as you're really just doing combos of addAll/drainAll. It's fine to leave it, but you could come up with other solutions using lists instead.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+        } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            throw new CatalogRuntimeException(e);
+        }
+    }
+
+    int getBufferSize() {
+        return queriedResultsBuffer.size();
+    }
+
+    QueryImpl copy(QueryImpl query) {
```
coyotesqrl: ❗️ Please reorder methods such that public ones are first, then protected, then package-private, then private.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Yes, as there may be extraneous query results from CatalogFramework which may exceed the pageSize, in which case they should be stored in the buffer until the next page is requested.

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
mweser: Currently taking a look at Aaron's unit tests and will make sure we test for the exception.

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
mweser: Currently in progress with merging my unit tests in with Aaron's to ensure maximum coverage.

catalog/opensearch/catalog-opensearch-endpoint/src/main/resources/ddf-os.xml
mweser: No 😶

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
```diff
@@ -0,0 +1,395 @@
...
+        catalogFrameworkMock = mock(CatalogFramework.class);
+        QueryRequestImpl queryRequestMock = mock(QueryRequestImpl.class);
+        queryResponseMock = mock(QueryResponseImpl.class);
+        queryMock = mock(QueryImpl.class);
+        when(catalogFrameworkMock.query(any(QueryRequest.class))).thenReturn(queryResponseMock);
+        when(queryRequestMock.getQuery()).thenReturn(queryMock);
+    }
+
+    @Test
+    public void testhasNext() throws Exception {
```
coyotesqrl: ❗️ I'd like to see a test that calls `hasNext()` repeatedly with no calls to `next()` between and ensures that the catalogFramework is ONLY queried once.

distribution/docs/src/main/jdocs/content/_endpoints/opensearch-endpoint-contents.adoc
```diff
@@ -179,12 +179,12 @@ Make sure to repeat the starting point as the last point to close the polygon.
 
 Sorting by date will sort the effective date.
 
-|`maxResults`
+|`pageSize`
```
coyotesqrl: ❓ 1. Can we change this or does changing this parameter count as a change to the public API? If so, I suggest we _add_ `pageSize` and ensure that the semantics between `pageSize` and `maxResults` are logically distinct.
❓ 2. If we're going to change (or add), then the abbreviated form should match. `ps` instead of `mr`.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+             return true;
+         }
+
+         if (isExhausted) {
+             return false;
+         }
+
+         //Buffer is empty, but query might not be exhausted.
+         fetchNext();
+         return notEmpty();
```
mweser: I don't see how that would be logically equivalent, as `isExhausted` is only set when there are no more results from the CatalogFramework, whereas the `notEmpty()` method checks if there are still results in the buffer. `hasNext()` should return true as long as there are results in the buffer, but not necessarily if the CatalogFramework has exhausted itself of results.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+     * @return Collection of results
+     * @throws NoSuchElementException if the iteration has no more elements
+     */
+    public List<Result> next() throws NoSuchElementException {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+
+        while (!isExhausted && getBufferSize() < pageSize) {
+            fetchNext();
```
mweser: I agree it can be cleaned up quite a bit, as it's pretty convoluted and does not have a great deal of focus.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
coyotesqrl: To get here, you had to call `fetchNext()` after having determined previously that the queue had been empty. After your call to `fetchNext()`, one of the following conditions will be true:

1. The queue IS empty AND `isExhausted` is true.
2. The queue IS NOT empty AND `isExhausted` is not true.

Unless you can show me a path by which you've gotten to this line and a different permutation of those two factors has occurred.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Agreed

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
lessarderic: Technically this is not part of an API bundle so I don't think it needs to be marked.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,104 @@
...
+import ddf.catalog.data.Result;
+import ddf.catalog.operation.impl.QueryImpl;
+
+/**
+ * A cursor over the iteratorResults of a filter that automatically pages through all iteratorResults
+ * <br/>
+ * Throws a {@link CatalogRuntimeException} if anything goes wrong during iteration or
+ * querying
+ */
+public class QueryResultIterable implements Iterable<Result> {
```
lessarderic: ✏️  Should make this description a little more generic, e.g., class that provides a way to iterate over individual catalog results?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private BlockingQueue<Result> queriedResultsBuffer;
+
+    /**
+     * QueryResultPaginator takes in the CatalogFramework object and a Query Request and
+     * queries for the number of results needed. This is done using the pageSize specified
+     * in the Query object. The Paginator will return a collection of results of the
+     * specified page size.
+     * <p>
+     * The startIndex is the first index in the results gathered from the CatalogFramework
```
lessarderic: ✏️  Probably should make it clear that `startIndex` comes from the `Query` object as well.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+     * <p>
+     * The endIndex is the last index in the results gathered from the CatalogFramework, which
```
lessarderic: ❗️ Don't think this applies anymore.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+     * may not be reached if the pageSize is smaller than the difference between the endIndex
+     * and startIndex.
+     * <p>
+     * The resultsRemaining are the total results not yet returned in a collection of the size
```
lessarderic: ❗️ Nor this.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+     * of a page. This value is tracked through the hasNextPage() method to inform the client that
+     * there are valid results to continue querying.
+     *
+     * @param catalogFramework
```
lessarderic: ❗️ We should add descriptions here. And we may want to put the `startIndex` and `pageSize` comments on the `query` parameter description instead of in the class description.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,186 @@
...
+        Validate.notNull(query);
+
+        this.catalogFramework = catalogFramework;
+        this.queryCopy = copy(query);
+
+        isExhausted = false;
+        pageSize = queryCopy.getPageSize();
+        indexOffset = queryCopy.getStartIndex();
+
+        Validate.inclusiveBetween(1,
```
lessarderic: ✏️  Should we move this validation at the top of the method with the other ones?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Removing this loop in fact causes our "testDifferentSizePages" test to fail. Will see if this is an issue with the test, but I believe that the test is fine and this is something that needs to remain.

coyotesqrl: Please walk through and determine exactly how that is happening. I'm quite curious about it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/CatalogRuntimeException.java
```diff
@@ -11,22 +11,18 @@
...
-package org.codice.ddf.commands.util;
+package ddf.catalog.util.impl;
 
-public class CatalogCommandRuntimeException extends RuntimeException {
-    public CatalogCommandRuntimeException() {
-        super();
-    }
-
-    public CatalogCommandRuntimeException(String message) {
+public class CatalogRuntimeException extends RuntimeException {
```
coyotesqrl: I think that's a better name.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,104 @@
...
+        return new ResultQueryIterator();
+    }
+
+    @Override
+    public Spliterator<Result> spliterator() {
+        int characteristics = Spliterator.DISTINCT;
+        return Spliterators.spliteratorUnknownSize(this.iterator(), characteristics);
+    }
+
+    public Stream<Result> stream() {
```
mweser: I'm going to go ahead and remove it.

mweser:
build now

mweser:
Passed swarm server build this morning, running again now in case changes broke anything.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/CatalogRuntimeException.java
```diff
@@ -11,22 +11,18 @@
...
+    public CatalogRuntimeException(String message) {
```
mweser: Removing this constructor breaks ImportCommand.java, so I have left it in https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java#L96

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -210,15 +210,15 @@ protected Object executeWithSubject() throws Exception {
         return null;
     }
 
-    private File initImportFile(String importFile) {
+    private File initImportFile(String importFile) throws IllegalArgumentException {
```
rzwiefel: ✏️ Declaring the runtime exception is not necessary

mweser:
build now

shaundmorris:
build now

mweser:
build now

mweser:
Successful on full swarm build.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+
+        return notEmpty();
+    }
+
+    public List<Result> next() throws NoSuchElementException {
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+
+        while (!noMoreCatalogFrameworkResultsToFetch && getBufferSize() < pageSize) {
```
coyotesqrl: ❓ On my last review, I had expressed confusion over this loop and @mweser had indicated that the `testDifferentSizePages` test failed if it were removed.

With the cleaner code, I now see that if you'd gotten more than `pageSize` results back from your previous query of the framework that you'd have a partially filled buffer at this point; that would mean that you'd return from `hasNext()` on the test of `notEmpty()`. So I can see that you might want to force a single execution of fetch here if these conditions are true. Probably clearer to change from a `while` to an `if`.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,111 @@
...
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.util.impl;
+
+import java.util.Iterator;
+import java.util.Spliterator;
+import java.util.Spliterators;
+
+import org.apache.commons.lang3.Validate;
```
brendan-hofmann: ❗️ Validate is available in lang 2.6 (which we export), so we could avoid embedding lang3 here.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    @Override
+    public Iterator<Result> iterator() {
+        return new QueryResultIterator(new QueryResultPaginator(catalog, query));
+    }
+
+    @Override
+    public Spliterator<Result> spliterator() {
+        int characteristics = Spliterator.DISTINCT;
+        return Spliterators.spliteratorUnknownSize(this.iterator(), characteristics);
```
brendan-hofmann: ❓ Should we respect the page size here so that clients with a performance need could benefit from the optimization, or provide an additional parameter for spliterator size?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Alrighty will change to `if` to make more clear.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+ * This class controls paging of {@link CatalogFramework} queries by batching them into lists of
+ * results, independent of the number of results the Catalog Framework returns with each query.
+ * This helps clients guarantee a particular batch size rather than getting fewer results than
+ * expected from the Catalog Framework.
+ */
+public class QueryResultPaginator {
+
+    private final CatalogFramework catalogFramework;
+
+    // Used to query the catalog framework and adjust the starting index of the query without
```
brendan-hofmann: ❓ why are these inline comments instead of javadoc comments? Personally I prefer the latter.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+    private final QueryImpl queryCopy;
+
+    // Used to track the start index of the next query and ensure that the correct query results
+    // are returned with no repetition of results.
+    private int indexOffset;
+
+    private int pageSize;
+
+    // Set to true when a query returns zero results and no more results can be fetched from
+    // Catalog Framework..
```
brendan-hofmann: ❓ extra period, or incomplete ellipsis? 😃

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
+
+/**
+ * This class controls paging of {@link CatalogFramework} queries by batching them into lists of
+ * results, independent of the number of results the Catalog Framework returns with each query.
+ * This helps clients guarantee a particular batch size rather than getting fewer results than
+ * expected from the Catalog Framework.
+ */
+public class QueryResultPaginator {
```
lessarderic: ✏️  We probably should make this package-private until we decide how clients do paging.

mweser:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+    // are returned with no repetition of results.
+    private int indexOffset;
+
+    private int pageSize;
+
+    // Set to true when a query returns zero results and no more results can be fetched from
+    // Catalog Framework..
+    private boolean noMoreCatalogFrameworkResultsToFetch;
+
+    // Used to buffer fetched results from the Catalog Framework and hold them until they can be
```
brendan-hofmann: ✏️ If this was a javadoc the reference to catalog framework could be an `@link` instead.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+
+    // Used to query the catalog framework and adjust the starting index of the query without
+    // modifying the original query.
+    private final QueryImpl queryCopy;
+
+    // Used to track the start index of the next query and ensure that the correct query results
+    // are returned with no repetition of results.
+    private int indexOffset;
+
+    private int pageSize;
```
lessarderic: ✏️  This can be made `final`.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    static class QueryResultIterator implements Iterator<Result> {
+
+        private QueryResultPaginator queryResultPaginator;
+
+        private Iterator<Result> queriedResults = null;
+
+        public QueryResultIterator(QueryResultPaginator queryResultPaginator) {
+            this.queryResultPaginator = queryResultPaginator;
```
brendan-hofmann: ✏️ It looks like we could validate non-null and make this field final.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
lessarderic: @coyotesqrl, the loop is needed to account for the scenario where the buffer is empty and we need to query the catalog framework more than once to get enough results to fulfill the page size requirement.
We'll add a unit test for that.

catalog/core/catalog-core-api/src/main/java/ddf/catalog/CatalogFramework.java
```diff
@@ -380,6 +380,12 @@ SourceInfoResponse getSourceInfo(SourceInfoRequest sourceInfoRequest)
      * <li/>Return the last {@link QueryResponse} to the caller.
      * </ol>
      * </p>
+     * <p>
+     * <b>Important:</b> Implementations are free to limit the number of results returned
```
brendan-hofmann: ❗️ This is dev docs/release notes worthy.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
lessarderic: @rzwiefel, given the fact that we may have to query the CF multiple times to get enough results to get a full page, we will need to merge multiple lists returned by the CF anyways. And we probably shouldn't be changing the returned list, so always making a copy might be a safer option.

catalog/spatial/geocoding/spatial-geocoding-plugin/pom.xml
brendan-hofmann: ❗️ I don't think we need to embed commons-lang here if we use lang 2.6 instead of lang3.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
lessarderic: We're working on the Spliterator part. It'll be included in a separate PR.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
lessarderic: Mostly because those are implementation details and I didn't feel like they should be part of the class documentation. If I add them there and the implementation changes, they may not get updated and become stale.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,111 @@
...
+        return new QueryResultIterator(new QueryResultPaginator(catalog, query));
+    }
+
+    @Override
+    public Spliterator<Result> spliterator() {
+        int characteristics = Spliterator.DISTINCT;
+        return Spliterators.spliteratorUnknownSize(this.iterator(), characteristics);
+    }
+
+    static class QueryResultIterator implements Iterator<Result> {
```
rzwiefel: ❓ What is the point of the `queryResultPaginator`?

This is still just an  inner class that takes the other class and embeds it inside of it. Why not just make this class contain all the logic instead of putting it elsewhere and just wrapping it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
rzwiefel: ❗️ This class still follows the iterator API almost exactly and yet does not implement iterator?

lessarderic: This will probably become clearer when we add the Spliterator implementation in the next PR. If we end up not using the Spliterator to do batching then we'll most likely find a way to combine this class with the other one.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
mweser: Most likely will be remaining as line comments, but if changed to JavaDoc, will add link.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
rzwiefel: Why would we need to query multiple times when someone is using this as an iterator though? We should be able to return the last as each result set comes back from the query.

rzwiefel: This seems like we're forcing a client to wait for possibly multiple queries to return in order to fill the (internal) buffer a certain amount before returning any results. If someone is simply iterating through the results of a query, it now could have to make multiple round trips just in order to return (possibly) a single result?

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultIteratorTest.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @Test
+    public void testNextResultsExistExpectResult() {
+
+        boolean isResultsListEmpty = false;
+
+        when(queriedResultsMock.hasNext()).thenReturn(!isResultsListEmpty);
+        when(queriedResultsMock.next()).thenReturn(resultMock);
+        queryIterator.setQueriedResults(queriedResultsMock);
```
lessarderic: ❗️ Instead of setting this here, we probably want to mock the call to the paginator and look at the sequence of calls from the iterator and paginator point of view.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            return queriedResults.next();
+        }
+
+        private boolean isQueriedResultListNullOrEmpty() {
+            return queriedResults == null || !queriedResults.hasNext();
+        }
+
+        void setQueriedResults(Iterator<Result> queriedResults) {
```
lessarderic: ❗️ We shouldn't need this method. Everything should be tested through the paginator mock and public API.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
coyotesqrl: It's unfortunate that we have to handle both the "more than pageSize returned" and "less than pageSize returned" scenarios.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,160 @@
...
+        if (!hasNext()) {
+            throw new NoSuchElementException();
+        }
+
+        while (!noMoreCatalogFrameworkResultsToFetch && getBufferSize() < pageSize) {
+            fetchNext();
+        }
+
+        List<Result> queriedResultsBufferCopy = new ArrayList<>(pageSize);
+        queriedResultsBuffer.drainTo(queriedResultsBufferCopy, pageSize);
```
bdeining: Could log the result of the `drainTo` to suppress this warning

coyotesqrl: That's just log noise, at any level. Better to mark it as a false positive in SonarQube.

catalog/core/catalog-core-commands/src/test/groovy/org/codice/ddf/commands/catalog/ExportCommandSpec.groovy
```diff
@@ -84,6 +80,14 @@ class ExportCommandSpec extends spock.lang.Specification {
 
     def "Test export no items"() {
         setup:
+        catalogFramework.query(_ as QueryRequest) >> { QueryRequest req ->
+            System.err.println("catalogFramework.query() called!")
```
bdeining: I know we try to avoid `System.out`, but not sure about `System.err`.  Is this necessary?

catalog/spatial/geocoding/spatial-geocoding-plugin/pom.xml
```diff
@@ -43,6 +45,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
```
bdeining: outside of scope, but i'm not sure this needs to be listed as a dependency anymore

catalog/core/catalog-core-commands/src/test/groovy/org/codice/ddf/commands/catalog/ExportCommandSpec.groovy
lessarderic: Nope. Left over from debugging the tests. Will remove. Good catch!

lessarderic:
@rzwiefel, I agree with the concern regarding the separate `QueryResultPaginator`. However, we have another PR to implement the `Spliterator` that will also be using the separate `QueryResultPaginator`. If it's ok with you, I'd like to merge this PR as-is and rework the design as part of the other PR if needed.

rzwiefel:
@lessarderic As long as we ensure the existing thing using this class still functions properly then thats fine by me. :+1:

catalog/core/catalog-core-commands/src/test/groovy/org/codice/ddf/commands/catalog/ExportCommandSpec.groovy
```diff
@@ -162,17 +167,20 @@ class ExportCommandSpec extends spock.lang.Specification {
         thrown(IllegalStateException)
     }
 
+    // TODO - Ignored until DDF-3123 has been addressed
```
vinamartin: ✏️  Always helpful to link this on the ticket so it won't get ignored forever.

lessarderic: Done.

mweser:
@rzwiefel The first thing we did was make sure our design of the paginator did not break the functionality of the ExportCommand. The point of this new class is to sit behind the scenes and not interfere with the iterator API. It simply guarantees the number of results a client would expect, as the Catalog Framework does not necessarily give all the results queried. A developer only needs to know how to plug an iterator into a foreach loop and the paginator does the work.

rzwiefel:
@mweser I'm aware that it **works**. My issue lies in that it's not an **efficient** iterator as is. However, that can be resolved later.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -145,7 +144,7 @@ protected Object executeWithSubject() throws Exception {
                     try {
                         metacard = transformer.transform(new UncloseableBufferedInputStreamWrapper(
                                 zipInputStream), id);
-                    } catch (IOException | CatalogTransformerException e) {
```
rzwiefel: ❓ Why was CatalogTransformerException removed?

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultIteratorTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        queryIterator = new QueryResultIterator(paginatorMock);
+    }
+
+    @Test
+    public void testNextResultsExistExpectResult() {
+
+        boolean isResultsListEmpty = false;
+
+        when(queriedResultsMock.hasNext()).thenReturn(!isResultsListEmpty);
```
mweser: Residual from abstracting the tests into a single method. Will clean up.

catalog/core/catalog-core-api/src/main/java/ddf/catalog/CatalogFramework.java
mweser: @ricklarsen What would be your take on how to approach this in the documentation?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultIterable.java
ahoffer: It does look the the wrapper class could go away. It would probably be most idiomatic to rename ... Paginator to ...Iterator. 

One other thing, how tired into the query request should this be? If kind of feels like the query request impl should return an object that implements iterable. I'm on my phone, and it's not easy to check if that even makes sense for these classes.

mweser:
Superseded by https://github.com/codice/ddf/pull/2133

PR #2083, opened 2017-06-21
DDF-3088 added password encryption/decryption to configuration files that are stored on disk

stustison:
@pklinef @lessarderic @beyelerb  @brendan-hofmann 
https://github.com/orgs/codice/teams/security/members
@coyotesqrl
@stustison
[DDF-3088](https://codice.atlassian.net/browse/DDF-3088)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/callables/GetConfigurationProperties.java
+++ b/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/callables/GetConfigurationProperties.java
@@ -13,7 +13,10 @@
  */
 package org.codice.ddf.itests.common.callables;
 
+import java.util.Arrays;
 import java.util.Dictionary;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.Callable;
 
 import org.osgi.service.cm.Configuration;
@@ -65,11 +68,18 @@ public class GetConfigurationProperties implements Callable<Dictionary<String, O
         String query = String.format("(%s=%s)", propertyName, propertyValue);
         Configuration[] configurations = configAdmin.listConfigurations(query);
 
-        if ((configurations == null) || (configurations.length == 0)) {
+        if (configurations == null) {
             return null;
         }
 
-        if (configurations.length > 1) {
+        //multiple configurations are returned because there are multiple persistence managers
+        Set<Configuration> configurationSet = new HashSet<>(Arrays.asList(configurations));
+
+        if (configurationSet.isEmpty()) {
+            return null;
+        }
+
+        if (configurationSet.size() > 1) {
             LOGGER.error(String.format("Multiple Configuration objects returned for query %s",
                     query));
             throw new IllegalArgumentException("Property name/value pair isn't unique");
--- a/platform/admin/admin-configuration-passwordencryption/pom.xml
+++ b/platform/admin/admin-configuration-passwordencryption/pom.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- /**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/ -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>admin</artifactId>
+        <groupId>ddf.admin</groupId>
+        <version>2.11.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <name>DDF :: Admin :: Configuration Password Encryption</name>
+    <artifactId>admin-configuration-passwordencryption</artifactId>
+    <packaging>bundle</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.felix</groupId>
+            <artifactId>org.apache.felix.configadmin</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>ddf.security.encryption</groupId>
+            <artifactId>security-encryption-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ddf.admin.core</groupId>
+            <artifactId>admin-core-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ddf.security</groupId>
+            <artifactId>ddf-security-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Export-Package />
+                    </instructions>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.64</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.88</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.57</minimum>
+                                        </limit>
+                                    </limits>
+                                </rule>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
+++ b/platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
@@ -1,0 +1,117 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.admin.configuration.password.encryption;
+
+import static org.osgi.framework.Constants.SERVICE_PID;
+import static org.osgi.service.cm.ConfigurationAdmin.SERVICE_FACTORYPID;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Dictionary;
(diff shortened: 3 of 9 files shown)


stustison:
build now

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+    @Override
+    public Dictionary load(String pid) throws IOException {
+        Dictionary dictionary = persistenceManager.load(pid);
+        updatePasswords(pid, dictionary, encryptionService::decrypt);
+        return dictionary;
+    }
+
+    @Override
+    public Enumeration getDictionaries() throws IOException {
+        return persistenceManager.getDictionaries();
```
paouelle: ❓ Are we not decrypting those dictionaries?

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public boolean exists(String pid) {
+        return persistenceManager.exists(pid);
+    }
+
+    @Override
+    public Dictionary load(String pid) throws IOException {
+        Dictionary dictionary = persistenceManager.load(pid);
```
paouelle: ❓ Are we not concerned that every time we do a load() or a store() below, we always create a new hashtable wether or not encryption/decryption is required?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/callables/GetConfigurationProperties.java
```diff
@@ -65,11 +68,18 @@ public GetConfigurationProperties(ConfigurationAdmin configAdmin, String propert
         String query = String.format("(%s=%s)", propertyName, propertyValue);
         Configuration[] configurations = configAdmin.listConfigurations(query);
 
-        if ((configurations == null) || (configurations.length == 0)) {
+        if (configurations == null) {
             return null;
         }
 
-        if (configurations.length > 1) {
```
stustison: It isn't 1 anymore. Because I added a new persistence manager, the size will always be at least 2. I just use the set to remove duplicates and then if the size is still greater than 1, we error out.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/callables/GetConfigurationProperties.java
```diff
@@ -65,11 +68,18 @@ public GetConfigurationProperties(ConfigurationAdmin configAdmin, String propert
...
+        //multiple configurations are returned because there are multiple persistence managers
+        Set<Configuration> configurationSet = new HashSet<>(Arrays.asList(configurations));
+
+        if (configurationSet.size() == 0) {
```
stustison: Yep, just wasn't thinking when I clobbered the array code.

platform/admin/admin-configuration-passwordencryption/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+        </dependency>
+        <dependency>
+            <groupId>ddf.security</groupId>
+            <artifactId>ddf-security-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>findbugs</artifactId>
+            <version>3.0.1</version>
```
stustison: I'll use the variable.

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+public class PasswordEncryptingPersistenceManager implements PersistenceManager {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(
+            PasswordEncryptingPersistenceManager.class);
+
+    private PersistenceManager persistenceManager;
+
+    //findbugs incorrectly flagged this as an unread field and there was no way to skip it with the annotations that worked
+    //findbugs is thus disabled for this class
+    private EncryptionService encryptionService;
```
stustison: That was my assumption as well. I tried everything I could think of to fix it without changing my code (because I like my code) or to just skip the single issue that it was flagging, but nothing worked.

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
stustison: Woops, no I probably need to. The funny thing is that it actually doesn't matter for this method. The persistence managers are wrapped with a caching manager, which basically overrides the load and getDictionaries methods anyways. The load one is hit during startup, but that is it. Also, how I managed to miss this method.

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
stustison: The load shouldn't create one anyways as far as I can tell. I'm not super worried about the store creating a new Hashtable unless you can think of a reason to be.

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+        String trimmedPid = pid.substring(0, (indexOfDot == -1) ? pid.length() - 1 : indexOfDot);
+        List<Service> services = null;
+        try {
+            services = getServices(trimmedPid);
+        } catch (PrivilegedActionException e) {
+            LOGGER.info(
+                    "Unable to find metatype for service call. Passwords in configuration could be unencrypted.",
+                    e);
+        }
+        if (services != null && services.size() == 1) {
```
stustison: That would be undefined, since you should only ever have a single service for any particular PID. They are _supposed_ to be unique.

platform/admin/admin-configuration-passwordencryption/src/test/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManagerTest.java
```diff
@@ -0,0 +1,173 @@
...
+        verify(persistenceManager, times(1)).exists("123");
+    }
+
+    @Test
+    public void testDelete() throws IOException {
+        passwordEncryptingPersistenceManager.delete("123");
+        verify(persistenceManager, times(1)).delete("123");
+    }
+
+    class ServiceImpl extends HashMap<String, Object> implements Service {
```
brendan-hofmann: ❓ Shouldn't we just mock the existing classes instead of declaring new ones here?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/callables/GetConfigurationProperties.java
```diff
@@ -65,11 +68,18 @@ public GetConfigurationProperties(ConfigurationAdmin configAdmin, String propert
         String query = String.format("(%s=%s)", propertyName, propertyValue);
         Configuration[] configurations = configAdmin.listConfigurations(query);
 
-        if ((configurations == null) || (configurations.length == 0)) {
+        if (configurations == null) {
```
stustison: There are two persistence managers now and can return the same object. Also, it is just test code, so I wasn't worrying too much about efficiency (it is only used in 2 tests).

stustison:
Not sure how I'd write an itest for it. You'd need to dig down into the bundle cache to see if they are encrypted. Easier to just verify that they are encrypted manually.

platform/admin/admin-configuration-passwordencryption/src/main/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public Enumeration getDictionaries() throws IOException {
+        return persistenceManager.getDictionaries();
+    }
+
+    @Override
+    public void store(String pid, Dictionary dictionary) throws IOException {
+        Dictionary tempDictionary = new Hashtable();
```
stustison: No. 👎

platform/admin/admin-configuration-passwordencryption/src/test/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManagerTest.java
stustison: Why? It is easier to create a new impl with a single line than it is to mock one out and mock all of the separate methods that are being used internally.

stustison:
build now

platform/admin/admin-configuration-passwordencryption/src/test/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManagerTest.java
brendan-hofmann: I agree it is easier, but the test is also now decoupled from the impls it is testing interactions with. If the real impls ever change, this test is less likely to catch that.

Admittedly that is highly unlikely for these specific cases so it is more of a general principle thing. In general it is a bad pattern, but I don't feel strongly about this specific instance.

stustison: I disagree, the test is only testing at the interface level and doesn't need to know anything about the actual impls, which is exactly why I can create my own impls here.

stustison:
build now

platform/admin/admin-configuration-passwordencryption/src/test/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManagerTest.java
```diff
@@ -0,0 +1,182 @@
...
+        encrypted.put("notpassword", "foo");
+        Hashtable noPassword = new Hashtable();
+        noPassword.put("service.factoryPid", "pid.uuid");
+        noPassword.put("random", "stuff");
+        Hashtable nullPassword = new Hashtable();
+        noPassword.put("notstring", new ArrayList<String>());
+        Enumeration enumeration = Collections.enumeration(Arrays.asList(encrypted,
+                noPassword,
+                nullPassword));
+        when(persistenceManager.load("123")).thenReturn(encrypted);
```
lessarderic: ✏️  Can we make the pids a little more descriptive instead of using `123`, `321`, etc., or use constants?

platform/admin/admin-configuration-passwordencryption/src/test/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManagerTest.java
```diff
@@ -0,0 +1,182 @@
...
+        passwordEncryptingPersistenceManager.exists("123");
+        verify(persistenceManager, times(1)).exists("123");
+    }
+
+    @Test
+    public void testDelete() throws IOException {
+        passwordEncryptingPersistenceManager.delete("123");
+        verify(persistenceManager, times(1)).delete("123");
+    }
+
```
lessarderic: ❓ Should we test the `PrivilegedActionException` in `updatePasswords`? If that exception isn't properly handled couldn't we end up with a vulnerability?

LinkMJB:
build now

LinkMJB:
build now

LinkMJB:
build now

platform/admin/admin-configuration-passwordencryption/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Export-Package />
+                    </instructions>
+                </configuration>
+            </plugin>
+            <plugin>
```
stustison: There is a comment in the code. If someone did decide to turn this back on, it wouldn't build.

platform/admin/admin-configuration-passwordencryption/src/test/java/org/codice/ddf/admin/configuration/password/encryption/PasswordEncryptingPersistenceManagerTest.java
stustison: The only way that we'd end up with an exception is if the system subject was not able to be retrieved. If that was the case, the system would most likely be unusable anyways.

stustison:
build now

platform/admin/admin-configuration-passwordencryption/pom.xml
lessarderic: But if we put the comment in the pom then people will know right away why it's been turned off instead of turning it on thinking it's a mistake, seeing the build fail, trying to figure out what's wrong and finally find the comment in the code :-)

stustison:
build now

stustison:
build now

paouelle:
The migration:export command should be modified to re-encrypted the passwords.

stustison:
build now

stustison:
To be clear, I'm not worrying about the configs in etc just yet and there are some karaf things that can encrypt what is in etc already. That will all be covered in a separate PR.

lessarderic:
Agreed with @stustison that remaining concerns will be addressed in separate PRs.

PR #2084, opened 2017-06-22
DDF-3101 changed regex to check for hostname only connection to STS, added test for constraint interceptor

stustison:
@ahoffer @lessarderic @brendan-hofmann 
@codice/security 
@clockard
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/etc/system.properties
+++ b/distribution/ddf-common/src/main/resources/etc/system.properties
@@ -284,7 +284,8 @@ karaf.secured.services=(&(osgi.command.scope=*)(osgi.command.function=*))
 #java.security.policy=${karaf.home}/etc/all.policy
 #org.osgi.framework.security=osgi
 #org.osgi.framework.trust.repositories=${karaf.home}/etc/trustStore.ks
-ws-security.subject.cert.constraints = .*
+# pipe (|) separated list of regex patterns for DNs that are allowed to connect to the Security Token Service
+ws-security.subject.cert.constraints = .*CN=${org.codice.ddf.system.hostname}.*
 security.audit.roles=group,admin,manager,viewer,system-admin,system-history,systembundles
 user.language="en"
 user.country="US"
--- a/platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/SubjectDNConstraintsInterceptor.java
+++ b/platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/SubjectDNConstraintsInterceptor.java
@@ -19,6 +19,7 @@ import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.cxf.interceptor.Fault;
 import org.apache.cxf.interceptor.security.AccessDeniedException;
 import org.apache.cxf.message.Message;
@@ -41,11 +42,10 @@ public class SubjectDNConstraintsInterceptor extends AbstractPhaseInterceptor<Me
     }
 
     /**
-     * Return true if the provided certificate matches the regular expression
+     * Checks if the provided certificate matches the regular expression
      * defined in the Subject DN Certificate Constraints.
      *
      * @param message
-     * @return true if the certificate matches the constraints
      */
     @Override
     public void handleMessage(Message message) throws Fault {
@@ -72,10 +72,10 @@ public class SubjectDNConstraintsInterceptor extends AbstractPhaseInterceptor<Me
         }
     }
 
-    public Collection<Pattern> setSubjectDNPatterns(String subjectDNConstraints) {
+    Collection<Pattern> setSubjectDNPatterns(String subjectDNConstraints) {
         ArrayList<Pattern> subjectDNPatterns = new ArrayList<>();
-        if (subjectDNConstraints != null) {
-            String[] patterns = subjectDNConstraints.split(",");
+        if (StringUtils.isNotEmpty(subjectDNConstraints)) {
+            String[] patterns = subjectDNConstraints.split("\\|");
             for (String pattern : patterns) {
                 Pattern p = Pattern.compile(pattern);
                 subjectDNPatterns.add(p);
--- a/platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
+++ b/platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
@@ -1,0 +1,113 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.security.sts;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.security.cert.X509Certificate;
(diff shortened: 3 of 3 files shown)


stustison:
build now

platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
```diff
@@ -0,0 +1,126 @@
...
+        when(message.get(WSHandlerConstants.SIG_SUBJECT_CERT_CONSTRAINTS)).thenReturn("CN=cert");
+        Request request = mock(Request.class);
+        when(message.get("HTTP.REQUEST")).thenReturn(request);
+        X509Certificate x509Certificate = mock(X509Certificate.class);
+        X500Principal x500Principal = new X500Principal("CN=cert");
+        when(x509Certificate.getSubjectX500Principal()).thenReturn(x500Principal);
+        when(request.getAttribute("javax.servlet.request.X509Certificate")).thenReturn(new X509Certificate[] {
+                x509Certificate});
+
+        subjectDNConstraintsInterceptor.handleMessage(message);
```
brendan-hofmann: ✏️ The javadoc says it returns true on success, but the signature is void. If the signature is correct, I guess there's nothing to really assert here, but maybe we should update the javadoc?

stustison: The signature is correct.

stustison:
build now

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -284,7 +284,8 @@ karaf.secured.services=(&(osgi.command.scope=*)(osgi.command.function=*))
 #java.security.policy=${karaf.home}/etc/all.policy
 #org.osgi.framework.security=osgi
 #org.osgi.framework.trust.repositories=${karaf.home}/etc/trustStore.ks
-ws-security.subject.cert.constraints = .*
+#ws-security.subject.cert.constraints = pipe (|) separated list of regex patterns for DNs that are allowed to connect to the Security Token Service
```
lessarderic: ✏️  Nit: To be consistent with the other comments in the file, should we aviod putting the property name in the comment and put a space between the `#` and the description?

platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
lessarderic: ❗️ We should update the `SubjectDNConstraintsInterceptor.handleMessage()` comment.

platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    SubjectDNConstraintsInterceptor subjectDNConstraintsInterceptor;
+
+    @Before
+    public void setup() {
+        subjectDNConstraintsInterceptor = new SubjectDNConstraintsInterceptor();
+    }
+
+    @Test
+    public void testHandleMessage() {
```
lessarderic: ✏️  Should we add `FromAllowedDn` to the name to clarify what this is testing?

platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
```diff
@@ -0,0 +1,126 @@
...
+        X500Principal x500Principal = new X500Principal("CN=cert");
+        when(x509Certificate.getSubjectX500Principal()).thenReturn(x500Principal);
+        when(request.getAttribute("javax.servlet.request.X509Certificate")).thenReturn(new X509Certificate[] {
+                x509Certificate});
+
+        subjectDNConstraintsInterceptor.handleMessage(message);
+    }
+
+    @Test
+    public void testHandleMessageCommaPattern() {
```
lessarderic: ✏️  Rename to `testHandleMessageWithMultipleCommaPatternsFromAllowedDN` or something along those lines?

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/SubjectDNConstraintsInterceptor.java
```diff
@@ -74,8 +75,8 @@ public void handleMessage(Message message) throws Fault {
 
     public Collection<Pattern> setSubjectDNPatterns(String subjectDNConstraints) {
```
lessarderic: ❗️ Is there a reason for this to be `public`? Couldn't find any other class calling this method.

lessarderic: ✏️  The name of this method is somewhat confusing. I was expecting it to set something on the object but it really just converts the `subjectDNConstraints` and returns a list of `Pattern`. Should we consider renaming it?

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/SubjectDNConstraintsInterceptor.java
```diff
@@ -74,8 +75,8 @@ public void handleMessage(Message message) throws Fault {
...
         ArrayList<Pattern> subjectDNPatterns = new ArrayList<>();
-        if (subjectDNConstraints != null) {
-            String[] patterns = subjectDNConstraints.split(",");
+        if (StringUtils.isNotEmpty(subjectDNConstraints)) {
```
lessarderic: ❓ Assuming we make this method `private`, can we replace the check on L56 with this one and get rid of this `if` statement? Or only have the check here, return an empty `List` if the constraint is `null` or empty and have the calling code trigger on that instead of the `subjectDNContraints` value? Just trying to avoid double `subjectDNContraints` validations that could become out-of-sync in the future.

platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
```diff
@@ -0,0 +1,126 @@
...
+        when(request.getAttribute("javax.servlet.request.X509Certificate")).thenReturn(new X509Certificate[] {
+                x509Certificate});
+
+        subjectDNConstraintsInterceptor.handleMessage(message);
+    }
+
+    @Test
+    public void testHandleMessageNoPattern() {
+        Message message = mock(Message.class);
+        when(message.get(WSHandlerConstants.SIG_SUBJECT_CERT_CONSTRAINTS)).thenReturn(null);
```
lessarderic: ❗️ We need to test the case where the constraints are empty.

platform/security/sts/security-sts-server/src/test/java/ddf/security/sts/SubjectDNConstraintsInterceptorTest.java
```diff
@@ -0,0 +1,126 @@
...
+        when(message.get(WSHandlerConstants.SIG_SUBJECT_CERT_CONSTRAINTS)).thenReturn("CN=certbad");
+        Request request = mock(Request.class);
+        when(message.get("HTTP.REQUEST")).thenReturn(request);
+        when(request.getAttribute("javax.servlet.request.X509Certificate")).thenReturn(null);
+
+        subjectDNConstraintsInterceptor.handleMessage(message);
+    }
+
+    @Test
+    public void testSetSubjectDNPatterns() {
```
lessarderic: ✏️  There's a lot of repeated setup code in this test class. We should consider extracting that code to a private method that takes parameters, use JUnit parametrized tests, or use Spock. Going with parametrized tests or Spock would help capture all the pattern tests in `testSetSubjectDNPatterns` (which really should be separate tests) and not have to expose the `setSubjectDNPatterns` method.

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/SubjectDNConstraintsInterceptor.java
lessarderic: ✏️  Per our coding standards, we should camel-case acronyms, so the class, variable, parameters and method names should `Dn` instead of `DN`. Not a big deal if it's too much trouble to rename all of those.

stustison:
build now

stustison:
build now

stustison:
build now

PR #2088, opened 2017-06-22
DDF-3071 refactor how function factories are handled and add divisibleBy function

ryeats:
#### What does this PR do?
Still have unit tests and itests to write but getting this PR up so i can get some feedback on my changes.

This PR adds a divisibleBy function for use in subsampling records with an ordinal. It also refactors how functions are plumbed. It should be possible to simply add a function to the GeoToolsFunctionFactory implement that function in SolrFilterDelegate, FilterToTextDelegate  and be able to do query catalog with cql or the csw endpoint with that function.

@tbatie @jrnorth @bdeining 
https://github.com/orgs/codice/teams/core-apis
@coyotesqrl
@pklinef
run the itests
[DDF-3071](https://codice.atlassian.net/browse/DDF-3071)
#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
@@ -16,6 +16,11 @@ package ddf.catalog.filter;
 import java.util.Date;
 
 /**
+ * <p>
+ * <b> This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ * </p>
+ *
  * Completes the fluent API to create a {@link org.opengis.filter.Filter} for a function that takes N parameters.
  */
 public interface ArgumentBuilder {
@@ -118,6 +123,14 @@ public interface ArgumentBuilder {
     ArgumentBuilder textArg(String text);
 
     /**
+     * Continues building the {@link ArgumentBuilder} based on a object value
+     *
+     * @param obj - {@link Object} argument to ArgumentBuilder on
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder objArg(Object obj);
+
+    /**
      * Continues building the {@link ArgumentBuilder} based on an attribute
      *
      * @param name - {@link String} argument to ArgumentBuilder on
@@ -130,7 +143,7 @@ public interface ArgumentBuilder {
      * <p>
      * Continue building a Filter with the "equal to" operator ( {@link org.opengis.filter.PropertyIsEqualTo})
      *
-     * @return {@link NumericalExpressionBuilder} to continue building the {@link org.opengis.filter.Filter}
+     * @return {@link EqualityExpressionBuilder} to continue building the {@link org.opengis.filter.Filter}
      */
     EqualityExpressionBuilder equalTo();
 
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
@@ -16,7 +16,7 @@ package ddf.catalog.filter;
 import java.util.Date;
 
 /**
- * Completes the fluent API to create a {@link org.opengis.filter.Filter} that takes N parameters.
+ * Completes the fluent API to create a {@link org.opengis.filter.Filter} for a function that takes N parameters.
  */
 public interface ArgumentBuilder {
 
@@ -126,11 +126,11 @@ public interface ArgumentBuilder {
     ArgumentBuilder attributeArg(String name);
 
     /**
-     * We only support propertyIsEqualTo(Function,val) currently so instead supporting ExpressionBuilder is() we just support equalTo
+     * Currently only propertyIsEqualTo(Function,val) is supported so instead supporting ExpressionBuilder is() the api just has equalTo
      * <p>
      * Continue building a Filter with the "equal to" operator ( {@link org.opengis.filter.PropertyIsEqualTo})
      *
-     * @return {@link NumericalExpressionBuilder} to continue building the {@link Filter}
+     * @return {@link NumericalExpressionBuilder} to continue building the {@link org.opengis.filter.Filter}
      */
     EqualityExpressionBuilder equalTo();
 
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
@@ -26,7 +26,7 @@ import java.util.Date;
 public interface ArgumentBuilder {
 
     /**
-     * continues building the {@link ArgumentBuilder} based on a float
+     * Continues building the {@link ArgumentBuilder} based on a float
      *
      * @param arg - float argument
      * @return {@link ArgumentBuilder}
@@ -101,7 +101,7 @@ public interface ArgumentBuilder {
     /**
      * Continues building the {@link ArgumentBuilder} based on a boolean value
      *
-     * @param arg - boolean value to ArgumentBuilder on
+     * @param arg - boolean argument
      * @return {@link ArgumentBuilder}
      */
     ArgumentBuilder boolArg(boolean arg);
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
@@ -1,0 +1,137 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.filter;
+
+import java.util.Date;
+
+/**
+ * Completes the fluent API to create a {@link org.opengis.filter.Filter} that takes N parameters.
+ */
+public interface ArgumentBuilder {
+
+    /**
+     * continues building the {@link ArgumentBuilder} based on a float
+     *
+     * @param arg - float argument
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder numberArg(float arg);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a double
+     *
+     * @param arg - double argument
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder numberArg(double arg);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a int
+     *
+     * @param arg - int argument
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder numberArg(int arg);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a short
+     *
+     * @param arg - short argument
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder numberArg(short arg);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a long
+     *
+     * @param arg - long argument
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder numberArg(long arg);
+
+    /*
+     * SPATIAL
+     */
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a WKT shape
+     *
+     * @param wkt - WKT-defined shape (2D)
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder wktArg(String wkt);
+
+    /*
+     * TEMPORAL
+     */
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a {@link Date}
+     *
+     * @param date - {@link Date}
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder dateArg(Date date);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a range defined by two {@link Date} instances
+     *
+     * @param begin - {@link Date} defining beginning of the range
+     * @param end   - {@link Date} defining end of the range
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder dateRangeArg(Date begin, Date end);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a boolean value
+     *
+     * @param arg - boolean value to ArgumentBuilder on
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder boolArg(boolean arg);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on byte value
+     *
+     * @param bytes byte array
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder bytesArg(byte[] bytes);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on a text value
+     *
+     * @param text - {@link String} argument to ArgumentBuilder on
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder textArg(String text);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on an attribute
+     *
+     * @param name - {@link String} argument to ArgumentBuilder on
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder attributeArg(String name);
+
+    /**
+     * We only support propertyIsEqualTo(Function,val) currently so instead supporting ExpressionBuilder is() we just support equalTo
+     * <p>
+     * Continue building a Filter with the "equal to" operator ( {@link org.opengis.filter.PropertyIsEqualTo})
+     *
+     * @return {@link NumericalExpressionBuilder} to continue building the {@link Filter}
+     */
+    EqualityExpressionBuilder equalTo();
+
+}
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
@@ -1255,29 +1255,4 @@ public abstract class FilterDelegate<T> {
                 "relative(String,long) not supported by org.opengis.filter.Filter Delegate.");
     }
 
-    /**
-     * Determines if the property contains search terms that are within a certain distance of each other.
-     *
-     * @param propertyName name of property to compare
-     * @param distance     distance (in words) between search terms
-     * @param searchTerm   search terms to compare
-     * @return
-     */
-    public T propertyIsInProximityTo(String propertyName, Integer distance, String searchTerm) {
-        throw new UnsupportedOperationException(
-                "propertyIsInProximityTo(String,Integer,String) not supported by org.opengis.filter.Filter Delegate.");
-    }
-
-    /**
-     * Determines if the property contains search terms that are not within a certain distance of each other.
-     *
-     * @param propertyName name of property to compare
-     * @param distance     distance (in words) between search terms
-     * @param searchTerm   search terms to compare
-     * @return
-     */
-    public T propertyIsNotInProximityTo(String propertyName, Integer distance, String searchTerm) {
-        throw new UnsupportedOperationException(
-                "propertyIsNotInProximityTo(String,Integer,String) not supported by org.opengis.filter.Filter Delegate.");
-    }
 }
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
@@ -268,7 +268,9 @@ public abstract class FilterDelegate<T> {
 
     /**
      * Compares the value function return value is equal to the value of a literal.
-     * Arguments and literal are objects but the function passed will dictate what they are required to be.
+     * For simplicity of the API arguments and literal are objects since filter delegates that
+     * support each function will have to be aware of what types they are since that is dictated by
+     * the function.
      * <p>
      * {@code function result == literal}
      *
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
@@ -267,10 +267,9 @@ public abstract class FilterDelegate<T> {
     }
 
     /**
-     * Compares the value function return value is equal to the value of a literal.
-     * For simplicity of the API arguments and literal are objects since filter delegates that
-     * support each function will have to be aware of what types they are since that is dictated by
-     * the function.
+     * Compares the function return value is equal to the value of a literal.
+     * Filter delegates will have to know the function argument types and return types for each specific
+     * function in order to support it, therefore to simplify the API they are just passed as objects.
      * <p>
      * {@code function result == literal}
      *
--- a/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
+++ b/catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
@@ -266,6 +266,22 @@ public abstract class FilterDelegate<T> {
                 "propertyIsEqualTo(String,Object) not supported by org.opengis.filter.Filter Delegate.");
     }
 
+    /**
+     * Compares the value function return value is equal to the value of a literal.
+     * Arguments and literal are objects but the function passed will dictate what they are required to be.
+     * <p>
+     * {@code function result == literal}
+     *
+     * @param functionName name of function
+     * @param arguments    list of arguments to pass to the function
+     * @param literal      value to compare
+     * @return result of equals operation between {@code functionName} result and {@code literal}
+     */
+    public T propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        throw new UnsupportedOperationException(
+                "function(String,List<Object> ) not supported by org.opengis.filter.Filter Delegate.");
+    }
+
     // PropertyIsNotEqualTo
 
     /***
@@ -1243,8 +1259,8 @@ public abstract class FilterDelegate<T> {
      * Determines if the property contains search terms that are within a certain distance of each other.
      *
      * @param propertyName name of property to compare
-     * @param distance distance (in words) between search terms
-     * @param searchTerm search terms to compare
+     * @param distance     distance (in words) between search terms
+     * @param searchTerm   search terms to compare
      * @return
      */
     public T propertyIsInProximityTo(String propertyName, Integer distance, String searchTerm) {
@@ -1256,13 +1272,12 @@ public abstract class FilterDelegate<T> {
      * Determines if the property contains search terms that are not within a certain distance of each other.
      *
      * @param propertyName name of property to compare
-     * @param distance distance (in words) between search terms
-     * @param searchTerm search terms to compare
+     * @param distance     distance (in words) between search terms
+     * @param searchTerm   search terms to compare
      * @return
      */
     public T propertyIsNotInProximityTo(String propertyName, Integer distance, String searchTerm) {
         throw new UnsupportedOperationException(
                 "propertyIsNotInProximityTo(String,Integer,String) not supported by org.opengis.filter.Filter Delegate.");
     }
-
 }
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
@@ -28,8 +28,8 @@ import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang.StringUtils;
 import org.apache.karaf.shell.api.action.Option;
 import org.apache.karaf.shell.api.action.lifecycle.Reference;
-import org.geotools.filter.text.cql2.CQL;
 import org.geotools.filter.text.cql2.CQLException;
+import org.geotools.filter.text.ecql.ECQL;
 import org.opengis.filter.Filter;
 
 import ddf.catalog.data.AttributeDescriptor;
@@ -146,7 +146,7 @@ public abstract class CqlCommands extends CatalogCommands {
         filterCurrentTime = System.currentTimeMillis();
 
         if (cqlFilter != null) {
-            return CQL.toFilter(cqlFilter);
+            return ECQL.toFilter(cqlFilter);
         }
 
         final long start = getFilterStartTime();
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang.StringUtils;
 import org.apache.karaf.shell.api.action.Option;
 import org.apache.karaf.shell.api.action.lifecycle.Reference;
+import org.geotools.filter.text.cql2.CQL;
 import org.geotools.filter.text.cql2.CQLException;
 import org.geotools.filter.text.ecql.ECQL;
 import org.opengis.filter.Filter;
@@ -146,7 +147,11 @@ public abstract class CqlCommands extends CatalogCommands {
         filterCurrentTime = System.currentTimeMillis();
 
         if (cqlFilter != null) {
-            return ECQL.toFilter(cqlFilter);
+            try {
+                return CQL.toFilter(cqlFilter);
+            } catch (CQLException e) {
+                return ECQL.toFilter(cqlFilter);
+            }
         }
 
         final long start = getFilterStartTime();
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
@@ -28,7 +28,6 @@ import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang.StringUtils;
 import org.apache.karaf.shell.api.action.Option;
 import org.apache.karaf.shell.api.action.lifecycle.Reference;
-import org.geotools.filter.text.cql2.CQL;
 import org.geotools.filter.text.cql2.CQLException;
 import org.geotools.filter.text.ecql.ECQL;
 import org.opengis.filter.Filter;
@@ -147,11 +146,7 @@ public abstract class CqlCommands extends CatalogCommands {
         filterCurrentTime = System.currentTimeMillis();
 
         if (cqlFilter != null) {
-            try {
-                return CQL.toFilter(cqlFilter);
-            } catch (CQLException e) {
-                return ECQL.toFilter(cqlFilter);
-            }
+            return ECQL.toFilter(cqlFilter);
         }
 
         final long start = getFilterStartTime();
--- a/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
+++ b/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
@@ -100,6 +100,12 @@ public class CopyFilterDelegate extends FilterDelegate<Filter> {
     }
 
     @Override
+    public Filter propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        throw new UnsupportedOperationException(
+                "propertyIsEqualTo(String,List,Object) not supported by CopyFilterDelegate.");
+    }
+
+    @Override
     public Filter propertyIsEqualTo(String propertyName, Date literal) {
         return filterBuilder.attribute(propertyName)
                 .equalTo()
--- a/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
+++ b/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
@@ -102,9 +102,9 @@ public class CopyFilterDelegate extends FilterDelegate<Filter> {
 
     @Override
     public Filter propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
-        //making an assumption that the first argument will be an attribute followed by N values
-        //this will work for most function if it doesn't we can add a switch based on function name
-        //to handle a specific function here
+        //Making an assumption that the first argument will be an attribute followed by N values.
+        //This will work for most functions. If it doesn't a switch based on function name can be added
+        //to handle a specific function here.
         ArgumentBuilder argBuilder = filterBuilder.function(functionName);
         for (int i = 0; i < arguments.size(); i++) {
             if (i == 0) {
--- a/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
+++ b/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
@@ -18,6 +18,7 @@ import java.util.List;
 
 import org.opengis.filter.Filter;
 
+import ddf.catalog.filter.ArgumentBuilder;
 import ddf.catalog.filter.FilterBuilder;
 import ddf.catalog.filter.FilterDelegate;
 
@@ -101,6 +102,45 @@ public class CopyFilterDelegate extends FilterDelegate<Filter> {
 
     @Override
     public Filter propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        //making an assumption that the first argument will be an attribute followed by N values
+        //this will work for most function if it doesn't we can add a switch based on function name
+        //to handle a specific function here
+        ArgumentBuilder argBuilder = filterBuilder.function(functionName);
+        for (int i = 0; i < arguments.size(); i++) {
+            if (i == 0) {
+                argBuilder.attributeArg(arguments.get(0)
+                        .toString());
+            } else {
+                argBuilder.objArg(arguments.get(i));
+            }
+        }
+
+        if (literal instanceof String) {
+            return argBuilder.equalTo()
+                    .text(literal.toString());
+        } else if (literal instanceof Boolean) {
+            return argBuilder.equalTo()
+                    .bool((Boolean) literal);
+        } else if (literal instanceof Long) {
+            return argBuilder.equalTo()
+                    .number((Long) literal);
+        } else if (literal instanceof Integer) {
+            return argBuilder.equalTo()
+                    .number((Integer) literal);
+        } else if (literal instanceof Double) {
+            return argBuilder.equalTo()
+                    .number((Double) literal);
+        } else if (literal instanceof Float) {
+            return argBuilder.equalTo()
+                    .number((Float) literal);
+        } else if (literal instanceof Short) {
+            return argBuilder.equalTo()
+                    .number((Short) literal);
+        } else if (literal instanceof Date) {
+            return argBuilder.equalTo()
+                    .date((Date) literal);
+        }
+
         throw new UnsupportedOperationException(
                 "propertyIsEqualTo(String,List,Object) not supported by CopyFilterDelegate.");
     }
--- a/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
+++ b/catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
@@ -25,9 +25,12 @@ public class FilterToTextDelegate extends FilterDelegate<String> {
     public String propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
         switch (functionName) {
         case "divisibleBy":
-            return propertyIsDivisibleBy(arguments) + '=' + literal;
+            return propertyIsDivisibleBy((String) arguments.get(0), (Long) arguments.get(1)) + '='
+                    + literal;
         case "proximity":
-            return propertyIsInProximityTo(arguments) + '=' + literal;
+            return propertyIsInProximityTo((String) arguments.get(0),
+                    (Integer) arguments.get(1),
+                    (String) arguments.get(2)) + '=' + literal;
         default:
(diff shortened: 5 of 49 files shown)


catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
```diff
@@ -100,6 +100,12 @@ public Filter propertyIsEqualTo(String propertyName, String literal, boolean isC
     }
 
     @Override
+    public Filter propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        throw new UnsupportedOperationException(
+                "propertyIsEqualTo(String,List,Object) not supported by CopyFilterDelegate.");
```
ryeats: Currently I am only passing the function name down so we can't recreate the filter exactly since it has no idea which parameters are literals and which are propertyNames. Conceivably it doesn't matter we could pass them around as literals since they get turned into strings anyway.

catalog/spatial/csw/spatial-csw-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/ExtendedGeotoolsFunctionFactory.java
```diff
@@ -24,20 +24,29 @@
...
+import ddf.catalog.impl.filter.DivisibleByFunction;
+import ddf.catalog.impl.filter.ProximityFunction;
+
 /**
  * The ExtendedGeotoolsFunctionFactory is used to provide the GeoTools CommonFactoryFinder a list of custom
  * functions.  This allows for DDF to extend the OGC Filter 1.1.0 schema and pass along the extension
  * changes to Geotools via a service loader .
  */
+//TODO (RWY) - can we get rid of this and use the one in geotools?
 public class ExtendedGeotoolsFunctionFactory implements FunctionFactory {
```
ryeats: I don't think their is any reason to have this class in addition to the GeotoolsFunctionFactory but I would have to move PropertyIsFuzzy down into geotools in order to keep fuzzy search in csw backwords compatible

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public DivisibleByFunction(List<Expression> parameters, Literal fallback) {
+        super(FUNCTION_NAME, fallback);
+
+        notNull(parameters, "Parameters are required");
+        isTrue(parameters.size() == NUM_PARAMETERS,
+                String.format("%s expression requires at least %s parameters", FUNCTION_NAME,
+                        NUM_PARAMETERS));
+
+        //TODO we could cast the arguments here to the correct object so that it doen't need to be done in the filter delegates
```
ryeats: this is the one place where we know what the parameters should be so we could do the conversion of the literal values and parameters here from expression strings to objects.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
```diff
@@ -272,6 +295,19 @@ public Object visit(PropertyIsBetween filter, Object extraData) {
 
     @Override
     public Object visit(PropertyIsEqualTo filter, Object extraData) {
+        FilterFactory2 factory = getFactory(extraData);
+
+        if (filter.getExpression1() instanceof Function) {
```
ryeats: This is the main reason I only implemented PropertyIsEqualTo support for functions every visitor would need to be updated to handle function expressions correctly

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/CswFilterFactory.java
```diff
@@ -835,4 +837,37 @@ private BinarySpatialOpType createBinarySpatialOpType(PropertyNameType propertyN
...
+    //default implementation just assume the first argument is a parameter name and the rest are values
+    public FilterType buildPropertyIsEqualTo(String functionName, List<Object> arguments,
+            Object literal) {
+        List<JAXBElement<?>> expressions = new ArrayList<>();
+        for (int i = 0; i < arguments.size(); i++) {
+            if (i == 0) {
+                expressions.add(createPropertyNameType(Arrays.asList(new Object[] {
+                        arguments.get(i)})));
+            } else {
+                expressions.add(createLiteralType(arguments.get(i)));
```
ryeats: may just have this do all literals since they just get turned into objects anyway it doesn't matter which are propertyNames and which are strings the delegate will just assume based on the parameter order

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
```diff
@@ -0,0 +1,75 @@
...
+                        NUM_PARAMETERS));
+
+        setName(FUNCTION_NAME);
+        setParameters(parameters);
+        setFallbackValue(fallback);
+        functionName = NAME;
+    }
+
+    public String getPropertyName() {
+        return ((PropertyName) params.get(0)).getPropertyName();
```
coyotesqrl: ✏️ I don't love these direct accesses to the protected params field. My preference - just a preference - would be to access them through the public `getParameters()` method of the super class.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
```diff
@@ -0,0 +1,75 @@
...
+        return Long.parseLong(((Literal) params.get(1)).getValue()
+                .toString());
+    }
+
+    public boolean equals(Object obj) {
+        return super.equals(obj);
+    }
+
+    //findbugs: parent class overrides equals() but not hashcode
+    public int hashCode() {
```
coyotesqrl: ❗️ These semantics are not the same as the `equals()` semantics of the super class. This could lead to some very strange conditions in collections or other cases.

Realistically, it is unlikely to happen or it would be happening already with the extant `FunctionExpressionImpl`s; however, I'm concerned about putting something that we know is incorrect in place just to silence a static analysis warning.

You might want to replace the super class's implementation of `equals()` with a custom one and have it and the `hashCode()` method match semantically.

catalog/core/catalog-core-impl/filter-proxy/pom.xml
```diff
@@ -64,6 +66,44 @@
...
+                                        </limit>
+                                        <limit>
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.75</minimum>
+                                        </limit>
+                                        <limit>
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.7</minimum>
```
coyotesqrl: ❗️ Set coverage values to two decimal places (and then leave at least a .03 margin because of floating point inconsistencies).

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -130,42 +129,48 @@ public Object visit(Divide expression, Object delegate) {
...
-            }
+        int argCount = expression.getFunctionName()
+                .getArgumentCount();
+
+        if (parameters == null || parameters.size() != argCount) {
+            throw new UnsupportedOperationException(
+                    Function.class.getSimpleName() + " requires " + argCount + " arguments.");
+        }
+        //unwrap the function arguments and return them
+        List<Object> ret = new ArrayList<>(argCount);
```
coyotesqrl: ✏️ There's usually no reason to pre-size an `ArrayList`, particularly when it is going to be short, but it obviously doesn't hurt to do so.

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -130,42 +129,48 @@ public Object visit(Divide expression, Object delegate) {
...
         if (expression.getValue() == null) {
             throw new UnsupportedOperationException(
                     Literal.class.getSimpleName() + " value must not be null.");
         }
+        //try and get the object as the requested class otherwise just return the object
+        if (clazz instanceof Class) {
+            Object ret = expression.evaluate(null, (Class) clazz);
+            if (ret != null) {
+                return ret;
+            } //could throw an exception saying it wasn't of the correct type?
```
coyotesqrl: ✏️ At least a debug log?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/builder/GeotoolsBuilder.java
```diff
@@ -44,9 +46,7 @@
 import ddf.catalog.impl.filter.JTSGeometryWrapper;
 
 /**
- *
  * @author michael.menousek@lmco.com
```
coyotesqrl: ❓ Is there value in leaving this `author` tag, especially as it is a defunct email address?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -161,6 +160,29 @@ private static double metersToDegrees(double distance) {
...
+        SolrQuery query;
+        switch (functionName) {
+        case "divisibleBy":
+
+            //the return type is boolean so cast the literal to boolean and in effect this is just a NOT so we will update the query as such9
+            not = (Boolean) literal ? "" : "!";
+            query = propertyIsDivisibleBy((String) arguments.get(0), (Long) arguments.get(1));
+            return query.setQuery(not + query.getQuery());
+        case "proximity":
+            not = (Boolean) literal ? "" : "!";
```
coyotesqrl: ✏️ Instead of duplicating this line, could you move it up to the declaration of `not`?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -161,6 +160,29 @@ private static double metersToDegrees(double distance) {
...
+            query = propertyIsInProximityTo((String) arguments.get(0),
+                    (Integer) arguments.get(1),
+                    (String) arguments.get(2));
+            return query.setQuery(not + query.getQuery());
```
coyotesqrl: ✏️ Likewise with these returns, could you just remove from here and have one below the `switch` statement?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
```diff
@@ -550,6 +572,22 @@ public SolrQuery propertyIsLessThanOrEqualTo(String propertyName, double literal
         return getLessThanOrEqualToQuery(propertyName, AttributeFormat.DOUBLE, literal);
     }
 
+    public SolrQuery propertyIsDivisibleBy(String propertyName, long divisor) {
+        //use the sort key for the field since divisible can't operate on multivalued fields (it will always be a single value).
+        List<String> solrExpressions = resolver.getAnonymousField(propertyName)
+                .stream()
+                .map(resolver::getSortKey)
+                .map(f -> "_val_:\"{!frange l=0 u=0}mod(" + f + "," + divisor + ")\"")
```
coyotesqrl: ✏️ 1. Favor `String.format()` over concatenation.
2. I think this is probably a complex enough lambda that it distracts from the form of the stream logic. It would read better if you extracted a method to convert the `sortKey` to this String representation and then in your map just called that method.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -161,8 +161,8 @@
     protected static final List<String> SERVICE_TYPE_VERSION =
             Collections.unmodifiableList(Arrays.asList(CswConstants.VERSION_2_0_2));
 
-    protected static final List<SpatialOperatorNameType> SPATIAL_OPERATORS =
-            Collections.unmodifiableList(Arrays.asList(SpatialOperatorNameType.BBOX,
+    protected static final Set<SpatialOperatorNameType> SPATIAL_OPERATORS =
+            Collections.unmodifiableSet(EnumSet.of(SpatialOperatorNameType.BBOX,
```
coyotesqrl: ❓ Is there a reason you're constructing this and the `COMPARISON_OPERATORS` list so differently?
1. Set vs. List
2. Explicitly adding enums to an interim `EnumSet` vs. `enumclass.values()`?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
```diff
@@ -119,6 +108,40 @@ public CswRecordMapperFilterVisitor(MetacardType metacardType,
...
+                Object userDataObj = geometry.getUserData();
+                if (userDataObj instanceof CoordinateReferenceSystem) {
+                    CoordinateReferenceSystem sourceCRS = (CoordinateReferenceSystem) userDataObj;
+                    Geometry convertedGeometry = null;
+                    try {
+                        convertedGeometry = GeospatialUtil.transformToEPSG4326LonLatFormat(geometry,
+                                sourceCRS);
+                        literalExpression.setValue(convertedGeometry);
+                    } catch (GeoFormatException e) {
+                        LOGGER.trace("Unable to convert geometry to EPSG:4326 format", e);
```
coyotesqrl: ❓ Just trace? Is this just something that will really never happen and if it does an admin will want to explicitly turn up logging on this class to figure out why?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
coyotesqrl: ❓ Is this introducing tech debt we'll want to pay down, then? Have you created tickets to address that?

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/CswFilterFactory.java
```diff
@@ -835,4 +837,38 @@ private BinarySpatialOpType createBinarySpatialOpType(PropertyNameType propertyN
...
     }
 
+    //default implementation just assume the first argument is a parameter name and the rest are values
+    public FilterType buildPropertyIsEqualTo(String functionName, List<Object> arguments,
+            Object literal) {
+        int propertyCount = 1;
+        List<JAXBElement<?>> expressions = new ArrayList<>();
+        for (int i = 0; i < arguments.size(); i++) {
+            if (i < propertyCount) {
+                expressions.add(createPropertyNameType(Arrays.asList(new Object[] {
```
coyotesqrl: ❓ This feels very fragile/brittle.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public DivisibleByFunction(List<Expression> parameters, Literal fallback) {
+        super(FUNCTION_NAME, fallback);
+
+        notNull(parameters, "Parameters are required");
+        isTrue(parameters.size() == NUM_PARAMETERS,
+                String.format("%s expression requires at least %s parameters", FUNCTION_NAME,
+                        NUM_PARAMETERS));
+
+        setName(FUNCTION_NAME);
```
jrnorth: Do you need to call `setName()` and `setFallbackValue()` since you're passing those to the super constructor?

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
```diff
@@ -0,0 +1,75 @@
...
+    public String getPropertyName() {
+        return ((PropertyName) params.get(0)).getPropertyName();
+    }
+
+    public Long getLiteral() {
+        return Long.parseLong(((Literal) params.get(1)).getValue()
+                .toString());
+    }
+
+    public boolean equals(Object obj) {
```
jrnorth: Don't need this if you're just using the parent's implementation.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
jrnorth: `@Override`

catalog/core/catalog-core-commons/src/test/java/ddf/catalog/impl/filter/GeoToolsFunctionFactoryTest.java
```diff
@@ -38,7 +38,7 @@ public void setUp() {
     @Test
     public void testGetFunctionNames() {
         List<FunctionName> functionNames = toTest.getFunctionNames();
-        assertThat(functionNames, hasSize(2));
+        assertThat(functionNames, hasSize(3));
         assertThat(functionNames.get(0).getName(), is(FuzzyFunction.NAME.getName()));
         assertThat(functionNames.get(1).getName(), is(ProximityFunction.NAME.getName()));
```
jrnorth: Should probably add `DivisibleByFunction` to the test.

catalog/core/catalog-core-impl/filter-proxy/pom.xml
jrnorth: Could you add a 0 at the end so we get more helpful error messages?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -975,55 +977,54 @@ private ExpressionValues getExpressions(Expression expression1, Expression expre
...
+        if (expression1 instanceof Function && expression2 instanceof Literal) {
+            functionName = ((Function) expression1).getName();
+            functionArgs = (List) expression1.accept(this, delegate);
+            literal = expression2.accept(this,
+                    ((Function) expression1).getFunctionName()
+                            .getReturn()
+                            .getType());
+            return new ExpressionValues(functionName, functionArgs, literal);
+        } else if (expression2 instanceof Function && expression1 instanceof Literal) {
+            literal = expression1.accept(this, delegate);
```
jrnorth: Should this match the expression for `literal` in the preceding case?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
```diff
@@ -272,6 +295,19 @@ public Object visit(PropertyIsBetween filter, Object extraData) {
...
+
+        if (filter.getExpression1() instanceof Function) {
+            Expression function = (Expression) visit((Function) filter.getExpression1(), extraData);
+            //given a function we can use its return value to infer the type of expression2
+            Expression typedExpression = factory.literal(filter.getExpression2()
+                    .evaluate(null,
+                            ((Function) filter.getExpression1()).getFunctionName()
+                                    .getReturn()
+                                    .getType()));
+            return factory.equals(function, visit((Expression) typedExpression, function));
```
jrnorth: Don't need to cast `typedExpression`.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -209,6 +216,32 @@
...
+                        new MediaAttributes(),
+                        new TopicAttributes(),
+                        new AssociationsAttributes(),
+                        new MetacardTypeImpl("TestDate",
+                                Collections.singleton(new AttributeDescriptorImpl("TestDate",
+                                        true,
+                                        true,
+                                        true,
+                                        false,
+                                        new AttributeType<Date>() {
```
jrnorth: `BasicTypes.DATE`?

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/CswFilterFactory.java
jrnorth: I don't think you need the `new Object[] { }`.

catalog/spatial/csw/spatial-csw-source-common/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/TestCswFilterDelegate.java
```diff
@@ -1011,6 +1050,19 @@ private String createComparisonFilterString(ComparisonOperator comparisonOp,
         return compString;
     }
 
+    private String createComparisonFunctionFilterString(ComparisonOperator comparisonOp,
+            String functionName, Object literal, Object... args) {
+
+        String compString =
```
jrnorth: Just return the string directly.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
```diff
@@ -20,6 +20,22 @@
...
 
+    // Custom functions
+    @Override
+    public String propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        switch (functionName) {
+        case "divisibleBy":
+            return propertyIsDivisibleBy((String) arguments.get(0), (Long) arguments.get(1)) + '='
+                    + literal;
+        case "proximity":
+            return propertyIsInProximityTo((String) arguments.get(0),
```
ryeats: Instead of doing the casting logic in every filter delegate I do it in the visit(Function) of the GeotoolsFilterAdapter. That being said right now if it can't coerce the value to the correct type it will just return it and would cause a casting error here( I can change it to throw an exception in the GeotoolsFilterAdapter).

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
ryeats: I should have been clearer. Not new tech debt(we didn't support functions in any visit method other than EqualTo in the first place). I am just not fully implementing Functions throughout since it doesn't make sense yet since Proximity and DivisibleBy only work with PropertyIsEqualTo Filters anyway. If we add a function that makes sense for PropertyIsGreater or PropertyIsBetween at that point we would have to plumb functions all the way through for those filter types.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
```diff
@@ -20,6 +20,22 @@
 
 public class FilterToTextDelegate extends FilterDelegate<String> {
 
+    // Custom functions
+    @Override
+    public String propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        switch (functionName) {
+        case "divisibleBy":
```
ryeats: Well we get it as a String from the FilterFunction I will at least change it to reference DisibleByFunction.FUNCTION_NAME.  They are sort of dynamic in the sense that a new FunctionFactory could be added to support a new function but that being said its an internal API so it wouldn't be an external implementor.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/impl/filter/DivisibleByFunction.java
ryeats: agreed, I actually don't need them so ill get rid of them.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
ryeats: hmm its not really common logic that literal could be anything its just for these functions the return type they dictate is a boolean and so I can do a not on their result.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrFilterDelegate.java
ryeats: I guess thats a question do I write the code so that someone could easily add a new function or since they happen to share logic now that is common to every statement currently in the switch do i drop that logic out?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
ryeats: oops need to revert this I am not adding a ComparisonOperatorType anymore. They should probably have both stay lists since despite them being enums they are passed around as lists.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
ryeats: I don't really know, I can change it but ill first blame my IDE for hoisting this method because I had clean code turned on when I committed.

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/CswFilterFactory.java
ryeats: I don't disagree here are some options:
A. (What I did) I can make an assumption that covers most cases so that this doesn't need to be implemented for most functions we add.
B. I can throw an exception to force this to be updated for every new function we implement if we remember to check if CSW federation works.
C. I can just treat them all as literals because a String from a property name is going to be treated (Mostly) the same as a string from a literal (when these functions are parsed from CQL the filter object has them as literals anyway so we will have this issue no matter what).
D. Rework everything to pass down more information from the FunctionFactory to the filter delegates about which values are PropertyNames and which are Values.

catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/ArgumentBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @param text - {@link String} argument to ArgumentBuilder on
+     * @return {@link ArgumentBuilder}
+     */
+    ArgumentBuilder textArg(String text);
+
+    /**
+     * Continues building the {@link ArgumentBuilder} based on an attribute
+     *
+     * @param name - {@link String} argument to ArgumentBuilder on
```
ryeats: yes

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
```diff
@@ -363,6 +379,12 @@ public String propertyIsLessThanOrEqualTo(String propertyName, Object literal) {
         return propertyName + "<=" + literal + "o";
     }
 
+    public String propertyIsDivisibleBy(String propertyName, long divisor) {
+        return divisor + "|" + propertyName;
+        //TODO (RWY) - which notation is more readable?
+        //        return propertyName + "%" + divisor + "i=0";
```
ryeats: I believe it just has to be human redable

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
ryeats: decided to throw an exception.

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -130,42 +129,48 @@ public Object visit(Divide expression, Object delegate) {
...
-    public Object visit(Literal expression, Object delegate) {
+    @Override
+    public Object visit(Literal expression, Object clazz) {
         if (expression.getValue() == null) {
             throw new UnsupportedOperationException(
                     Literal.class.getSimpleName() + " value must not be null.");
         }
+        //try and get the object as the requested class otherwise just return the object
+        if (clazz instanceof Class) {
+            Object ret = expression.evaluate(null, (Class) clazz);
```
ryeats: How do we feel about relying on the underlying geotools literal implementation to do the type conversion?

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
ryeats: changed GeotoolsFilterAdapter to throw an exception if it couldn't find a convert to convert the types

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/CqlCommands.java
```diff
@@ -146,7 +147,11 @@ protected Filter getFilter(FilterBuilder filterBuilder) throws CQLException, Par
         filterCurrentTime = System.currentTimeMillis();
 
         if (cqlFilter != null) {
-            return CQL.toFilter(cqlFilter);
+            try {
+                return CQL.toFilter(cqlFilter);
+            } catch (CQLException e) {
```
ryeats: They are pretty much the same so you can't tell the difference but I am going to look into if the backwards compatibility issues are mitigated then I can just do ECQL.

ryeats:
build now

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/TestCswRecordMapperFilterVisitor.java
```diff
@@ -259,6 +263,38 @@ public void testVisitPropertyIsEqualTo() {
...
+                Arrays.asList(attrExpr, factory.literal(1), factory.literal("Marry little")),
+                null);
+
+        PropertyIsEqualTo filter = factory.equals(function, val);
+
+        PropertyIsEqualTo duplicate = (PropertyIsEqualTo) visitor.visit(filter, null);
+        assertThat(duplicate.getExpression1(), Matchers.instanceOf(Function.class));
+        Function dupFunction = (Function) duplicate.getExpression1();
+        assertThat(dupFunction.getFunctionName(), is(function.getFunctionName()));
+        assertThat(dupFunction.getParameters()
```
jrnorth: You don't need these assertions since your `assertThat(dupFunction.getParameters(), is(function.getParameters()))` will check that the lists are equal (and therefore that the corresponding pairs of elements in the two lists are equal).

distribution/test/itests/test-itests-common/src/main/resources/csw-function-query.xml
```diff
@@ -0,0 +1,34 @@
...
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+
+<csw:GetRecords resultType="results" outputFormat="$outputFormat$"
```
jrnorth: ✏️  The formatting of this file is weird.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/FilterToTextDelegate.java
```diff
@@ -363,6 +381,10 @@ public String propertyIsLessThanOrEqualTo(String propertyName, Object literal) {
         return propertyName + "<=" + literal + "o";
     }
 
+    public String propertyIsDivisibleBy(String propertyName, long divisor) {
+        return propertyName + "%" + divisor + "i=0";
```
ryeats: should be an 'L' following the naming scheme above which uses a letter to describe the type of literal

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
ryeats: okay going to do the same thing i did for CswFilterDelegate. I will assume the first arg is a attributeName and the rest are values. Even if it isn't correct for all function it should still work since the filter adapter should still be able to convert to the correct object.

ryeats:
build now

ryeats:
build now

catalog/core/catalog-core-metricsplugin/src/main/java/ddf/catalog/metrics/QueryTypeFilterDelegate.java
```diff
@@ -98,6 +102,11 @@ public Boolean propertyIsFuzzy(String propertyName, String literal) {
         return isComparison = true;
     }
 
+    @Override
+    public Boolean propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        return isFunction = true;
```
coyotesqrl: Oh, Sonarqube. You're so silly.

❗️ That should be `return isFunction == true;`, correct?

ryeats: actually i think its right for once i can split them out `isFunction = true; return isFunction;` i was just following what the other methods were doing.

catalog/core/catalog-core-commons/src/main/java/ddf/catalog/filter/delegate/CopyFilterDelegate.java
```diff
@@ -100,6 +101,51 @@ public Filter propertyIsEqualTo(String propertyName, String literal, boolean isC
     }
 
     @Override
+    public Filter propertyIsEqualTo(String functionName, List<Object> arguments, Object literal) {
+        //making an assumption that the first argument will be an attribute followed by N values
+        //this will work for most function if it doesn't we can add a switch based on function name
```
bdeining: nit : `most function**s**` and a run-on sentence

catalog/core/catalog-core-api/src/main/java/ddf/catalog/filter/FilterDelegate.java
```diff
@@ -266,6 +266,24 @@ public T propertyIsEqualTo(String propertyName, Object literal) {
                 "propertyIsEqualTo(String,Object) not supported by org.opengis.filter.Filter Delegate.");
     }
 
+    /**
+     * Compares the value function return value is equal to the value of a literal.
+     * For simplicity of the API arguments and literal are objects since filter delegates that
```
bdeining: nit : run-on sentence

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/mappings/CswRecordMapperFilterVisitor.java
```diff
@@ -119,6 +108,40 @@ public CswRecordMapperFilterVisitor(MetacardType metacardType,
...
+                Object userDataObj = geometry.getUserData();
+                if (userDataObj instanceof CoordinateReferenceSystem) {
+                    CoordinateReferenceSystem sourceCRS = (CoordinateReferenceSystem) userDataObj;
+                    Geometry convertedGeometry = null;
+                    try {
+                        convertedGeometry = GeospatialUtil.transformToEPSG4326LonLatFormat(geometry,
+                                sourceCRS);
+                        literalExpression.setValue(convertedGeometry);
+                    } catch (GeoFormatException e) {
+                        LOGGER.debug("Unable to convert geometry to EPSG:4326 format", e);
```
bdeining: Should we also log `geometry` or `valueObj` here?

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -209,6 +215,20 @@
...
     private static List<MetacardType> metacardTypeList;
 
+    public static MetacardType getCswMetacardType() {
+        return new MetacardTypeImpl(CswConstants.CSW_METACARD_TYPE_NAME,
+                Arrays.asList(new ContactAttributes(),
+                        new LocationAttributes(),
+                        new MediaAttributes(),
+                        new TopicAttributes(),
+                        new AssociationsAttributes(),
+                        new MetacardTypeImpl("TestDate",
```
bdeining: Is it necessary to make a new test attribute wrapped in a new metacard type?

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/CswFilterDelegate.java
```diff
@@ -97,12 +96,21 @@ public CswFilterDelegate(Operation getRecordsOp, FilterCapabilities filterCapabi
...
+        cswFilterFactory = new CswFilterFactory(cswSourceConfiguration.getCswAxisOrder(),
                 cswSourceConfiguration.isSetUsePosList());
         updateAllowedOperations(filterCapabilities);
     }
 
     @Override
+    public FilterType propertyIsEqualTo(String functionName, List<Object> arguments,
+            Object literal) {
+        //this assumes the first argument is a property name followed by 0-N literals if a new
+        //function is added that doesn't conform to this then a other cases can be added. Most
```
bdeining: nit : `then a other cases` and a run-on sentence

catalog/spatial/csw/spatial-csw-source-common/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/TestCswFilterDelegate.java
```diff
@@ -989,6 +1008,26 @@ private String createComparisonFilterStringWithoutHeaderAndFooter(
         return expression;
     }
 
+    private String createComparisonFunctionFilterStringWithoutHeaderAndFooter(
+            ComparisonOperator comparisonOp, String functionName, Object literal, Object[] args) {
+        String expression = "<ns3:" + comparisonOp;
```
bdeining: Could use `StringBuilder` here

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
```diff
@@ -597,6 +598,47 @@ public void testCswUtmQuery() {
     }
 
     @Test
+    public void testCswCQLFunctionQuery() {
+        String id = ingest(getFileContent("metacard5.xml"), "text/xml");
+        try {
+
+            given().header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_XML)
+                    .get(CSW_PATH.getUrl()
+                            + "?service=CSW&version=2.0.2&request=GetRecords&outputFormat=application/xml&outputSchema=http://www.opengis.net/cat/csw/2.0.2&NAMESPACE=xmlns(csw=http://www.opengis.net/cat/csw/2.0.2)&resultType=results&typeNames=csw:Record&ElementSetName=brief&ConstraintLanguage=CQL_TEXT&constraint=proximity(metadata,2,'All Hail Our SysAdmin')=true")
```
bdeining: Is this line formatted?

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
ryeats: this got changed during formatting but it looks like we can just use CoreAttributes.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestCatalog.java
ryeats: yes but ill break it up so its easier to read

ryeats:
build now

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -130,42 +130,52 @@ public Object visit(Divide expression, Object delegate) {
...
+            ret.add(arg.accept(this, type)); //the type would only be used by literals
         }
 
-        return expression;
+        return ret;
     }
 
-    public Object visit(Literal expression, Object delegate) {
+    @Override
+
```
jrnorth: Extra line

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -130,42 +130,52 @@ public Object visit(Divide expression, Object delegate) {
...
+    public Object visit(Literal expression, Object clazz) {
         if (expression.getValue() == null) {
             throw new UnsupportedOperationException(
                     Literal.class.getSimpleName() + " value must not be null.");
         }
+        //try and get the object as the requested class otherwise just return the object
+        if (clazz instanceof Class) {
```
jrnorth: Should the parameter be called `clazz` if it might not be a `Class`?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -194,6 +204,7 @@ public Object visit(ExcludeFilter filter, Object delegate) {
 
     public Object visit(IncludeFilter filter, Object delegate) {
         return ((FilterDelegate<?>) delegate).include();
+
```
jrnorth: Extra line

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -975,55 +982,56 @@ private ExpressionValues getExpressions(Expression expression1, Expression expre
...
+                    ((Function) expression1).getFunctionName()
+                            .getReturn()
+                            .getType());
+            return new ExpressionValues(functionName, functionArgs, literal);
+        } else if (expression2 instanceof Function && expression1 instanceof Literal) {
+            literal = expression2.accept(this, ((Function) expression2).getFunctionName()
+                            .getReturn()
+                            .getType());
+            functionName = ((Function) expression2).getName();
+            functionArgs = (List) expression1.accept(this, delegate);
```
jrnorth: `expression2`?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -975,55 +982,56 @@ private ExpressionValues getExpressions(Expression expression1, Expression expre
...
+        if (expression1 instanceof Function && expression2 instanceof Literal) {
+            functionName = ((Function) expression1).getName();
+            functionArgs = (List) expression1.accept(this, delegate);
+            literal = expression2.accept(this,
+                    ((Function) expression1).getFunctionName()
+                            .getReturn()
+                            .getType());
+            return new ExpressionValues(functionName, functionArgs, literal);
+        } else if (expression2 instanceof Function && expression1 instanceof Literal) {
+            literal = expression2.accept(this, ((Function) expression2).getFunctionName()
```
jrnorth: `expression1.accept`?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
ryeats: Its to document what this visitor expects as an extraData object for other developers since the ExpressionVisitor interface requires that it be an object.

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
ryeats: seeing as how I messed this up twice I will break each step out to clean it up more

ryeats:
build now

ryeats:
build now

PR #2089, opened 2017-06-22
DDF-2486 Created a command to list active transformers (catalog:transformers)

josephthweatt:
#### What does this PR do?

This is a new feature to assist the user in finding available transformers. It is called with `catalog:transformers`, and displays three details about each transformer:

- id
- schema
- mime-type (multiple values)

Schema & mime-type can have no values ("N/A"), but id must be included.

@vinamartin 
@emmberk 
@coyotesqrl 
@clockard
@coyotesqrl 
1. Build and run DDF
2. Load the catalog commands by setting up the Admin Console
3. In DDF, type in `catalog:transformers`
4. Review the output. Verify that the formatting is correct and that all of your available transformers are shown in the output.
5. Go to the Admin Console and select Catalog. In the features tab, uninstall some transformers and execute `catalog:transformers` again to see if the number of "Active Transformers" decreases. 

[2486](https://codice.atlassian.net/browse/)
screen shot 2017-06-22 at 4 27 25 pm

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.commands.catalog;
+
+import java.text.MessageFormat;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.karaf.shell.api.action.Command;
+import org.apache.karaf.shell.api.action.lifecycle.Service;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+
+import ddf.catalog.Constants;
+import ddf.catalog.transform.InputTransformer;
+
+/**
+ * Provides information on available transformers
+ */
+@Service
+@Command(scope = CatalogCommands.NAMESPACE, name = "transformers", description = "Provides information on available transformers.")
+public class TransformersCommand extends CatalogCommands {
+
+    // output strings
+
+    static final String ACTIVE_TRANSFORMERS_HEADER = "Active Transformers: ";
+
+    static final String NO_ACTIVE_TRANSFORMERS = "There are no active transformers";
+
+    private static final String LINE = "------------------------";
+
+    // Transformer Properties
+
+    private static final String MIME_TYPE = "mime-type";
+
+    private static final String SCHEMA = "schema";
+
+    private static final String NOT_AVAILABLE = "N/A";
+
+    @Override
+    protected Object executeWithSubject() throws Exception {
+
+        // ServiceReferences are needed to get property information from InputTransformers
+        List<ServiceReference<InputTransformer>> serviceReferences =
+                (List<ServiceReference<InputTransformer>>) bundleContext.getServiceReferences(
+                        InputTransformer.class,
+                        "(id=*)");
+
+        List<TransformerProperties> transformersProperties = serviceReferences.stream()
+                .map(TransformerProperties::new)
+                .collect(Collectors.toList());
+
+        int activeTransformers = transformersProperties.size();
+
+        if (activeTransformers == 0) {
+            console.printf("%s%n%n", NO_ACTIVE_TRANSFORMERS);
+            return null;
+        }
+
+        console.printf("%n%s%d%n%s%n%n", ACTIVE_TRANSFORMERS_HEADER, activeTransformers, LINE);
+
+        Iterator<TransformerProperties> tpIterator = transformersProperties.iterator();
+        TransformerProperties tp;
+
+        while (tpIterator.hasNext()) {
+            tp = tpIterator.next();
+            console.printf("%s", tp.printProperties());
+
+            if (tpIterator.hasNext()) {
+                console.printf("%n%s%n", StringUtils.repeat(LINE, 3));
+            }
+            console.printf("%n");
+        }
+
+        return null;
+    }
+
+    private static class TransformerProperties {
+
+        private String id;
+
+        private String schema;
+
+        private List<String> mimeTypes;
+
+        public TransformerProperties(ServiceReference ref) {
+
+            this.id = getTransformerPropertyString(ref, Constants.SERVICE_ID);
+            this.schema = getTransformerPropertyString(ref, SCHEMA);
+            this.mimeTypes = getTransformerMimeTypes(ref);
+        }
+
+        public String printProperties() {
+
+            StringBuilder s = new StringBuilder(MessageFormat.format(
+                    "{0}: {1}\n\n\t{2}: {3}\n\t{4}s: {5}\n",
+                    Constants.SERVICE_ID,
+                    id,
+                    SCHEMA,
+                    schema,
+                    MIME_TYPE,
+                    mimeTypes.remove(0)));
+
+            for (String mimeType : mimeTypes) {
+                s.append("\t\t    ")
+                        .append(mimeType)
+                        .append("\n");
+            }
+            return s.toString();
+        }
+
+        private String getTransformerPropertyString(ServiceReference ref, String property) {
+
+            return Optional.ofNullable(ref.getProperty(property))
+                    .map(String.class::cast)
+                    .filter(String.class::isInstance)
+                    .orElse(NOT_AVAILABLE);
+        }
+
+        private List<String> getTransformerMimeTypes(ServiceReference ref) {
+
+            List<String> mimeProperties;
+
+            if (ref.getProperty(MIME_TYPE) instanceof List) {
+                mimeProperties = (ArrayList<String>) ref.getProperty(MIME_TYPE);
+            } else {
+                mimeProperties = new ArrayList<>();
+                mimeProperties.add(getTransformerPropertyString(ref, MIME_TYPE));
+            }
+            return mimeProperties;
+        }
+    }
+
+    // Used for testing TransformersCommand
+    protected void setBundleContext(BundleContext bundleContext) {
+        this.bundleContext = bundleContext;
+    }
+}
--- a/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
+++ b/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
@@ -1,0 +1,181 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.commands.catalog;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.StringContains.containsString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
+
+import com.google.common.collect.ImmutableList;
+
+import ddf.catalog.Constants;
+import ddf.catalog.transform.InputTransformer;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TransformersCommandTest extends ConsoleOutputCommon {
+
+    private static final String SERVICE_ID = "id";
+
+    private static final String MIME_TYPE = "mime-type";
+
+    private static final String SCHEMA = "schema";
+
+    @Mock
+    BundleContext bundleContext;
+
+    TransformersCommand transformersCommand;
+
+    @Before
+    public void setUp() {
+
+        transformersCommand = new TransformersCommand();
+        transformersCommand.setBundleContext(bundleContext);
+    }
+
+    @Test
+    public void testNullIds() throws Exception {
+
+        List<ServiceReference<InputTransformer>> serviceReferences = new ArrayList<>();
+        ServiceReference<InputTransformer> serviceReference = mock(ServiceReference.class);
+        serviceReferences.add(serviceReference);
+
+        List<ServiceReference<InputTransformer>> serviceReferencesFiltered =
+                serviceReferences.stream()
+                        .filter(ref -> ref.getProperty(Constants.SERVICE_ID) != null)
+                        .collect(Collectors.toList());
+
+        when(serviceReference.getProperty(SERVICE_ID)).thenReturn(null);
+        when(bundleContext.getServiceReferences(InputTransformer.class, "(id=*)")).thenReturn(
+                serviceReferencesFiltered);
+
+        transformersCommand.executeWithSubject();
+        assertThat(consoleOutput.getOutput(),
+                containsString(TransformersCommand.NO_ACTIVE_TRANSFORMERS));
+    }
+
+    @Test
+    public void testNullAndNotNullIds() throws Exception {
+
+        int nullIds = 5;
+        int notNullIds = 7;
+
+        List<ServiceReference<InputTransformer>> serviceReferences = new ArrayList<>();
+
+        // mock service refs with null and not null ids
+        for (int i = 0; i < nullIds + notNullIds; i++) {
+            ServiceReference<InputTransformer> serviceReference = mock(ServiceReference.class);
+            serviceReferences.add(serviceReference);
+
+            if (i < nullIds) {
+                when(serviceReference.getProperty(SERVICE_ID)).thenReturn(null);
+            } else {
+                when(serviceReference.getProperty(SERVICE_ID)).thenReturn("Test");
+            }
+        }
+
+        List<ServiceReference<InputTransformer>> serviceReferencesFiltered =
+                serviceReferences.stream()
+                        .filter(ref -> ref.getProperty(Constants.SERVICE_ID) != null)
+                        .collect(Collectors.toList());
+
+        when(bundleContext.getServiceReferences(InputTransformer.class, "(id=*)")).thenReturn(
+                serviceReferencesFiltered);
+
+        transformersCommand.executeWithSubject();
+        assertThat(consoleOutput.getOutput(),
+                containsString(String.format("%s%d",
+                        TransformersCommand.ACTIVE_TRANSFORMERS_HEADER,
+                        notNullIds)));
+    }
+
+    @Test
+    public void testActiveTransformerCount() throws Exception {
+
+        List<ServiceReference<InputTransformer>> serviceReferences;
+
+        ServiceReference<InputTransformer> serviceReference = mock(ServiceReference.class);
+        when(serviceReference.getProperty(SERVICE_ID)).thenReturn("Test");
+
+        serviceReferences = ImmutableList.of(serviceReference, serviceReference, serviceReference);
+
+        when(bundleContext.getServiceReferences(InputTransformer.class, "(id=*)")).thenReturn(
+                serviceReferences);
+
+        transformersCommand.executeWithSubject();
+        assertThat(consoleOutput.getOutput(),
+                containsString(String.format("%s3",
+                        TransformersCommand.ACTIVE_TRANSFORMERS_HEADER)));
+    }
+
+    @Test
+    public void testNullPropertiesWithValidId() throws Exception {
+
+        List<ServiceReference<InputTransformer>> serviceReferences = new ArrayList<>();
+
+        ServiceReference<InputTransformer> serviceReference = mock(ServiceReference.class);
+
+        when(serviceReference.getProperty(SERVICE_ID)).thenReturn("Test");
+        when(serviceReference.getProperty(SCHEMA)).thenReturn(null);
+        when(serviceReference.getProperty(MIME_TYPE)).thenReturn(null);
+
+        serviceReferences.add(serviceReference);
+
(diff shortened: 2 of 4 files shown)


shaundmorris:
@josephthweatt nice start, can you please add some unit testing?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+@Service
+@Command(scope = CatalogCommands.NAMESPACE, name = "transformers", description = "Provides information on available transformers.")
+public class TransformersCommand extends CatalogCommands {
+
+    private static final String MIME_TYPE = "mime-type";
+    private static final String SCHEMA = "schema";
+
+    private static final String NOT_AVAILABLE = "N/A";
+    private static final String LINE = "------------------------";
```
vinamartin: ❓ I wonder if Karaf has some kind of utility for this, I'd try and see what other commands are doing for their lines

josephthweatt: Would you know what other commands are using separators like this?

bdeining: Resource Management has a command that outputs a table : https://github.com/codice/ddf/blob/master/catalog/resourcemanagement/resourcemanagement-querymonitor-plugin/src/main/java/org/codice/ddf/resourcemanagement/query/plugin/ActiveSearch.java#L71.  This might be of some use, idk

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private HashMap<String, Pair<String, ArrayList<String>>> transformersProperties;
```
vinamartin: ✏️  While a pair works out now for what we're displaying currently, I wonder if there's a way we can make this more generic if we ever wanted to display more information.
No change is needed here, just food for thought.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
vinamartin: I think there's some code in Karaf that does this, let me search for it.

josephthweatt: I see that `config:list` uses lines, anyone know where that is?

josephthweatt: Okay I looked at the `config:list` [source code](https://github.com/apache/karaf/blob/4888b6a1788d4d83c90f601ec00e386706a6703a/config/src/main/java/org/apache/karaf/config/command/ListCommand.java) and they are just using:
```System.out.println("----------------------------------------------------------------");```
Not sure if you were able to find anything, but if karaf isn't using a special utility for their own features I'm not sure it exists. Let me know if you find anything.

Comment (vinamartin) [inline catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java]: Yeah, that's exactly what I was looking at, hah, okay, your solution here is definitely better. No action required!

File: catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java [hunk 3]
```diff
@@ -0,0 +1,111 @@
...
+    private static final String NOT_AVAILABLE = "N/A";
+    private static final String LINE = "------------------------";
+
+    HashMap<String, Pair<String, ArrayList<String>>> transformersProperties;
+    ArrayList<ServiceReference<InputTransformer>> serviceReferences;
+    
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        transformersProperties = new HashMap<>();
+        serviceReferences = (ArrayList<ServiceReference<InputTransformer>>) bundleContext.getServiceReferences(InputTransformer.class, null);
```
vinamartin: ✏️  It's rare, but I think this reference list could come back null. I wonder if we should protect against that.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+@Command(scope = CatalogCommands.NAMESPACE, name = "transformers", description = "Provides information on available transformers.")
+public class TransformersCommand extends CatalogCommands {
+
+    private static final String MIME_TYPE = "mime-type";
+    private static final String SCHEMA = "schema";
+
+    private static final String NOT_AVAILABLE = "N/A";
+    private static final String LINE = "------------------------";
+
+    HashMap<String, Pair<String, ArrayList<String>>> transformersProperties;
```
coyotesqrl: ❗️ Don't use `Pair` for this. That's a horrible misuse of a graphics library.

✏️ Look at [Guava's Multimap](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Multimap.html) instead. 

What you have here is a `Map<String, Map<String, List<String>>>` where you want to constrain the inner map to a single entry. I suppose you could make a `Map<String, Map.Entry<String, List<String>>>` but that would be overkill. So, using a Multimap, you can create a `Map<String, Multimap<String>>` and work with it more easily.

Also, even though you're using these internally and not returning them, you should get in the habit of declaring things abstractly: e.g. `Map<...>` instead of `HashMap<...>`.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+    ArrayList<ServiceReference<InputTransformer>> serviceReferences;
```
coyotesqrl: Again, declare as a `List<...>`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+import javafx.util.Pair;
+
+/**
+ * Provides information on available transformers
+ */
+@Service
+@Command(scope = CatalogCommands.NAMESPACE, name = "transformers", description = "Provides information on available transformers.")
+public class TransformersCommand extends CatalogCommands {
+
+    private static final String MIME_TYPE = "mime-type";
```
coyotesqrl: ❗️ You need to run the formatter over this class.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
coyotesqrl: ❗️ Both `transformersProperties` and `serviceReferences` should be declared in the `executeWithSubject()` method.

While it is vanishingly unlikely to happen in this particular case, it is possible that two users could be logged into the console at the same time and invoke this command concurrently. Depending on how the threads progressed, you could end up in an incomplete and incorrect state. Since these fields have no use outside the `executeWithSubject()` call stack, there's no reason for them to be represented as shared state.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+                console.printf("%n");
+            }
+        }
+
+        return null;
+    }
+
+    private String getTransformerProperty(ServiceReference ref, String property) {
+        if (ref.getProperty(property) != null) {
```
coyotesqrl: ✏️ Alternatively, you could write this as
```java
        return Optional.ofNullable(ref.getProperty(property))
                .map(String.class::cast)
                .orElse(NOT_AVAILABLE);
```

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+        int activeTransformers = transformersProperties.size();
+
+        if (activeTransformers == 0) {
+            console.printf("There are no active transformers%n%n");
+
+        } else {
+            console.printf("%nActive Transformers: %d%n%s%n%n", activeTransformers, LINE);
+
+            int entryCount = 1;
+            for (Map.Entry<String, Pair<String, ArrayList<String>>> t: transformersProperties.entrySet()) {
```
coyotesqrl: ❗️ Now looking at how you're using this collection, I'm leaning very strongly toward you creating a private static inner class to hold the id, schema, and mime types. This would allow you to expand it in the future if needed and would clarify your access here. Instead of calls like `t.getKey()` and `t.getValue().getValue()`, you could call `t.getId()` and `t.getMimeTypes()`.

✏️ In fact, you could let the static class have methods that encapsulate some or all of this String formatting logic and then really simplify your code here.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                console.printf("\t%s: %s%n",
+                        SCHEMA, t.getValue().getKey());
+                console.printf("\t%s: %s%n",
+                        MIME_TYPE + "s", t.getValue().getValue().remove(0));
+
+                for (String mimeType : t.getValue().getValue()) {
+                    console.printf("\t\t    %s%n", mimeType);
+                }
+
+                if (entryCount++ != activeTransformers) {
```
coyotesqrl: ✏️ You could also consider using an iterator over the collection and checking `hasNext()` here instead of maintaining a counter.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+                    console.printf("%n%s%s%s%s%n", LINE, LINE, LINE, LINE);
```
coyotesqrl: ✏️ Look at [Apache's `StringUtils.repeat()`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#repeat-java.lang.String-int-) method.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private String getTransformerProperty(ServiceReference ref, String property) {
+        if (ref.getProperty(property) != null) {
+            return (String) ref.getProperty(property);
+        } else {
+            return NOT_AVAILABLE;
+        }
+    }
+
+    private ArrayList<String> getTransformerMimeProperty(ServiceReference ref) {
```
coyotesqrl: ❗️ Return `List<String>`.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
```diff
@@ -0,0 +1,6 @@
+package org.codice.ddf.commands.catalog;
+
+public class TransformersCommandTest extends ConsoleOutputCommon {
+
```
coyotesqrl: ❓ I feel as though something is missing here...

josephthweatt: Yeah, tests are WIP. I'll have them out next week.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Override
+    protected Object executeWithSubject() throws Exception {
+        List<TransformerProperties> transformersProperties = new ArrayList<>();
+
+        // ServiceReferences are needed to get property information from InputTransformers
+        List<ServiceReference<InputTransformer>> serviceReferences =
+                (ArrayList<ServiceReference<InputTransformer>>) bundleContext.getServiceReferences(
+                        InputTransformer.class,
+                        null);
```
coyotesqrl: ✏️ If you add a filter here - `"(id=*)"` - then you can simplify your processing below and not check for a null id. The filter will only return those `InputTransformer`s that have an id.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,147 @@
...
+
+        for (ServiceReference ref : serviceReferences) {
+            String id = getTransformerProperty(ref, Constants.SERVICE_ID);
+
+            if (!id.equals(NOT_AVAILABLE)) {
+                String schema = getTransformerProperty(ref, SCHEMA);
+                List<String> mimeType = getTransformerMimeProperty(ref);
+
+                transformersProperties.add(new TransformerProperties(id, schema, mimeType));
```
coyotesqrl: ✏️ You can make the ctor for `TransformerProperties` smarter, passing in the `ServiceReference` and letting it extract the fields it needs. Then you can replace this loop with something simpler like...
```java
List<TransformerProperties> transformersProperties = serviceReferences.stream()
    .map(TransformerProperties::new)
```

josephthweatt:
Just pushed up the Tests and Documentation changes.

josephthweatt:
build now

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        private String getTransformerProperty(ServiceReference ref, String property) {
+            return Optional.ofNullable(ref.getProperty(property))
+                    .map(String.class::cast)
+                    .orElse(NOT_AVAILABLE);
+        }
+
+        private List<String> getTransformerMimeProperty(ServiceReference ref) {
+            List<String> mimeTypes;
```
coyotesqrl: ❗️ Definitely do this.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,141 @@
...
+                s.append("\t\t    ")
+                        .append(mimeType)
+                        .append("\n");
+            }
+            return s.toString();
+        }
+
+        private String getTransformerProperty(ServiceReference ref, String property) {
+            return Optional.ofNullable(ref.getProperty(property))
+                    .map(String.class::cast)
```
coyotesqrl: ✏️ I know you've applied the suggestion I made in an earlier review and I do think it's sufficient here, but generally, you'll want to filter prior to the cast to prevent class cast exceptions with
```java
                    .filter(String.class::isInstance)
```
It's probably best to make the change now lest someone see this in the future and use it as a model without understanding that the instance check was intentionally elided.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
```diff
@@ -0,0 +1,136 @@
...
+        List<ServiceReference<InputTransformer>> serviceReferences = new ArrayList<>();
+        ServiceReference<InputTransformer> serviceReference = mock(ServiceReference.class);
+        serviceReferences.add(serviceReference);
+
+        when(serviceReference.getProperty(SERVICE_ID)).thenReturn(null);
+        when(bundleContext.getServiceReferences(InputTransformer.class, null)).thenReturn(
+                serviceReferences);
+
+        transformersCommand.executeWithSubject();
+        assertThat(consoleOutput.getOutput(), containsString("There are no active transformers"));
```
coyotesqrl: ✏️ It might be a bit less brittle for testing purposes if you convert the string messages you'll be testing for here into package-private consts in the `TransformersCommand` class. This way, if the wording of the messages ever changes, your unit tests will remain in sync.

We don't traditionally do that, but commands are also an outlier in that they're the odd birds in which we test output like this.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Test
+    public void testActiveTransformerCount() throws Exception {
+
+        List<ServiceReference<InputTransformer>> serviceReferences;
+
+        ServiceReference<InputTransformer> serviceReference = mock(ServiceReference.class);
+        when(serviceReference.getProperty(SERVICE_ID)).thenReturn("Test");
+
+        serviceReferences = new ArrayList<>(Arrays.asList(serviceReference,
```
coyotesqrl: ✏️ You might consider using Guava for this:
```java
serviceReferences = ImmutableList.of(serviceReference, serviceReference, serviceReference);
```

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
```diff
@@ -0,0 +1,136 @@
...
+    TransformersCommand transformersCommand;
+
+    @Before
+    public void setUp() {
+        transformersCommand = new TransformersCommand();
+        transformersCommand.setBundleContext(bundleContext);
+    }
+
+    @Test
+    public void testNullIds() throws Exception {
```
coyotesqrl: ✏️ I'd like to see a test where you return N service references with ids and M refs without and then confirm you print out only N as your count.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/TransformersCommandTest.java
josephthweatt: @coyotesqrl Should I make a separate file to hold that? Where would I put it?

coyotesqrl: Nah. I'd just make them package-private constants in `TransformersCommand`.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+
+        Iterator<TransformerProperties> tpIterator = transformersProperties.iterator();
+        TransformerProperties tp;
+
+        while (tpIterator.hasNext()) {
+            tp = tpIterator.next();
+            console.printf("%s", tp.printProperties());
+
+            if (tpIterator.hasNext()) {
+                console.printf("%n%s%n", StringUtils.repeat(LINE, 4));
```
clockard: My personal preference would be to leave the `---` line out and just do a `\n\n`. Since we don't have the width of the console it is very likely that the dashes will wrap which is kind of annoying

josephthweatt: The line is only 96 characters long, so it would take a very narrow terminal to wrap the line. The line helps provide visual assistance to show where a new transformer is getting displayed, so I think its worth the risk. Do you know any terminal that might have a width less than 96 chars?

vinamartin: A width of 80 is pretty common, if I recall correctly.

josephthweatt: I'll bump the size of the line to 72 (LINE * 3). The lines than `mvn` uses when they are building modules is 73, so that might be more appropriate.

josephthweatt:
build now

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+
+    // Transformer Properties
+
+    private static final String MIME_TYPE = "mime-type";
+
+    private static final String SCHEMA = "schema";
+
+    private static final String NOT_AVAILABLE = "N/A";
+
+
```
emmberk: ✏️ Extra lines here

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+    private static class TransformerProperties {
+
+        private String id;
+
+        private String schema;
+
+        private List<String> mimeTypes;
+
+        public TransformerProperties(ServiceReference ref) {
```
emmberk: ✏️  In the method above you have a blank line as the first line in the method. These methods (and all methods in the unit test class) could be consistent.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        private String getTransformerProperty(ServiceReference ref, String property) {
+            return Optional.ofNullable(ref.getProperty(property))
+                    .map(String.class::cast)
+                    .filter(String.class::isInstance)
+                    .orElse(NOT_AVAILABLE);
+        }
+
+        private List<String> getTransformerMimeProperty(ServiceReference ref) {
```
emmberk: ✏️  This method could be renamed to `getTransformerMimeProperties`, or maybe even better `getTransformerMimeTypes`, because it will return multiple mime types, right?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+
+            for (String mimeType : mimeTypes) {
+                s.append("\t\t    ")
+                        .append(mimeType)
+                        .append("\n");
+            }
+            return s.toString();
+        }
+
+        private String getTransformerProperty(ServiceReference ref, String property) {
```
emmberk: ✏️  You could rename this method to `getTransformerPropertyStringValue` or something like that to clarify that it will only work for properties that are strings, as opposed to, for example, the mime type string list property.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+                mimeProperties = (ArrayList<String>) ref.getProperty(MIME_TYPE);
+            } else {
+                mimeProperties = new ArrayList<>();
+                mimeProperties.add(getTransformerProperty(ref, MIME_TYPE));
+            }
+            return mimeProperties;
+        }
+    }
+
+    // Used for testing TransformersCommand
```
emmberk: 👍

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/TransformersCommand.java
```diff
@@ -0,0 +1,151 @@
...
+    protected void setBundleContext(BundleContext bundleContext) {
+        super.bundleContext = bundleContext;
```
emmberk: ✏️  Will `this.bundleContext = bundleContext;` work here instead?

josephthweatt:
build now

PR #2090, opened 2017-06-23
DDF-3105 Create Component test for REST endpoint

lessarderic:
Includes initial REST endpoint component test and infrastructure code.

@AzGoalie 
@tbatie 
@clockard
@shaundmorris

@coyotesqrl
@brendan-hofmann 

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:


libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/KarafOptions.java
```diff
@@ -0,0 +1,104 @@
...
+                            logLevel))
+                    .collect(Collectors.toList());
+            return composite(options.toArray(new Option[0]));
+        }
+
+        return null;
+    }
+
+    private Option getDebugOptions() {
+        return when(Boolean.getBoolean("debugEnabled")).useOptions(debugConfiguration());
```
AzGoalie: Should we change this option to be the same as the itest debug option `isDebugEnabled` or change the itests option. I think there should consistency for enabling test debugging.

catalog/rest/catalog-rest-endpoint/pom.xml
```diff
@@ -24,7 +24,15 @@
     <artifactId>catalog-rest-endpoint</artifactId>
     <packaging>bundle</packaging>
     <name>DDF :: Catalog :: REST :: Endpoint</name>
+
+    <properties>
+        <pax.exam.version>4.11.0</pax.exam.version> <!-- TODO: Move to root pom file -->
```
AzGoalie: Are we not able to use the versions from the component-test pom? I'm not too familiar with maven property propagation.

catalog/rest/catalog-rest-endpoint/pom.xml
```diff
@@ -127,6 +125,138 @@
...
+            <artifactId>catalog-core-api</artifactId>
+            <classifier>features</classifier>
+            <type>xml</type>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- Pax-Exam -->
+        <dependency>
+            <groupId>org.ops4j.pax.exam</groupId>
```
AzGoalie: Are these not brought in by the component-test dependency?

catalog/rest/catalog-rest-endpoint/src/test/java/org/codice/ddf/endpoints/rest/RestEndpointIT.java
```diff
@@ -0,0 +1,235 @@
...
+    @Rule
+    public final ServiceRegistrationRule serviceRegistrationRule = new ServiceRegistrationRule();
+
+    @Inject
+    ConfigurationAdmin configAdmin;
+
+    @Inject
+    private BundleContext bundleContext;
+
+    @MockOsgiService
```
AzGoalie: Love these annotations 😍

catalog/rest/catalog-rest-endpoint/src/test/java/org/codice/ddf/endpoints/rest/RestEndpointIT.java
```diff
@@ -0,0 +1,235 @@
...
+                .statusCode(200)
+                .header(HttpHeaders.CONTENT_LENGTH, is(String.valueOf(contentBytes.length)))
+                .header(HttpHeaders.ACCEPT_RANGES, "bytes")
+                .contentType(contentType)
+                .body(is(contentText));
+    }
+
+    @Override
+    protected ApplicationOptions getApplicationOptions() {
+        return new DdfOptions(portFinder) {
```
AzGoalie: I like this usage better than mine, nice!

libs/component-tests/src/main/java/org/codice/ddf/test/component/AbstractComponentTest.java
```diff
@@ -0,0 +1,64 @@
...
+                getTestBundleOptions().build());
+    }
+
+    protected ContainerOptions getContainerOptions() {
+        return new KarafOptions();
+    }
+
+    protected abstract ApplicationOptions getApplicationOptions();
+
+    // TODO - See if we can put those in a feature file instead
```
AzGoalie: For putting these in a feature file, could add one to this bundle.

libs/component-tests/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+                        <id>attach-artifacts</id>
+                        <phase>package</phase>
+                        <inherited>false</inherited>
+                        <goals>
+                            <goal>attach-artifact</goal>
+                        </goals>
+                        <configuration>
+                            <artifacts>
+                                <artifact>
+                                    <file>target/classes/features.xml</file>
```
AzGoalie: I don't see a feature file for this bundle, where is it getting it from?

catalog/core/catalog-core-api/src/main/resources/features.xml
```diff
@@ -0,0 +1,29 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+ -->
+<features name="${project.artifactId}-${project.version}"
+          xmlns="http://karaf.apache.org/xmlns/features/v1.3.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://karaf.apache.org/xmlns/features/v1.3.0 http://karaf.apache.org/xmlns/features/v1.3.0">
+
+    <feature name="catalog-core-api" install="manual" version="${project.version}"
```
brendan-hofmann: ❓ Does this really belong here if it's only used by the component test? Or is it the start of greater things to come? 😻

catalog/rest/catalog-rest-endpoint/pom.xml
brendan-hofmann: It's hierarchical inheritance. In other words, you can only use properties that belong to  yourself or a parent of yourself. There are workarounds like importing other poms directly but I'd rather just move it up to root like the todo says.

libs/component-tests/pom.xml
brendan-hofmann: ❓ is it being pulled from one of the bundles under test?

libs/component-tests/src/main/java/org/codice/ddf/test/component/UrlBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+                System.getProperty(portSystemProperty),
+                System.getProperty(securePortSystemProperty),
+                System.getProperty(rootContextSystemProperty));
+    }
+
+    public UrlBuilder add(String contextPath) {
+        String cleanContextPath = cleanContextPath(contextPath);
+
+        try {
+            return new UrlBuilder(//
```
brendan-hofmann: ❓ This seems like more manipulation than necessary to build up the URL. Why not just keep track of the context path and build the secure or public URL when it is requested? If you wanted to get really fancy you could even cache the URL and invalidate it when `add()` is called. For sanity you could prohibit slashes, e.g. `add(services).add(catalog)...` That feels more buildery to me anyways.

This is probably premature optimization, but it seems like it could be an easy thing to trip over in loops or as we start to use this class more throughout the codebase.

libs/component-tests/src/main/java/org/codice/ddf/test/component/UrlBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+    private UrlBuilder(URL rootContextUrl, URL secureRootContextUrl) {
+        this.rootContextUrl = rootContextUrl;
+        this.secureRootContextUrl = secureRootContextUrl;
+        LOGGER.debug("UrlBuilder created for {}, {}", rootContextUrl, secureRootContextUrl);
+    }
+
+    private UrlBuilder(String hostname, String port, String securePort, String rootContext) {
+
+        try {
+            this.rootContextUrl = new URL(String.format("http://%s:%s/%s",
```
brendan-hofmann: ❓ Some of the refactoring I mentioned above could also clean up this code down here. We shouldn't need to do the same construction, validation, and logging in multiple places.

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/DdfOptions.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.editConfigurationFilePut;
+
+import org.codice.ddf.test.component.configurators.BundleOptionBuilder.BundleOption;
+import org.codice.ddf.test.component.configurators.FeatureOptionBuilder.FeatureOption;
+import org.ops4j.pax.exam.Option;
+import org.ops4j.pax.exam.util.PathUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DdfOptions implements ApplicationOptions {
```
brendan-hofmann: ❗️ These constants look familiar. Deliberate tech debt? 😺  https://github.com/codice/ddf/blob/db54111591e3cde03507c8f7b718304aeae36e66/libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java#L20

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/KarafOptions.java
brendan-hofmann: ❗️ Agreed for consistency's sake. I have no strong preference save that `isDebugEnabled` was there first, so changing this one has a lower cognitive overhead.

libs/component-tests/src/main/resources/logback.xml
```diff
@@ -0,0 +1,30 @@
...
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<configuration>
```
brendan-hofmann: ❓ Do you really need this file?

libs/component-tests/src/main/java/org/codice/ddf/test/component/rules/TestFailureLogger.java
```diff
@@ -0,0 +1,109 @@
...
+            new ImmutableMap.Builder<Integer, String>().put(Bundle.UNINSTALLED, "UNINSTALLED")
+                    .put(Bundle.INSTALLED, "INSTALLED")
+                    .put(Bundle.RESOLVED, "RESOLVED")
+                    .put(Bundle.STARTING, "STARTING")
+                    .put(Bundle.STOPPING, "STOPPING")
+                    .put(Bundle.ACTIVE, "ACTIVE")
+                    .build();
+
+    @Override
+    protected void failed(Throwable e, Description description) {
```
brendan-hofmann: ❗️ It looks like this doesn't log the stacktrace anywhere. Can we do something like https://github.com/codice/ddf/blob/4cb34066fcbd9cf9a885c41c6336374dc4e8f0f0/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/utils/LoggingUtils.java#L29 instead?

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/DdfOptions.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private final PortFinder portFinder;
+
+    public DdfOptions(PortFinder portFinder) {
+        this.portFinder = portFinder;
+    }
+
+    @Override
+    public Option get() {
+        return composite(//
```
peterhuffer: Are this comment and one on line 81 here on purpose?

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
```diff
@@ -912,6 +910,12 @@ CreateInfo parseAttachments(List<Attachment> contentParts, String transformerPar
         return createInfo;
     }
 
+    // Package-private for unit testing
```
coyotesqrl: ❓ Wouldn't you want some package-private setter so you can inject a mock?

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/DdfOptions.java
coyotesqrl: ☝️ They block the IJ code formatter from making stupid/ugly choices.

libs/component-tests/src/main/java/org/codice/ddf/test/component/rules/MockOsgiServiceField.java
```diff
@@ -0,0 +1,90 @@
...
+            throw new UnsupportedOperationException(errorMessage, e);
+        }
+    }
+
+    void cleanup() {
+        serviceRegistration.unregister();
+    }
+
+    private Dictionary<String, Object> getServiceProperties(MockOsgiService annotation) {
+        return Arrays.stream(annotation.properties())
```
coyotesqrl: ✏️ Alternatively (but not much more efficient because of the wrap into a `Dictionary` implementation)
```java
        return new Hashtable<>(Arrays.stream(annotation.properties())
                .collect(Collectors.toMap(MockOsgiService.Property::key, MockOsgiService.Property::value)));
```

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/KarafOptions.java
lessarderic: Renamed and created a constant for it too.

catalog/rest/catalog-rest-endpoint/pom.xml
```diff
@@ -137,22 +245,48 @@
...
                             javax.ws.rs.*,
                             *
                         </Import-Package>
                     </instructions>
                 </configuration>
             </plugin>
             <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <!-- Need to use an older version of failsafe for versionAsInProject() to work -->
+                <version>2.17</version>
```
shaundmorris: ❓ should we use a property for the failsafe version so we can use the same version throughout?

libs/component-tests/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>
+                            ${project.groupId}.${project.artifactId}
+                        </Bundle-SymbolicName>
+                        <Bundle-Name>${project.name}</Bundle-Name>
+                        <Import-Package>
+                            *;resolution:=optional
```
brendan-hofmann: I could see an argument for leeway given that this is a component test. A justifying comment wouldn't hurt in that case, as it is normally an antipattern.

tbatie: This has bit us several times in the past with the test-itests-common bundle

brendan-hofmann: Valid point, I'm convinced. The failure case is much clearer if we are explicit here.

lessarderic:
@coyotesqrl, @AzGoalie, @shaundmorris, @brendan-hofmann, @tbatie, @peterhuffer, @clockard: PR has been updated and is ready for another round of review. Please let me know how you feel about the stuff I moved from test-itests-common to test-common.

Still TBD: fix the itests and look into the TODOs I left.

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

catalog/rest/catalog-rest-endpoint/pom.xml
lessarderic: Done

libs/component-tests/src/main/java/org/codice/ddf/test/component/UrlBuilder.java
lessarderic: Good input @brendan-hofmann. I wanted to put that out there so we can start iterating on it. I'm all for improving it but don't want to hold up this PR any longer.

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/DdfOptions.java
lessarderic: Good catch! Removed duplicated constants.

libs/component-tests/src/main/resources/logback.xml
lessarderic: It helps reduce the logging noise and prevents logging errors at startup.

libs/component-tests/src/main/java/org/codice/ddf/test/component/configurators/DdfOptions.java
lessarderic: Removed those since we've changed formatter since then.

catalog/rest/catalog-rest-endpoint/pom.xml
lessarderic: Looks like we're using different versions of failsafe throughout the project so we would need to make sure everything still works as expected if we wanted to make that change.

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

libs/test-common/src/main/java/org/codice/ddf/test/common/rules/MockOsgiServiceField.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  /** Cleans up the mock service registration. */
+  void cleanup() {
+    serviceRegistration.unregister();
+  }
+
+  private Dictionary<String, Object> getServiceProperties(MockOsgiService annotation) {
+    return new Hashtable<>(
```
coyotesqrl: ✏️ Might also consider using the `org.codice.ddf.configuration.DictionaryMap` here:
```java
    return Arrays.stream(annotation.properties())
        .collect(
            Collectors.toMap(
                MockOsgiService.Property::key,
                MockOsgiService.Property::value,
                (v1, v2) -> v2,
                DictionaryMap::new));
```
It's unfortunately a bit more verbose, as a merge function needs to be provided, but avoids creating a map just to use it to construct a `Hashtable`.

catalog/rest/catalog-rest-endpoint/pom.xml
```diff
@@ -137,20 +241,47 @@
...
                             javax.ws.rs.*,
                             *
                         </Import-Package>
                     </instructions>
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <!-- Need to use an older version of failsafe for versionAsInProject() to work -->
+                <version>2.17</version>
```
brendan-hofmann: ✏️ Since we need this workaround in a few places, can we extract the variable as something like `failsafe.versionAsInProjectWorks.version`?

catalog/rest/catalog-rest-endpoint/src/test/java/org/codice/ddf/endpoints/rest/RestEndpointIT.java
```diff
@@ -0,0 +1,249 @@
...
+                when().head(restEndpointUrlBuilder.build()).then().assertThat().statusCode(200);
+              } catch (Throwable e) {
+                return false;
+              }
+
+              return true;
+            });
+  }
+
+  // TODO - Change to make sure all the bundles of the feature under test are active and move to
```
brendan-hofmann: ❓ Will you create a ticket for this or address it in this PR?

lessarderic:
@coyotesqrl, my hope was that those `feature.xml` files wouldn't be used only for testing, but could be used to actually install and start the bundle related to the Maven module it's contained in, and eventually be used as feature dependencies in other `features.xml` files that contain higher-level features/apps.

The reason I called them `feature.xml` is because they contain only one feature. Also because I couldn't keep the name `features.xml` since that created false positives when pomfix ran.

I'm open the suggestions here though. If we feel this isn't the right approach, I can rename them to something else for now.

libs/test-common/src/main/java/org/codice/ddf/test/common/rules/MockOsgiServiceField.java
lessarderic: Done

catalog/rest/catalog-rest-endpoint/pom.xml
lessarderic: Added `<failsafe.pax-exam-compatible.version>` property to root pom file and updated other pom files accordingly.

catalog/rest/catalog-rest-endpoint/src/test/java/org/codice/ddf/endpoints/rest/RestEndpointIT.java
lessarderic: Created https://codice.atlassian.net/browse/DDF-3547 and updated TODO.

lessarderic:
build now

PR #2097, opened 2017-06-26
DDF-3034 Paginator and paginator unit tests

ahoffer:
#### What does this PR do?
1. Create a new class, QueryResultsPaginator, to return a certain number of records (one page worth of results) until there are not more results left.
2. Modify ExportCommand to use the paginator.
@mweser @lessarderic @rzwiefel @AzGoalie @emanns95 @jhunzik 
@coyotesqrl
@lessarderic
For manual testing:
1. Ingest between 129-191 documents into a catalog (export command's page size is 64).
2. From the search UI, do a wildcard search. Note the exact number of results. For example, say there were "130" search results.
3. Edit a metacard, save the changes, and check the History tab for that metcard. There should be one entry in that card's history.
4. From the DDF console, execute `catalog:export --delete=false`.
5. Look at how many metacards were exported.  The expected results it that It should equal the number of search results plus the number of historical entries, 130+1=131. So look for "Number of metacards exported: 131" in the output from the export command.

The paginator is part of a larger effort championed by @lessarderic and @mweser 
#### Checklist:
- [X ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
@@ -74,7 +74,6 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
      */
     public boolean hasNext() {
 
-        boolean hasMoreResults;
         if (notEmpty()) {
             return true;
         } else {
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
@@ -43,7 +43,7 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
 
     private final CatalogFramework catalogFramework;
 
-    private final QueryRequestImpl queryRequestCopy;
+    private final QueryImpl copyOfQuery;
 
     private int currentIndex;
 
@@ -54,15 +54,13 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
 
     private BlockingQueue<Result> buffer;
 
-    public QueryResultPaginator(CatalogFramework catalogFramework, QueryRequestImpl queryRequest) {
+    public QueryResultPaginator(CatalogFramework catalogFramework, QueryImpl query) {
         Validate.notNull(catalogFramework);
-        Validate.notNull(queryRequest);
+        Validate.notNull(query);
         this.catalogFramework = catalogFramework;
-        this.queryRequestCopy = clone(queryRequest);
-        pageSize = queryRequestCopy.getQuery()
-                .getPageSize();
-        currentIndex = queryRequest.getQuery()
-                .getStartIndex();
+        this.copyOfQuery = copy(query);
+        pageSize = copyOfQuery.getPageSize();
+        currentIndex = copyOfQuery.getStartIndex();
         Validate.inclusiveBetween(1,
                 Integer.MAX_VALUE,
                 currentIndex,
@@ -76,14 +74,21 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
      */
     public boolean hasNext() {
 
-        if (exhausted && buffer.isEmpty()) {
-            return false;
+        boolean hasMoreResults;
+        if (notEmpty()) {
+            return true;
+        } else {
+            if (exhausted) {
+                return false;
+            } else {
+                //Buffer is empty, but query might not be exhausted.
+                fetchNext();
+                return notEmpty();
+            }
         }
+    }
 
-        if (!exhausted) {
-            fetchNext();
-        }
-
+    protected boolean notEmpty() {
         return !buffer.isEmpty();
     }
 
@@ -112,14 +117,13 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
         return buffer.size();
     }
 
-    QueryRequestImpl clone(QueryRequestImpl queryRequest) {
-        QueryImpl query = (QueryImpl) queryRequest.getQuery();
-        return new QueryRequestImpl(new QueryImpl(query.getFilter(),
+    QueryImpl copy(QueryImpl query) {
+
+        return new QueryImpl(query.getFilter(),
                 query.getStartIndex(),
                 query.getPageSize(),
                 query.getSortBy(),
-                query.requestsTotalResultsCount(),
-                query.getTimeoutMillis()));
+                query.requestsTotalResultsCount(), query.getTimeoutMillis());
     }
 
     int getCurrentIndex() {
@@ -135,7 +139,7 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
      * If no results were returned, assume there are no more results possible.
      */
     private void fetchNext() {
-        ((QueryImpl) queryRequestCopy.getQuery()).setStartIndex(currentIndex);
+        copyOfQuery.setStartIndex(currentIndex);
         List<Result> results = queryCatalogFramework();
         if (results.isEmpty()) {
             exhausted = true;
@@ -147,7 +151,7 @@ public class QueryResultPaginator implements Iterator<List<Result>> {
 
     private List<Result> queryCatalogFramework() {
         try {
-            return catalogFramework.query(queryRequestCopy)
+            return catalogFramework.query(new QueryRequestImpl(copyOfQuery))
                     .getResults();
         } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
             throw new CatalogRuntimeException(e);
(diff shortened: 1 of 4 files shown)


ahoffer:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,156 @@
...
+        buffer.drainTo(copy, pageSize);
+        return copy;
+    }
+
+    int getBufferSize() {
+        return buffer.size();
+    }
+
+    QueryRequestImpl clone(QueryRequestImpl queryRequest) {
+        QueryImpl query = (QueryImpl) queryRequest.getQuery();
```
coyotesqrl: ❗️ I can see no way that you can know the `Query` returned here is castable to a `QueryImpl`. Is there a reason you're not calling the `QueryRequestImpl` ctor that takes a `Query`?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,156 @@
...
+        List<Result> copy = new ArrayList<>(pageSize);
+        buffer.drainTo(copy, pageSize);
+        return copy;
+    }
+
+    int getBufferSize() {
+        return buffer.size();
+    }
+
+    QueryRequestImpl clone(QueryRequestImpl queryRequest) {
```
coyotesqrl: ❗️ It's mildly bothersome that you've used the method name `clone` here as it implies different semantics and looks out of place operating on a different class instead of on itself. I would rename to something like `copy`.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,156 @@
...
+        Validate.notNull(catalogFramework);
+        Validate.notNull(queryRequest);
+        this.catalogFramework = catalogFramework;
+        this.queryRequestCopy = clone(queryRequest);
+        pageSize = queryRequestCopy.getQuery()
+                .getPageSize();
+        currentIndex = queryRequest.getQuery()
+                .getStartIndex();
+        Validate.inclusiveBetween(1,
+                Integer.MAX_VALUE,
```
coyotesqrl: ❓ Do we really want to use max_int as a valid start index? I mean...it's _valid_, I suppose.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Get the next batch of results.
+     * Assume the current index is pointed to the first result
+     * that has not been retrieved.
+     * Update the current index after the results are fetched.
+     * Add the fetched results to the buffer.
+     * If no results were returned, assume there are no more results possible.
+     */
+    private void fetchNext() {
+        ((QueryImpl) queryRequestCopy.getQuery()).setStartIndex(currentIndex);
```
coyotesqrl: ❗️ Again, I'm not clear how you can know with any certainty that this is a `QueryImpl`. Although, if your `clone/copy` method were to explicitly create a `QueryImpl` that might help.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,156 @@
...
+                "Query request start index must be greater than zero");
+        exhausted = false;
+        buffer = new LinkedBlockingDeque<>();
+    }
+
+    /**
+     * @return true if there are more results for the client
+     */
+    public boolean hasNext() {
+
```
coyotesqrl: ❗️ It appears that there's nothing to protect against multiple calls to `hasNext()` without calls to `next` between them. For very large result sets, this would allow bad/malicious code to eat up all memory resources by loading the entire result set into the blocking queue.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Catalog Framework queried for results until current page is full of results
+     * or endIndex is reached.
+     *
+     * @return Collection of results
+     * @throws NoSuchElementException if the iteration has no more elements
+     */
+    public List<Result> next() throws NoSuchElementException {
+        if (!hasNext()) {
```
coyotesqrl: ❗️ This call to `hasNext()`, unless I'm mistaken, will double-load the queue if the calling code invokes 
```java
hasNext();
next();
```
in succession.

catalog/core/catalog-core-api-impl/src/test/java/ddf/catalog/util/impl/QueryResultPaginatorTest.java
```diff
@@ -0,0 +1,174 @@
...
+        queryRequestMock = mock(QueryRequestImpl.class);
+        queryResponseMock = mock(QueryResponseImpl.class);
+        queryMock = mock(QueryImpl.class);
+        when(catalogFrameworkMock.query(queryRequestMock)).thenReturn(queryResponseMock);
+        when(catalogFrameworkMock.query(queryRequestMock)).thenReturn(queryResponseMock);
+        when(queryRequestMock.getQuery()).thenReturn(queryMock);
+    }
+
+    @Test
+    public void testhasNextIdempotency() throws Exception {
```
coyotesqrl: ✏️ Not exactly idempotent as it does force the underlying queue to be filled more on each call. You need a test to ensure that does not happen.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -91,30 +91,22 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ExportCommand.class);
 
+    private static final int PAGE_SIZE = 64;
+
     private static final SimpleDateFormat ISO_8601_DATE_FORMAT;
 
+    private static Supplier<String> fileNamer;
```
coyotesqrl: ❗️ It's fine to move the initialization of this into the static block (though I'm not sure I see the advantage of it), but it should still be `final`.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -132,6 +124,26 @@
...
+    private StorageProvider storageProvider;
+
+    /**
+     * Generates stateful predicate to filter distinct elements by a certain key in the object.
+     *
+     * @param keyExtractor Function to pull the desired key out of the object
+     * @return the stateful predicate
+     */
+    private static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
+        Map<Object, Boolean> seen = new ConcurrentHashMap<>();
```
coyotesqrl: ❗️ This is going to create a new `ConcurrentHashMap` on each invocation. If the purpose of the map is to ensure uniqueness of the keys, this won't work.

I see that this code existed before; it is tech debt.

You might consider changing the `contentItemsToExport` collection in `doContentExport` to be a Set and provide a custom comparator to assert equality on these keys.

More simply, you can change this method to be:
```java
private static <T> Predicate<T> distinctByKey(Map<Object, Boolean, seen, Function<? super T, ?> keyExtractor) {
    return t -> seen.putIfAbsent(keyExtractor.apply(t), Boolean .TRUE) == null;
}
```
And then create the ConcurrentHashMap right above your stream method and include it in the params passed in.

coyotesqrl: Ignore the above comment. It was a too-quick, bad reading on my part.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryResultPaginator.java
ahoffer: I really only wanted to validate the lower bound. But the Apache library didn't have a `Validate.greateThan` method.  I could go old school:
`if(currentIndex < 1) throw new ...`

Maybe that would be clearer?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
rzwiefel: @ahoffer , Perhaps we can add an additional comment below the first line of the main javadoc comment along the lines of the following:
```
</br>
The method will be called once per stream to generate the Predicate which will close over the
map. The Predicate itself will then be called N times (using the closed-over map) to perform
the test.
```

ahoffer:
build now

mweser:
Can close this PR as this is the updated version: https://github.com/codice/ddf/pull/2080

lessarderic:
Closing this PR. Changes have been ported to https://github.com/codice/ddf/pull/2080 and comments will be addressed as part of that PR.

PR #2102, opened 2017-06-27
DDF-3038 several map-layer configuration ui improvements

djblue:
#### What does this PR do?

- rename `May Layers ` to `Map Layers`
- display valid alpha values in label
- detect invalid map layers configuration and allow users to fix or
  discard
- added yarn ignore proxy setting on build

#### Who is reviewing it?

@tbatie 
@garrettfreibott 

@andrewkfiedler
@lessarderic

- Full build of `catalog-admin-module-maplayers` module
- Insert invalid map layers config through config admin and make sure map layers ui allows users to fix or discard those configs

[DDF-3038](https://codice.atlassian.net/browse/DDF-3038)

screen shot 2017-06-27 at 1 46 54 pm

#### Checklist:
- [ ] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
+++ b/catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
@@ -1,11 +1,12 @@
 import React from 'react'
 
-import { Map } from 'immutable'
+import { Map, fromJS } from 'immutable'
 import { connect } from 'react-redux'
 
 import CircularProgress from 'material-ui/CircularProgress'
 import ContentAdd from 'material-ui/svg-icons/content/add'
 import DeleteIcon from 'material-ui/svg-icons/action/delete'
+import Dialog from 'material-ui/Dialog'
 import FlatButton from 'material-ui/FlatButton'
 import Flexbox from 'flexbox-react'
 import FloatingActionButton from 'material-ui/FloatingActionButton'
@@ -28,18 +29,23 @@ import 'brace/theme/github'
 import options from './options'
 import {
   // actions
+  set,
   fetch,
   update,
   save,
   validate,
+  validateJson,
+  validateStructure,
   reset,
   message,
+  setInvalid,
 
   // selectors
   getProviders,
   isLoading,
   hasChanges,
-  getMessage
+  getMessage,
+  getInvalid
 } from './reducer'
 
 const submittingStyle = {
@@ -117,7 +123,7 @@ const ProviderEditor = ({ provider, onUpdate, buffer, onEdit, error = Map() }) =
         floatingLabelText='Provider URL' />
     </div>
     <Flexbox flex='1' style={{ padding: '0 16px' }}>
-      <Flexbox flex='5' style={{ marginRight: 20 }}>
+      <Flexbox flex='3' style={{ marginRight: 20 }}>
         <div style={{ width: '100%' }}>
           <SelectField
             floatingLabelText='Provider Type'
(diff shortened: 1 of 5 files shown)


catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/reducer.js
```diff
@@ -45,7 +47,20 @@ export const fetch = () => (dispatch) => {
     .then((res) => res.json())
     .then((json) => {
       const config = fromJS(json).getIn(configPath)
-        .update('imageryProviders', (providers) => fromJS(JSON.parse(providers || '[]')))
+        .update('imageryProviders', (providers) => {
+          try {
+            const parsed = JSON.parse(providers)
+            const err = validateStructure(parsed)
+            if (err !== undefined) {
+              throw Error(err)
```
lessarderic: ❓ Throwing this error here and catching it on L58 feels like using the error as control flow. Instead of having the `if (err != undefined)` check here, could we have `validateStructure()` throw the `Error` instead and never return `undefined`?

djblue:
build now

catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/reducer.js
rzwiefel: @lessarderic Talked with @djblue about this, unfortunately because JSON.parse already throws exceptions  (and I would argue throwing more errors by changing validateStructure is not the right path) , the simplest way to deal with this is to have the error check and throw. You could not throw the error and just deal with it inline, but then you get duplicated code and a little deeper nesting. Even with abstracting the error handling to a method it doesn't look much cleaner (Feels like splitting up the related code far apart.. at least without changing some of these to be vars instead of consts) 
```javascript
try {
  const parsed = JSON.parse(providers)
  const err = validateStructure(parsed)
  if (err !== undefined) {
    return emptyAndDispatch(e, providers)
  }
} catch (e) {
  return emptyAndDispatch(e, providers)
}
```

catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
```diff
@@ -237,7 +243,49 @@ let FixedHeader = ({ muiTheme: { palette }, disabled = true, onSave, onReset })
...
+          enableBasicAutocompletion
+          name='json-editor'
+          value={buffer}
+          onChange={onUpdate} />
+      </Error>
+    </Dialog>
+  )
+}
+
+const MapLayers = ({ onFetch, onUpdate, onSave, onReset, onMessage, onSetInvalid, onSet, disabled, providers = [], errors, loading, message, invalid }) => (
```
rzwiefel: Doesn't your formatter have any respect for line length?  Some of us use fixed width IDE's ya know... ༼ つ ◕_◕ ༽つ 😉 
but you should look into setting a line wrap option if they allow it.

djblue: `standard` doesn't have any configuration options. I didn't notice how long it was.

catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/reducer.js
djblue: I think if we hit this code path, things have gone seriously wrong, which is kind of exceptional. The whole PR is dealing with an edge case where the backend returns malformed map-layers configuration, which it should really never do, but just in case, the user should be allowed fix/discard the configuration.

djblue:
build now

djblue:
build now

PR #2118, opened 2017-07-06
DDF-3028 - updated CertificateCommand and CertNew script to support s…

paouelle:
#### What does this PR do?
This PR add an optional argument to the certificate command to allow subject alternative names to be added to the generated certificate. Format supported on the command line is similar to the OpenSSL X509 configuration where multiple SANs can be specified separated by commas.

@Lambeaux @brjeter @lessarderic 

https://github.com/orgs/codice/teams/security/members

@figliold
@stustison

Execute the CertNew.sh or CertNew.bat to generate a new certificate and specify a -san option with a set of subject alternative names (e.g. -san DNS:localhost,IP:127.0.0.1) then verify the certificate in the keystore is defined with the SANs specified. Also verify without using the -san option for backward compatibility.

#### Checklist:
- [x] Documentation Updated
- [ ] Change Log Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
+++ b/distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
@@ -1,5 +1,5 @@
 REM Usage:
-REM   CertNew.sh [-cn <cn>|-dn <dn>] [-san <tag:name,tag:name,...>]
+REM   CertNew.sh (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]
 REM
 REM where:
 REM <cn> represents a fully qualified common name (e.g. "<FQDN>", where <FQDN> could be something like cluster.yoyo.com)
@@ -18,8 +18,6 @@ REM Create new certificate and certificate chain signed by Demo Certificate Auth
 REM The new certificate chain and private key are installed in the keystore.
 REM The alias will be the same as the common name.
 REM The localhost key will be deleted from the keystore.
-REM If no arguments specified on the command line, `hostname -f` is used as the
-REM the common-name for the certificate.
 REM Adds the specified subject alternative names if any.
 REM
 REM NOTE: Execute from the <DDF_HOME>/etc/certs directory.
--- a/distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
+++ b/distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
@@ -1,5 +1,6 @@
+@echo off
 REM Usage:
-REM   CertNew.sh (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]
+REM   CertNew.sh (-cn <cn>|-dn <dn>) [-san "<tag:name,tag:name,...>"]
 REM
 REM where:
 REM <cn> represents a fully qualified common name (e.g. "<FQDN>", where <FQDN> could be something like cluster.yoyo.com)
@@ -22,8 +23,6 @@ REM Adds the specified subject alternative names if any.
 REM
 REM NOTE: Execute from the <DDF_HOME>/etc/certs directory.
 REM NOTE: Defaults to Java Keystore file type.
-
-@echo off
 
 REM This next line captures the output of the dir command and stores it in a variable
 for /f "delims=" %%a in ('dir /S /B ..\..\security-certificate-generator*.jar') do @set JARFILE=%%a
--- a/distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
+++ b/distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
@@ -1,12 +1,26 @@
-REM Usage: CertNew -cn some.computer.com
-REM    or  CertNew -dn "cn=some.computer.com,o=some company,ou=some department"
-REM The first usage creates a certificate with a subject/common name of some.computer.com.
-REM The second usage creates a certificate with the FQDN provided and extracts the common name
+REM Usage:
+REM   CertNew.sh [-cn <cn>|-dn <dn>] [-san <tag:name,tag:name,...>]
+REM
+REM where:
+REM <cn> represents a fully qualified common name (e.g. "<FQDN>", where <FQDN> could be something like cluster.yoyo.com)
+REM <dn> represents a distinguished name as a comma-delimited string (e.g. "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>")
+REM <tag:name,tag:name,...> represents optional subject alternative names to be added to the generated certificate
+REM    (e.g. "DNS:<FQDN>,DNS:node1.<FQDN>,DNS:node2.<FQDN>"). The format for subject alternative names
+REM    is similar to the OpenSSL X509 configuration format. Supported tags are:
+REM      email - email subject
+REM      URI - uniformed resource identifier
+REM      RID - registered id
+REM      DNS - hostname
+REM      IP - ip address (V4 or V6)
+REM      dirName - directory name
 REM
 REM Create new certificate and certificate chain signed by Demo Certificate Authority.
 REM The new certificate chain and private key are installed in the keystore.
 REM The alias will be the same as the common name.
 REM The localhost key will be deleted from the keystore.
+REM If no arguments specified on the command line, `hostname -f` is used as the
+REM the common-name for the certificate.
+REM Adds the specified subject alternative names if any.
 REM
 REM NOTE: Execute from the <DDF_HOME>/etc/certs directory.
 REM NOTE: Defaults to Java Keystore file type.
--- a/distribution/ddf-common/src/main/resources/etc/certs/CertNew.sh
+++ b/distribution/ddf-common/src/main/resources/etc/certs/CertNew.sh
@@ -37,13 +37,14 @@ fi
 if [[ $1 && $2 ]]; then
     PARAM1="$1"
     PARAM2="$2"
+    shift
+    shift
 else
     PARAM1="-cn"
     PARAM2="$(hostname -f)"
 fi
-
 echo "--IGNORE SLF4J ERRORS"--
-$(java -Djavax.net.ssl.keyStore="$KEYFILE" -Djavax.net.ssl.keyStorePassword="$PASSWORD" -Djavax.net.ssl.keyStoreType="$KEYTYPE" -jar "$JARFILE" "$PARAM1" "$PARAM2")
+$(java -Djavax.net.ssl.keyStore="$KEYFILE" -Djavax.net.ssl.keyStorePassword="$PASSWORD" -Djavax.net.ssl.keyStoreType="$KEYTYPE" -jar "$JARFILE" "$PARAM1" "$PARAM2" "$@")
 
 
 if [[ $? == 0 ]]; then
--- a/distribution/ddf-common/src/main/resources/etc/certs/CertNew.sh
+++ b/distribution/ddf-common/src/main/resources/etc/certs/CertNew.sh
@@ -1,16 +1,31 @@
 #! /bin/bash 
 #
 # Usage:
-#   CertNew.sh
-#   CertNew.sh -cn <common-name>
-#   CertNew.sh -dn <distinguished name>
+#   CertNew.sh [-cn <cn>|-dn <dn>] [-san <tag:name,tag:name,...>]
+#
+# where:
+# <cn> represents a fully qualified common name (e.g. "<FQDN>", where <FQDN> could be something like cluster.yoyo.com)
+# <dn> represents a distinguished name as a comma-delimited string (e.g. "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>")
+# <tag:name,tag:name,...> represents optional subject alternative names to be added to the generated certificate
+#    (e.g. "DNS:<FQDN>,DNS:node1.<FQDN>,DNS:node2.<FQDN>"). The format for subject alternative names
+#    is similar to the OpenSSL X509 configuration format. Supported tags are:
+#      email - email subject
+#      URI - uniformed resource identifier
+#      RID - registered id
+#      DNS - hostname
+#      IP - ip address (V4 or V6)
+#      dirName - directory name
 #
 # Create new certificate and certificate chain signed by Demo Certificate Authority.  
 # The new certificate chain and private key are installed in the keystore.
 # The alias will be the same as the common name.
 # The localhost key will be deleted from the keystore.
 # If no arguments specified on the command line, `hostname -f` is used as the
-# the  the common-name for the certificate.
+# the common-name for the certificate.
+# Adds the specified subject alternative names if any.
+#
+# NOTE: Execute from the <DDF_HOME>/etc/certs directory.
+# NOTE: Defaults to Java Keystore file type.
 
 #Assume script is in <DDF>/etc/certs/ directory
 cd `dirname $0`
--- a/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
+++ b/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
@@ -83,7 +83,7 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
 
 For *NIX, use the `CertNew.sh` script.
 
-`sh CertNew.sh (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]`
+`sh CertNew.sh [-cn <cn>|-dn <dn>] [-san <tag:name,tag:name,...>]`
 
 where:
 
@@ -97,6 +97,7 @@ where:
 ** `IP` - ip address (V4 or V6)
 ** `dirName` - directory name
 
+If no arguments specified on the command line, `hostname -f` is used as the common-name for the certificate.
 ****
 
 .Windows Demo CA Script
--- a/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
+++ b/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
@@ -83,28 +83,47 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
 
 For *NIX, use the `CertNew.sh` script.
 
-`sh CertNew.sh <FQDN>`
+`sh CertNew.sh <FQDN> [-san <tag:name,tag:name,...>]`
+
+where `<subject tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+
+* `email`
+* `URI` - uniformed resource identifier
+* `RID` - registered id
+* `DNS`
+* `IP` - ip address (V4 or V6)
+* `dirName` - directory name
 
 Alternatively, a distinguished name can be provided to the script with a comma-delimited string.
 
-`sh CertNew.sh -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>"`
+`sh CertNew.sh -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <tag:name,tag:name,...>]`
 ****
 
 .Windows Demo CA Script
 ****
 For Windows, use the `CertNew.cmd` script.
 
-`CertNew -cn <FQDN>`
+`CertNew -cn <FQDN> [-san <tag:name,tag:name,...>]`
+
+where `<subject tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+
+* `email`
+* `URI` - uniformed resource identifier
+* `RID` - registered id
+* `DNS`
+* `IP` - ip address (V4 or V6)
+* `dirName` - directory name
 
 Alternatively, a distinguished name can be provided to the script with a comma-delimited string.
 
-`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>"`
+`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <subject tag:name,tag:name,...>]`
 ****
 
 The `CertNew` scripts:
 
 * Create a new entry in the server keystore.
 * Use the hostname as the fully qualified domain name (FQDN) when creating the certificate.
+* Adds the specified subject alternative names if any
 * Use the Demo Certificate Authority to sign the certificate so that it will be trusted by the default configuration.
 
 To install a certificate signed by a different Certificate Authority, see <<_managing_keystores,Managing Keystores>>.
--- a/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
+++ b/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
@@ -83,40 +83,40 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
 
 For *NIX, use the `CertNew.sh` script.
 
-`sh CertNew.sh <FQDN> [-san <tag:name,tag:name,...>]`
+`sh CertNew.sh (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]`
 
-where `<tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+where:
 
-* `email` - email subject
-* `URI` - uniformed resource identifier
-* `RID` - registered id
-* `DNS` - hostname
-* `IP` - ip address (V4 or V6)
-* `dirName` - directory name
+* `<cn>` represents a fully qualified common name (e.g. "<FQDN>", where <FQDN> could be something like cluster.yoyo.com)
+* `<dn>` represents a distinguished name as a comma-delimited string (e.g. "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>")
+* `<tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate (e.g. "DNS:<FQDN>,DNS:node1.<FQDN>,DNS:node2.<FQDN>"). The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+** `email` - email subject
+** `URI` - uniformed resource identifier
+** `RID` - registered id
+** `DNS` - hostname
+** `IP` - ip address (V4 or V6)
+** `dirName` - directory name
 
-Alternatively, a distinguished name can be provided to the script with a comma-delimited string.
-
-`sh CertNew.sh -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <tag:name,tag:name,...>]`
 ****
 
 .Windows Demo CA Script
 ****
 For Windows, use the `CertNew.cmd` script.
 
-`CertNew -cn <FQDN> [-san <tag:name,tag:name,...>]`
+`CertNew (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]`
 
-where `<tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+where:
 
-* `email` - email subject
-* `URI` - uniformed resource identifier
-* `RID` - registered id
-* `DNS` - hostname
-* `IP` - ip address (V4 or V6)
-* `dirName` - directory name
+* `<cn>` represents a fully qualified common name (e.g. "<FQDN>", where <FQDN> could be something like cluster.yoyo.com)
+* `<dn>` represents a distinguished name as a comma-delimited string (e.g. "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>")
+* `<tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate (e.g. "DNS:<FQDN>,DNS:node1.<FQDN>,DNS:node2.<FQDN>"). The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+** `email` - email subject
+** `URI` - uniformed resource identifier
+** `RID` - registered id
+** `DNS` - hostname
+** `IP` - ip address (V4 or V6)
+** `dirName` - directory name
 
-Alternatively, a distinguished name can be provided to the script with a comma-delimited string.
-
-`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <tag:name,tag:name,...>]`
 ****
 
 The `CertNew` scripts:
--- a/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
+++ b/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
@@ -123,7 +123,7 @@ The `CertNew` scripts:
 
 * Create a new entry in the server keystore.
 * Use the hostname as the fully qualified domain name (FQDN) when creating the certificate.
-* Adds the specified subject alternative names if any
+* Adds the specified subject alternative names if any.
 * Use the Demo Certificate Authority to sign the certificate so that it will be trusted by the default configuration.
 
 To install a certificate signed by a different Certificate Authority, see <<_managing_keystores,Managing Keystores>>.
--- a/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
+++ b/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
@@ -85,12 +85,12 @@ For *NIX, use the `CertNew.sh` script.
 
 `sh CertNew.sh <FQDN> [-san <tag:name,tag:name,...>]`
 
-where `<subject tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+where `<tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
 
-* `email`
+* `email` - email subject
 * `URI` - uniformed resource identifier
 * `RID` - registered id
-* `DNS`
+* `DNS` - hostname
 * `IP` - ip address (V4 or V6)
 * `dirName` - directory name
 
@@ -105,18 +105,18 @@ For Windows, use the `CertNew.cmd` script.
 
 `CertNew -cn <FQDN> [-san <tag:name,tag:name,...>]`
 
-where `<subject tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
+where `<tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
 
-* `email`
+* `email` - email subject
 * `URI` - uniformed resource identifier
 * `RID` - registered id
-* `DNS`
+* `DNS` - hostname
 * `IP` - ip address (V4 or V6)
 * `dirName` - directory name
 
 Alternatively, a distinguished name can be provided to the script with a comma-delimited string.
 
-`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <subject tag:name,tag:name,...>]`
+`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <tag:name,tag:name,...>]`
 ****
 
 The `CertNew` scripts:
--- a/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
+++ b/distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
@@ -104,7 +104,7 @@ If no arguments specified on the command line, `hostname -f` is used as the comm
 ****
 For Windows, use the `CertNew.cmd` script.
 
-`CertNew (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]`
+`CertNew (-cn <cn>|-dn <dn>) [-san "<tag:name,tag:name,...>"]`
 
 where:
 
--- a/platform/security/certificate/security-certificate-generator/pom.xml
+++ b/platform/security/certificate/security-certificate-generator/pom.xml
@@ -73,6 +73,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>findbugs</artifactId>
+            <version>${findbugs.version}</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateAuthority.java
+++ b/platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateAuthority.java
@@ -20,6 +20,7 @@ import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 
 import org.apache.commons.lang.Validate;
+import org.bouncycastle.cert.CertIOException;
 import org.bouncycastle.cert.X509CertificateHolder;
 import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
 import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder;
@@ -58,13 +59,17 @@ public class CertificateAuthority {
     }
 
     public KeyStore.PrivateKeyEntry sign(CertificateSigningRequest csr) {
-        //Converters, holders, and builders! Oh my!
-        JcaX509v3CertificateBuilder builder = csr.newCertificateBuilder(getCertificate());
-        X509CertificateHolder holder = builder.build(getContentSigner());
-        JcaX509CertificateConverter converter = newCertConverter();
         X509Certificate signedCert;
+
         try {
+            //Converters, holders, and builders! Oh my!
+            JcaX509v3CertificateBuilder builder = csr.newCertificateBuilder(getCertificate());
+            X509CertificateHolder holder = builder.build(getContentSigner());
+            JcaX509CertificateConverter converter = newCertConverter();
+
             signedCert = converter.getCertificate(holder);
+        } catch (CertIOException e) {
+            throw new CertificateGeneratorException("Could not create signed certificate.", e);
         } catch (CertificateException e) {
             throw new CertificateGeneratorException("Could not create signed certificate.",
                     e.getCause());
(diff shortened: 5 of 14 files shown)


figliold:
❗️ Don't we also need to update CertNew.cmd for Windows developers?

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,228 @@
...
+        final KeyStoreFile ksf = CertificateCommand.getKeyStoreFile();
+
+        assertThat(ksf.aliases()
+                .size(), is(2));
+        if (hasSAN) {
+            validateCertificateHasSANs(ksf, alias);
+        } else {
+            validateCertificateHasNoSAN(ksf, alias);
+        }
+    }
```
brjeter: ✏️ Private methods should be moved to the bottom.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,228 @@
...
+        System.setProperty(SecurityConstants.KEYSTORE_PATH, systemKeystoreFile.getAbsolutePath());
+        System.setProperty(SecurityConstants.KEYSTORE_PASSWORD, "changeit");
+    }
+
+    @Test
+    public void testConfigureDemoCertWithoutSAN() throws Exception {
+        final KeyStoreFile ksf = CertificateCommand.getKeyStoreFile();
+
+        assertThat(ksf.aliases()
+                .size(), is(2));
```
brjeter: ❓ Isn't this call redundant since validateKeyStore runs this assertion as well? (applies to all testConfigureDemoCert methods)

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/PkiTools.java
```diff
@@ -260,6 +263,55 @@ public static X500Name convertDistinguishedName(String... tuples) {
...
+     * @param name a string representing an X509 general name in an openssl format
+     * @return the corresponding parsed X509 general name
+     * @throws IllegalArgumentException if the string is <code>null</code>, empty or cannot be parsed properly
+     * @see <a href="https://www.openssl.org/docs/man1.0.2/apps/x509v3_config.html">OpenSSL X509 V3 Configuration</a>
+     * @see <a href="https://en.wikipedia.org/wiki/SubjectAltName">Subject Alternative Names on Wikipedia'</a>
+     */
+    public static GeneralName makeGeneralName(String name) {
+        Validate.isTrue(name != null, "Certificate general name cannot be null");
+
+        assert name != null;
```
figliold: ❓ Do you need this assert after the Validate?

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -106,7 +144,7 @@ protected static KeyStoreFile getKeyStoreFile() {
                         .toCharArray());
     }
 
-    private static String configureCert(String commonName, CertificateSigningRequest csr) {
+    static String configureCert(String commonName, CertificateSigningRequest csr) {
```
brjeter: ❓ Why is this changed from private? I don't see it being used anywhere.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,228 @@
...
+        assertThat(CertificateCommand.configureDemoCertWithDN(new String[] {"cn=John Whorfin",
+                        "o=Yoyodyne", "l=San Narciso", "st=California", "c=US"},
+                CertificateCommandTest.SANs),
+                equalTo("CN=John Whorfin,O=Yoyodyne,L=San Narciso,ST=California,C=US"));
+
+        validateKeyStore("john whorfin", true);
+    }
+
+    @Test
+    public void testMainWithCNAndWithoutSAN() throws Exception {
```
brjeter: ✏️  These tests bother me because they are all ultimately testing the same exact logic as the above tests plus the happy path for the main method. I don't think you need to do it as part of this PR, but I think CertificateCommand's methods (other than main) should be broken out to a separate class. Then this test class could just test the logic of the main method.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,228 @@
...
+        validateKeyStore("john whorfin", true);
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testMainWithTooFewArguments() throws Exception {
+        CertificateCommand.main(new String[] { "something" });
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testMainWithTooFewArgumentsWithSAN() throws Exception {
```
brjeter: ❗️ What about tests with too many arguments?

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
```diff
@@ -109,22 +166,104 @@ public void subjectName() throws Exception {
...
+                subjectName,
                 containsString("o=police box"));
     }
 
     @Test(expected = IllegalArgumentException.class)
     public void badSerialNumber() {
         csr.setSerialNumber(-1);
     }
+
+    public void testNewCertificateBuilderWithoutSAN() throws Exception {
```
brjeter: ❗️ Where's the Test annotation?

paouelle:
CertNew.cmd is already designed to pass all arguments to the Java main() method so no need to change it.

distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
```diff
@@ -83,28 +83,47 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
 
 For *NIX, use the `CertNew.sh` script.
 
-`sh CertNew.sh <FQDN>`
+`sh CertNew.sh <FQDN> [-san <tag:name,tag:name,...>]`
+
+where `<subject tag:name,tag:name,...>` represents optional subject alternative names to be added to the generated certificate. The format for subject alternative names is similar to the OpenSSL X509 configuration format. Supported tags are:
```
paouelle: left over from the previous version I made, will remove ;-)

distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
```diff
@@ -83,28 +83,47 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
...
+* `URI` - uniformed resource identifier
+* `RID` - registered id
+* `DNS`
+* `IP` - ip address (V4 or V6)
+* `dirName` - directory name
 
 Alternatively, a distinguished name can be provided to the script with a comma-delimited string.
 
-`sh CertNew.sh -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>"`
+`sh CertNew.sh -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <tag:name,tag:name,...>]`
```
paouelle: The example already provided for the distinguished name is not a fully concrete one. It still references <FQDN>. I didn't want to explode this section with now 4 examples (total combinations of all options) which is why I opted for documenting the SANs this way. What would you suggest?

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -28,66 +30,102 @@
...
+                        .map(String::trim)
+                        .toArray(String[]::new);
+                mainIndex = 2;
+                expected = 4;
+            } else if (args[2].trim()
+                    .equalsIgnoreCase("-san")) {
+                sans = Arrays.stream(args[3].split("[,]"))
+                        .map(String::trim)
+                        .toArray(String[]::new);
+                expected = 4;
```
paouelle: I think it is meant to be called from the command line window before you start DDF. It could be simplified with apache command but I didn't want to introduce it here just for that. I would say we should upgrade to the apache library if we ever need to add more arguments than that.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -28,66 +30,102 @@
...
+            }
+        }
+        if (args.length != expected) {
             String canonicalName = CertificateCommand.class.getCanonicalName();
+            String exSan = "-san DNS:localhost,IP:127.0.0.1";
```
paouelle: sure

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -28,66 +30,102 @@
...
      *
-     * @param dn String params in the form {@code attrKey=attrVal} composing a distinguished name.
-     *           e.g. {@code configureDemoCertWithDN("cn=John Whorfin", "o=Yoyodyne", "l=San Narciso", "st=California", "c=US")}
+     * @param dn              String params in the form {@code attrKey=attrVal} composing a distinguished name.
+     *                        e.g. {@code configureDemoCertWithDN(new String[] {"cn=John Whorfin", "o=Yoyodyne", "l=San Narciso", "st=California", "c=US"), null}
+     * @param subjectAltNames names in the form {@code tag:name} (format similar to OpenSSL X509 configuration)
      * @return the string used as the common name in the new certificate
      */
-    public static String configureDemoCertWithDN(String... dn) {
+    public static String configureDemoCertWithDN(String[] dn, @Nullable String[] subjectAltNames) {
```
paouelle: I think the original name was to differentiate with the method above. Now both have optional support for SANs so not sure renaming it brings any value.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
paouelle: Oops! I think it was done for my previous version of unit tests.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/PkiTools.java
paouelle: Agreed. I just modeled it after the previous method that was already there. I will remove it as I don't like that anyway.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: I am running the formatter and because they are static, it leaves them at the top

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: Notice that this validation is done one the original keystore and not on the modified one to validate pre-conditions.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: I had to add them to increase coverage. There are different paths in the main() method with these different arguments.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: I can add it; it was simply not testing any new paths or conditions in the code so I left it out.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
paouelle: it's invisible 😆

distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
```diff
@@ -83,28 +83,47 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
...
 
-`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>"`
+`CertNew -dn "c=US, st=California, o=Yoyodyne, l=San Narciso, cn=<FQDN>" [-san <tag:name,tag:name,...>]`
 ****
 
 The `CertNew` scripts:
 
 * Create a new entry in the server keystore.
 * Use the hostname as the fully qualified domain name (FQDN) when creating the certificate.
+* Adds the specified subject alternative names if any
```
ricklarsen: `.` at the end of the line.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/PkiTools.java
```diff
@@ -260,6 +263,53 @@ public static X500Name convertDistinguishedName(String... tuples) {
...
+        final int i = name.indexOf(':');
+
+        Validate.isTrue(i != -1, "General name components must be in the format tag:value");
+        final String tag = name.substring(0, i)
+                .trim();
+        final String gname = name.substring(i + 1)
+                .trim();
+
+        Validate.isTrue(!gname.isEmpty(), "General name cannot be empty");
+        switch (tag) {
```
lessarderic: ✏️  How about putting the `tag` names in a `static final Map<String, Integer>` and do?
```
Integer tagNo = tags.get(tag);

if (tagNo == null) {
    throw new ...
}

return new GeneralName(tagNo, gname)
```

If we keep the `switch`, should we define constants for the different tag names?

platform/security/certificate/security-certificate-generator/pom.xml
```diff
@@ -73,6 +73,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
lessarderic: ❓ Is this needed for the `@Nullable` annotation?

distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
paouelle: I didn't like it either as I prefer something like you proposed which is typical in Linux for command usage. However, the documentation was set like that so I tried to keep it the same as it was. Will change it.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
paouelle: It would complicate the script code and that would have to be done in Windows and Linux versions. I prefer dealing with that in Java. Eventually, we should modify that code to use apache's  command library which is better designed for this. However, doing so is still a bit more complicated right now then what we have here for now. As for the current support, it is backward compatible and allows you to call it the same way as before, pass the alternative arguments either before or after.

distribution/ddf-common/src/main/resources/etc/certs/CertNew.sh
```diff
@@ -37,13 +37,14 @@ fi
 if [[ $1 && $2 ]]; then
     PARAM1="$1"
     PARAM2="$2"
+    shift
```
paouelle: What is the problem with keeping PARAM1 and PARAM2, they represent respectively the first and second parameters retrieved from the parameters?

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateAuthority.java
```diff
@@ -58,13 +59,17 @@ public CertificateAuthority(X509Certificate certificate, PrivateKey privateKey)
...
+
         try {
+            //Converters, holders, and builders! Oh my!
+            JcaX509v3CertificateBuilder builder = csr.newCertificateBuilder(getCertificate());
+            X509CertificateHolder holder = builder.build(getContentSigner());
+            JcaX509CertificateConverter converter = newCertConverter();
+
             signedCert = converter.getCertificate(holder);
+        } catch (CertIOException e) {
+            throw new CertificateGeneratorException("Could not create signed certificate.", e);
```
paouelle: The second catch deals with the wrapper exception by propagating the cause. I do not believe there is any value here in keeping that wrapper around. As for catching RuntimeException, I any occurs then they should be runtime errors and should be left to bubble out.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -26,68 +28,105 @@
 
 public class CertificateCommand {
 
+    private static final String SAN_EXAMPLE = "-san DNS:localhost,IP:127.0.0.1";
+
     /**
      * Pass in a string to use as the common name of the certificate to be generated.
-     * Exception thrown if 0 arguments or more than 1 argument.
+     * Exception thrown if 0 arguments or more than 2 argument (or more than 4 when -san is used).
```
paouelle: Not sure what you mean about the first sentence. Done for the second one.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -26,68 +28,105 @@
...
      *
      * @param args
      */
     public static void main(String args[]) {
-        if (args.length != 2) {
+        // try to extract -san if provided
+        int expected = 2;
+        String[] sans = null;
```
paouelle: now that the arguments are documented and since this is a local variable limited to this method, I think we are ok keeping it simple.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -26,68 +28,105 @@
...
             String canonicalName = CertificateCommand.class.getCanonicalName();
             String exCn = String.format("java %s -cn \"John Smith\"", canonicalName);
-
             String exDn = String.format(
                     "java %s -dn \"cn=John Whorfin, o=Yoyodyne, l=San Narciso, st=California, c=US\"",
                     canonicalName);
             String usage = String.format(
-                    "%nUsage: java %s [-cn <common name>] | [-dn <distinguished name>]%n Examples: %n%s%n%s",
-                    canonicalName, exCn, exDn);
+                    "%nUsage: java %s [-san <subject alt names>] [-cn <common name>] | [-dn <distinguished name>]%n Examples: %n%s%n%s%n%s %s%n%s %s",
```
paouelle: I was not a fan f the way the usage was generated first and didn't want to change it originally. Changed it to align with doc changes done above

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -26,68 +28,105 @@
...
      * entry has the new certificate chain that connects the server to the Demo CA. The matching
      * private key is also stored in the entry.
      *
-     * @param commonName string to use as the common name in the new certificate.
+     * @param commonName      string to use as the common name in the new certificate.
+     * @param subjectAltNames names in the form {@code tag:name} (format similar to OpenSSL X509 configuration)
      * @return the string used as the common name in the new certificate
      */
-    public static String configureDemoCert(String commonName) {
+    public static String configureDemoCert(String commonName, @Nullable String[] subjectAltNames) {
```
paouelle: restored previous version and overloaded it.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateGenerator.java
```diff
@@ -54,7 +54,7 @@ public String configureDemoCertWithDefaultHostname() {
      * @return the string used as the common name in the new certificate
      */
     public String configureDemoCert(String commonName) {
-        return CertificateCommand.configureDemoCert(commonName);
+        return CertificateCommand.configureDemoCert(commonName, null);
```
paouelle: restored previous version and overloaded it.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequest.java
```diff
@@ -114,9 +122,21 @@ public void setSubjectKeyPair(KeyPair keyPair) {
         subjectKeyPair = keyPair;
     }
 
-    JcaX509v3CertificateBuilder newCertificateBuilder(X509Certificate certificate) {
-        return new JcaX509v3CertificateBuilder(certificate, serialNumber, notBefore.toDate(),
-                notAfter.toDate(), subjectName, getSubjectPublicKey());
+    JcaX509v3CertificateBuilder newCertificateBuilder(X509Certificate certificate)
```
paouelle: First the method is package private and second, the signature hasn't changed.

platform/security/certificate/security-certificate-generator/pom.xml
paouelle: Yup

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequest.java
```diff
@@ -62,6 +68,8 @@
 
     private X500Name subjectName;
 
+    private Set<GeneralName> subjectAltNames = new LinkedHashSet<>(8); // to preserve order
```
lessarderic: ❓ Why `8`?

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequest.java
lessarderic: We added the `throws`. But you're right, it's package-private so we're 👍

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/PkiTools.java
lessarderic: ❓ The tags are all case-sensitive, right? We don't need to lowercase them or something?

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -18,42 +18,83 @@
...
+                sans = Arrays.stream(args[3].split("[,]"))
+                        .map(String::trim)
+                        .toArray(String[]::new);
+                expected = 4;
+            }
         }
-
-        if (args[0].trim()
+        if (args.length != expected) {
+            throw new RuntimeException(String.format(
```
lessarderic: ✏️ `IllegalArgumentException` instead of `RuntimeException`?

lessarderic:
build now

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequest.java
paouelle: Missed that. 👍

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/PkiTools.java
paouelle: They can be whatever we want. In the OpenSSL tool, I believe they are case sensitive which is why I didn't bother with it.

distribution/docs/src/main/jdocs/content/_configuring/managing-certificates-contents.adoc
```diff
@@ -83,28 +83,47 @@ To use the scripts, run them out of the `<INSTALL_HOME>/etc/certs` directory.
 
 For *NIX, use the `CertNew.sh` script.
 
-`sh CertNew.sh <FQDN>`
+`sh CertNew.sh (-cn <cn>|-dn <dn>) [-san <tag:name,tag:name,...>]`
```
paouelle: Good catch. I copied and pasted everywhere when only the CertsNew.sh supports -cn and -dn to be optional. I corrected it everywhere including the doc.

distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
```diff
@@ -1,12 +1,26 @@
...
+REM      RID - registered id
+REM      DNS - hostname
+REM      IP - ip address (V4 or V6)
+REM      dirName - directory name
 REM
 REM Create new certificate and certificate chain signed by Demo Certificate Authority.
 REM The new certificate chain and private key are installed in the keystore.
 REM The alias will be the same as the common name.
 REM The localhost key will be deleted from the keystore.
+REM If no arguments specified on the command line, `hostname -f` is used as the
```
paouelle: Removed it from the Windows command as it doesn't do that. Updated the doc for the Linux command to document it. BTW, this is not a new feature added by this PR 😉

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -18,42 +18,83 @@
...
 import ddf.security.SecurityConstants;
 
 public class CertificateCommand {
-
     /**
      * Pass in a string to use as the common name of the certificate to be generated.
-     * Exception thrown if 0 arguments or more than 1 argument.
+     * Exception thrown if 0 arguments or more than 2 argument (or more than 4 when -san is used).
+     * <pre>
+     * Arguments to this program are: <code>[-cn &lt;cn&gt;|-dn &lt;dn&gt;] [-san &lt;tag:name,tag:name,...&gt;]</code>
```
paouelle: I know ... I am just used to doing it that way. I also thought that {@code} generates the block differently in the javadoc then <pre> though.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -18,42 +18,83 @@
...
+     *
+     * @param args arguments to the certificate program (see above for description)
      */
     public static void main(String args[]) {
-        if (args.length != 2) {
-            String canonicalName = CertificateCommand.class.getCanonicalName();
-            String exCn = String.format("java %s -cn \"John Smith\"", canonicalName);
+        // try to extract -san if provided
+        int expected = 2;
+        String[] sans = null;
```
paouelle: That would work but I would now need to check if the array is empty later in the internal methods. I also did that before I knew it was whitelisted as I would have preferred making it Optional instead. That being said, are you ok with it the way it is?

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
paouelle: That is the original code and I didn't want to change the original behavior unless you think we should. I would never throw RuntimeException on my own 😉

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -82,12 +142,33 @@ public static String configureDemoCert(String commonName) {
      * private key is also stored in the entry.
      *
      * @param dn String params in the form {@code attrKey=attrVal} composing a distinguished name.
-     *           e.g. {@code configureDemoCertWithDN("cn=John Whorfin", "o=Yoyodyne", "l=San Narciso", "st=California", "c=US")}
+     *           e.g. {@code configureDemoCertWithDN(new String[] {"cn=John Whorfin", "o=Yoyodyne", "l=San Narciso", "st=California", "c=US"), null}
```
paouelle: Introduced because we had to overload the methods 😉 and I copied and pasted the javadoc.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
```diff
@@ -82,12 +142,33 @@ public static String configureDemoCert(String commonName) {
...
+     * Generates new signed certificate. The input parameter is the full set of attributes of the
+     * distinguished name for the cert and optional subject alternative names. It must include a
+     * single {@code CN} value for the common name.
+     * <p>
+     * Postcondition is the server keystore is updated to include a private entry. The private
+     * entry has the new certificate chain that connects the server to the Demo CA. The matching
+     * private key is also stored in the entry.
+     *
+     * @param dn              String params in the form {@code attrKey=attrVal} composing a distinguished name.
+     *                        e.g. {@code configureDemoCertWithDN(new String[] {"cn=John Whorfin", "o=Yoyodyne", "l=San Narciso", "st=California", "c=US"), null}
```
paouelle: I prefer to keep the Nullable and allow either without the creation of an empty array. Or as I said before, I would change this to be an Optional.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequest.java
paouelle: I don't expect we will ever have a lot of SANs. 8 makes it more optimal then the default size used by the default ctors which tends to allocate too much memory for nothing.

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
lessarderic: I usually prefer to avoid `null` when there's another option but this is all internal so no big deal.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+    private static final String[] SANs = new String[] {"IP:1.2.3.4", "DNS:A"};
+
+    private static final String SANs_ARG = "IP:1.2.3.4,DNS:A";
+
+    @Rule
+    public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+    private File systemKeystoreFile = null;
+
+    private static void validateCertificateHasNoSAN(KeyStoreFile ksf, String alias)
```
lessarderic: ✏️ We usually put the `@Before`, `@Test` and `@After` methods first so people don't have to search for them, and keep the `private` methods at the bottom of the class.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+        System.setProperty(SecurityConstants.KEYSTORE_PASSWORD, "changeit");
+    }
+
+    @Test
+    public void testConfigureDemoCertWithoutSAN() throws Exception {
+        final KeyStoreFile ksf = CertificateCommand.getKeyStoreFile();
+
+        assertThat(ksf.aliases()
+                .size(), is(2));
+        assertThat(ksf.isKey("my-fqdn"), is(false));
```
lessarderic: ❗️ We should have a separate test to check that `getKeyStoreFile()` does the right thing and not have these repeated assertions in all the methods.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
lessarderic: ✏️ Our convention is to camel-case acronyms, so `DN` should be `Dn`, `SAN` should be `San`, etc.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+    private static void validateCertificateHasNoSAN(KeyStoreFile ksf, String alias)
+            throws Exception {
+        final KeyStore.Entry ke = ksf.getEntry(alias);
+
+        assertThat(ke, instanceOf(KeyStore.PrivateKeyEntry.class));
+        final KeyStore.PrivateKeyEntry pke = (KeyStore.PrivateKeyEntry) ke;
+        final Certificate c = pke.getCertificate();
+        final X509CertificateHolder holder = new X509CertificateHolder(c.getEncoded());
+
+        MatcherAssert.assertThat("There should be no subject alternative name extension",
```
lessarderic: ❓ Can't we use `assertThat` here?

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+        final KeyStore.Entry ke = ksf.getEntry(alias);
+
+        assertThat(ke, instanceOf(KeyStore.PrivateKeyEntry.class));
+        final KeyStore.PrivateKeyEntry pke = (KeyStore.PrivateKeyEntry) ke;
+        final Certificate c = pke.getCertificate();
+        final X509CertificateHolder holder = new X509CertificateHolder(c.getEncoded());
+        final org.bouncycastle.asn1.x509.Extension csn =
+                holder.getExtension(org.bouncycastle.asn1.x509.Extension.subjectAlternativeName);
+
+        MatcherAssert.assertThat(csn.getParsedValue()
```
lessarderic: ❓ Can't we use `assertThat` here?

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+                        .toASN1Primitive()
+                        .getEncoded(ASN1Encoding.DER),
+                equalTo(new GeneralNamesBuilder().addName(new GeneralName(GeneralName.iPAddress,
```
lessarderic: ✏️ Might be clearer to extract the encoded `GeneralName` to a constant.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
```diff
@@ -36,33 +53,72 @@
 
     CertificateSigningRequest csr;
 
+    private static KeyPair makeKeyPair() throws Exception {
```
lessarderic: ❗️ Tests should be first.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
```diff
@@ -36,33 +53,72 @@
...
     @Test
+    public void testAddSubjectAlternativeNames() {
+        assertThat("CSR should not have any SAN by default",
+                csr.getSubjectAlternativeNames(),
+                emptyCollectionOf(GeneralName.class));
+        csr.addSubjectAlternativeNames("IP:1.2.3.4", "DNS:A");
+        assertThat(csr.getSubjectAlternativeNames(),
+                contains(new GeneralName(GeneralName.iPAddress, "1.2.3.4"),
+                        new GeneralName(GeneralName.dNSName, "A")));
+        csr.addSubjectAlternativeNames("RID:0.2.1.4", "DNS:A");
```
lessarderic: ✏️ Not a big deal, but we should try to follow the given/when/then structure and keep the `csr.addSubjectAlternativeNames()` together first and the `assertThat()` together after.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
```diff
@@ -109,22 +166,105 @@ public void subjectName() throws Exception {
...
                 containsString("o=police box"));
     }
 
     @Test(expected = IllegalArgumentException.class)
     public void badSerialNumber() {
         csr.setSerialNumber(-1);
     }
+
+    @Test
+    public void testNewCertificateBuilderWithoutSAN() throws Exception {
```
lessarderic: ✏️ `San`.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
```diff
@@ -109,22 +166,105 @@ public void subjectName() throws Exception {
...
+        assertThat(holder.getNotBefore(), equalTo(new Time(start.toDate()).getDate()));
+        assertThat(holder.getNotAfter(), equalTo(new Time(end.toDate()).getDate()));
+        assertThat(holder.getSubject()
+                .toString(), equalTo("cn=A"));
+        assertThat("Unable to validate public key",
+                holder.getSubjectPublicKeyInfo(),
+                equalTo(SubjectPublicKeyInfo.getInstance(kp.getPublic()
+                        .getEncoded())));
+        assertThat("There should be no subject alternative name extension",
+                holder.getExtension(org.bouncycastle.asn1.x509.Extension.subjectAlternativeName),
```
lessarderic: Love those assert messages btw!

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: For the private methods request, I am using the codice code formatter and it located the static private methods at the top like that. Didn't know about the acronyms, will change. BTW, I cannot change the one in CertificateCommand as it was part of the existing API.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: hum! It should not have been that one. I guess I didn't pay attention to the auto importer.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
paouelle: Hum! These are used here to validate (or assert) the result of the tests below by verifying the content of the generated keystore and not necessarily to test the getKeyStoreFile().

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
paouelle: Again, I am using the codice formatter and it moves it to that location every time because it is static.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateSigningRequestTest.java
paouelle: Agreed when mocking, here I am not mocking so this seemed simpler.

paouelle:
build now

paouelle:
build now

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
lessarderic: Regarding the method ordering, it's strange cause I just checked my settings and it shouldn't re-order methods, just fields. Let me know and we can compare our settings to see what's different.

platform/security/certificate/security-certificate-generator/src/test/java/org/codice/ddf/security/certificate/generator/CertificateCommandTest.java
lessarderic: As a general rule we try to follow the given/when/then structure and not repeat the same assertions in multiple test cases. Makes the tests easier to follow, ensures that they are testing only one thing (https://codice.atlassian.net/wiki/display/DDF/Unit+Testing#UnitTesting-Assertions), reduces repetition (DRY) and prevents tests from failing because of something else.

paouelle:
build now

paouelle:
build now

paouelle:
build now

platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateCommand.java
paouelle: Changed anyway to appease Codacy 😄

paouelle:
build now

paouelle:
build now

distribution/ddf-common/src/main/resources/etc/certs/CertNew.cmd
```diff
@@ -1,12 +1,24 @@
...
+REM      dirName - directory name
 REM
 REM Create new certificate and certificate chain signed by Demo Certificate Authority.
 REM The new certificate chain and private key are installed in the keystore.
 REM The alias will be the same as the common name.
 REM The localhost key will be deleted from the keystore.
+REM Adds the specified subject alternative names if any.
 REM
 REM NOTE: Execute from the <DDF_HOME>/etc/certs directory.
 REM NOTE: Defaults to Java Keystore file type.
```
oconnormi: we should move the `@echo off` call above these `REM` statements since there is really no point in printing them (even though this predates this PR)

paouelle:
build now

PR #2127, opened 2017-07-07
DDF-3139 Update CrlChecker to refresh CRL on nextUpdate

kcwire:
#### What does this PR do?
Updates the CrlChecker to attempt to refresh the CRL when the nextUpdate time has passed.
#### Who is reviewing it? 
@rzwiefel 
@codice/security 
@beyelerb
@jlcsmith
System configured with a valid CRL. After start up login with a specific user's PKI. Add that certificate to the CRL and verify the CRL is refreshed on the correct interval.
Currently the CRL will only be refreshed on system restart (or bundle restart of the security-handler-pki)
[DDF-3139](https://codice.atlassian.net/browse/ddf-3139)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
+++ b/platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
@@ -1,10 +1,10 @@
 /**
  * Copyright (c) Codice Foundation
- * <p>
+ * <p/>
  * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
  * General Public License as published by the Free Software Foundation, either version 3 of the
  * License, or any later version.
- * <p>
+ * <p/>
  * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
  * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
  * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
@@ -22,9 +22,15 @@ import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.cert.CRL;
 import java.security.cert.CertificateFactory;
+import java.security.cert.X509CRL;
 import java.security.cert.X509Certificate;
+import java.util.Calendar;
 import java.util.Properties;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.commons.lang.time.DateUtils;
 import org.apache.wss4j.common.crypto.Merlin;
 import org.codice.ddf.configuration.AbsolutePathResolver;
 import org.slf4j.Logger;
@@ -42,14 +48,16 @@ public class CrlChecker {
     private static String encryptionPropertiesLocation = new AbsolutePathResolver(
             "etc/ws-security/server/encryption.properties").getPath();
 
-    private CRL crl = null;
+    private AtomicReference<CRL> crlCache = new AtomicReference<>();
+
+    private CrlRefresh refresh = new CrlRefresh();
 
     /**
      * Constructor method. Reads encryption.properties and sets CRL location
      */
     public CrlChecker() {
-        Properties encryptionProperties = loadProperties(encryptionPropertiesLocation);
-        setCrlLocation(encryptionProperties.getProperty(CRL_PROPERTY_KEY));
+        Executors.newScheduledThreadPool(1)
+                .scheduleWithFixedDelay(refresh, 0, 1, TimeUnit.HOURS);
     }
 
     /**
@@ -59,7 +67,7 @@ public class CrlChecker {
      * @return true if one of the certs passes the CRL or if CRL revocation is disabled, false if they are all revoked.
      */
     public boolean passesCrlCheck(X509Certificate[] certs) {
-        if (crl == null) {
+        if (crlCache.get() == null) {
             String errorMsg = "CRL is not set. Skipping CRL check";
             LOGGER.trace(errorMsg);
             return true;
@@ -78,7 +86,8 @@ public class CrlChecker {
         if (certs != null) {
             LOGGER.debug("Got {} certificate(s) in the incoming request", certs.length);
             for (X509Certificate curCert : certs) {
-                if (crl.isRevoked(curCert)) {
+                if (crlCache.get() != null && crlCache.get()
+                        .isRevoked(curCert)) {
                     SecurityLogger.audit("Denying access for subject DN: " + curCert.getSubjectDN()
                             + " due to certificate being revoked by CRL.");
                     return false;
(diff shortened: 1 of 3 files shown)


platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -44,12 +51,18 @@
...
+    private CrlRefresh refresh = new CrlRefresh();
+
     /**
      * Constructor method. Reads encryption.properties and sets CRL location
      */
     public CrlChecker() {
-        Properties encryptionProperties = loadProperties(encryptionPropertiesLocation);
-        setCrlLocation(encryptionProperties.getProperty(CRL_PROPERTY_KEY));
+        scheduler = Executors.newScheduledThreadPool(1);
+        scheduler.scheduleWithFixedDelay(refresh, 0, 1, TimeUnit.HOURS);
```
coyotesqrl: :question: Should we make this configurable?

kcwire: I debated about this as well. There isn't any configuration in this bundle currently and I feel like it would just be configuration for the sake of configuration.  I'm thinking of the scenario where you know the CRL updates every 24 hours. If you schedule it to every 24 hours, it likely won't align because the timer will start when the instance is created. I think I see more value in a forced update after some set interval (like once a day?) in case the integrator just sets the nextUpdate to some point in the future and actually updates the CRL prior to nextUpdate.

The RFC I've been reading for reference: https://www.ietf.org/rfc/rfc5280.txt

kcwire:
@coyotesqrl updated as suggested.

kcwire:
build now

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -154,4 +131,74 @@ Properties loadProperties(String location) {
...
+    private class CrlRefresh implements Runnable {
+
+        private Calendar start = Calendar.getInstance();
+
+        public void run() {
+            Properties encryptionProperties = loadProperties(encryptionPropertiesLocation);
+            CRL crl = crlCache.get();
+            if (crl instanceof X509CRL) {
+                X509CRL x509Crl = (X509CRL) crl;
+                if (1 == Calendar.getInstance()
```
coyotesqrl: ✏️ Clearer to do 
```java
if (Calendar.getInstance().getTime().after(x509Crl.getNextUpdate()))
```
Also, there's no guarantee that the `compareTo` call will return a `1` for after; it's just "some int greater than 0".

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -154,4 +131,74 @@ Properties loadProperties(String location) {
...
+         *                 check certificate revocation.
+         */
+        public void setCrlLocation(String location) {
+            CRL current = crlCache.get();
+            if (location == null) {
+                LOGGER.info("CRL property in {} is not set. Certs will not be checked against a CRL",
+                        encryptionPropertiesLocation);
+                crlCache.compareAndSet(current, null);
+            } else {
+                crlCache.compareAndSet(current, createCrl(location));
```
coyotesqrl: ❓ Should we prevent nulling out the CRL if we had a value before but got a null back from `createCrl`? @stustison what do you think about that?

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
kcwire: fixed as suggested.

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
stustison: Ah, yeah that is a good point. If we had a valid value previously, it should probably keep it. If they want to really remove the CRL for some reason then we could require a restart.

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -154,4 +132,82 @@ Properties loadProperties(String location) {
...
+            URL url = urlFromPath(location);
+
+            //If we get a URL, use URL, otherwise use as local file path
+            try (InputStream is = url != null ? url.openStream() : new FileInputStream(new File(
+                    location))) {
+
+                CertificateFactory cf = CertificateFactory.getInstance("X.509");
+                return cf.generateCRL(is);
+            } catch (IOException e) {
+                LOGGER.debug("An error occurred while accessing {}", location, e);
```
jlcsmith: It looks like this exception could occur if the CRL location is invalid.  I'm thinking the condition should be logged at a higher level to bring this to the attention of the administrator.

kcwire: Updated to `WARN`.

kcwire:
build now

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -154,4 +132,82 @@ Properties loadProperties(String location) {
...
+
+        public void run() {
+            Properties encryptionProperties = loadProperties(encryptionPropertiesLocation);
+            CRL crl = crlCache.get();
+            if (crl instanceof X509CRL) {
+                X509CRL x509Crl = (X509CRL) crl;
+                if (Calendar.getInstance()
+                        .getTime()
+                        .after(x509Crl.getNextUpdate()) || rolledOverDay()) {
+                    setCrlLocation(encryptionProperties.getProperty(CRL_PROPERTY_KEY));
```
beyelerb: It would be nice to have a log statement indicating that the CRL has been/is being updated. At least at the DEBUG level.

kcwire: Log statement has been added.

kcwire:
build now

kcwire:
build now

kcwire:
build now

kcwire:
build now

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -78,7 +86,8 @@ private boolean passesCrl(X509Certificate[] certs) {
         if (certs != null) {
             LOGGER.debug("Got {} certificate(s) in the incoming request", certs.length);
             for (X509Certificate curCert : certs) {
-                if (crl.isRevoked(curCert)) {
+                if (crlCache.get()
```
rzwiefel: The crlCache reference could have been nulled out since the previous check to `crlCache.get()` on line 70. We should not rely on it staying the same for any period of time after it is checked. 
We should either capture the .get90 on line 70 into a variable and pass it as a parameter to passesCrl or we should re fetch here and check again for null.

rzwiefel:
build now

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
kcwire: updated.

kcwire:
build now

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -42,14 +48,16 @@
...
+    private CrlRefresh refresh = new CrlRefresh();
 
     /**
      * Constructor method. Reads encryption.properties and sets CRL location
      */
     public CrlChecker() {
-        Properties encryptionProperties = loadProperties(encryptionPropertiesLocation);
-        setCrlLocation(encryptionProperties.getProperty(CRL_PROPERTY_KEY));
+        Executors.newScheduledThreadPool(1)
+                .scheduleWithFixedDelay(refresh, 0, 1, TimeUnit.HOURS);
```
ahoffer: ✏️   `scheduleWithFixedDelay` is a static method on `Executors`, so:
`Executors.scheduleWithFixedDelay(refresh, 0, 1, TimeUnit.HOURS);`

--

rzwiefel:
build now

rzwiefel:
Looks like failed for coverage checks @kcwire 
```
project security-handler-pki: Coverage checks have not been met.
[WARNING] Rule violated for bundle DDF :: Security :: Handler :: PKI: complexity covered ratio is 0.73, but expected minimum is 0.74
```

kcwire:
build now

rzwiefel:
build now

rzwiefel:
build now

PR #2129, opened 2017-07-10
DDF-3141 Publish to remote registry when registry url is updated

kcrowe01:
#### What does this PR do?
Changing the remote registry url will publish to the new url.
#### Who is reviewing it? 
@mcalcote @vinamartin @emmberk 
@clockard
@lessarderic 
To Test:
* Build the DDF
* Setup three DDFs, one as a registry (registry), the others as remote DDFs (ddf1, ddf2).
* Configure the first remote DDF's Remote Registry CSW endpoint to point at itself, e.g., https://ddf1:8993/services/csw
* Update the first remote DDF's Remote Registry CSW endpoint to point at the registry, e.g, https://registry:8993/services/csw
* Update the second remote DDF's Remote Registry CSW endpoint to point at the registry, e.g., https://registry:8993/services/csw
* Refresh the Catalog App's Node information tab on all the DDFs.
* Notice ddf1 should show up in the Remote Node sections of both the registry DDF and ddf2.
[DDF-3141](https://codice.atlassian.net/browse/DDF-3141)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStorePublisher.java
+++ b/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStorePublisher.java
@@ -13,7 +13,9 @@
  */
 package org.codice.ddf.registry.api.impl;
 
+import java.util.Collections;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -55,7 +57,7 @@ public class RegistryStorePublisher implements EventHandler {
 
     private static final String UNPUBLISH = "unpublish";
 
-    private ConcurrentHashMap<String, Boolean> registryStoreMap = new ConcurrentHashMap<>();
+    private Set<String> registryStoreSet = Collections.newSetFromMap(new ConcurrentHashMap<>());
 
     private RegistryPublicationService registryPublicationService;
 
@@ -100,8 +102,7 @@ public class RegistryStorePublisher implements EventHandler {
 
         RegistryStore registryStore = bundleContext.getService(reference);
 
-        Boolean previousState = registryStoreMap.put(registryStore.getConfigurationPid(),
-                registryStore.isAutoPush());
+        registryStoreSet.add(registryStore.getConfigurationPid());
 
         SourceMonitor storeRegisteredSourceMonitor = new SourceMonitor() {
             @Override
@@ -117,9 +118,7 @@ public class RegistryStorePublisher implements EventHandler {
         };
 
         if (registryStore.isAvailable(storeRegisteredSourceMonitor) && registryStore.isAutoPush()) {
-            if (previousState == null || !previousState) {
-                registryPublish(registryStore, PUBLISH);
-            }
+            registryPublish(registryStore, PUBLISH);
         }
     }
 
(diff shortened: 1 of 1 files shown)


catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStorePublisher.java
```diff
@@ -210,7 +210,7 @@ public void handleEvent(Event event) {
                 (RegistryStore) bundleContext.getService(((ServiceEvent) event.getProperty(
                         EventConstants.EVENT)).getServiceReference());
 
-        if (!previousAutoPush && registryStore.isAutoPush()) {
+        if (registryStore.isAutoPush()) {
```
kcrowe01: This will now publish to the remote registry on any update if AutoPush is enabled and the remote registry has not already been published to (this is checked in the publish function).

kcrowe01:
build now

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStorePublisher.java
mcalcote: With the change to this conditional statement, you should be able to remove ` && !registryStore.isAutoPush()` from the else statement.

vinamartin:
Manually verified with @emmberk ✅

clockard:
build now

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStorePublisher.java
lessarderic: ✏️  We should also be able to remove the `previousAutoPush` variable and change lines L199-L201 to:
```
if (registryStoreMap.get(pid)) {
    return;
}
```

lessarderic: ❓ So this used to publish and un-publish only when the auto push changed from `true` to `false` or vice-versa. Now it will always publish or un-publish no matter what. I'm curious to know why this actually fixes the problem since I would expect that we only care about state changes.

kcrowe01: Previously, this was only publishing or un-publishing when auto push changed. The problem with this was that when the url was updated without changing auto push, it would not publish to the new url. This now attempts to publish on any change _if_ the url has not already been published to (the publish function checks published locations before attempting a new publish). It looks like the un-publish function also performs this check so I will take out `previousAutoPush` as it is not needed.

kcrowe01: This will attempt to publish on any update when Auto-Push is enabled, and try to un-publish on any update if Auto-Push is disabled. Not sure if that is the behavior we want though.

mcalcote: That sounds right like the functionality we expect 👍

lessarderic: ❓ Since it doesn't look like the `Boolean` stored in the `registryStoreMap` was only used for this, could we simplify the code a little by replacing the `Map` with a `Set` that contains the list of pids we know about?

vinamartin:
build now

taz77:
build now

PR #2133, opened 2017-07-12

lessarderic:
#### What does this PR do?
Provides two generic `Iterable` classes that can be used to iterate over all the results returned by the catalog framework regardless of any page size limitations enforced. The iterations can be done one `Result` at a time or in batches.

#### Who is reviewing it? 

@mweser 
@rzwiefel 
@brendan-hofmann 
@ahoffer 

@coyotesqrl
@pklinef

These classes are not currently used so a simple passing build is sufficient.

These utilities are needed to address many potential paging problems that currently exist in the application.

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-api-impl/pom.xml
+++ b/catalog/core/catalog-core-api-impl/pom.xml
@@ -30,17 +30,6 @@
             <artifactId>catalog-core-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.geotools</groupId>
-            <artifactId>gt-main</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
-            <version>${logback.classic.version}</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>commons-collections</groupId>
             <artifactId>commons-collections</artifactId>
         </dependency>
@@ -55,6 +44,23 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-main</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>${logback.classic.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ddf.lib</groupId>
+            <artifactId>spock-shaded</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
     <build>
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryFunction.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryFunction.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p/>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p/>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.util.impl;
+
+import ddf.catalog.federation.FederationException;
+import ddf.catalog.operation.QueryRequest;
+import ddf.catalog.operation.SourceResponse;
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
+
+/**
+ * Functional interface used to abstract away methods that take in a {@link QueryRequest} and
+ * return a {@link SourceResponse}. Used by classes such as {@link ResultIterable} to support
+ * interfaces other than {@link ddf.catalog.CatalogFramework}.
+ */
+@FunctionalInterface
+public interface QueryFunction {
+    /**
+     * Runs a query.
+     *
+     * @param queryRequest request to use
+     * @return source response
+     * @throws SourceUnavailableException if a required {@link ddf.catalog.source.Source} is unavailable
+     * @throws UnsupportedQueryException  if the {@link ddf.catalog.operation.Query} can not be
+     *                                    evaluated by this {@link ddf.catalog.CatalogFramework} or
+     *                                    any of its {@link ddf.catalog.source.Source}s.
+     * @throws FederationException        if the {@link QueryRequest} includes
+     *                                    {@link ddf.catalog.source.FederatedSource}s and there is
+     *                                    either a problem connecting to a
+     *                                    {@link ddf.catalog.source.FederatedSource} or a
+     *                                    {@link ddf.catalog.source.FederatedSource}
+     *                                    cannot evaluate the {@link ddf.catalog.operation.Query}
+     */
+    SourceResponse query(QueryRequest queryRequest)
+            throws UnsupportedQueryException, SourceUnavailableException, FederationException;
+}
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
@@ -1,0 +1,184 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p/>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p/>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.util.impl;
+
+import static org.apache.commons.lang.Validate.notNull;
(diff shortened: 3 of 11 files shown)


catalog/core/catalog-core-api-impl/pom.xml
```diff
@@ -56,6 +45,23 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-main</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
```
brendan-hofmann: ❓ do we really still need logback?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
```diff
@@ -0,0 +1,124 @@
...
+     *                         results.
+     * @param queryRequest     request used to retrieve the results.
+     */
+    public ResultListIterable(CatalogFramework catalogFramework, QueryRequest queryRequest) {
+        notNull(catalogFramework, "Catalog framework reference cannot be null");
+        notNull(queryRequest, "Query request cannot be null");
+        isTrue(queryRequest.getQuery()
+                .getPageSize() > 1, "Page size must be greater than 1");
+
+        this.queryFunction = catalogFramework::query;
```
brendan-hofmann: This is slick. 👍

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -0,0 +1,372 @@
...
+        then:
+        1.upto(nextCalls, { results.get(it - 1).distanceInMeters == (Double) it })
+
+        where:
+        nextCalls | pageSize | resultSize | expectedQueries
+        1         | 1        | 1          | 1
+        2         | 1        | 2          | 1
+        3         | 1        | 1          | 3
+        2         | 2        | 1          | 2
+        2         | 0        | 1          | 2
```
rzwiefel: beautiful

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -0,0 +1,372 @@
...
+        then:
+        1.upto(resultSize + 1, { results.get(it - 1).distanceInMeters == (Double) it })
+    }
+
+    def "next() when number of results from catalog varies"() {
+        setup:
+        def totalResults = 6
+        3 * catalogFramework.query(_ as QueryRequest) >> {
+            QueryRequest queryRequest -> buildQueryResponse(queryRequest, 2, totalResults)
+        } >> {
```
rzwiefel: This seems really quite strange... but i don't think we really have our formatter for groovy enforced (or even setup) so It probably doesn't matter for now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
ahoffer: It is slick. I had to stare at that for a while.
What is the advantage? With the function reference, invoking the query looks like:

` SourceResponse response = queryFunction.query(queryRequestCopy);`

without the function reference, the invocation looks like:

` SourceResponse response = catalogFramework.query(queryRequestCopy);`

The class imports CatalogFramework in either case. Is there some advantage I'm missing?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryFunction.java
```diff
@@ -0,0 +1,31 @@
...
+import ddf.catalog.source.UnsupportedQueryException;
+
+/**
+ * Functional interface used to abstract away methods that take in a {@link QueryRequest} and
+ * return a {@link QueryResponse}. Used by classes such as {@link ResultIterable} to support
+ * interfaces other than {@link ddf.catalog.CatalogFramework}.
+ */
+@FunctionalInterface
+public interface QueryFunction {
+    QueryResponse query(QueryRequest queryRequest)
```
coyotesqrl: ✏️ I know in our earlier pre-review of this code I had suggested this interface should be updated to return `QueryResponse` instead of the more generic `SourceResponse`; however, if someone were to want to use the new iterables with a query against `CatalogFacade`, this would no longer work. I think it should go back to `SourceResponse`.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
```diff
@@ -0,0 +1,124 @@
...
+     *
+     * @param catalogFramework reference to the {@link CatalogFramework} to call to retrieve the
+     *                         results.
+     * @param queryRequest     request used to retrieve the results.
+     */
+    public ResultListIterable(CatalogFramework catalogFramework, QueryRequest queryRequest) {
+        notNull(catalogFramework, "Catalog framework reference cannot be null");
+        notNull(queryRequest, "Query request cannot be null");
+        isTrue(queryRequest.getQuery()
+                .getPageSize() > 1, "Page size must be greater than 1");
```
coyotesqrl: ❓ Is there maybe some value in logging a debug message with the page size of the query? Or conditionally logging it if it exceeds some pre-determined value? Maybe > 1024 or 2048 or something like that?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -0,0 +1,372 @@
...
+
+        when:
+        def hasNext = resultIterator.hasNext()
+
+        then:
+        hasNext.is true
+    }
+
+    @Unroll
+    def '''test that #nextCalls results are returned and catalog is queried #expectedQueries times
```
coyotesqrl: Ryan loved the data table (I do too), but I'm going to be honest and say I might love this long, fully descriptive test name even more. 👍

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
brendan-hofmann: The advantage is that `queryFunction` is a functional interface. While the default behavior just queries `CatalogFramework`, clients can provide any `queryFunction` of their own to support custom behaviors while still benefiting from the pagination logic.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -0,0 +1,165 @@
...
+            fetchNextResults();
+
+            if (!results.hasNext()) {
+                throw new NoSuchElementException("No more results match the specified query");
+            }
+
+            return results.next();
+        }
+
+        private void fetchNextResults() {
```
rzwiefel: ❗️ This looks like its going to alway send out a query -- purely to check if there's any more results. If the response already came back with 0 results (or our current index has reached the resultCount) then we should **know** there are non left to fetch and short circuit.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
ahoffer: I saw Ryan's comment about the unit testing. Makes that easier.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
ahoffer: If it is a federated query, does the consumer have to use the method `CatalogFramework. query(QueryRequest fedQueryRequest)`? That does not match the signature of the functional interface because of the second parameter.

coyotesqrl: That's a valid point. We could add a terminal boolean.

1. `boolean terminated = false;`
2. update `hasNext()` to short-circuit if `terminated == true`
3. update `next()` to short-circuit if `terminated == true`
4. update `fetchNextResults` to set `terminated == response.getResults().size() == 0;`

catalog/core/catalog-core-api-impl/pom.xml
lessarderic: Saves us the following three output logs when running the tests:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
lessarderic: I added that because I realized some catalog commands were using their own `CatalogFacade` class that didn't implement `CatalogFramework` but had a `query()` method and couldn't use these `Iterable`s. With the `QueryFunction`, they now can :-)

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/QueryFunction.java
lessarderic: Done.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
lessarderic: Glad it's understandable and makes you happy :-)

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultListIterable.java
lessarderic: Added debug log that contains `QueryRequest` and `Query`.

lessarderic:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
lessarderic: Changed code to assign the `results` iterator to `RESULTS_EXHAUSTED` instead of introducing an extra `boolean` flag. Let me know what you all think.

Also added unit tests for this.

lessarderic:
build now

lessarderic:
build now

lessarderic:
build now

PR #2130, opened 2017-07-10
DDF 3128 Add Decanter and alert/solr appenders base for future expansion

clockard:
#### What does this PR do?
Add Decanter and alert/solr appenders base for future expansion. Also includes
- Update PersistenceStore to support batch add
- Add AND/OR capability to persistent store cql queries

#### Who is reviewing it? 
(please choose AT LEAST two reviewers that need to approve the PR before it can get merged)
@emmberk @vinamartin 
#### Select relevant component teams: 
https://github.com/orgs/codice/teams
#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)
@coyotesqrl 
#### How should this be tested? (List steps with links to updated documentation)
As far as the alerting goes there is nothing testable right now. The Decanter event collector can be started along with the solr appender to verify that those aspects work. 
#### What are the relevant tickets?
[DDF-3128](https://codice.atlassian.net/browse/DDF-3128)
#### Screenshots (if appropriate)

screen shot 2017-07-11 at 11 52 27 am

#### Checklist:
- [X] Documentation Updated
- [X] Update / Add Unit Tests

#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2141, opened 2017-07-13
DDF 3143 Update alerting in the admin UI

clockard:
#### What does this PR do?
Updates the admin ui alerting mechanism to use the new alerting methodology in https://github.com/codice/ddf/pull/2130 
This cannot be merged or tested before  DDF-3128 is merged.
The UI portion is primarily pulled from the DDF 2.10.x branch
#### Who is reviewing it? 
@emmberk @mcalcote @vinamartin 

@figliold 
@stustison 
Verify you can add, view and dismiss events. An example is
Events can be added from the DDF console `event:send decanter/alert/test title='System reboot at 5pm EST' source=admin-notice priority=2 details="System restart tonight to upgrade hardware"`
[DDF-3143](https://codice.atlassian.net/browse/DDF-3143)
screen shot 2017-07-13 at 10 16 23 am

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests
#### Review Comment Legend:


platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
```diff
@@ -0,0 +1,144 @@
...
+    @Test
+    public void testDismissAlertNoSubject() throws Exception {
+        when(subject.getPrincipals()).thenReturn(null);
+        ThreadContext.bind(subject);
+        adminAlert.dismissAlert("myId");
+        verify(eventAdmin, never()).postEvent(any());
+    }
+
+    @Test
+    public void testDismissAlertNoId() throws Exception {
```
vinamartin: ❓ Is it worth making a test for the no details case?

platform/admin/ui/src/main/webapp/js/views/Alerts.view.js
```diff
@@ -16,36 +16,71 @@
...
                 var model = this.model;
-                model.get('button') === 'Show' ? model.set('button', 'Hide') : model.set('button', 'Show');
+                model.get('details-button-action') === 'Show' ? model.set('details-button-action', 'Hide') : model.set('details-button-action', 'Show');
                 model.get('collapse') === 'out' ? model.set('collapse', 'in') : model.set('collapse', 'out');
             },
             serializeData: function () {
                 var json = this.model.toJSON();
                 json.collapseId = 'alertCollapse_' + parseInt((Math.random() * Math.pow(2, 32)), 10);
+                switch (this.model.get('priority')) {
+                    case 2:
```
vinamartin: ❓  There's one priority not here, right? If someone uses that, what will happen?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AbstractMBean.java
```diff
@@ -0,0 +1,78 @@
...
+        mbeanServer = ManagementFactory.getPlatformMBeanServer();
+
+        try {
+            objectName = new ObjectName(objectNameString);
+        } catch (MalformedObjectNameException e) {
+            LOGGER.debug("Exception while creating object name: " + objectNameString, e);
+        }
+
+        try {
+            try {
```
vinamartin: ❓  What's up with this nested try?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+        } catch (PersistenceException pe) {
+            LOGGER.error("Error retrieving system alert.", pe);
+        }
+        return alerts;
+    }
+
+    @Override
+    public void dismissAlert(String id) {
+        if (StringUtils.isEmpty(id)) {
```
vinamartin: ✏️  We log when the subject name is null, but should we log anything if the id is empty?

platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    AdminAlertImpl adminAlert;
+
+    Map<String, Object> solrMap;
+
+    @Before
+    public void setup() throws Exception {
+        adminAlert = new AdminAlertImpl(persistentStore, eventAdmin);
+        solrMap = new HashMap<>();
+        solrMap.put("id_txt", "myId");
```
vinamartin: ✏️  I find it a little weird that below you have `SystemNotice.SYSYTEM_NOTICE_DETAILS_KEY + "_txt"` but here you have it inlined.

vinamartin:
✏️  ❓  Just a thought, if it's not too much more effort, should we consider sorting the results when we get the alerts so that we can show the highest priority results toward the top of the list? 
The reason I bring this up is because when testing this, I would notice a custom alert that I created would start at the bottom of the list, and when another event occurred that triggered my alert again it would move toward the top (since it was more newly updated). 
Sorting the results would create consistency every time the admin console is loaded.

clockard:
@vinamartin I agree that we should sort them at least by their priority.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * BasicMBean implements the standard init and destroy methods all of our mbeans use. Other mbean
+ * implementations can extend this BasicMBean and only need to implement their interface without the
+ * need to worry about the bean registration.
+ */
+public class BasicMBean extends StandardMBean {
```
mcalcote: ✏️ I wish I had a suggestion for a better name... but I don't

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
```diff
@@ -0,0 +1,83 @@
...
+    public void init() {
+        mbeanServer = ManagementFactory.getPlatformMBeanServer();
+
+        try {
+            objectName = new ObjectName(objectNameString);
+        } catch (MalformedObjectNameException e) {
+            LOGGER.debug("Exception while creating object name: " + objectNameString, e);
+        }
+
+        try {
```
mcalcote: ❓ Do you think it be any cleaner to have `try { unregister } catch try { register } catch` ?

vinamartin:
Manually tested this by creating alerts from the command line and also in DDF-3169 where I've implemented an alert appender 👍

platform/admin/ui/src/main/webapp/main.js
```diff
@@ -159,33 +159,35 @@
...
-            }
-        });
+        // setup alert banners
+        app.addInitializer(function () {
+            var alerts = new Backbone.Collection([]);
 
-        //refresh the session if keypress  or mouse click events occur
-        SessionRefresherUtil(60000);
+            var backendAlerts = new AlertsModel.BackendAlerts();
+            backendAlerts.fetch();
```
mcalcote: ❓ Is there any chance this backendAlerts collection would be substantial in size or have any kind of delay while being fetched?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
clockard: It's not great but will work till we can think of something better.

platform/admin/ui/src/main/webapp/main.js
clockard: It could be large but shouldn't be in real life cases. This is an async call so it shouldn't keep the ui from loading.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
clockard: Probably so. Will change.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+                    //strings/sets of strings. Single value sets will come back as strings.
+                    if (item.get(SystemNotice.SYSTEM_NOTICE_DETAILS_KEY) instanceof String) {
+                        item.put(SystemNotice.SYSTEM_NOTICE_DETAILS_KEY,
+                                Collections.singleton(item.get(SystemNotice.SYSTEM_NOTICE_DETAILS_KEY)));
+                    }
+                    alerts.add(item);
+                }
+            }
+        } catch (PersistenceException pe) {
+            LOGGER.error("Error retrieving system alert.", pe);
```
stustison: This needs to tell the admin what they can possibly do about this situation and it needs to explain more than simply "error" or put it at < info

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        return alerts;
+    }
+
+    @Override
+    public void dismissAlert(String id) {
+        if (StringUtils.isEmpty(id)) {
+            LOGGER.warn("Cannot dismiss alert without an id");
```
stustison: Again, need to explain to an admin what they can do or move it to < info.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+            return;
+        }
+
+        String subjectName = SubjectUtils.getName(SecurityUtils.getSubject());
+        if (subjectName == null) {
+            LOGGER.warn("No Subject Name! Could not dismiss alert {}", id);
```
stustison: Same as above.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+                Alert.ALERT_ACTIVE_STATUS,
+                InetAddress.getLocalHost()
+                        .getHostName());
+    }
+
+    @Override
+    public List<Map<String, Object>> getAlerts() {
+        List<Map<String, Object>> alerts = new ArrayList<>();
+        try {
+            List<Map<String, Object>> results = persistentStore.get("alerts", activeAlertQuery);
```
stustison: If any of these results contain data that was obtained from a user on the system, it would be possible to execute a XSS attack against an administrator user. We need to escape or encode these results to avoid executing any scripts that they might contain since we are pushing them up to a handlebars template that uses the {{{}}} braces and doesn't automatically escape HTML for us.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private EventAdmin eventAdmin;
+
+    public AdminAlertImpl(PersistentStore persistentStore, EventAdmin eventAdmin)
+            throws NotCompliantMBeanException, UnknownHostException {
+        super(AdminAlertMBean.class, AdminAlertMBean.OBJECT_NAME);
+        this.persistentStore = persistentStore;
+        this.eventAdmin = eventAdmin;
+        activeAlertQuery = String.format("status='%s' AND hostName='%s'",
+                Alert.ALERT_ACTIVE_STATUS,
```
figliold: ❓ It seems like using the local host might be a problem when running in a clustered environment.  Wouldn't we want to use the cluster address?  We also might need a way to identify which node of a cluster the alert belongs to.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+import ddf.security.SubjectUtils;
+
+public class AdminAlertImpl extends BasicMBean implements AdminAlertMBean {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AdminAlertImpl.class);
+
+    private final String activeAlertQuery;
+
+    private PersistentStore persistentStore;
```
figliold: ❓ final?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private EventAdmin eventAdmin;
```
figliold: ❓ final?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(BasicMBean.class);
+
+    private MBeanServer mbeanServer;
+
+    private ObjectName objectName;
+
+    private String objectNameString;
+
+    private Class classInterface;
```
figliold: ❓ final?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class BasicMBean extends StandardMBean {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(BasicMBean.class);
+
+    private MBeanServer mbeanServer;
+
+    private ObjectName objectName;
+
+    private String objectNameString;
```
figliold: ❓ final?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
```diff
@@ -0,0 +1,83 @@
...
+        this.classInterface = classInterface;
+        this.objectNameString = objectNameString;
+    }
+
+    public void init() {
+        mbeanServer = ManagementFactory.getPlatformMBeanServer();
+
+        try {
+            objectName = new ObjectName(objectNameString);
+        } catch (MalformedObjectNameException e) {
```
figliold: ❓ Should this really be at debug? If it fails, can we proceed?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        try {
+            try {
+                mbeanServer.registerMBean(new StandardMBean(this, classInterface), objectName);
+            } catch (InstanceAlreadyExistsException e) {
+                mbeanServer.unregisterMBean(objectName);
+                mbeanServer.registerMBean(new StandardMBean(this, classInterface), objectName);
+            }
+        } catch (Exception e) {
```
figliold: ❓ Should this only be at info? If we get an exception, can we proceed?

platform/admin/core/admin-core-insecuredefaults/src/main/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsCollector.java
```diff
@@ -83,31 +80,17 @@
...
             "javax.net.ssl.trustStorePassword";
 
-    public static final String MBEAN_NAME =
-            InsecureDefaultsServiceBean.class.getName() + ":service=insecure-defaults-service";
-
-    private ObjectName objectName;
+    private List<Validator> validators = new ArrayList<>();
 
-    private MBeanServer mBeanServer;
+    private EventAdmin eventAdmin;
```
figliold: ❓ final?

platform/admin/core/admin-core-insecuredefaults/src/main/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsCollector.java
```diff
@@ -83,31 +80,17 @@
     private static final String TRUSTSTORE_PASSWORD_SYSTEM_PROPERTY =
             "javax.net.ssl.trustStorePassword";
 
-    public static final String MBEAN_NAME =
-            InsecureDefaultsServiceBean.class.getName() + ":service=insecure-defaults-service";
-
-    private ObjectName objectName;
+    private List<Validator> validators = new ArrayList<>();
```
figliold: ❓ final?

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
clockard: Each instance should have it's own hostname which is stored in the alert. (This is also how all the decanter collectors work) If you used a cluster address it wouldn't match any of the alerts and if you used the cluster address in the alert you wouldn't be able to distinguish which instance the alert came from.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
clockard: The system can proceed but there wouldn't be any alertes shown in the admin ui. I would prefer to log a error but since I don't know that there is anything the admin could do about it the guidance has been to log it at debug.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/BasicMBean.java
clockard: See comment above. These log levels are consistent with existing mbeans.

platform/admin/core/admin-core-insecuredefaults/src/main/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsCollector.java
clockard: Added final

platform/admin/core/admin-core-insecuredefaults/src/main/java/org/codice/ddf/admin/insecure/defaults/service/InsecureDefaultsCollector.java
clockard: Added final

clockard:
build now

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -0,0 +1,120 @@
...
+                    alerts.add(item);
+                }
+            }
+        } catch (PersistenceException pe) {
+            LOGGER.debug("Error retrieving system alert.", pe);
+        }
+        return alerts;
+    }
+
+    private void escapeStrings(Map<String, Object> map) {
```
SmithJosh: ✏️ I'm not sure if there's a good reason we use the {{{ }}} braces in the alerts template, but if not, it would be nice to escape the alerts there so input sanitization isn't an issue in the future.

clockard: Changed the {{{ to just {{ in the alert template

clockard:
build now

PR #2149, opened 2017-07-21
DDF-3167 Changes the timing for the IdP server & client initialization of metadata.

coyotesqrl:
#### What does this PR do?
Previously, these operations occurred on injection of the metadata locators. That led to an intermittent timing issue, as the two bundles cross-query HTTP endpoints. Now, they execute the queries and parse the data once, on initial access. This allows both services to be fully initialized before any queries are executed.

This change also includes some minor cleanup to the `MetadataConfigurationParser` class. That cleanup had been associated with an earlier approach to resolving the timing issue and is not necessary; however, it does make for smaller, more concise methods.

#### Who is reviewing it? 
@rzwiefel @tbatie @oconnormi 

@pklinef
@stustison

As this resolves an intermittent problem, any definitive test is going to be difficult to manage. Some environments have proven to be more likely to demonstrate the issue than others based on speed/performance of equipment. The best way to test this is to repeatedly unzip, startup, and then attempt to login to the admin console. On my laptop, it would fail within fewer than six iterations. On other machines, it might take more successful login attempts to increase testing confidence.

N/A

[DDF-3167](https://codice.atlassian.net/browse/DDF-3167)

N/A

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSingleSignOn.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSingleSignOn.java
@@ -58,6 +58,7 @@ import org.codice.ddf.itests.common.AbstractIntegrationTest;
 import org.codice.ddf.itests.common.annotations.BeforeExam;
 import org.codice.ddf.itests.common.utils.LoggingUtils;
 import org.codice.ddf.security.common.jaxrs.RestSecurity;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.ops4j.pax.exam.junit.PaxExam;
@@ -119,6 +120,25 @@ public class TestSingleSignOn extends AbstractIntegrationTest {
         METADATA, PROTOCOL
     }
 
+    @Before
+    public void setup() throws Exception {
+        // Get all of the metadata
+        String metadata = get(SERVICE_ROOT + "/idp/login/metadata").asString();
+        String ddfSpMetadata = get(SERVICE_ROOT + "/saml/sso/metadata").asString();
+
+        // Make sure all the metadata is valid before we set it
+        validateSaml(metadata, SamlSchema.METADATA);
+        validateSaml(ddfSpMetadata, SamlSchema.METADATA);
+
+
+        // The IdP server can point to multiple Service Providers and as such expects an array.
+        // Thus, even though we are only setting a single item, we must wrap it in an array.
+        setConfig("org.codice.ddf.security.idp.client.IdpMetadata", "metadata", metadata);
+        setConfig("org.codice.ddf.security.idp.server.IdpEndpoint",
+                "spMetadata",
+                new String[] {ddfSpMetadata});
+    }
+
     @BeforeExam
     public void beforeTest() throws Exception {
         try {
@@ -127,21 +147,6 @@ public class TestSingleSignOn extends AbstractIntegrationTest {
             // We need to start the Search UI to test that it redirects properly
             getServiceManager().startFeature(true, "security-idp");
             getServiceManager().waitForAllBundles();
-
-            // Get all of the metadata
-            String metadata = get(SERVICE_ROOT + "/idp/login/metadata").asString();
-            String ddfSpMetadata = get(SERVICE_ROOT + "/saml/sso/metadata").asString();
-
-            // Make sure all the metadata is valid before we set it
-            validateSaml(metadata, SamlSchema.METADATA);
-            validateSaml(ddfSpMetadata, SamlSchema.METADATA);
-
-            // The IdP server can point to multiple Service Providers and as such expects an array.
-            // Thus, even though we are only setting a single item, we must wrap it in an array.
-            setConfig("org.codice.ddf.security.idp.client.IdpMetadata", "metadata", metadata);
-            setConfig("org.codice.ddf.security.idp.server.IdpEndpoint",
-                    "spMetadata",
-                    new String[] {ddfSpMetadata});
 
             ingest(getFileContent(JSON_RECORD_RESOURCE_PATH + "/SimpleGeoJsonRecord"),
                     "application/json");
--- a/platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpMetadata.java
+++ b/platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpMetadata.java
@@ -21,21 +21,21 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-
-import javax.xml.stream.XMLStreamException;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.commons.lang.StringUtils;
-import org.apache.wss4j.common.ext.WSSecurityException;
-import org.opensaml.saml.common.SAMLException;
 import org.opensaml.saml.saml2.metadata.Endpoint;
 import org.opensaml.saml.saml2.metadata.EntityDescriptor;
 import org.opensaml.saml.saml2.metadata.IDPSSODescriptor;
 import org.opensaml.saml.saml2.metadata.KeyDescriptor;
 import org.opensaml.security.credential.UsageType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import ddf.security.samlp.MetadataConfigurationParser;
 
 public class IdpMetadata {
+    private static final Logger LOGGER = LoggerFactory.getLogger(IdpMetadata.class);
 
     private static final String SAML_2_0_PROTOCOL = "urn:oasis:names:tc:SAML:2.0:protocol";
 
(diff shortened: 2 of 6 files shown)


coyotesqrl:
And yes. I'm aware that the exponential & IOException back-off in `MetadataConfigurationParser` _should_ have successfully managed this under the old implementation. My best guess right now is that there is a bug either in the Google client library (possible but not likely) or in our implementation of the back-off (more likely, but nowhere I can see).

Even if there is a bug there and it is subsequently found/resolved, I believe the initialization on first access is still better and safer than initialization on injection.

Lambeaux:
I was repeatedly running into this issue last week. I'll give it a try on Monday.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
```diff
@@ -207,14 +211,15 @@ public IdpEndpoint(String signaturePropertiesPath, String encryptionPropertiesPa
...
                         "/templates/submitForm.handlebars");
                 InputStream redirectPageStream = IdpEndpoint.class.getResourceAsStream(
                         "/templates/redirect.handlebars");
                 InputStream soapMessageStream = IdpEndpoint.class.getResourceAsStream(
-                        "/templates/soap.handlebars");
+                        "/templates/soap.handlebars")
+                //
         ) {
             indexHtml = IOUtils.toString(indexStream);
             submitForm = IOUtils.toString(submitFormStream);
```
stustison: We should be specifying the character encoding here for these toStrings.

coyotesqrl: Got all of them in this class.

coyotesqrl:
build now

coyotesqrl:
build now

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSingleSignOn.java
```diff
@@ -119,6 +120,25 @@ public String toString() {
...
+        String ddfSpMetadata = get(SERVICE_ROOT + "/saml/sso/metadata").asString();
+
+        // Make sure all the metadata is valid before we set it
+        validateSaml(metadata, SamlSchema.METADATA);
+        validateSaml(ddfSpMetadata, SamlSchema.METADATA);
+
+
+        // The IdP server can point to multiple Service Providers and as such expects an array.
+        // Thus, even though we are only setting a single item, we must wrap it in an array.
+        setConfig("org.codice.ddf.security.idp.client.IdpMetadata", "metadata", metadata);
```
coyotesqrl: Needed to move these configuration changes to the setup method instead of running once; otherwise, the maps of metadata in IdP server and client could be blank based on the order of tests.

coyotesqrl:
build now

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSingleSignOn.java
coyotesqrl: This change, and the itests that were broken prior to making it, expose a secondary bug that was lurking in the old implementation. Nothing was clearing out old metadata from the internal maps in `IdpEndpoint` and `IdpMetadata` on re-injection of their associated lookup properties. That's resolved as well now.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpMetadata.java
```diff
@@ -53,18 +53,18 @@
...
 
-    public void setMetadata(String metadata)
-            throws WSSecurityException, XMLStreamException, SAMLException, IOException {
-        MetadataConfigurationParser metadataConfigurationParser = new MetadataConfigurationParser(
-                Collections.singletonList(metadata),
-                ed -> entryDescriptions.put(ed.getEntityID(), ed));
-        entryDescriptions.putAll(metadataConfigurationParser.getEntryDescriptions());
+    public void setMetadata(String metadata) {
+        this.metadata = metadata;
+        entryDescriptions.set(null);
```
rzwiefel: :question: Why instantiate this to null? By default `new AtomicReference<>();` is going to have a null reference already.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
```diff
@@ -228,37 +233,61 @@ public void init() {
...
+        if (spMap == null) {
+            spMap = parseServiceProviderMetadata();
+
+            boolean updated = serviceProviders.compareAndSet(null, spMap);
+            if (!updated) {
+                spMap = serviceProviders.get();
             }
         }
+
+        return Collections.unmodifiableMap(spMap);
```
rzwiefel: ❗️ This could throw a NPE under the following conditions:
line 250: serviceProviders is not null, so it is not set and updated is returned as false
Line 251: The If evaluates to true to execute the branch
Someone sets the Metadata and the setSpMetadata method is called and the ServiceProviders object is set to null (Line 1314)
Line 252: The serviceProviders.get() method returns the newly set null object 
Line 256: Collections.UnmodifiableMap(map) does not tolerate null arguments and throws a NPE

coyotesqrl: Your analysis is definitely correct, but I'm not sure it matters in the real world. For this condition to occur, a thread would have to be updating the property concurrent with the first access. It isn't impossible, but the odds of it happening are vanishingly small. Setting the injected property is an action that is driven by administration of the system. Someone would need to be _resetting it_ while multiple users are logging in. Not initial set, but reset.

I can certainly catch the unlikely NPE and throw a different exception, but I think anything else is either overkill or is going to drag the system performance down.

rzwiefel: If it is improbable to happen,  then you should simply return the _improbably_ out-of-date information instead of trying to re-fetch, then and risk throwing an _improbable_ NPE.

rzwiefel:
``` java
    private Map<String, EntityInformation> getServiceProvidersMap() {
         Map<String, EntityInformation> spMap = serviceProviders.get();
         if (spMap == null) {
             spMap = parseServiceProviderMetadata();
             serviceProviders.compareAndSet(null, spMap);
          }
 
         return Collections.unmodifiableMap(spMap);
```

coyotesqrl:
Small changes made to the `IdpMetadata.getEntityDescriptor()` and `IdpEndpoint.getServiceProvidersMap()` methods to deal with vanishingly small possibility of concurrent lookup and clearing of internal maps.

coyotesqrl:
build now

coyotesqrl:
build now

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpMetadata.java
```diff
@@ -162,7 +162,25 @@ private void initCertificates() {
...
+                LOGGER.debug("Error parsing SSO metadata", e);
+                return null;
+            }
+
+            boolean updated = entryDescriptions.compareAndSet(null, edMap);
+            if (!updated) {
+                LOGGER.debug("Safe but concurrent update to serviceProviders map; using processed value");
+            }
+        }
+        if (edMap == null) {
```
coyotesqrl: Nice catch, Sonar! 👍

platform/security/idp/security-idp-server/pom.xml
```diff
@@ -22,7 +22,7 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>security-idp-server</artifactId>
-    <name>DDF :: Security :: IDP :: Server</name>
+    <name>DDF :: Security :: IdP :: Server</name>
```
coyotesqrl: Fixed for @millerw8 ;)

rzwiefel: giphy

coyotesqrl:
build now

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
```diff
@@ -207,19 +211,20 @@ public IdpEndpoint(String signaturePropertiesPath, String encryptionPropertiesPa
     }
 
     public void init() {
-        try (
+        try (//
```
rzwiefel: They are there to appease the formatter. Without them it globs this into a nasty unreadable chunk

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpMetadata.java
coyotesqrl: Could certainly change it to a `getAndSet` to get the volatile read semantics as well.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
coyotesqrl: Correct. Formatter hints.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
```diff
@@ -228,37 +233,61 @@ public void init() {
...
+                    });
+
+            spMap.putAll(metadataConfigurationParser.getEntryDescriptions()
+                    .entrySet()
+                    .stream()
+                    .map(e -> Maps.immutableEntry(e.getKey(),
+                            new EntityInformation.Builder(e.getValue(),
+                                    SUPPORTED_BINDINGS).build()))
+                    .filter(e -> nonNull(e.getValue()))
+                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
```
coyotesqrl: I don't see how that's possible. They keys making it to the `Collector` are the same keys that were in the `metadataConfigurationParser.getEntryDescriptions` map. They're distinct on the start of the stream.

coyotesqrl:
build now

coyotesqrl:
This still keeps the lazy load on access for the client and server metadata but uses the safe `Futures.addCallback` method to run in the delegate executor's thread pool.

PR #2148, opened 2017-07-19
DDF-3161 Results displayed by Intrigue can be incomplete because the Catalog UI endpoint doesn't handle paging

josephthweatt:
#### What does this PR do?
Uses ResultIterable to retrieve Metacards from the catalog. After DDF-2782, there was a possibility of a user only receiving a subset of the actual results.

#### Who is reviewing it? 
@mweser 
@emmberk 

#### Select relevant component teams: 
https://github.com/orgs/codice/teams
#### Choose 2 committers to review/merge the PR. 
@clockard 
@coyotesqrl 

#### How should this be tested? (List steps with links to updated documentation)
1. Build DDF and unzip
2. Before running DDF, edit the **setenv** inside **ddf-2.11.0-SNAPSHOT/bin** so that the last line reads: `export JAVA_OPTS=-"server -Dcatalog.maxPageSize=3 .......`
3. Run DDF
4. In DDF's terminal, install the Catalog UI using `feature:install catalog-ui`
5. Ingest a metacard through Catalog UI
6. Using the Catalog UI, select the metacard and expand it. In the details tab, make four changes to the title. Save after every change
7. Select the History tab, verify that there are four changes to the metacard

#### What are the relevant tickets?
[DDF-3161](https://codice.atlassian.net/browse/)
[DDF-2782](https://codice.atlassian.net/browse/)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests
#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2150, opened 2017-07-21
DDF-3165 catalog:remove command now uses ResultIterable to batch query responses for metacard deletion

jhunzik:
#### What does this PR do?
Refactors RemoveCommand.java to utilize ResultIterable from the generic utility found in DDF-3034
#### Who is reviewing it? 
@josephthweatt 
@Joy603 
@johnsliao
@brendan-hofmann 
@mweser
@emanns95 
@coyotesqrl 
@clockard
@shaundmorris
1. Build DDF
2. Ingest metacards into the DDF (more than the default batch size of 250). 
3. Attempt to delete all metacards using `catalog:remove --cql "title like '*'"
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
@@ -13,23 +13,28 @@
  */
 package org.codice.ddf.commands.catalog;
 
+import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
 import ddf.catalog.operation.DeleteResponse;
-import ddf.catalog.operation.SourceResponse;
+import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.impl.DeleteRequestImpl;
 import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.util.impl.ResultIterable;
 import java.io.Serializable;
+import java.text.ParseException;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Objects;
+import java.util.Set;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.karaf.shell.api.action.Argument;
 import org.apache.karaf.shell.api.action.Command;
 import org.apache.karaf.shell.api.action.Option;
 import org.apache.karaf.shell.api.action.lifecycle.Service;
-import org.codice.ddf.commands.catalog.facade.CatalogFacade;
+import org.geotools.filter.text.cql2.CQLException;
 import org.slf4j.LoggerFactory;
 
 /** Deletes records by ID. */
@@ -45,14 +50,16 @@ public class RemoveCommand extends CqlCommands {
 
   private static final String IDS_LIST_ARGUMENT_NAME = "IDs";
 
+  private static final int BATCH_SIZE = 250;
+
   @Argument(
     name = IDS_LIST_ARGUMENT_NAME,
-    description = "The id(s) of the document(s) by which to filter.",
+    description = "The ID(s) of documents to remove",
     index = 0,
     multiValued = true,
     required = false
   )
-  List<String> ids = null;
+  Set<String> ids = new HashSet<>();
 
   @Option(
     name = "--cache",
@@ -66,9 +73,8 @@ public class RemoveCommand extends CqlCommands {
   protected Object executeWithSubject() throws Exception {
     if (CollectionUtils.isEmpty(ids) && !hasFilter()) {
       printErrorMessage(
-          "Nothing to remove. Either the "
-              + IDS_LIST_ARGUMENT_NAME
-              + " argument or another filter must be specified. Please, use the catalog:removeall command if the goal is to delete all records from the Catalog.");
+          "No IDs or filter provided"
+              + "\nPlease use the catalog:removeall command if the goal is to remove all records from the Catalog.");
       return null;
     }
 
@@ -80,70 +86,83 @@ public class RemoveCommand extends CqlCommands {
   }
 
   private Object executeRemoveFromCache() throws Exception {
-    String[] idsArray = new String[ids.size()];
-    idsArray = ids.toArray(idsArray);
-    getCacheProxy().removeById(idsArray);
 
-    List idsList = Arrays.asList(ids);
+    if (hasFilter()) {
+      printErrorMessage("Cache does not support filtering. Remove filter parameter.");
+      return null;
+    }
 
-    printSuccessMessage(idsList + " successfully removed from cache.");
-
-    LOGGER.info(idsList + " removed from cache using catalog:remove command");
+    if (CollectionUtils.isNotEmpty(ids)) {
+      getCacheProxy().removeById(ids.toArray(new String[0]));
+      printSuccessMessage(ids + " successfully removed from cache");
+      LOGGER.info("{} removed from cache by catalog:remove command", ids);
+    }
 
     return null;
   }
 
+  private int deletedIdsPassedAsArguments() throws Exception {
+
+    DeleteRequestImpl deleteRequest = new DeleteRequestImpl(ids.toArray(new String[0]));
+
+    LOGGER.debug("Attempting to delete {} metacards by ID", ids.size());
+    DeleteResponse deleteResponse = getCatalog().delete(deleteRequest);
+    return deleteResponse.getDeletedMetacards().size();
+  }
+
   private Object executeRemoveFromStore() throws Exception {
-    CatalogFacade catalogProvider = getCatalog();
+
+    int batchCount = 0;
+    int deletedCount = 0;
+    if (CollectionUtils.isNotEmpty(ids) && !hasFilter()) {
+      deletedCount = deletedIdsPassedAsArguments();
+    }
 
     if (hasFilter()) {
-      QueryImpl query = new QueryImpl(getFilter());
+      QueryRequestImpl queryRequest = new QueryRequestImpl(getQuery(), false);
+      String[] idsToDelete = getNextQueryBatch(queryRequest);
+      while (idsToDelete.length > 0) {
+        if (CollectionUtils.isNotEmpty(ids)) {
+          idsToDelete =
+              Arrays.asList(idsToDelete)
+                  .stream()
+                  .filter(id -> ids.contains(id))
+                  .toArray(String[]::new);
+        }
+        DeleteRequestImpl deleteRequest = new DeleteRequestImpl(idsToDelete);
+        LOGGER.debug(
+            "Attempting to delete {} metacards from batch {}", idsToDelete.length, ++batchCount);
+        DeleteResponse deleteResponse = catalogFramework.delete(deleteRequest);
+        deletedCount += deleteResponse.getDeletedMetacards().size();
 
-      query.setRequestsTotalResultsCount(true);
-      query.setPageSize(-1);
-
-      Map<String, Serializable> properties = new HashMap<>();
-      properties.put("mode", "native");
-
-      SourceResponse queryResponse = catalogProvider.query(new QueryRequestImpl(query, properties));
-
-      final List<String> idsFromFilteredQuery =
-          queryResponse
-              .getResults()
-              .stream()
-              .map(result -> result.getMetacard().getId())
-              .collect(Collectors.toList());
-
-      if (ids == null) {
-        ids = idsFromFilteredQuery;
-      } else {
-        ids =
-            ids.stream()
-                .filter(id -> idsFromFilteredQuery.contains(id))
-                .collect(Collectors.toList());
+        idsToDelete = getNextQueryBatch(queryRequest);
       }
     }
-
-    final int numberOfMetacardsToRemove = ids.size();
-    if (numberOfMetacardsToRemove > 0) {
-      printSuccessMessage("Found " + numberOfMetacardsToRemove + " metacards to remove.");
+    if (deletedCount > 0) {
+      printSuccessMessage(deletedCount + " documents successfully deleted.");
+      LOGGER.debug("{} documents removed using catalog:remove command", deletedCount);
     } else {
-      printErrorMessage("No records found meeting filter criteria.");
-      return null;
+      printErrorMessage("No documents match provided IDs or filter");
+      LOGGER.debug("No documents deleted using the catalog:remove command");
     }
-
-    DeleteRequestImpl request =
-        new DeleteRequestImpl(ids.toArray(new String[numberOfMetacardsToRemove]));
-    DeleteResponse response = catalogProvider.delete(request);
-
-    if (response.getDeletedMetacards().size() > 0) {
-      printSuccessMessage(ids + " successfully deleted.");
-      LOGGER.info(ids + " removed using catalog:remove command");
-    } else {
-      printErrorMessage(ids + " could not be deleted.");
-      LOGGER.info(ids + " could not be deleted using catalog:remove command");
-    }
-
     return null;
+  }
+
+  private String[] getNextQueryBatch(QueryRequest queryRequest) {
+    return ResultIterable.resultIterable(catalogFramework, queryRequest, BATCH_SIZE)
+        .stream()
+        .filter(Objects::nonNull)
+        .map(Result::getMetacard)
+        .filter(Objects::nonNull)
+        .map(Metacard::getId)
+        .distinct()
+        .toArray(String[]::new);
+  }
+
+  private QueryImpl getQuery() throws InterruptedException, ParseException, CQLException {
+    QueryImpl query = new QueryImpl(getFilter());
+    Map<String, Serializable> properties = new HashMap<>();
+    properties.put("mode", "native");
+    return query;
   }
 }
--- a/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
+++ b/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
@@ -13,114 +13,255 @@
  */
 package org.codice.ddf.commands.catalog;
 
-import static org.hamcrest.Matchers.containsString;
+import static java.util.stream.Collectors.toList;
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.hamcrest.core.StringContains.containsString;
 import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.isA;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
+import com.google.common.collect.ImmutableSet;
+import ddf.catalog.CatalogFramework;
 import ddf.catalog.cache.SolrCacheMBean;
 import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
 import ddf.catalog.data.impl.MetacardImpl;
+import ddf.catalog.data.impl.ResultImpl;
+import ddf.catalog.operation.DeleteRequest;
+import ddf.catalog.operation.DeleteResponse;
+import ddf.catalog.operation.QueryRequest;
+import ddf.catalog.operation.QueryResponse;
+import ddf.catalog.operation.impl.QueryResponseImpl;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Queue;
+import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Stream;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.stubbing.Answer;
 
 public class RemoveCommandTest extends ConsoleOutputCommon {
 
-  private List<Metacard> metacardList = getMetacardList(5);
+  public static final String DUMMY_FILTER = "title like '*'";
+
+  private static final int BATCH_SIZE = 250;
+
+  private List<Metacard> metacardList;
+
+  private RemoveCommand removeCommand;
+
+  private DeleteResponse deleteResponse;
+
+  private List<QueryResponse> queryResponseBatch;
+
+  private int sizeOfDeleteRequest;
 
   @Before
   public void setup() {
-    metacardList = getMetacardList(5);
-  }
-
-  @Test
-  public void testSingleItemList() throws Exception {
-    final SolrCacheMBean mbean = mock(SolrCacheMBean.class);
-    RemoveCommand removeCommand =
+    metacardList = populateMetacardList(5);
+    SolrCacheMBean solrCacheMock = mock(SolrCacheMBean.class);
+    deleteResponse = mock(DeleteResponse.class);
+    removeCommand =
         new RemoveCommand() {
           @Override
           protected SolrCacheMBean getCacheProxy() {
-            return mbean;
+            return solrCacheMock;
           }
         };
-
-    List<String> ids = new ArrayList<>();
-    ids.add(metacardList.get(0).getId());
-
-    removeCommand.ids = ids;
-    removeCommand.cache = true;
-
-    removeCommand.executeWithSubject();
-
-    String[] idsArray = new String[ids.size()];
-    idsArray = ids.toArray(idsArray);
-    verify(mbean, times(1)).removeById(idsArray);
+    removeCommand.catalogFramework = mock(CatalogFramework.class);
+    removeCommand.cache = false;
   }
 
   @Test
-  public void testMultipleItemList() throws Exception {
-    final SolrCacheMBean mbean = mock(SolrCacheMBean.class);
-    RemoveCommand removeCommand =
-        new RemoveCommand() {
-          @Override
-          protected SolrCacheMBean getCacheProxy() {
-            return mbean;
-          }
-        };
-
-    List<String> ids = new ArrayList<>();
-    ids.add(metacardList.get(0).getId());
-    ids.add(metacardList.get(1).getId());
-    ids.add(metacardList.get(2).getId());
-
+  public void testDeleteSingleMetacardFromCache() throws Exception {
+    Set<String> ids = oneIdToDelete();
     removeCommand.ids = ids;
     removeCommand.cache = true;
-
     removeCommand.executeWithSubject();
-
     String[] idsArray = new String[ids.size()];
     idsArray = ids.toArray(idsArray);
-    verify(mbean, times(1)).removeById(idsArray);
+    verify(removeCommand.getCacheProxy(), times(1)).removeById(idsArray);
+  }
+
+  @Test
+  public void testDeleteMultipleMetacardsFromCache() throws Exception {
+    Set<String> ids = threeIdsToDelete();
+    removeCommand.ids = ids;
+    removeCommand.cache = true;
+    removeCommand.executeWithSubject();
+    String[] idsArray = new String[ids.size()];
+    idsArray = ids.toArray(idsArray);
+    verify(removeCommand.getCacheProxy(), times(1)).removeById(idsArray);
   }
 
   /**
-   * Tests the {@Link RemoveCommand} when passed a null list of ids
-   *
-   * @throws Exception
+   * Tests condition where number of results returned by each catalog framework query is more than
+   * batchSize being used.
    */
   @Test
-  public void testNullList() throws Exception {
-    final SolrCacheMBean mbean = mock(SolrCacheMBean.class);
-    RemoveCommand removeCommand =
-        new RemoveCommand() {
-          @Override
-          protected SolrCacheMBean getCacheProxy() {
-            return mbean;
-          }
-        };
-
-    removeCommand.ids = null;
-
+  public void testQueryGreaterThanBatchSize() throws Exception {
+    int sizeOfDeleteRequest = 255;
+    configureQueryMock(sizeOfDeleteRequest);
+    configureDeleteMock();
+    removeCommand.cqlFilter = DUMMY_FILTER;
     removeCommand.executeWithSubject();
-
-    assertThat(consoleOutput.getOutput(), containsString("Nothing to remove."));
+    ArgumentCaptor<DeleteRequest> deleteRequestArgumentCaptor =
+        ArgumentCaptor.forClass(DeleteRequest.class);
+    verify(removeCommand.catalogFramework, times(2)).delete(deleteRequestArgumentCaptor.capture());
+    List<DeleteRequest> values = deleteRequestArgumentCaptor.getAllValues();
+    int totalDeleted = 0;
+    for (DeleteRequest request : values) {
+      totalDeleted += request.getAttributeValues().size();
+    }
+    assertThat(totalDeleted, equalTo(sizeOfDeleteRequest));
   }
 
-  private java.util.List<Metacard> getMetacardList(int amount) {
-    List<Metacard> metacards = new ArrayList<>();
+  @Test
+  public void testNullList() throws Exception {
+    removeCommand.ids = null;
+    removeCommand.executeWithSubject();
+    assertThat(consoleOutput.getOutput(), containsString("No IDs or filter provided"));
+  }
 
-    for (int i = 0; i < amount; i++) {
-      String id = UUID.randomUUID().toString();
+  @Test
+  public void testDeleteSingleMetacardFromStore() throws Exception {
+    sizeOfDeleteRequest = 1;
+    configureQueryMock(sizeOfDeleteRequest);
+    configureDeleteMock();
+    removeCommand.ids = oneIdToDelete();
+    removeCommand.executeWithSubject();
+    ArgumentCaptor<DeleteRequest> deleteRequestArgumentCaptor =
+        ArgumentCaptor.forClass(DeleteRequest.class);
+    verify(removeCommand.catalogFramework, times(1)).delete(deleteRequestArgumentCaptor.capture());
+    List<DeleteRequest> values = deleteRequestArgumentCaptor.getAllValues();
+    int totalDeleted = 0;
+    for (DeleteRequest request : values) {
+      totalDeleted += request.getAttributeValues().size();
+    }
+    assertThat(totalDeleted, equalTo(sizeOfDeleteRequest));
+  }
+
+  @Test
+  public void testFilterMatchesNothing() throws Exception {
+    sizeOfDeleteRequest = 0;
+    configureQueryMock(sizeOfDeleteRequest);
+    configureDeleteMock();
+    removeCommand.cqlFilter = "title like 'fake-filter'";
+    removeCommand.executeWithSubject();
+    ArgumentCaptor<DeleteRequest> deleteRequestArgumentCaptor =
+        ArgumentCaptor.forClass(DeleteRequest.class);
+    verify(removeCommand.catalogFramework, times(0)).delete(deleteRequestArgumentCaptor.capture());
+    List<DeleteRequest> values = deleteRequestArgumentCaptor.getAllValues();
+    int totalDeleted = 0;
+    for (DeleteRequest request : values) {
+      totalDeleted += request.getAttributeValues().size();
+    }
+    assertThat(totalDeleted, equalTo(sizeOfDeleteRequest));
+  }
+
+  @Test
+  public void testCacheAndFilter() throws Exception {
+    removeCommand.cqlFilter = DUMMY_FILTER;
+    removeCommand.cache = true;
+    removeCommand.executeWithSubject();
+    assertThat(consoleOutput.getOutput(), containsString("Cache does not support filtering."));
+  }
+
+  @Test
+  public void testBothIdsAndFilter() throws Exception {
+    configureQueryMock(3);
+    configureDeleteMock();
+    removeCommand.ids = ImmutableSet.of("1");
+    removeCommand.cqlFilter = DUMMY_FILTER;
+    removeCommand.executeWithSubject();
+    ArgumentCaptor<DeleteRequest> deleteRequestArgumentCaptor =
+        ArgumentCaptor.forClass(DeleteRequest.class);
+    verify(removeCommand.catalogFramework, times(1)).delete(deleteRequestArgumentCaptor.capture());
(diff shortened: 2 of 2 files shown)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -85,6 +89,11 @@ public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest) {
         return new QueryResultIterator(queryFunction, queryRequest);
     }
 
+    public Stream<Result> stream() {
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(iterator(),
+                +Spliterator.ORDERED), false);
```
josephthweatt: `+` Before the Spliterator is a typo I'm guessing?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -84,19 +88,15 @@ private Object executeRemoveFromStore() throws Exception {
...
+                            1,
+                            PAGE_SIZE,
+                            SortBy.NATURAL_ORDER,
+                            true,
+                            TimeUnit.SECONDS.toMillis(30)), false));
+
+            final List<String> idsFromFilteredQuery = resultIterable.stream()
                     .map(result -> result.getMetacard()
                             .getId())
                     .collect(Collectors.toList());
```
jhunzik: Running out of memory here and looping. Not sure as to why.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
jhunzik: Good catch

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -45,6 +47,8 @@
 
     private static final String IDS_LIST_ARGUMENT_NAME = "IDs";
 
+    private static final int PAGE_SIZE = 250;
```
mweser: What is the rationale for this number?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -84,19 +88,15 @@ private Object executeRemoveFromStore() throws Exception {
...
-            properties.put("mode", "native");
-
-            SourceResponse queryResponse = catalogProvider.query(new QueryRequestImpl(query,
-                    properties));
-
-            final List<String> idsFromFilteredQuery = queryResponse.getResults()
-                    .stream()
+            ResultIterable resultIterable = new ResultIterable(catalogFramework,
+                    new QueryRequestImpl(new QueryImpl(getFilter(),
+                            1,
```
mweser: Looks like a magic number that might benefit from being set as a constant.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
mweser: Perhaps restructure to the old-fashioned foreach loop and see if looping continues indefinitely. Also reference https://github.com/codice/ddf/pull/2148#discussion-diff-129152361R134 (line 134) where the last expected result in the test mock is an emptyList (simulates when results are exhausted).

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -118,7 +118,8 @@ private Object executeRemoveFromStore() throws Exception {
             return null;
         }
 
-        DeleteRequestImpl request = new DeleteRequestImpl(ids.toArray(new String[numberOfMetacardsToRemove]));
+        DeleteRequestImpl request =
+                new DeleteRequestImpl(ids.toArray(new String[numberOfMetacardsToRemove]));
```
mweser: Might consider breaking this into more steps/utility method to preserve readability.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
     }
 
+    @Test
+    public void testOverMaxItemList() throws Exception {
+        ConsoleOutput consoleOutput = mock(ConsoleOutput.class);
+        int totalResultsToDelete = 100;
+        int resultsPerCFQuery = 22;
+        int numCFCalls = 5;
+        List<String> ids = new ArrayList<>();
+        List<Metacard> mcList = getMetacardList(totalResultsToDelete);
```
mweser: Also might consider making the test vars into declared class fields if they are reused across more future tests. Then you can instantiate them in a Setup method and remove some redundancy. 

Agree with @josephthweatt on removing the abbreviations from the variable names to help with readability.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+        for(int i = 0; i < mcList.size(); i++) {
+            ids.add(mcList.get(i).getId());
+        }
+
+        RemoveCommand removeCommand = new RemoveCommand();
+
+        final CatalogFramework catalogFramework = mock(CatalogFramework.class);
```
mweser: Why is this mock "final"?

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
 
+    @Test
+    public void testOverMaxItemList() throws Exception {
+        ConsoleOutput consoleOutput = mock(ConsoleOutput.class);
+        int totalResultsToDelete = 100;
+        int resultsPerCFQuery = 22;
+        int numCFCalls = 5;
+        List<String> ids = new ArrayList<>();
+        List<Metacard> mcList = getMetacardList(totalResultsToDelete);
+        for(int i = 0; i < mcList.size(); i++) {
```
mweser: Might consider refactoring to foreach loop to make more concise and remove need for index.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+            ids.add(mcList.get(i).getId());
+        }
+
+        RemoveCommand removeCommand = new RemoveCommand();
+
+        final CatalogFramework catalogFramework = mock(CatalogFramework.class);
+
+        QueryResponse queryResponse = mock(QueryResponse.class);
```
mweser: Consider placing mock instantiations into a setup method as well as any repeated when statements to clean up this test method. That is the role of the setup. Refer to https://github.com/codice/ddf/pull/2148/files#diff-84f7749e6d7b0fc4c56387a17d2799a8R82 for an example of how to set up class/setup method for better reusability.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+        DeleteResponse deleteResponse = mock(DeleteResponse.class);
+        when(deleteResponse.getDeletedMetacards()).thenReturn(getMetacardList(resultsPerCFQuery));
+        when(catalogFramework.delete(isA(DeleteRequest.class))).thenReturn(deleteResponse);
+
+        removeCommand.catalogFramework = catalogFramework;
+        removeCommand.cqlFilter = "title like 'a*'";
+        removeCommand.ids = ids;
+        removeCommand.cache = false;
+
+        verify(catalogFramework, times(numCFCalls)).delete(isA(DeleteRequest.class));
```
mweser: Verify should go after assert statement to ensure that extra call wasn't made in process of assertion.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+        assertThat(consoleOutput.getOutput(), containsString("100" + " metacards to remove."));
```
mweser: Consider extracting expected results strings to the top of the method for better readability and reusability.

mweser: Also extract string returned from `.getOutput()` to var and then compare in second assertThat argument with `resultString.contains(...)`. Right now simply calling `containsString("")` with a string literal does not seem to check much of anything.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -128,4 +171,21 @@ protected SolrCacheMBean getCacheProxy() {
 
         return metacards;
     }
+
+    private java.util.List<Result> getResultList(int amount) {
```
mweser: Comment or rename to indicate operation of adding random UUID to each result in the list.

Perhaps also changing "amount" to "size" would be a bit more consistent with how we usually write these kinds of list population methods. 

Reference: https://github.com/codice/ddf/pull/2148/files#diff-84f7749e6d7b0fc4c56387a17d2799a8R176

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -128,4 +171,21 @@ protected SolrCacheMBean getCacheProxy() {
...
+        java.util.List<Result> results = new ArrayList<>();
+
+        for (int i = 0; i < amount; i++) {
+
+            String id = UUID.randomUUID()
+                    .toString();
+            MetacardImpl metacard = new MetacardImpl();
```
mweser: Can a mock work here instead of MetacardImpl and ResultImpl instances? If changed to use mocks, make sure to add `when` clauses before adding to list.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -128,4 +171,21 @@ protected SolrCacheMBean getCacheProxy() {
 
         return metacards;
     }
+
+    private java.util.List<Result> getResultList(int amount) {
+        java.util.List<Result> results = new ArrayList<>();
```
mweser: Should be able to remove `java.util.` section of List type. Might need to put cursor on "List" in IntelliJ and then press `ALT+ENTER` to add import.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+        removeCommand.cqlFilter = "title like 'a*'";
+        removeCommand.ids = ids;
+        removeCommand.cache = false;
+
+        verify(catalogFramework, times(numCFCalls)).delete(isA(DeleteRequest.class));
+        assertThat(consoleOutput.getOutput(), containsString("100" + " metacards to remove."));
+    }
+
     /**
      * Tests the {@Link RemoveCommand} when passed
```
mweser: Perhaps more details on the test. Method name might be descriptive enough to render this comment unnecessary.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
         verify(mbean, times(1)).removeById(idsArray);
     }
 
+    @Test
+    public void testOverMaxItemList() throws Exception {
```
mweser: For readability, try to separate inputs to test method from declarations. This can be done as mentioned in another of my comments about restructuring into class fields and setup method.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+        QueryResponse queryResponse = mock(QueryResponse.class);
+        when(queryResponse.getResults()).thenReturn(getResultList(totalResultsToDelete));
+        when(catalogFramework.query(isA(QueryRequest.class))).thenReturn(queryResponse);
+
+        DeleteResponse deleteResponse = mock(DeleteResponse.class);
+        when(deleteResponse.getDeletedMetacards()).thenReturn(getMetacardList(resultsPerCFQuery));
+        when(catalogFramework.delete(isA(DeleteRequest.class))).thenReturn(deleteResponse);
+
+        removeCommand.catalogFramework = catalogFramework;
+        removeCommand.cqlFilter = "title like 'a*'";
```
mweser: Consider using local variable to store this at the top of the method rather than passing in literal. Same with the boolean value below. May help with readability.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -91,6 +101,39 @@ protected SolrCacheMBean getCacheProxy() {
...
+        int totalResultsToDelete = 100;
+        int resultsPerCFQuery = 22;
+        int numCFCalls = 5;
+        List<String> ids = new ArrayList<>();
+        List<Metacard> mcList = getMetacardList(totalResultsToDelete);
+        for(int i = 0; i < mcList.size(); i++) {
+            ids.add(mcList.get(i).getId());
+        }
+
+        RemoveCommand removeCommand = new RemoveCommand();
```
mweser: Try to make this one of the only `new` operators in this test class and replace others with `mock` whenever possible.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -84,19 +88,15 @@ private Object executeRemoveFromStore() throws Exception {
...
-                    properties));
-
-            final List<String> idsFromFilteredQuery = queryResponse.getResults()
-                    .stream()
+            ResultIterable resultIterable = new ResultIterable(catalogFramework,
+                    new QueryRequestImpl(new QueryImpl(getFilter(),
+                            1,
+                            PAGE_SIZE,
+                            SortBy.NATURAL_ORDER,
+                            true,
```
mweser: A nit, but perhaps store the `true` to a more descriptive local variable to better illustrate what this value is used for.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -84,19 +88,15 @@ private Object executeRemoveFromStore() throws Exception {
...
+                            TimeUnit.SECONDS.toMillis(30)), false));
```
mweser: Perhaps extract timeout value to constant higher up in class so that it is easier to change in the future.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
clockard: Doing `ids.toArray(new String[0])` gives slightly better performance. Counter intuitive I know. Something I learned at uberconf.

rzwiefel: Just for everyones knowledge who missed the session...Technically just `toArray()` and the `Object[]` versions are the fastest, but you would have to cast the result. I doubt its worth obfuscating this code though since this is probably the least of our performance concerns. 😄

rzwiefel: screenshot 2017-07-26 12 42 37
Credit Douglas Hawkins Uberconf 2017 talk https://uberconf.com/s/slides/2017/speaker/Douglas_Hawkins/java_performance_puzzlers/Performance_Puzzlers.pdf

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
lessarderic: ❗️ The `Iterable` will loop over _every_ metacard that matches the filter and will load them all in memory. The same thing would have happened in the past since we were asking for all (`-1`) metacards, but now that we're capping the number, no more than 1,000 metacards will be returned.

What we need to do here is either use the `ResultListIterable` to process things in batches, or keep using the `QueryResponse` in a loop until the number of metacards returned is 0. You can look at this PR for an example of how this can be done: https://github.com/codice/ddf/pull/2204.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -118,7 +124,8 @@ private Object executeRemoveFromStore() throws Exception {
             return null;
         }
 
-        DeleteRequestImpl request = new DeleteRequestImpl(ids.toArray(new String[numberOfMetacardsToRemove]));
+        DeleteRequestImpl request =
+                new DeleteRequestImpl(ids.toArray(new String[numberOfMetacardsToRemove]));
         DeleteResponse response = catalogProvider.delete(request);
 
         if (response.getDeletedMetacards()
```
lessarderic: ❗️ We should change the success and error messages. Since the number of IDs could be huge if the filter is very broad, printing out all of them is probably not a good idea. We could only keep track of the number of metacards deleted instead and only report that number.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
lessarderic: @jhunzik, the `ResultListIterable` will be removed. If you need to process results in batches, you should use Guava's [Iterables.partitions()](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Iterables.html#partition-java.lang.Iterable-int-) instead.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -88,7 +111,50 @@ protected SolrCacheMBean getCacheProxy() {
...
+    public void testMultipleCatalogCalls() throws Exception {
+
+        int numCatalogCalls = 6;
+        int totalResultsToDelete = 100;
+
+        String expectedResult = totalResultsToDelete + " metacards successfully deleted.";
+
+        filter = "title like '*'";
+
+        setQueryAndDeleteResponseMocks(19, 18, 17, 19, 22, 5, 0);
```
jhunzik: ResultIterable in RemoveCommand.java only gets 18 results instead of 100 and I feel like it's coming from the implementation of this method.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -99,33 +165,99 @@ protected SolrCacheMBean getCacheProxy() {
...
+
+        List<List<Result>> queryResponseMockList = new ArrayList<>();
+
+        for (int numResultsForQuery : numResultsPerQueryList) {
+
+            List<Result> resultList = populateResultList(numResultsForQuery);
+            queryResponseMockList.add(resultList);
+
+            // add two copies of first result list as implementation of RemoveAllCommand
+            // calls queryResponse.getResults() three times per catalog framework query
```
emanns95: This command is not `removeall` and does not do that. I don't think the first result should be added three times

jhunzik: Oops, I need to have Weser remind me why it's necessary to add 3 times in `removeall`

emanns95: This seems like the tests are way too tightly coupled to the implementation 😞

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -81,53 +91,50 @@ private Object executeRemoveFromCache() throws Exception {
     }
 
     private Object executeRemoveFromStore() throws Exception {
+
         CatalogFacade catalogProvider = getCatalog();
 
+        int batchCount = 1;
```
ahoffer: Where is `batchCount` used?

jhunzik: It's used for logging the current batch in debug. Not really important. I'll keep it in if there if I don't see any objections.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -13,67 +13,93 @@
...
     public void setup() {
-        metacardList = getMetacardList(5);
-    }
+        metacardList = populateMetacardList(5);
 
-    @Test
-    public void testSingleItemList() throws Exception {
-        final SolrCacheMBean mbean = mock(SolrCacheMBean.class);
-        RemoveCommand removeCommand = new RemoveCommand() {
+        mbeanMock = mock(SolrCacheMBean.class);
```
lessarderic: ✏️  We may want to give this variable a better name, e.g., `solrCacheMBeanMock` or just `solrCacheMock`.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -88,7 +114,46 @@ protected SolrCacheMBean getCacheProxy() {
...
+     * Fourth:      73                 (19 results returned)
+     * Fifth:       95                 (22 results returned)
+     * Sixth:      100                 (5 results returned)
+     * <p>
+     * Final:       0 results returned, therefore done.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testMultipleCatalogCalls() throws Exception {
```
lessarderic: ❗️ We probably need to clarify what we're trying to test here and break up this test into multiple ones, or use JUnit's `@Paramterized`.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -81,53 +91,50 @@ private Object executeRemoveFromCache() throws Exception {
...
+        int deletedCount = 0;
+
         if (hasFilter()) {
             QueryImpl query = new QueryImpl(getFilter());
-
             query.setRequestsTotalResultsCount(true);
-            query.setPageSize(-1);
-
+            query.setTimeoutMillis(TimeUnit.SECONDS.toMillis(10));
```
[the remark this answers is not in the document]
jhunzik: This is an old addition and I don't think it's necessary. Master doesn't have it.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -45,8 +53,10 @@
 
     private static final String IDS_LIST_ARGUMENT_NAME = "IDs";
 
+    int batchSize = 250;
+
     @Argument(name = IDS_LIST_ARGUMENT_NAME, description = "The id(s) of the document(s) by which to filter.", index = 0, multiValued = true, required = false)
-    List<String> ids = null;
+    Set<String> ids = new HashSet<>();
```
emanns95: Might want to double check removal by ID in case it's not okay with a set here instead of a list

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
emanns95: Update: The ResultIterable also will call 3 `getResults`, so it's unfortunate. Maybe there's a better way to do this without knowing all of that

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -81,53 +91,50 @@ private Object executeRemoveFromCache() throws Exception {
...
-                ids = idsFromFilteredQuery;
-            } else {
-                ids = ids.stream()
-                        .filter(id -> idsFromFilteredQuery.contains(id))
-                        .collect(Collectors.toList());
+            ResultIterable resultIterable = ResultIterable.resultIterable(catalogFramework,
+                    new QueryRequestImpl(query, false));
+
+            Set<String> idsToDelete = resultIterable.stream()
+                    .filter(Objects::nonNull)
```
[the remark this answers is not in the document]
jhunzik: A null result or a null metacard within a valid result

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -81,53 +90,49 @@ private Object executeRemoveFromCache() throws Exception {
...
 
     private Object executeRemoveFromStore() throws Exception {
+
         CatalogFacade catalogProvider = getCatalog();
 
+        int batchCount = 1;
+        int deletedCount = 0;
+
         if (hasFilter()) {
             QueryImpl query = new QueryImpl(getFilter());
```
rzwiefel: @emanns95 @lessarderic Sorry I missed the convo but I don't think we want the default timeout... because the default is infinite.

emanns95: I thought we updated the default to be 5 minutes?
https://codice.atlassian.net/browse/DDF-2862

jhunzik:
build now

emanns95:
build now

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -45,17 +49,19 @@
...
 
     @Option(name = "--cache", required = false, multiValued = false, description = "Only remove cached entries.")
     boolean cache = false;
 
     @Override
     protected Object executeWithSubject() throws Exception {
         if (CollectionUtils.isEmpty(ids) && !hasFilter()) {
-            printErrorMessage("Nothing to remove. Either the " + IDS_LIST_ARGUMENT_NAME
-                    + " argument or another filter must be specified. Please, use the catalog:removeall command if the goal is to delete all records from the Catalog.");
+            printErrorMessage("Nothing to remove. No IDs detected and no filter detected."
```
AzGoalie: ✏️ Nit: Could be rewritten as "No IDs or filter detected"

emanns95:
To hero: 
1. Ingest hundreds of records (700+)
2. Perform a remove command from the catalog that will remove all of them (` catalog:remove --cql "title like '*'"`)
3. Verify all records are removed

emanns95:
I tried to manually verify this but it would only remove 1000 records at a time

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -67,69 +73,74 @@ protected Object executeWithSubject() throws Exception {
...
-        getCacheProxy().removeById(idsArray);
-
-        List idsList = Arrays.asList(ids);
-
-        printSuccessMessage(idsList + " successfully removed from cache.");
-
-        LOGGER.info(idsList + " removed from cache using catalog:remove command");
-
+        if (CollectionUtils.isNotEmpty(ids)) {
+            getCacheProxy().removeById(ids.toArray(new String[ids.size()]));
```
AzGoalie: ✏️ Nit: `new String[0]` would actually be faster than `new String[id.size()]`

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -67,69 +73,74 @@ protected Object executeWithSubject() throws Exception {
...
-        CatalogFacade catalogProvider = getCatalog();
-
-        if (hasFilter()) {
-            QueryImpl query = new QueryImpl(getFilter());
+    private int deletedIdsPassedAsArguments() throws Exception {
 
-            query.setRequestsTotalResultsCount(true);
-            query.setPageSize(-1);
+        DeleteRequestImpl deleteRequest =
+                new DeleteRequestImpl(ids.toArray(new String[ids.size()]));
```
AzGoalie: Same comment as above

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -67,69 +73,74 @@ protected Object executeWithSubject() throws Exception {
...
-        final int numberOfMetacardsToRemove = ids.size();
-        if (numberOfMetacardsToRemove > 0) {
-            printSuccessMessage("Found " + numberOfMetacardsToRemove + " metacards to remove.");
+        if (deletedCount > 0) {
+            printSuccessMessage(deletedCount + " metacards successfully deleted.");
+            LOGGER.debug(deletedCount + " metacards removed using catalog:remove command");
         } else {
-            printErrorMessage("No records found meeting filter criteria.");
-            return null;
+            printErrorMessage("No metacard IDs and no metacards found that match search criteria");
```
AzGoalie: ✏️ Nit: Could remove mention of metacard IDs as it is redundant.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -88,7 +116,29 @@ protected SolrCacheMBean getCacheProxy() {
...
+        totalResultsToDelete = 255;
+        String expectedResult = totalResultsToDelete + " metacards successfully deleted.";
+        filter = "title like '*'";
+
+        configureMocks(250, 5, 0);
+        setUpRemoveCommand();
+        removeCommand.executeWithSubject();
+
+        String actualResult = consoleOutput.getOutput();
+        assertThat(actualResult, containsString(expectedResult));
```
clockard: asserting on the output string seems a little brittle. I think for what you are trying to test the verify check is probably sufficient.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
ahoffer: Poor Josh. I was the one who change it from `new String[0]` to `new String[id.size()]`. I did not know the first was faster. Why is it so?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -67,69 +73,74 @@ protected Object executeWithSubject() throws Exception {
...
-        List idsList = Arrays.asList(ids);
-
-        printSuccessMessage(idsList + " successfully removed from cache.");
-
-        LOGGER.info(idsList + " removed from cache using catalog:remove command");
-
+        if (CollectionUtils.isNotEmpty(ids)) {
+            getCacheProxy().removeById(ids.toArray(new String[ids.size()]));
+            printSuccessMessage(ids + " successfully removed from cache");
+            LOGGER.info(ids + " removed from cache using catalog:remove command");
```
shaundmorris: ✏️  you could use slf4j parameterized logging for `ids` on this logger statement: https://www.slf4j.org/faq.html#logging_performance

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
coyotesqrl: Because when the array is `malloc`'d Java needs to null all the slots in it...and then the `toArray` call is just going to blow it away and make a new array anyway.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveCommand.java
```diff
@@ -67,69 +73,74 @@ protected Object executeWithSubject() throws Exception {
...
+                idsToDelete = getNextQueryBatch(queryRequest);
             }
         }
-
-        final int numberOfMetacardsToRemove = ids.size();
-        if (numberOfMetacardsToRemove > 0) {
-            printSuccessMessage("Found " + numberOfMetacardsToRemove + " metacards to remove.");
+        if (deletedCount > 0) {
+            printSuccessMessage(deletedCount + " metacards successfully deleted.");
+            LOGGER.debug(deletedCount + " metacards removed using catalog:remove command");
```
coyotesqrl: ✏️ Use parameter interpolation for logging.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -88,7 +116,29 @@ protected SolrCacheMBean getCacheProxy() {
...
+        verify(catalogFrameworkMock, times(2)).delete(isA(DeleteRequest.class));
```
lessarderic: ❗️For the new tests, we may want to capture the parameter being passed to `delete` using an `ArgumentCaptor` and at least make sure we're calling the catalog framework with the right number of IDs.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -99,33 +149,99 @@ protected SolrCacheMBean getCacheProxy() {
...
+        configureMocks(1, 0);
+        setUpRemoveCommand();
+        removeCommand.ids = ImmutableSet.of(metacardList.get(0)
+                .getId());
         removeCommand.executeWithSubject();
+        verify(catalogFrameworkMock, times(1)).delete(isA(DeleteRequest.class));
+    }
 
-        assertThat(consoleOutput.getOutput(), containsString("Nothing to remove."));
+    private void configureMocks(int... totalNumberOfMetacards) throws Exception {
```
lessarderic: ❗️We should add one test where the filter returns no IDs (e.g., `configureMocks(0)`).
❗️We should also have a test where both, a list of IDs and a filter are provided.

ahoffer:
Manually verified correct behavior for commit 7bd1b8a6eb4b613552353db9cb73757cd1582eb1. Deleted 1,0001 with the command `catalog:remove --cql "title like '*'"

ahoffer:
build now

ahoffer:
build now

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveCommandTest.java
```diff
@@ -13,119 +13,233 @@
...
+            queryResponseMockList.add(resultList);
+
+            // add two copies of first result list as implementation of RemoveCommand
+            // calls queryResponse.getResults() three times per catalog framework query
+            // Calling getResults does NOT advance index of queryResponse
+            queryResponseMockList.add(resultList);
+            queryResponseMockList.add(resultList);
         }
+        when(queryResponse.getHits()).thenReturn(hits);
+        when(queryResponse.getResults()).thenReturn(queryResponseMockList.get(0),
```
emanns95: I think we're too closely tied to the implementation here. We should mock out the response from calling the `query` method on the catalog framework instead of the `getResults` on the actual mocked response. This will free us from being tied to the number of calls to `getResults`. See https://github.com/codice/ddf/commit/68f352ed2dd2d1554d6256977dbd6936023f452a (TestCswEndpoint) for an example

jhunzik:
I'm trying to uncouple the testing from the implementation but ran into issues with the delete response in `configureMocks`.

jhunzik:
build now

jhunzik:
build now

mweser:
Completed hero, and works as expected.

jhunzik:
build now

jhunzik:
build now

jhunzik:
build now

jhunzik:
build now

jhunzik:
build now

PR #2131, opened 2017-07-11
DDF-3073 Adds failover log back to logging config

vinamartin:
#### What does this PR do?
Adds failover log back to logging configuration.
With the Karaf upgrade, there is an issue with the pax logging configuration that makes it unavailable to use through our default configuration, however, when using the log4j2.config.xml file failover seems to be functional (though errors populate the log). 

#### Who is reviewing it? 
@emmberk 
@Lambeaux 
@josephthweatt 
@pklinef 
@coyotesqrl 

#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)
@clockard
@lessarderic

#### How should this be tested? (List steps with links to updated documentation)

- Create a disk image and mount it (Here's an example of how to do so on a mac: https://support.apple.com/kb/PH22247?locale=en_US&viewlocale=en_US)
- Unzip the DDF
- Configure the DDF's logging configuration per the documentation and setup the `securityMain` appender to write to a log file inside of the disk image
- Start up the DDF 
- Verify logs are written to the security log inside the disk image
- Begin to fill up the disk image (for instance: http://www.skorks.com/2010/03/how-to-quickly-generate-a-large-file-on-the-command-line-with-linux/)
- Verify that when the disk is full, the security backup log is written to

#### What are the relevant tickets?
[DDF-3073](https://codice.atlassian.net/browse/DDF-3073)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2154, opened 2017-07-24
DDF-3169 Adds alerts for failover log failure

vinamartin:
#### What does this PR do?
Alerts admins when failover logging fails. 

_Original PR_ : 
[DDF-2912 Implemented System Administrator Alerts](https://github.com/codice/ddf/pull/1829) by @emmberk 
_Changes_: 
Log4j2 formatting is different from Log4j
Had to add in a time check to the appender because of how fast the logs come in

_Must be built using_ : 
[DDF 3128 Add Decanter and alert/solr appenders base for future expansion](https://github.com/codice/ddf/pull/2130) by @clockard 
[DDF 3143 Update alerting in the admin UI](https://github.com/codice/ddf/pull/2141) by @clockard 

#### Who is reviewing it? 
@emmberk @clockard @ahoffer @josephthweatt 

@clockard
@figliold 

Follow the directions here: https://github.com/codice/ddf/pull/2131
but while configuring the log4j2.config.xml config, also configure the failover appender look like the following:
```
        <Failover name="securityFailover" primary="securityMain">
            <Failovers>
                <AppenderRef ref="osgi-adminAlertAppender"/>
            </Failovers>
        </Failover>
```
This will make it so when a failover occurs it will not use the failover file and will immediately use the alerter. Ensure that an alert appears on the admin console once the disk is full and the main audit log has failed. 

#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.logging.cfg
+++ b/distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.logging.cfg
@@ -31,7 +31,8 @@ log4j2.pattern = %-5.5p | %-16.16t | %-32.32c{1} | %-32.32C %4L | %X{bundle.id} 
 # Root logger
 log4j2.rootLogger.level = INFO
 log4j2.rootLogger.appenderRef.out.ref = out
-log4j2.rootLogger.appenderRef.osgi-all.ref = osgi-all
+log4j2.rootLogger.appenderRef.osgi-platformLogging.ref = osgi-platformLogging
+log4j2.rootLogger.appenderRef.osgi-karafConsole.ref = osgi-karafConsole
 
 # Loggers configuration
 
@@ -106,10 +107,10 @@ log4j2.appender.osgi-platformLogging.type = PaxOsgi
 log4j2.appender.osgi-platformLogging.name = osgi-platformLogging
 log4j2.appender.osgi-platformLogging.filter = org.codice.ddf.platform.logging.LoggingService
 
-# osgi-all
-log4j2.appender.osgi-all.type = PaxOsgi
-log4j2.appender.osgi-all.name = osgi-all
-log4j2.appender.osgi-all.filter = *
+# osgi-karafConsole
+log4j2.appender.osgi-karafConsole.type = PaxOsgi
+log4j2.appender.osgi-karafConsole.name = osgi-karafConsole
+log4j2.appender.osgi-karafConsole.filter = VmLogAppender
 
 # syslog
 log4j2.appender.syslog.type = Syslog
@@ -196,5 +197,3 @@ log4j2.appender.artemis.policies.size.type = SizeBasedTriggeringPolicy
 log4j2.appender.artemis.policies.size.size = 20MB
 log4j2.appender.artemis.strategy.type = DefaultRolloverStrategy
 log4j2.appender.artemis.strategy.max = 10
-
-# Setting custom configurations using log:set from the console will append those configurations below this line
--- a/distribution/ddf-common/src/main/resources/etc/log4j2.config.xml
+++ b/distribution/ddf-common/src/main/resources/etc/log4j2.config.xml
@@ -23,7 +23,10 @@
         <PaxOsgi name="osgi-platformLogging"
                  filter="org.codice.ddf.platform.logging.LoggingService"/>
 
-        <PaxOsgi name="osgi-all" filter="*"/>
+        <PaxOsgi name="osgi-adminAlertAppender"
+                 filter="ddf.security.common.audit.LogFailoverAlertAppender"/>
+
+        <PaxOsgi name="osgi-karafConsole" filter="VmLogAppender"/>
 
         <Syslog name="syslog" facility="AUTH" host="localhost" protocol="UDP" port="514"/>
 
@@ -75,6 +78,7 @@
         <Failover name="securityFailover" primary="securityMain">
             <Failovers>
                 <AppenderRef ref="securityBackup"/>
+                <AppenderRef ref="osgi-adminAlertAppender"/>
             </Failovers>
         </Failover>
 
@@ -146,7 +150,8 @@
 
         <Root level="info">
             <AppenderRef ref="out"/>
-            <AppenderRef ref="osgi-all"/>
+            <AppenderRef ref="osgi-karafConsole"/>
+            <AppenderRef ref="osgi-platformLogging"/>
         </Root>
     </Loggers>
 </Configuration>
--- a/platform/security/platform-security-core-api/pom.xml
+++ b/platform/security/platform-security-core-api/pom.xml
@@ -140,6 +140,16 @@
             <artifactId>platform-util</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.ops4j.pax.logging</groupId>
+            <artifactId>pax-logging-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.codice.ddf</groupId>
+            <artifactId>alerts</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
     <dependencyManagement>
         <dependencies>
@@ -189,7 +199,8 @@
                             httpcore,
                             google-http-client,
                             opensaml-soap-impl,
-                            common-system
+                            common-system,
+                            alerts
                         </Embed-Dependency>
                         <Export-Package>
                             ddf.security*;version=${project.version}
--- a/platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/LogFailoverAlertAppender.java
+++ b/platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/LogFailoverAlertAppender.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p/>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p/>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.security.common.audit;
+
+import static org.apache.commons.lang.Validate.notNull;
+
+import java.util.Arrays;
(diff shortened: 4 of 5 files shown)


platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public void doAppend(PaxLoggingEvent event) {
+        if(lastUpdated == null){
+            lastUpdated = DateTime.now();
+        } else if (lastUpdated.minusSeconds(5).isAfter(lastUpdated.toInstant())) {
+            return;
+        }
+        SystemNotice notice = new SystemNotice(AdminAlertAppender.class.toString(),
+                NoticePriority.CRITICAL.value(),
+                "Failover Appender Failure",
```
clockard: Maybe `Audit Logging Failure` since this is going to be displayed to the admin.

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,52 @@
...
+                Collections.emptySet());
```
clockard: Do we want to add an entry here that gives some explanation as to what happened with potential causes?

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
vinamartin: That sounds good, I was hoping someone would come up with better phrasing.

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -104,10 +105,10 @@ log4j2.appender.osgi-platformLogging.type = PaxOsgi
 log4j2.appender.osgi-platformLogging.name = osgi-platformLogging
 log4j2.appender.osgi-platformLogging.filter = org.codice.ddf.platform.logging.LoggingService
 
-# osgi-all
+# osgi-karafConsole
 log4j2.appender.osgi-all.type = PaxOsgi
 log4j2.appender.osgi-all.name = osgi-all
-log4j2.appender.osgi-all.filter = *
+log4j2.appender.osgi-all.filter = VmLogAppender
```
emmberk: I'd think this section would need to be updated to 
```
# osgi-karafConsole
log4j2.appender.osgi-karafConsole.type = PaxOsgi
log4j2.appender.osgi-karafConsole.name = osgi-karafConsole
log4j2.appender.osgi-karafConsole.filter = VmLogAppender```

Comment (vinamartin) [inline distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.logging.cfg]: Oops good catch!

File: platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java (hunk 2 above)
Comment (vinamartin) [inline platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java]: "Audit Logging Failure"
"Writing to the backup audit log has failed. This could occur if there are insufficient permissions to write to the log directory, a full disk, or the primary or backup log configurations are incorrect."

Thoughts?

Comment (clockard) [inline platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java]: Sounds good to me.

File: platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java [hunk 4]
```diff
@@ -0,0 +1,52 @@
...
+    public AdminAlertAppender(EventAdmin eventAdmin) {
+        notNull(eventAdmin, "eventAdmin may not be null");
+        this.eventAdmin = eventAdmin;
+    }
+
+    @Override
+    public void doAppend(PaxLoggingEvent event) {
+        if(lastUpdated == null){
+            lastUpdated = DateTime.now();
+        } else if (lastUpdated.minusSeconds(5).isAfter(lastUpdated.toInstant())) {
```
figliold: :question: What case does this test?

vinamartin: Good question! Every time a log would have been written to the backup log (and fails), this appender gets called. Sometimes there is a huge list of these logs that get fired very quickly. 
Since the alerts work by querying Solr for an alert from the same place, there's the possibility for a timing issue which will make several of these alerts all at once. 
This check is supposed to mean, _"If I've already sent this alert in the last 5 seconds, don't send it again"_

vinamartin: Ah just realized there's an issue here. I meant to say "If this has already been sent in the last 5 seconds, don't send it again" but now I see that's not what this is doing.

In any case, there's a timing issue with how these alerts work because of how fast the log gets spammed.

josephthweatt: I think the logic for this could also be a bit clearer. The statement reads `(time - 5) > lastTime` when it might be better to do `time > lastTime + 5`

vinamartin: Huh, now that I look at it again, don't understand why I was doing minus when there's a plus operation. 🤔

distribution/ddf-common/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -196,5 +197,3 @@ log4j2.appender.artemis.policies.size.type = SizeBasedTriggeringPolicy
 log4j2.appender.artemis.policies.size.size = 20MB
 log4j2.appender.artemis.strategy.type = DefaultRolloverStrategy
 log4j2.appender.artemis.strategy.max = 10
-
-# Setting custom configurations using log:set from the console will append those configurations below this line
```
emmberk: Nice change! I confirmed that custom configurations using `log:set` from the console are appended after the last line of configuration in this file, not appended to the end of the file.

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,56 @@
...
+        if (lastUpdated == null) {
+            lastUpdated = DateTime.now();
+        } else if (DateTime.now()
+                .isBefore(lastUpdated.plusSeconds(5))) {
+            // don't spam the admin console when multiple log messages come in at once
+            return;
+        }
+        SystemNotice notice = new SystemNotice(AdminAlertAppender.class.toString(),
+                NoticePriority.CRITICAL,
+                "Failover Appender Failure",
```
clockard: I thought we had changed this name and added a description line?

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.codice.ddf.system.alerts.NoticePriority;
+import org.codice.ddf.system.alerts.SystemNotice;
+import org.joda.time.DateTime;
+import org.ops4j.pax.logging.spi.PaxAppender;
+import org.ops4j.pax.logging.spi.PaxLoggingEvent;
+import org.osgi.service.event.Event;
+import org.osgi.service.event.EventAdmin;
+
+public final class AdminAlertAppender implements PaxAppender {
```
ahoffer: From the class name, it looks like the AdminAlertAppender could be appending any kind of alert. But from the doAppend, it looks specific to failure of the audit log. If it is that specific, consider making the class name more specific.

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,56 @@
...
+        this.eventAdmin = eventAdmin;
+    }
+
+    @Override
+    public void doAppend(PaxLoggingEvent event) {
+        if (lastUpdated == null) {
+            lastUpdated = DateTime.now();
+        } else if (DateTime.now()
+                .isBefore(lastUpdated.plusSeconds(5))) {
+            // don't spam the admin console when multiple log messages come in at once
```
ahoffer: I think of the admin console as the Karaf shell. Are these notices begin printed to the Karaf shell, or only the Admin UI?

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
```diff
@@ -0,0 +1,56 @@
...
+        notNull(eventAdmin, "eventAdmin may not be null");
+        this.eventAdmin = eventAdmin;
+    }
+
+    @Override
+    public void doAppend(PaxLoggingEvent event) {
+        if (lastUpdated == null) {
+            lastUpdated = DateTime.now();
+        } else if (DateTime.now()
+                .isBefore(lastUpdated.plusSeconds(5))) {
```
ahoffer: If the event parameter always the same even, this works fine-- subsequent messages are likely to be duplicates. Is there any chance of different kinds of events being passed into the method? For example, if a logging event for audit failure was followed 1 second later by a logging event for denial-of-service, then the second event would not become an alert.

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
vinamartin: Weird, looks like I forgot to add the file to the last commit

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
vinamartin: You're right, I just copy pasted this so I can go ahead and rename it to something more specific

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
vinamartin: ^ Following on from your previous comment above, I designed this to only do this on failover which is why this check exists (because of the nature of logs)

platform/security/platform-security-core-api/src/main/java/ddf/security/common/audit/AdminAlertAppender.java
vinamartin: Admin UI

vinamartin:
build now

vinamartin:
build now

alexaabrd:
successfully heroed

vinamartin:
build now

PR #2155, opened 2017-07-25
DDF-2863 Added ability to set servlet filter initialization parameters

emmberk:
#### What does this PR do?
This PR
- Adds an experimental `SecurityService` marker interface.
- Updates the `DelegateServletFilter` to
  - gather services that implement the `SecurityService` interface instead of the `javax.servlet.Filter` interface.
  - only `init` each filter once instead of on every `doFilter` call.
- Updates current filters to
  - implement the `SecurityService` interface instead of the `javax.servlet.Filter` interface.
  - use the Pax Web Extender filter registration notation. See [Pax Web Extender Whiteboard](https://ops4j1.jira.com/wiki/display/ops4j/Pax+Web+Extender+-+Whiteboard), [`org.ops4j.pax.web.extender.whiteboard.internal.tracker.ServletTracker`](https://github.com/ops4j/org.ops4j.pax.web/blob/web-6.0.3/pax-web-extender-whiteboard/src/main/java/org/ops4j/pax/web/extender/whiteboard/internal/tracker/ServletTracker.java), and [`org.ops4j.pax.web.extender.whiteboard`](https://github.com/ops4j/org.ops4j.pax.web/blob/web-6.0.3/pax-web-extender-whiteboard/src/main/java/org/ops4j/pax/web/extender/whiteboard/ExtenderConstants.java).
  - remove servlet-specific properties because the `DelegateServletFilter` adds the filters to every servlet, so these properties we never being used.
- Removes the `ServiceComparator` and `SortedServiceList` classes from the `platform-filter-delegate` module because they are not needed.

#### Who is reviewing it?
@mcalcote @vinamartin @josephthweatt @alexaabrd 
@codice/security @stustison
@clockard
@pklinef
@stustison
- Confirm unit tests and itests succeed.
- Add
  ```
  log4j2.logger.org_codice_ddf_platform_filter_delegate_delegateservletfilter.level = DEBUG
  log4j2.logger.org_codice_ddf_platform_filter_delegate_delegateservletfilter.name = 
  org.codice.ddf.platform.filter.delegate.DelegateServletFilter
  ```
  to `etc/org.ops4j.pax.logging.cfg` before starting DDF. Confirm log statements for initializing `SecurityFilter`s in the `DelegateServletFilter` are correct when DDF is running.
- Confirm manual install succeeds.
[DDF-2863](https://codice.atlassian.net/browse/DDF-2863)
#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:


platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,155 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+            LOGGER.error(
+                    "DelegateServletFilter was initialized with a null filterConfig. Initializing SecurityFilter {} with null ServletContext",
+                    ServicePropertiesUtils.getStringProperty(securityFilterServiceReference,
+                            HttpWhiteboardConstants.HTTP_WHITEBOARD_FILTER_NAME));
+            servletContext = null;
+        } else {
+            servletContext = this.filterConfig.getServletContext();
+        }
+
+        bundleContext.getService(securityFilterServiceReference)
```
clockard: After sleeping on it I see what is going on here. It still feels a little weird to call an init method on a service.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,92 +14,77 @@
...
-                referenceCollection = context.getServiceReferences(Filter.class, null);
-            } catch (InvalidSyntaxException ise) {
-                LOGGER.debug("Should not get this exception as there is no filter being passed.");
-            }
-        }
-
-        if (referenceCollection != null && !referenceCollection.isEmpty()) {
-            LOGGER.debug("Found {} filter(s), now filtering...", referenceCollection.size());
-
+        if (!CollectionUtils.isEmpty(sortedAndInitializedSecurityFilterServiceReferences)) {
```
clockard: Could we just add another check here to see if everything is initialized and do a pass through if not? That way we don't need the notYetInitializedSecurityFilterServiceReferences set and can just put all references directly in the sortedAndInitializedSecurityFilterServiceReferences when adding filters and reduce the complexity of the class.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,155 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
     @Override
     public void init(FilterConfig filterConfig) throws ServletException {
         this.filterConfig = filterConfig;
+        hasBeenInitialized = true;
+
+        // now that the DelegateServletFilter has been initialized with a filterConfig, initialize all SecurityFilters that have not yet been initialized using that filterConfig
+        for (ServiceReference<SecurityFilter> securityFilterServiceReference : notYetInitializedSecurityFilterServiceReferences) {
```
clockard: If we remove the notYetInitializedSecurityFilterServiceReferences we could just run through the sortedAndInitializedSecurityFilterServiceReferences list and intilialize those. If you did that you would probably want to rename the variable too. You could ignore my other comment about the hasBeenInitialized and set that to true at the end of this method and then use that variable in the do filter check above.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,155 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+    public void addSecurityFilter(
+            final ServiceReference<SecurityFilter> securityFilterServiceReference)
+            throws ServletException {
+        if (notYetInitializedSecurityFilterServiceReferences.contains(securityFilterServiceReference)
+                || sortedAndInitializedSecurityFilterServiceReferences.contains(
+                securityFilterServiceReference)) {
+            LOGGER.warn("SecurityFilter has already been added. Not adding again.");
+        } else {
+            // only initialize SecurityFilters after the DelegateServletFilter has been initialized with a filterConfig
+            if (hasBeenInitialized) {
```
clockard: Ignore this comment if you make the other recommended changes.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,92 +14,77 @@
...
-
+        if (!CollectionUtils.isEmpty(sortedAndInitializedSecurityFilterServiceReferences)) {
             ProxyFilterChain chain = new ProxyFilterChain(filterChain);
 
-            LinkedList<ServiceReference<Filter>> sortedFilters =
-                    new LinkedList<ServiceReference<Filter>>(referenceCollection);
-            // natural ordering of service references is to sort by service ranking
-            Collections.sort(sortedFilters);
-
             // because we're inserting these one at a time (inserting at index 0),
```
emmberk: I'm not sure what you mean. The `SecurityFilter`s are inserted into the chain one at a time.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: It's the `init(FilterConfig filterConfig)` method for the `javax.servlet.Filter` interface. We need to initialize the `Filter`s with the init params stored in the service properties.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,155 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+                    filter.getProperty(Constants.SERVICE_RANKING),
+                    ServicePropertiesUtils.getStringProperty(filter,
+                            HttpWhiteboardConstants.HTTP_WHITEBOARD_FILTER_NAME));
+        }
+    }
+
+    public void removeSecurityFilter(
+            final ServiceReference<SecurityFilter> securityFilterServiceReference) {
+        sortedAndInitializedSecurityFilterServiceReferences.remove(securityFilterServiceReference);
+        bundleContext.getService(securityFilterServiceReference)
```
emmberk: :+1: Removed

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,155 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+                ServiceReference<SecurityFilter> securityFilterServiceReference,
+                ServletContext servletContext) {
+            this.securityFilterServiceReference = securityFilterServiceReference;
+            this.servletContext = servletContext;
+            initParams = createInitParams(securityFilterServiceReference);
+        }
+
+        /**
+         * This code to get the filter name from a {@link ServiceReference<Filter>} is copied from {@link org.ops4j.pax.web.extender.whiteboard.internal.tracker.ServletTracker#createWebElement(ServiceReference, javax.servlet.Servlet)}.
+         * See the pax-web Whiteboard documentation and {@link HttpWhiteboardConstants#HTTP_WHITEBOARD_FILTER_NAME} for how to configure {@link Filter} services with a filter name.
```
josephthweatt: Maybe consider wrapping these comments so they don't extend so far to the right

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,155 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+        }
+
+        @Override
+        public Enumeration<String> getInitParameterNames() {
+            return Collections.enumeration(initParams.keySet());
+        }
+
+        /**
+         * This code to create the init params from a {@link ServiceReference<Filter>} is copied from {@link org.ops4j.pax.web.extender.whiteboard.internal.tracker.ServletTracker#createWebElement(ServiceReference, javax.servlet.Servlet)}.
+         * See the pax-web Whiteboard documentation and {@link ExtenderConstants#PROPERTY_INIT_PREFIX} for how to configure {@link Filter} services with init params.
```
josephthweatt: Same comment as above

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: I added ` destroy-method="destroy"` to every `SecurityFilter` blueprint definition instead of calling `destroy` inside of `removeSecurityFilter`.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/SecurityFilter.java
```diff
@@ -0,0 +1,20 @@
...
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.platform.filter;
+
+import javax.servlet.Filter;
+
+public interface SecurityFilter extends Filter {
```
stustison: Also please mark this as experimental.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,92 +14,77 @@
...
+    // used to only init SecurityFilters after the DelegateServletFilter has been initialized with a filterConfig
+    final Set<ServiceReference<SecurityFilter>> notYetInitializedSecurityFilterServiceReferences;
+
+    private Boolean hasBeenInitialized;
+
+    private final BundleContext bundleContext;
+
+    FilterConfig filterConfig;
+
+    public DelegateServletFilter(final BundleContext bundleContext) {
```
stustison: You cannot pass the bundle context in from blueprint and hold onto it like this. These bundle contexts go stale, which is why we were using FrameworkUtil previously. You can use and discard a bundle context within a constructor, but you cannot save the reference.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
mcalcote: What about the line below about them being "in order"?

emmberk: The `TreeSet` will add `SecurityFilter`s in order. The `testDoFilterWithSecurityFiltersInCorrectOrder` tests that this works as expected.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,165 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+        hasBeenInitialized = true;
+    }
+
+    public void addSecurityFilter(
+            final ServiceReference<SecurityFilter> securityFilterServiceReference)
+            throws ServletException {
+        if (hasBeenInitialized) {
+            initAndAddToFilterList(securityFilterServiceReference, filterConfig);
+        } else {
+            sortedSecurityFilterServiceReferences.add(securityFilterServiceReference);
```
clockard: Could move this out of the if/else so you always add to the tree and then only do the init if hasBeenInitialized is true.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +97,165 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+                            HttpWhiteboardConstants.HTTP_WHITEBOARD_FILTER_NAME));
+            servletContext = null;
+        } else {
+            servletContext = this.filterConfig.getServletContext();
+        }
+
+        getContext().getService(securityFilterServiceReference)
+                .init(new SecurityFilterInitFilterConfig(securityFilterServiceReference,
+                        servletContext));
+        sortedSecurityFilterServiceReferences.add(securityFilterServiceReference);
```
clockard: If you make the above change you can remove this line to add to the tree.

emmberk:
I have addressed remaining comments and squashed commits.

emmberk:
build now

platform/platform-filter-delegate/pom.xml
```diff
@@ -91,23 +118,23 @@
                             <rules>
                                 <rule>
                                     <element>BUNDLE</element>
+                                    <!--These ratios are low because the FilterInjector class has low unit test coverage.-->
                                     <limits>
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.32</minimum>
+                                            <minimum>0.99</minimum>
```
stustison: Lets bump these down to 97 or so, to allow for some discrepancies between operating systems.

platform/platform-filter-delegate/pom.xml
```diff
@@ -91,23 +118,23 @@
                             <rules>
                                 <rule>
                                     <element>BUNDLE</element>
+                                    <!--These ratios are low because the FilterInjector class has low unit test coverage.-->
```
stustison: Can probably remove this comment now ;)

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,92 +14,82 @@
...
-                referenceCollection = context.getServiceReferences(Filter.class, null);
-            } catch (InvalidSyntaxException ise) {
-                LOGGER.debug("Should not get this exception as there is no filter being passed.");
-            }
-        }
-
-        if (referenceCollection != null && !referenceCollection.isEmpty()) {
-            LOGGER.debug("Found {} filter(s), now filtering...", referenceCollection.size());
-
+        if (!CollectionUtils.isEmpty(sortedSecurityFilterServiceReferences)) {
```
stustison: After looking at this some more, I believe there is an issue with hanging onto the service references in this way. Just like the bundle contexts, service references can also go stale. I believe that is why we were grabbing them and sorting for every request.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +102,168 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+     * {@link SecurityFilter}'s service properties containing init params and the same
+     * {@link ServletContext} as the {@link DelegateServletFilter}. The {@link FilterConfig} is used
+     * to initialize the {@link SecurityFilter}.
+     * <p>
+     * The logic of this inner class is copied from the
+     * {@link org.ops4j.pax.web.extender.whiteboard} feature.
+     */
+    private class SecurityFilterInitFilterConfig implements FilterConfig {
+
+        private final ServiceReference<SecurityFilter> securityFilterServiceReference;
```
stustison: You'll need to not hang onto this reference. Pull the name and properties off of it as you go through the constructor.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +102,168 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+
+    public void addSecurityFilter(
+            final ServiceReference<SecurityFilter> securityFilterServiceReference)
+            throws ServletException {
+        LOGGER.trace("Adding SecurityFilter ranking={} name={}",
+                securityFilterServiceReference.getProperty(Constants.SERVICE_RANKING),
+                ServicePropertiesUtils.getStringProperty(securityFilterServiceReference,
+                        HttpWhiteboardConstants.HTTP_WHITEBOARD_FILTER_NAME));
+
+        if (filtersHaveBeenInitialized) {
```
coyotesqrl: ❓ Is there a reason that you've chosen only to init the filters here in the case where the `init` method has already been called? Wouldn't it work the same and simplify everything if you just dumped this boolean flag and _always_ called `initFilter` in this method and never looped and called it in the `init` method?

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +102,168 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+        private final ServletContext servletContext;
+
+        private final Map<String, String> initParams;
+
+        SecurityFilterInitFilterConfig(
+                ServiceReference<SecurityFilter> securityFilterServiceReference,
+                ServletContext servletContext) {
+            this.securityFilterServiceReference = securityFilterServiceReference;
+            this.servletContext = servletContext;
+            initParams = createInitParams(securityFilterServiceReference);
```
coyotesqrl: ✏️ I don't see much point in passing the reference into this `createInitParams` method. If you were to make that method static, then it would make sense to pass this on the stack; however, as it's an instance method, it already has access to the state of the class.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +102,168 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+
+            // make all the service parameters available as initParams
+            Map<String, String> initParams = new HashMap<>();
+            for (String key : initParamKeys) {
+                String value = securityFilterServiceReference.getProperty(key) == null ?
+                        "" :
+                        securityFilterServiceReference.getProperty(key)
+                                .toString();
+
+                // if the prefix is null or empty, match is true, otherwise its
```
coyotesqrl: ❓ This comment doesn't appear to match the conditional code below. I don't see how you're matching true on an empty prefix.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/ProxyFilterChain.java
```diff
@@ -14,84 +14,65 @@
...
-     *
-     * @param filters
-     *            The servlet filters to add.
-     */
-    public void addFilters(List<Filter> filters) {
         if (iterator != null) {
             throw new IllegalStateException("Cannot add filter to current running chain.");
         }
 
         if (filters != null) {
```
coyotesqrl: ✏️ `filters` can't be null at this point.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/ProxyFilterChain.java
```diff
@@ -14,84 +14,65 @@
...
 
-    /**
-     * Adds a list of filters to the start of the local filter chain but before
-     * the initialized chain.
-     *
-     * @param filters
-     *            The servlet filters to add.
-     */
-    public void addFilters(List<Filter> filters) {
         if (iterator != null) {
```
coyotesqrl: ✏️ I know it's outside the scope of this change, but there is no concurrent access to this instance of the `ProxyFilterChain` (and if there were, this would not be anywhere near a sufficient test). While it's not specifically improper to check that the iterator is null and no chain is in progress, it does imply some things that aren't exactly true when viewed out of context.

The only value in the test is ensuring that no caller attempts to reuse this instance. That does hold some value, but I would add an inline comment to that effect here and update the javadocs on the class to make that clear.

platform/platform-filter-delegate/pom.xml
emmberk: I accidentally left the ratios high. I lowered them back.

platform/platform-filter-delegate/pom.xml
emmberk: I accidentally left the ratios high. I lowered them back.

platform/platform-filter-delegate/pom.xml
emmberk: I accidentally left the ratios high. I lowered them back.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: The `SecurityFilter`s need a `ServletContext` when they are `init`ed. The `ServletContext` that we want to to use is the one that comes from the `filterConfig` in `DelegateServletFilter#init(FilterConfig filterConfig)`.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: This code and comments are copied from [org.ops4j.pax.web.extender.whiteboard.internal.tracker.ServletTracker#createWebElement](https://github.com/ops4j/org.ops4j.pax.web/blob/web-6.0.3/pax-web-extender-whiteboard/src/main/java/org/ops4j/pax/web/extender/whiteboard/internal/tracker/ServletTracker.java#L76-L224).

coyotesqrl: Just because the lack of clarity isn't ours doesn't mean we can't fix it. 😆

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
coyotesqrl: I think the point was that there is no attempt at "inserting at index 0" as it is now a sorted set.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -112,6 +102,168 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+
+        // now that the DelegateServletFilter has been initialized with a filterConfig, initialize all SecurityFilters using that filterConfig
+        for (ServiceReference<SecurityFilter> securityFilterServiceReference : sortedSecurityFilterServiceReferences) {
+            initFilter(securityFilterServiceReference, filterConfig);
+        }
+
+        filtersHaveBeenInitialized = true;
+    }
+
+    public void addSecurityFilter(
```
coyotesqrl: ❗️ Considering @stustison's comment about keeping around stale references and looking a little more at the implementation here...this method should definitely go away. At _most_ it should remain as a package-local method for testing purposes, but I believe even then you'd be side-stepping the true functionality that should be present.

- The `init` method should revert to what it was previously, just setting the `filterConfig` field.
- This method should be removed and you should mock/spy in your unit tests
- `doFilter` should retrieve the `SecurityFilter`s on-demand and put them into a stack-local sorted set for processing.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: The comment is talking about the index in the ProxyFilterChain#addSecurityFilter method. I have updated the comment to say
```java
            // Insert the SecurityFilters into the chain one at a time (from lowest service ranking
            // to highest service ranking). The SecurityFilter with the highest service-ranking will
            // end up at index 0, which means that the SecurityFilters will run in order of highest
            // to lowest service ranking.
```

emmberk:
I have address all review comments this far except I am still working on finding a solution to not embedding `pax-web-extender-whiteboard` and `pax-web-api`.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,104 +14,276 @@
...
+            LOGGER.warn(
+                    "Unable to remove SecurityFilter {}. Try restarting the system or turning up logging to monitor current SecurityFilters.");
+        }
+    }
+
+    @Nonnull
+    private static String getFilterKey(
+            final ServiceReference<SecurityFilter> securityFilterServiceReference,
+            final BundleContext bundleContext) {
+        return getFilterName(securityFilterServiceReference, bundleContext);
```
emmberk: I'm not sure what is a good unique identifier for `SecurityFilter`s. Any ideas?

emmberk:
I have address all review comments this far.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
stustison: What you're doing seems fine.

emmberk:
I rebased and squashed commits.

emmberk:
build now

emmberk:
build now

emmberk:
build now

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,104 +14,286 @@
...
- *
+ * {@link DelegateServletFilter} is meant to detect any {@link SecurityFilter}s and doFilter through
+ * those before continuing on its {@link FilterChain}. If no {@link SecurityFilter}s are found, it
+ * becomes a pass-through filter.
  */
 public class DelegateServletFilter implements Filter {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DelegateServletFilter.class);
 
+    private CopyOnWriteArraySet<String> keysOfInitializedSecurityFilters;
```
coyotesqrl: ❓ This is going to change infrequently? I recognize that on initialization, it's going to update multiple times for each `SecurityFilter`, but after that period of chaos is over the will be relatively stable?

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,104 +14,286 @@
...
     public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
             FilterChain filterChain) throws IOException, ServletException {
+        TreeSet<ServiceReference<SecurityFilter>> sortedSecurityFilterServiceReferences = null;
+        final BundleContext bundleContext = getContext();
 
-        Collection<ServiceReference<Filter>> referenceCollection = null;
-        BundleContext context = getContext();
-
-        if (null != context) {
+        if (bundleContext != null) {
```
coyotesqrl: ✏️ You can invert this condition to reduce cyclomatic complexity.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,104 +14,286 @@
...
+            bundleContext.getService(securityFilterServiceReference)
+                    .destroy();
+        } else {
+            LOGGER.warn(
+                    "Unable to remove SecurityFilter {}. Try restarting the system or turning up logging to monitor current SecurityFilters.");
+        }
+    }
+
+    @Nonnull
+    private static String getFilterKey(
```
coyotesqrl: ❓ Do you need to keep this method and `getFilterName` since this delegates to that and both are private?

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,104 +14,286 @@
...
+                    .getCanonicalName();
+        } else {
+            return filterNameFromTheServiceProperty;
+        }
+    }
+
+    private static String getStringProperty(ServiceReference<?> serviceReference, String key) {
+        Object value = serviceReference.getProperty(key);
+        if (value != null && !(value instanceof String)) {
+            LOGGER.warn("Service property [" + key + "] value must be a String");
```
coyotesqrl: ❗️ To rephrase Sonarqube...use variable interpolation instead of concatenation for your log message.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -14,104 +14,286 @@
...
+            final String propertyInitPrefix = "init-prefix";
+            String initPrefixProp = getStringProperty(securityFilterServiceReference,
+                    propertyInitPrefix);
+            if (initPrefixProp == null) {
+                final String defaultInitPrefixProp = "init.";
+                initPrefixProp = defaultInitPrefixProp;
+            }
+
+            // make all the service parameters available as initParams
+            Map<String, String> initParams = new HashMap<>();
```
coyotesqrl: ❗️ Concur with Sonarqube.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: Yes, `SecurityFilter`s are only added or removed when specific bundles are redeployed, which is not common. There are only about 5 `SecurityFIlter`s that are there by default.

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
emmberk: I was trying to be clear that the filter name is what it being used as the unique identifier for each `SecurityFilter`. This method could be refactored to use something else for the key, and now it's clear exactly what places in the code want the key (which is used for `keysOfInitializedSecurityFilters`) and what places actually want the filter name.

emmberk:
build now

emmberk:
build now

emmberk:
build now

emmberk:
build now

clockard:
Manually verified. LGTM :+1:

clockard:
build now

emmberk:
build now

emmberk:
build now

emmberk:
I have confirmed that itests succeed for downstream projects.

PR #2158, opened 2017-07-26
DDF-3069 Fix loading issues for catalog-ui-search

albany551:
#### What does this PR do?
Catalog-search-ui has an implicit dependency on the spatial-app, which can cause it to fail during installation if it is initialized first. 

This change declares the spatial-app feature as a dependency of the catalog-ui feature.

#### Who is reviewing it? 
@bdeining 
@brendan-hofmann 

@millerw8
@pklinef

Installation with standard profile should complete successfully - both spatial-app and catalog search ui should be installed.

https://codice.atlassian.net/browse/DDF-3069

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/search-ui-app/src/main/resources/features.xml
+++ b/catalog/ui/search-ui-app/src/main/resources/features.xml
@@ -18,6 +18,7 @@
           xsi:schemaLocation="http://karaf.apache.org/xmlns/features/v1.3.0 http://karaf.apache.org/xmlns/features/v1.3.0">
 
     <feature name="catalog-ui" install="auto" version="${project.version}" description="Catalog UI">
+        <feature prerequisite="true">spatial-app</feature>
         <feature prerequisite="true">camel-http</feature>
         <feature prerequisite="true">search-ui-app</feature>
         <configfile finalname="/etc/org.codice.ddf.catalog.ui.config.config" override="false">


coyotesqrl:
On quick glance, this looks like a big, brute force solution to a problem taken without enough analysis and planning as to what the solution should be. Moving service interfaces - experimental though they may be - around from one place to another in order to resolve a dependency issue strikes me as a sledgehammer approach.

Maybe it's right that the service be moved out of `catalog-ui`; I'm not convinced that dumping it into `spatial` is necessarily the right move.

Was this determined to be the best solution after analyzing the problem or was it just the most expeditious solution?

albany551:
@coyotesqrl 
This PR came from collaborating with @bdeining and @brendan-hofmann on this catalog loading issue and I can't take credit for any ideas here. It's confusing because all the moving around of interfaces is actually unrelated to fixing the underlying problem. 
My assumption was that in the OSGi world, a blueprint could have an optional `<reference>` to an interface without actually having that interface class loaded in the bundle (figuring that the service registry only needs the class path, not the actual class). So I moved the geographic feature-related things I had added to the catalog down into spatial bundles.
That doesn't work of course, because the catalog blueprint still needs that spatial-geocoding-api package to declare the reference. I left this commit in because I agree these things should probably live somewhere besides `catalog-ui`. I can certainly remove it from the PR.
It was only after that where the idea became "just make the dependency between catalog and spatial explicit". I don't have much context around the way ddf and its bundles are organized, so I'm not too familiar with the downsides there. My vague understanding is that some users won't use spatial features at all, but they still certainly need the search ui to function?

coyotesqrl:
Many other users won't use the catalog ui but will use spatial. That definitely helps the argument that the functionality might best be moved out of the ui; I'm just not sure spatial is the right target for it.

Also, as the purpose of this PR is to resolve the dependency issue which is currently impacting master, I think we should limit the change to the minimum and address the moving of services in a follow-up.

That's my take, anyway.

albany551:
Refactorings have been forcibly removed :+1:

bdeining:
build now

adimka:
Manually verified startup of Intrigue and Standard Search 👍  LGTM

PR #2161, opened 2017-07-26
DDF-3144 Fix removeAll command not deleting all metacards when batchSize exceeds catalog page size

mweser:
#### What does this PR do?
Fixes issue with RemoveAll command where not all metacards are deleted when the batchSize is greater than the page size the catalog framework uses. 

#### Who is reviewing it? 
@ahoffer @emanns95 @emmberk @jhunzik  @josephthweatt 

@bdeining
@coyotesqrl

Running a build with itests and unit tests ought to be sufficient to verify correctness.

Catalog Framework maintains an internal page size, but not all commands function with an awareness of this. Therefore, paging has to be accounted for.

[](https://codice.atlassian.net/browse/3144)

#### Checklist:
- [x] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
@@ -77,7 +77,7 @@ public class RemoveAllCommand extends CatalogCommands {
 
     @Argument(name = "Batch size", description =
             "Number of Metacards to delete at a time until completion. Change this argument based on system memory and Catalog limits. "
-                    + "Must be a positive integer.", index = 0, multiValued = false, required = false)
+                    + "Must be a positive integer.\nNOTE: Batch size may not be honored given system constraints.", index = 0, multiValued = false, required = false)
     int batchSize = DEFAULT_BATCH_SIZE;
 
     @Option(name = "-e", required = false, aliases = {
@@ -189,7 +189,8 @@ public class RemoveAllCommand extends CatalogCommands {
             console.flush();
 
             // Break out if there are no more records to delete
-            if (amountDeleted < batchSize || batchSize < 1) {
+            // Prevents additional query to catalog framework
+            if (response.getResults().isEmpty()) {
                 break;
             }
 
--- a/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
+++ b/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
@@ -13,6 +13,7 @@
  */
 package org.codice.ddf.commands.catalog;
 
+import static java.util.stream.Collectors.toList;
 import static org.codice.ddf.commands.catalog.CommandSupport.ERROR_COLOR;
 import static org.hamcrest.Matchers.startsWith;
 import static org.junit.Assert.assertThat;
@@ -23,10 +24,13 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import org.fusesource.jansi.Ansi;
+import org.junit.Before;
 import org.junit.Test;
 
 import ddf.catalog.CatalogFramework;
@@ -43,6 +47,18 @@ import ddf.catalog.operation.QueryResponse;
 
 public class RemoveAllCommandTest extends ConsoleOutputCommon {
 
+    private int batchSize;
+
+    private int numCatalogCalls;
+
+    private boolean forceCommand;
+
+    private CatalogFramework catalogFrameworkMock;
+
+    private QueryResponse queryResponse;
+
+    private DeleteResponse deleteResponse;
+
     static final String DEFAULT_CONSOLE_COLOR = Ansi.ansi()
             .reset()
             .toString();
@@ -51,6 +67,14 @@ public class RemoveAllCommandTest extends ConsoleOutputCommon {
             .fg(ERROR_COLOR)
             .toString();
 
+    @Before
+    public void setUp() throws Exception {
+
+        catalogFrameworkMock = mock(CatalogFramework.class);
+        queryResponse = mock(QueryResponse.class);
+        deleteResponse = mock(DeleteResponse.class);
+    }
+
     /**
      * If it is possible to give bad batch size, this test checks the proper outcome.
      *
@@ -58,6 +82,7 @@ public class RemoveAllCommandTest extends ConsoleOutputCommon {
      */
     @Test
     public void testBadBatchSize() throws Exception {
+
         // given
         RemoveAllCommand command = new RemoveAllCommand();
         command.batchSize = 0;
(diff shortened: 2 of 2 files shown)


mweser:
build now

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
```diff
@@ -110,7 +188,9 @@ public void testExecuteWithSubject() throws Exception {
      */
     @Test
     public void testExecuteWithSubjectWithCache() throws Exception {
-        // given
+
+        int numCFCalls = 1;
+
```
josephthweatt: Does CF mean catalogFramework? Should that be spelled out?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
```diff
@@ -189,7 +189,8 @@ private Object executeRemoveAllFromStore() throws Exception {
             console.flush();
 
             // Break out if there are no more records to delete
-            if (amountDeleted < batchSize || batchSize < 1) {
```
emanns95: Should we still be breaking if we have a batchSize less than 1?

emanns95:
It's not obvious to me why there are such significant coverage declines. Maybe run the tests and check the jacoco reports

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
mweser: I'll check on that.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
mweser: I seem to remember saying something like that on one of your PRs. I wouldn't want to be a hypocrite now :)

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
mweser: I added the `batchsize < 1` back in as it does not break the tests this PR is focusing on. @lessarderic might this be part of a future task with removing the magic numbers? I feel for the time being we'll leave it in and later change if it's deemed unnecessary.

mweser:
build now

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
```diff
@@ -43,6 +45,28 @@
 
 public class RemoveAllCommandTest extends ConsoleOutputCommon {
 
+    int resultsPerCatalogQuery;
```
bdeining: any reason not to make these private?

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
emanns95: I don't feel like this is a magic number, but a logical case. If the batch size is asking for less than one record, there's nothing to do. There are no records to perform operations on in each batch.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
```diff
@@ -189,7 +189,7 @@ private Object executeRemoveAllFromStore() throws Exception {
...
+            if (response.getResults().isEmpty() || batchSize < 1) {
```
emanns95: I looked further up the code and I see we are in a while loop with the condition `response.getResults().size() > 0`. The Response is not modified in the loop, so I don't see this first condition `response.getResults().isEmpty()` ever happening unless I'm missing some piece of the logic.

The older logic here makes a little more sense to me, since if the amount of metacards it deleted is less than the batch size, then there are no more results and there is no point in sending out another query.

And actually I don't think there's any point in having the batchSize condition in here. It should be in the condition for the while loop, since the batch size never changes and it seems like a batchSize of 0 should essentially be a no-op, right? Right now I think it will delete one page of results and that's it. I'm not sure if that's the intended behavior

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
```diff
@@ -122,43 +204,46 @@ protected SolrCacheMBean getCacheProxy() {
...
+    private void populateMockLists() {
+        resultMockList = populateResultList(resultsPerCatalogQuery);
+        metacardMockList = populateMetacardList(resultsPerCatalogQuery);
+    }
 
-            String id = UUID.randomUUID()
-                    .toString();
-            MetacardImpl metacard = new MetacardImpl();
-            metacard.setId(id);
+    private List<Result> populateResultList(int size) {
```
mweser: @coyotesqrl I hope you like this bit of refactoring :)

mweser:
build now

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
```diff
@@ -72,35 +108,79 @@ public void testBadBatchSize() throws Exception {
     }
 
     /**
+     * Tests that if a current batch to be removed is less than the batch size,
+     * but items still remain to be deleted, command will not terminate prematurely.
+     *
+     * Catalog Fra
```
coyotesqrl: ✏️ I believe this line is erroneous.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
```diff
@@ -72,35 +108,79 @@ public void testBadBatchSize() throws Exception {
...
+     * Second:      100
+     * Third:       150
+     * Fourth:      200 (stops at 200)
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testDeleteFewerThanBatchSizeExpectMultipleCatalogCalls() throws Exception {
+
+        resultsPerCatalogQuery = 50;
```
coyotesqrl: ✏️ I would rather we not use such _pretty_ numbers, prone to fencepost errors. I think this test would be more clear if you had numbers that did not evenly divide into one another.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java
```diff
@@ -72,35 +108,77 @@ public void testBadBatchSize() throws Exception {
...
+     * Fourth:      212 (stops at 201)
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testDeleteFewerThanBatchSizeExpectMultipleCatalogCalls() throws Exception {
+
+        resultsPerCatalogQuery = 53;
+        batchSize = 101;
+        numCatalogCalls = 6;
```
coyotesqrl: ✏️ I think you need more comments to explain this test. I fundamentally don't understand it. The `delete` method is invoked six times? Why is it six and not four? Why do you return the `metacardMockList` three times followed by an empty list? Shouldn't it be three times plus another list of 42?

This test is intimately aware of and tied to the internals - which may change in the future - of the way the catalog framework works. It's fragile and not clear.

mweser:
build now

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/RemoveAllCommand.java
coyotesqrl: I agree with @emanns95. Not every use of a number in code is a magic red flag. Sometimes they really do just represent obvious, logical assumptions.

PR #2165, opened 2017-07-27
DDF-3172 bad.files and bad.file.extensions in system.properties is not being respected

johnsliao:
#### What does this PR do?
- Fix the issue where mime type improperly used the original filename and overwrote the sanitized filename.
- Fix a bug where if `files.bad`, `files.bad.extension` in `system.properties` contained any capitalization, to-be sanitized files would not be properly identified
- Minor refactoring of `generateMetacardAndContentItems()`
- Add ability to ignore files (`ignore.files` in `system.properties`)
- Update documentation as needed

#### Who is reviewing it? 
@ahoffer 
@brendan-hofmann 
@emanns95 
@AzGoalie 
@jhunzik 
@mweser 

@coyotesqrl
@stustison

1. Build ddf
2. Add a  dummy filename to `ignore.files` in `etc/system.properties`
3. Stand up ddf
4. Ingest the attached test files
5. `crossdomain.xml` and `thUmbs.db` will be successfully ingested as `file.bin`
6. `bad_file.cgi` will be successfully ingested as `bad_file.bin`
7. Try to ingest a file with that filename added in step 2. It will not be successful

[DDF-3172.TestFiles.zip](https://github.com/codice/ddf/files/1180825/DDF-3172.TestFiles.zip)

DDF-2151
#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
@@ -103,7 +103,7 @@ public class OperationsMetacardSupport {
                                 "Could not copy bytes of content message.  Message was NULL.");
                     }
 
-                    if (!InputValidation.isFileNameClientSideVulnerable(fileName)) {
+                    if (!InputValidation.isFileNameClientSideSafe(fileName)) {
                         throw new IngestException("Ignored filename found.");
                     }
 
@@ -133,7 +133,7 @@ public class OperationsMetacardSupport {
                 String mimeTypeRaw = contentItem.getMimeTypeRawData();
                 mimeTypeRaw = guessMimeType(mimeTypeRaw, fileName, tmpPath);
 
-                if (!InputValidation.isMimeTypeClientSideVulnerable(mimeTypeRaw)) {
+                if (!InputValidation.isMimeTypeClientSideSafe(mimeTypeRaw)) {
                     throw new IngestException("Unsupported mime type.");
                 }
 
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
@@ -103,7 +103,7 @@ public class OperationsMetacardSupport {
                                 "Could not copy bytes of content message.  Message was NULL.");
                     }
 
-                    if (!InputValidation.checkForClientSideVulnerableFileName(fileName)) {
+                    if (!InputValidation.isFileNameClientSideVulnerable(fileName)) {
                         throw new IngestException("Ignored filename found.");
                     }
 
@@ -133,7 +133,7 @@ public class OperationsMetacardSupport {
                 String mimeTypeRaw = contentItem.getMimeTypeRawData();
                 mimeTypeRaw = guessMimeType(mimeTypeRaw, fileName, tmpPath);
 
-                if (!InputValidation.checkForClientSideVulnerableMimeType(mimeTypeRaw)) {
+                if (!InputValidation.isMimeTypeClientSideVulnerable(mimeTypeRaw)) {
                     throw new IngestException("Unsupported mime type.");
                 }
 
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
@@ -94,15 +94,20 @@ public class OperationsMetacardSupport {
         for (ContentItem contentItem : incomingContentItems) {
             try {
                 Path tmpPath = null;
+                String fileName;
                 long size;
                 try (InputStream inputStream = contentItem.getInputStream()) {
+                    fileName = contentItem.getFilename();
                     if (inputStream == null) {
                         throw new IngestException(
                                 "Could not copy bytes of content message.  Message was NULL.");
                     }
 
-                    String sanitizedFilename =
-                            InputValidation.sanitizeFilename(contentItem.getFilename());
+                    if (!InputValidation.checkForClientSideVulnerableFileName(fileName)) {
+                        throw new IngestException("Ignored filename found.");
+                    }
+
+                    String sanitizedFilename = InputValidation.sanitizeFilename(fileName);
                     tmpPath = Files.createTempFile(FilenameUtils.getBaseName(sanitizedFilename),
                             FilenameUtils.getExtension(sanitizedFilename));
                     Files.copy(inputStream, tmpPath, StandardCopyOption.REPLACE_EXISTING);
@@ -126,13 +131,20 @@ public class OperationsMetacardSupport {
                     throw new IngestException("Could not copy bytes of content message.", e);
                 }
                 String mimeTypeRaw = contentItem.getMimeTypeRawData();
-                mimeTypeRaw = guessMimeType(mimeTypeRaw, contentItem.getFilename(), tmpPath);
+                mimeTypeRaw = guessMimeType(mimeTypeRaw, fileName, tmpPath);
 
                 if (!InputValidation.checkForClientSideVulnerableMimeType(mimeTypeRaw)) {
                     throw new IngestException("Unsupported mime type.");
                 }
 
-                String fileName = updateFileExtension(mimeTypeRaw, contentItem.getFilename());
+                // If any sanitization was done, rename file name to sanitized file name.
+                if (!InputValidation.sanitizeFilename(fileName)
+                        .equals(fileName)) {
+                    fileName = InputValidation.sanitizeFilename(fileName);
+                } else {
+                    fileName = updateFileExtension(mimeTypeRaw, fileName);
+                }
+
                 Metacard metacard = metacardFactory.generateMetacard(mimeTypeRaw,
                         contentItem.getId(),
                         fileName,
@@ -140,7 +152,9 @@ public class OperationsMetacardSupport {
                 metacardMap.put(metacard.getId(), metacard);
 
                 ContentItem generatedContentItem = new ContentItemImpl(metacard.getId(),
-                        StringUtils.isNotEmpty(contentItem.getQualifier()) ? contentItem.getQualifier() : "",
+                        StringUtils.isNotEmpty(contentItem.getQualifier()) ?
+                                contentItem.getQualifier() :
+                                "",
                         com.google.common.io.Files.asByteSource(tmpPath.toFile()),
                         mimeTypeRaw,
                         fileName,
--- a/distribution/ddf-common/src/main/resources/etc/system.properties
+++ b/distribution/ddf-common/src/main/resources/etc/system.properties
@@ -122,13 +122,13 @@ jdk.tls.ephemeralDHKeySize=matched
 # Files uploaded with these bad file extensions or names will have their file names sanitized
 # before being saved
 bad.file.extensions=.exe,.jsp,.html,.js,.php,.phtml,.php3,.php4,.php5,.phps,.shtml,.jhtml,.pl,.py,.cgi,.msi,.com,.scr,.gadget,.application,.pif,.hta,.cpl,.msc,.jar,.kar,.bat,.cmd,.vb,.vbs,.vbe,.jse,.ws,.wsf,.wsc,.wsh,.ps1,.ps1xml,.ps2,.ps2xml,.psc1,.psc2,.msh,.msh1,.msh2,.mshxml,.msh1xml,.msh2xml,.scf,.lnk,.inf,.reg,.dll,.vxd,.cpl,.cfg,.config,.crt,.cert,.pem,.jks,.p12,.p7b,.key,.der,.csr,.jsb,.mhtml,.mht,.xhtml,.xht,.tmp
-bad.files=crossdomain.xml,clientaccesspolicy.xml,.htaccess,.htpasswd,hosts,passwd,group,resolv.conf,nfs.conf,ftpd.conf,ntp.conf,web.config,robots.txt,Thumbs.db
+bad.files=crossdomain.xml,clientaccesspolicy.xml,.htaccess,.htpasswd,hosts,passwd,group,resolv.conf,nfs.conf,ftpd.conf,ntp.conf,web.config,robots.txt
 
 # Files uploaded matching these mime types will be excluded from being stored
 bad.mime.types=text/html,text/javascript,text/x-javascript,application/x-shellscript,text/scriptlet,application/x-msdownload,application/x-msmetafile
 
 # Files uploaded matching these file names will be excluded from being stored
-ignore.files=.DS_Store
+ignore.files=.DS_Store,Thumbs.db
 
 #
 # Basic Authentication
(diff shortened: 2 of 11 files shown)


platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/InputValidationTest.java
```diff
@@ -44,14 +46,21 @@
...
 
+    private static final String IGNORE_FILE = ".DS_Store";
+
+    private static final String IGNORE_FILE_CASE_INSENSITIVE = ".dS_sTorE";
+
+    private static final String IGNORE_FILE_KNOWN_GOOD = "valid_ddms_record.xml";
+
     @Before
     public void setup() {
         System.setProperty("bad.files",
```
ahoffer: I recommend using this library to manage system properties for testing. 
http://stefanbirkner.github.io/system-rules/

Using it ensures that the system properties are restored to their original values.

johnsliao: Thanks for the feedback. Added

johnsliao:
Pushed change to use http://stefanbirkner.github.io/system-rules/ instead of `System.setProperty()` in the unit tests

johnsliao:
build now

distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -127,6 +127,8 @@ bad.files=crossdomain.xml,clientaccesspolicy.xml,.htaccess,.htpasswd,hosts,passw
 # Files uploaded matching these mime types will be excluded from being stored
 bad.mime.types=text/html,text/javascript,text/x-javascript,application/x-shellscript,text/scriptlet,application/x-msdownload,application/x-msmetafile
 
+# Files uploaded matching these file names will be excluded from being stored
+ignore.files=.DS_Store
```
mweser: ❓ Were there DS_store files gumming up the works to need this ignore?

distribution/docs/src/main/jdocs/content/_configuring/configuring-global-settings-contents.adoc
```diff
@@ -372,7 +372,8 @@ nfs.conf,
 ftpd.conf,
 ntp.conf,
 web.config,
-robots.txt
+robots.txt,
+Thumbs.db
```
mweser: ❓ What is potentially dangerous about Thumbs.db? Just general curiosity.

platform/util/platform-util/pom.xml
```diff
@@ -70,6 +70,12 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.16.1</version>
+            <scope>test</scope>
+        </dependency>
```
mweser: ❓ Do we use this external dependency elsewhere? I know we're usually pretty leery about this sort of thing.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
```diff
@@ -31,14 +31,22 @@
     private static final String DEFAULT_FILE = "file" + DEFAULT_EXTENSION;
 
     private static final List<String> BAD_FILES = Arrays.asList(System.getProperty("bad.files")
+            .toLowerCase()
```
mweser: ❓ Was there an issue with some file names not being all lowercase? Why does this need to happen?

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
```diff
@@ -97,4 +105,22 @@ public static boolean checkForClientSideVulnerableMimeType(String mimetype) {
...
         return true;
     }
+
+    /**
+     * Checks for filenames that have been disallowed by the system.
+     *
+     * @param filename
+     * @return true if the filename is acceptable
+     */
+    public static boolean checkForClientSideVulnerableFileName(String filename) {
```
mweser: ✏️  Might this method name benefit from following the naming convention of starting with a form of "to be", as in `isClientSideVulnerableFileName`? Might help with readability by future developers where this is used so they know it is a boolean return.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/InputValidationTest.java
```diff
@@ -44,15 +48,30 @@
...
+            new ProvideSystemProperty("bad.file.extensions",
+                    ".exe,.jsp,.html,.js,.php,.phtml,.php3,.php4,.php5,.phps,.shtml,.jhtml,.pl,.py,.cgi,.msi,.com,.scr,.gadget,.application,.pif,.hta,.cpl,.msc,.jar,.kar,.bat,.cmd,.vb,.vbs,.vbe,.jse,.ws,.wsf,.wsc,.wsh,.ps1,.ps1xml,.ps2,.ps2xml,.psc1,.psc2,.msh,.msh1,.msh2,.mshxml,.msh1xml,.msh2xml,.scf,.lnk,.inf,.reg,.dll,.vxd,.cpl,.cfg,.config,.crt,.cert,.pem,.jks,.p12,.p7b,.key,.der,.csr,.jsb,.mhtml,.mht,.xhtml,.xht");
+
+    @ClassRule
+    public static final ProvideSystemProperty badMimeTypesSystemProperties =
+            new ProvideSystemProperty("bad.mime.types",
+                    "text/html,text/javascript,text/x-javascript,application/x-shellscript,text/scriptlet,application/x-msdownload,application/x-msmetafile");
+
+    @ClassRule
+    public static final ProvideSystemProperty ignoreFilesSystemProperties =
```
mweser: ❓ Are these classrules applied under the hood for the rest of the unit tests?

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/InputValidationTest.java
```diff
@@ -95,4 +114,30 @@ public void testCheckForClientSideVulnerableMimeTypeGood() {
         boolean result = InputValidation.checkForClientSideVulnerableMimeType(GOOD_MIME);
         assertTrue(result);
     }
+
+    @Test
+    public void testSanitizeFilenameKnownBadCaseInsensitive() {
+        String sanitizedName = InputValidation.sanitizeFilename(BAD_FILE3);
+        assertThat(sanitizedName, is(DEFAULT_FILE));
```
mweser: ✏️ Perhaps use the Hamcrest `assertThat` method which takes a String reason and the boolean check to help provide better logging in case of failure. Same with the other tests.

distribution/ddf-common/src/main/resources/etc/system.properties
emanns95: Yes right now the CDM ingests those and it's pretty annoying

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OperationsMetacardSupport.java
```diff
@@ -94,15 +94,20 @@ void generateMetacardAndContentItems(List<ContentItem> incomingContentItems,
...
+                    fileName = contentItem.getFilename();
                     if (inputStream == null) {
                         throw new IngestException(
                                 "Could not copy bytes of content message.  Message was NULL.");
                     }
 
-                    String sanitizedFilename =
-                            InputValidation.sanitizeFilename(contentItem.getFilename());
+                    if (!InputValidation.checkForClientSideVulnerableFileName(fileName)) {
+                        throw new IngestException("Ignored filename found.");
```
emanns95: Maybe I'm losing some context in the scope of the diff, but is this failing the ingest here? The documentation says that these names will be sanitized if they're in the `bad.files` or `bad.file.extensions` lists. Only the `bad.mime.types` should be rejected

emanns95: Ah I read a little further. I see ignored files are different from bad.files

distribution/docs/src/main/jdocs/content/_configuring/configuring-global-settings-contents.adoc
bdeining: It's not dangerous, but on Windows machines I believe this gets auto generated and will continuously be created and ingested through the CDM

emanns95:
I think there needs to be an itest that actually monitors a directory with bad files, bad file extensions, bad file mimetypes, and an ignored file and shows that the first two are sanitized and the last two rejected.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
johnsliao: Yes. If there is capitalization in the bad.files/bad.file.extensions, they wouldn't have been matched properly.

e.g. a filename as `thumbs.db` would not be matched with `Thumbs.db` in system.properties.

Although I suppose using regex would be the cleanest solution. It's just that the legacy code identified them by converting filenames to lowercase.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/InputValidationTest.java
johnsliao: Yes, for all the `InputValidationTest.java` unit tests. If I understand your question correctly.

I thought this was helpful:
https://stackoverflow.com/questions/41121778/junit-rule-and-classrule

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
johnsliao: Good point. Thanks.

platform/util/platform-util/pom.xml
johnsliao: I took a look. 

This dependency exists @`/platform/security/sts/security-sts-ldaplogin/pom.xml`. Are there drawbacks on moving it to a common `pom.xml`?

taz77:
build now

distribution/docs/src/main/jdocs/content/_configuring/configuring-global-settings-contents.adoc
stustison: We should probably move Thumbs.db to the "ignored" list.

mweser:
Manual testing verified 100% 👍

johnsliao:
build now

johnsliao:
Pushed up the following changes:

- Added integration tests
- Moved `Thumbs.db` from `bad.files` to `ignore.files`
- Minor function changes

distribution/docs/src/main/jdocs/content/_configuring/configuring-global-settings-contents.adoc
emanns95: Looks like the docs still need to be updated to move this to the ignored list

platform/util/platform-util/pom.xml
emanns95: You could move the version up to platform's `pom` as a property to ensure they stay on the same version (if we want them to). I'm not sure we'd want to move that dependency up but I'm not sure about the pros/cons of that.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/InputValidationTest.java
emanns95: Could also use `assertThat("This is my failure reason", sanitizedName, is(DEFAULT_FILE));` if you still want to use matchers

johnsliao: Good point. Thanks

johnsliao:
build now

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
brendan-hofmann: Regex would be nice because we could ignore all hidden files on *nix systems by default. `\..*`

johnsliao:
build now

johnsliao:
build now

johnsliao:
build now

platform/util/platform-util/pom.xml
mweser: @coyotesqrl or @lessarderic might have better direction on this.

coyotesqrl: As it's only used in scope-test, I see no reason not to put it in the root pom.

johnsliao:
Summary of pushed changes:
- Fixed up other unit tests
- I-test was flakey when ingesting 2 files via rest endpoint. Broke the itest into two separate itests
- Fixed up documentation

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
```diff
@@ -87,7 +95,7 @@ public static String sanitizeFilename(String filename) {
      * @param mimetype
      * @return true if the mime type is acceptable
      */
-    public static boolean checkForClientSideVulnerableMimeType(String mimetype) {
+    public static boolean isMimeTypeClientSideVulnerable(String mimetype) {
```
coyotesqrl: ❗️ This name change makes this method make no sense. The test asks: "is this vulnerable?" and if it is, you return false. If you want to change the name, either make it something like `isMimeTypeClientSideSafe` and keep its current semantics or switch the semantics and update all the calling locations.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
```diff
@@ -97,4 +105,22 @@ public static boolean checkForClientSideVulnerableMimeType(String mimetype) {
...
         return true;
     }
+
+    /**
+     * Checks for filenames that have been disallowed by the system.
+     *
+     * @param filename
+     * @return true if the filename is acceptable
+     */
+    public static boolean isFileNameClientSideVulnerable(String filename) {
```
coyotesqrl: ❗️ The same mismatch between method name and method semantics is present in this new method. Either switch the name or switch the semantics.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
mweser: @coyotesqrl I suggested this name change to capture the spirit of using forms of "to be" for boolean-returning methods, but totally agree that the semantics of it were reversed. Good catch. We could always rename it to something like `isMimeTypeClientSideNotVulnerable` to really confuse people :)

johnsliao:
build now

johnsliao:
build now

johnsliao:
build now

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/InputValidation.java
```diff
@@ -97,4 +105,22 @@ public static boolean checkForClientSideVulnerableMimeType(String mimetype) {
...
     }
+
+    /**
+     * Checks for filenames that have been disallowed by the system.
+     *
+     * @param filename
+     * @return true if the filename is acceptable
+     */
+    public static boolean isFileNameClientSideSafe(String filename) {
+        filename = filename.toLowerCase();
```
coyotesqrl: ✏️ Agree with static analysis recommendation.

johnsliao:
That CI build failure is a result of a bad commit and `build now`. I reset hard back to the Job Success CI build.

johnsliao:
build now

johnsliao:
build now

johnsliao:
build now

johnsliao:
build now

PR #2178, opened 2017-08-01
DDF-3166 Add Unit Test to Ensure catalog:seed Does Not Skip Resources

josephthweatt:
#### What does this PR do?
The current implementation of the catalog:seed command takes a maximum number of resources to download and uses that number to page through the results:
https://github.com/codice/ddf/blob/088e81f76c19291ba3a6e953b25700b08bb217db/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SeedCommand.java#L115-L119
Now that   DDF-2872  has been added to address potential Solr and security problems, the number of results returned is limited to a maximum, which means that the command could skip some resources when it needs to page because of some download failures or already cached resources.

This PR added a unit test to verify that results would not be skipped.

#### Who is reviewing it? 
@mweser 
@vinamartin 
@jhunzik 
@clockard 
@bdeining 

1. Build DDF & unzip the snapshot
2. In the snapshot, modify the last line of **bin/setenv** by adding `-Dcatalog.maxPageSize=100` to `export JAVA_OPTS="`
3. Run DDF
4. Run another DDF, either on your own computer or on another, and make 200 files available for the original DDF to seed. Add the new DDF as a source to the original.
5. Run the `seed` command as follows: `seed -rl 200`
6. If successful, the program should display that 200 downloads have started.

[DDF-3166](https://codice.atlassian.net/browse/)
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
+++ b/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
@@ -30,8 +30,6 @@ import static org.mockito.Mockito.when;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Sets.newHashSet;
 
-import com.google.common.collect.Lists;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
@@ -46,6 +44,8 @@ import org.junit.Before;
 import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.ArgumentMatcher;
+
+import com.google.common.collect.Lists;
 
 import ddf.catalog.CatalogFramework;
 import ddf.catalog.data.Metacard;
@@ -212,17 +212,25 @@ public class SeedCommandTest extends CommandCatalogFrameworkCommon {
         verify(catalogFramework, times(2)).query(any(QueryRequest.class));
     }
 
+    /**
+     * This test will verify that seed does not skip resources when the maximum page
+     * size is less than the specified resource limit. This concern was raised after
+     * query results became limited to a maximum in DDF-2872.
+     *
+     * @throws Exception
+     */
     @Test
     public void testDoesNotSkipResources() throws Exception {
 
-        int maxPageSize = NumberUtils.toInt(System.getProperty("catalog.maxPageSize"), 1000);
+        int maxPageSize = getMaxPageSize();
         int resourceLimit = maxPageSize * 2;
 
-        mockMultipleQueries(maxPageSize, resourceLimit);
+        String expected = resourceLimit + " resource download(s) started.";
+
+        mockMultipleCatalogFrameworkQueries(maxPageSize, resourceLimit);
 
         seedCommand.executeWithSubject();
-        assertThat(consoleOutput.getOutput(),
-                containsString(resourceLimit + " resource download(s) started."));
+        assertThat(consoleOutput.getOutput(), containsString(expected));
     }
 
     private void runCommandAndVerifyResourceRequests(int expectedResourceRequests,
@@ -306,17 +314,22 @@ public class SeedCommandTest extends CommandCatalogFrameworkCommon {
         }
     }
 
-    private void mockMultipleQueries(int pageSize, int resourceLimit) throws Exception {
+    private int getMaxPageSize() {
+        return NumberUtils.toInt(System.getProperty("catalog.maxPageSize"), 1000);
+    }
+
+    private void mockMultipleCatalogFrameworkQueries(int pageSize, int resourceLimit)
+            throws Exception {
 
         mockResourceResponse();
         QueryResponse queryResponseMock = mock(QueryResponse.class);
 
-        // have as many results as the resource limit will allow
+        // Populate list of mock results, sized at resource limit
         List<Result> resultsList = populateResultMockList(resourceLimit);
         List<Result>[] queriedResultsList = mockQueryGetResults(resultsList, pageSize);
 
-        when(queryResponseMock.getResults()).thenReturn(queriedResultsList[0], queriedResultsList);
         when(catalogFramework.query(any())).thenReturn(queryResponseMock);
+        when(queryResponseMock.getResults()).thenReturn(queriedResultsList[0], queriedResultsList);
 
         seedCommand.resourceLimit = resourceLimit;
     }
@@ -338,18 +351,20 @@ public class SeedCommandTest extends CommandCatalogFrameworkCommon {
     }
 
     /**
-     *     Divides results list by the page size. Used for thenReturn statements.
-     *     The array being passed will have doubles for every query after the first,
-     *     for example:
-     *          [ query1, query2, query2, ... ]
-     *     This is because the query result will need to be returned twice for every
-     *     iteration, and the first query is get inserted in the first parameter
-     *     of thenReturn.
-     *     i.e. thenReturn(query1, [ query1, query2, query2 ]).
+     * Divides results list by the page size. Used for thenReturn statements.
+     * The array being passed will have doubles for every query after the first,
+     * for example:
+     * [ query1, query2, query2, ... ]
+     * This is because the query result will need to be returned twice for every
+     * iteration, and the first query is get inserted in the first parameter
+     * of thenReturn.
+     * i.e. thenReturn(query1, [ query1, query2, query2 ]).
      */
     private List<Result>[] mockQueryGetResults(List<Result> results, int pageSize) {
 
-        int queries = (int) Math.ceil(results.size() / pageSize);
+        int queries = results.size() / pageSize;
+        queries += (results.size() % pageSize > 0) ? 1 : 0;
+
         List<List<Result>> pageResults = Lists.partition(results, pageSize);
         List[] pageArray = new List[queries * 2 - 1];
 
--- a/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
+++ b/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
@@ -223,15 +223,31 @@ public class SeedCommandTest extends CommandCatalogFrameworkCommon {
 
         int maxPageSize = NumberUtils.toInt(System.getProperty("catalog.maxPageSize"), 1000);
         int resourceLimit = maxPageSize * 2;
-        int resourceCount = resourceLimit * 2;
 
-        List<Result> resultsList = populateResultMockList(resourceCount);
+        // have as many results as the resource limit will allow
+        List<Result> resultsList = populateResultMockList(resourceLimit);
         List<Result>[] queriedResultsList = queryResults(resultsList, maxPageSize);
 
         QueryResponse responseMock = mock(QueryResponse.class);
 
-        when(responseMock.getResults()).thenReturn(queriedResultsList[0],
-                Arrays.copyOfRange(queriedResultsList, 1, queriedResultsList.length));
+        ResourceResponse resourceResponseMock = mock(ResourceResponse.class);
+
+        Resource resourceMock = mock(Resource.class);
+
+        InputStream inputStreamMock = mock(InputStream.class);
+
+        try {
+            doReturn(-1).when(inputStreamMock)
+                    .read(any(byte[].class), anyInt(), anyInt());
+            Mockito.doNothing().when(inputStreamMock).close();
+            when(resourceMock.getInputStream()).thenReturn(inputStreamMock);
+            when(resourceResponseMock.getResource()).thenReturn(resourceMock);
+
+            when(catalogFramework.getResource(any(), anyString())).thenReturn(resourceResponseMock);
+        } catch (IOException | ResourceNotFoundException | ResourceNotSupportedException e) {
+        }
+
+        when(responseMock.getResults()).thenReturn(queriedResultsList[0], queriedResultsList);
         when(catalogFramework.query(any())).thenReturn(responseMock);
 
         seedCommand.resourceLimit = resourceLimit;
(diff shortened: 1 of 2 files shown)


catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -62,6 +65,10 @@
 
     private CatalogFramework catalogFramework;
 
+    Metacard metacardMock;
+
+    Result resultMock;
```
emmberk: ✏️ These should probably be `private`.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -205,6 +212,19 @@ public void testDoesNotDownloadCachedResource() throws Exception {
         verify(catalogFramework, times(2)).query(any(QueryRequest.class));
     }
 
+    @Test
+    public void testDoesNotSkipResources() throws Exception {
```
mweser: ✏️ Perhaps some JavaDoc comments above this test briefly explaining what is being tested and what the inputs mean.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -205,6 +212,19 @@ public void testDoesNotDownloadCachedResource() throws Exception {
...
+
+        int maxPageSize = NumberUtils.toInt(System.getProperty("catalog.maxPageSize"), 1000);
```
mweser: ✏️ Using the single layer of abstraction principle, perhaps pull the `NumberUtils....` section into a separate method to better describe how/why this maxPageSize is being set to what it's being set to.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -205,6 +212,19 @@ public void testDoesNotDownloadCachedResource() throws Exception {
...
+        int resourceLimit = maxPageSize * 2;
```
mweser: ❓ Why is this limit set to double the maxPageSize? Could this be any value, or is 2 special? Something to include in the JavaDoc to explain the methodology of this test.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -205,6 +212,19 @@ public void testDoesNotDownloadCachedResource() throws Exception {
...
+
+        mockMultipleQueries(maxPageSize, resourceLimit);
```
mweser: ✏️ A more descriptive method name would be beneficial here for future developers curious about how this is functioning.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -205,6 +212,19 @@ public void testDoesNotDownloadCachedResource() throws Exception {
...
+
+        seedCommand.executeWithSubject();
+        assertThat(consoleOutput.getOutput(),
+                containsString(resourceLimit + " resource download(s) started."));
```
mweser: ✏️ Often it helps to place an `expected` string at the start of the unit test and avoid constructing a string inside the assert statement (construct to its own local variable in the line above it might be a bit clearer).

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -285,4 +305,61 @@ public boolean matches(Object o) {
             return test.test((QueryRequest) o);
         }
     }
+
+    private void mockMultipleQueries(int pageSize, int resourceLimit) throws Exception {
+
+        mockResourceResponse();
+        QueryResponse queryResponseMock = mock(QueryResponse.class);
+
+        // have as many results as the resource limit will allow
```
mweser: ✏️ Perhaps clarify this comment a bit to something more like "Populate list of mock results, sized at max resource limit" to help readers know the purpose of the lists.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -285,4 +305,61 @@ public boolean matches(Object o) {
...
+     *     for example:
+     *          [ query1, query2, query2, ... ]
+     *     This is because the query result will need to be returned twice for every
+     *     iteration, and the first query is get inserted in the first parameter
+     *     of thenReturn.
+     *     i.e. thenReturn(query1, [ query1, query2, query2 ]).
+     */
+    private List<Result>[] mockQueryGetResults(List<Result> results, int pageSize) {
+
+        int queries = (int) Math.ceil(results.size() / pageSize);
```
rzwiefel: Dividing two integers should result in an integer. (you can try it here online for quick verification http://www.javarepl.com/term.html) Thus, Math.ceil is having no effect here as the result is already `floor`ed. So both the ceil and cast to int should be able to be removed, though you will have to ensure your logic works with the floor or account for the +1 somehow.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
josephthweatt: The resource limit should be higher than the max page size to prove the test is valid. So it doesn't necessarily need to be double, but I felt it was a simple way to convey that. The JavaDoc I'm writing will explain this

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -285,4 +305,61 @@ public boolean matches(Object o) {
...
+        List<List<Result>> pageResults = Lists.partition(results, pageSize);
+        List[] pageArray = new List[queries * 2 - 1];
+
+        pageArray[0] = pageResults.get(0);
+
+        for (int i = 1, j = 1; j < pageResults.size(); i += 2, j++) {
+            pageArray[i] = pageResults.get(j);
+            pageArray[i + 1] = pageResults.get(j);
+        }
```
rzwiefel: You might be able to make copies and put into arrays like this if you wish
```List<List<Result>> pageResults = new ArrayList<>();
        List[] lists = pageResults.stream()
                .sequential()
                .map(resultList -> Collections.nCopies(2, resultList))
                .flatMap(Collection::stream)
                .collect(Collectors.toList())
                .toArray(new List[0]);```

Comment (josephthweatt) [inline catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java]: I would prefer to do that, but would this affect the structure of the array I am trying to return? The first query only needs to be in the array once, and I'm not sure if that stream solution would cover that

Comment (rzwiefel) [inline catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java]: You could do something such as 
```java
List[] lists = Stream.concat(pageResults.get(0),
                pageResults.stream()
                .sequential()
                .skip(1)
                .map(resultList -> Collections.nCopies(2, resultList))
                .flatMap(Collection::stream)
                .collect(Collectors.toList())
                .toArray(new List[0]);
```

rzwiefel: Though at this point I suppose its starting to get uglier and longer, so it may not be the best option. If you can find a way to succinctly express it then you may, but the final choice is up to you.

josephthweatt: I think for the moment I will keep the traditional for loop, just because the loop does not seem overly complicated at the moment.

bdeining: Just curious : what is the benefit of using a stream instead of a for-loop here?

rzwiefel: There may not be one 😄  I was just highlighting the option so it was known.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -205,6 +212,27 @@ public void testDoesNotDownloadCachedResource() throws Exception {
...
+
+        int maxPageSize = getMaxPageSize();
+        int resourceLimit = maxPageSize * 2;
+
+        String expected = resourceLimit + " resource download(s) started.";
+
+        mockMultipleCatalogFrameworkQueries(maxPageSize, resourceLimit);
+
+        seedCommand.executeWithSubject();
+        assertThat(consoleOutput.getOutput(), containsString(expected));
```
bdeining: Is there a way to write this test without asserting on console output?  In general, I think this should be avoided because of its brittleness.

josephthweatt: @bdeining  I can't think of any other way at the moment. **seedCommand.executeWithSubject()** uses a local int to count the amount of successful downloads. I would like to check that int directly, but I couldn't without changing the scope

bdeining: Ok, just curious.  We have this type of test in other places anyway

bdeining:
build now

clockard:
build now

josephthweatt:
build now

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java
```diff
@@ -285,4 +313,64 @@ public boolean matches(Object o) {
...
+
+        mockResourceResponse();
+        QueryResponse queryResponseMock = mock(QueryResponse.class);
+
+        // Populate list of mock results, sized at resource limit
+        List<Result> resultsList = populateResultMockList(resourceLimit);
+        List<Result>[] queriedResultsList = mockQueryGetResults(resultsList, pageSize);
+
+        when(catalogFramework.query(any())).thenReturn(queryResponseMock);
+        when(queryResponseMock.getResults()).thenReturn(queriedResultsList[0], queriedResultsList);
```
coyotesqrl: ✏️ I believe the second value in `thenReturn` should be `queriedResultsList[1]`. Otherwise, the mock behavior would be equivalent to:
```java
when(queryResponseMock.getResults())
    .thenReturn(queriedResultsList[0])
    .thenReturn(queriedResultsList[0])
    .thenReturn(queriedResultsList[1]);
```

josephthweatt: The purpose of `mockQueryGetResults` is to make an array that provides `thenReturn` with two queries. The reason being that `queryResponseMock.getResults()` is called twice per iteration, and so it will need to get any given query twice. The desired affect is
```when(queryResponseMock.getResults())
    .thenReturn(queriedResultsList[0])
    .thenReturn(queriedResultsList[0])
    .thenReturn(queriedResultsList[1])
    .thenReturn(queriedResultsList[1]);```

Comment (coyotesqrl) [inline catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java]: Oh. That's very non-obvious _here_, despite the javadoc _there_.
But of course the semantics on `thenReturn` don't allow you to just pass in a list so I guess I see why you'd do it that way. I don't love it, to be honest.

Comment (mweser) [inline catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java]: @josephthweatt This is something that Richard _did_ like along the lines of internal query calls being abstracted out to be more readable: https://github.com/codice/ddf/blob/b132cafeb6b770730c0c243dcd988f57b0580b36/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java#L123

😄

Comment (mweser) [inline catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java]: @josephthweatt On this line, you can see the comment that indicates a similar idea, where three calls were needed per query: https://github.com/codice/ddf/blob/b132cafeb6b770730c0c243dcd988f57b0580b36/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java#L213

File: catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/SeedCommandTest.java [hunk 12]
```diff
@@ -38,12 +38,15 @@
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
+import org.apache.commons.lang3.math.NumberUtils;
```
mweser: ❗️ Delete as this is unused: causing Checkstyle error when running full build. Probably a residual from when you were using a ceiling function.

mweser:
Full swarm build passes once NumberUtils unused import is removed.

clockard:
build now

PR #2185, opened 2017-08-02
DDF-3191 Intrigue incorrectly adds wildcards to all queries by default

josephthweatt:
#### What does this PR do?
The intrigue UI always adds a wildcard contextual parameter. There does not appear to be any way to perform a strictly temporal or location-based search.

This PR makes it so that the wildcard is only in the query when no other specifications have been given.

#### Who is reviewing it? 
@djblue 
@emmberk 
@rzwiefel 
@vinamartin 
@andrewkfiedler 
@clockard
@coyotesqrl 
1. Build & run DDF. Install the catalog commands through admin.
2. Go to intrigue and create the following searches:

    - An empty search with no parameters specified. When you save this search, you should see the text field with a wildcard (*) in its place
    - A temporal search. Only specify the time range of the search. Here, the text field should be blank.
    - A type search. Only specify the type of information you are expecting. Again, the text field should be blank.

[DDF-3191](https://codice.atlassian.net/browse/)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
@@ -73,7 +73,7 @@ define([
         var propertyValueMap = {};
         var downConversion = false;
 
-        if (filter.filters && isAnyDate(filter)) {
+        if (filter.filters !== null && isAnyDate(filter)) {
            propertyValueMap['anyDate'] = propertyValueMap['anyDate'] || [];
            if (propertyValueMap['anyDate'].filter(function(existingFilter){
                    return existingFilter.type === filter.filters[0].type;
@@ -82,7 +82,7 @@ define([
            }
         }
 
-        if (filter.filters){
+        if (filter.filters !== null){
             filter.filters.forEach(function(filter){
                if (filter.filters === null){
                    propertyValueMap[CQLUtils.getProperty(filter)] = propertyValueMap[CQLUtils.getProperty(filter)] || [];
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
@@ -84,7 +84,7 @@ define([
 
         if (filter.filters){
             filter.filters.forEach(function(filter){
-               if (!filter.filters){
+               if (filter.filters === null){
                    propertyValueMap[CQLUtils.getProperty(filter)] = propertyValueMap[CQLUtils.getProperty(filter)] || [];
                    if (propertyValueMap[CQLUtils.getProperty(filter)].filter(function(existingFilter){
                            return existingFilter.type === filter.type;
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
@@ -73,7 +73,7 @@ define([
         var propertyValueMap = {};
         var downConversion = false;
 
-        if (filter.filters !== null && isAnyDate(filter)) {
+        if (filter.filters && isAnyDate(filter)) {
            propertyValueMap['anyDate'] = propertyValueMap['anyDate'] || [];
            if (propertyValueMap['anyDate'].filter(function(existingFilter){
                    return existingFilter.type === filter.filters[0].type;
@@ -82,9 +82,9 @@ define([
            }
         }
 
-        if (filter.filters !== null){
+        if (filter.filters){
             filter.filters.forEach(function(filter){
-               if (filter.filters === null){
+               if (!filter.filters){
                    propertyValueMap[CQLUtils.getProperty(filter)] = propertyValueMap[CQLUtils.getProperty(filter)] || [];
                    if (propertyValueMap[CQLUtils.getProperty(filter)].filter(function(existingFilter){
                            return existingFilter.type === filter.type;
@@ -524,12 +524,9 @@ define([
             }
 
             var text = this.basicText.currentView.model.getValue()[0];
+            text = text === "" ? '*' : text;
 
-            if (filters.length == 0) {
-                text = '*'
-            }
-
-            if (text !== "") {
+            if (filters.length == 0 || text !== '*') {
                 var matchCase = this.basicTextMatch.currentView.model.getValue()[0];
                 filters.unshift(CQLUtils.generateFilter(matchCase, 'anyText', text));
             }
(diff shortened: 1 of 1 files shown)


vinamartin:
@djblue should take a look at this as well

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
```diff
@@ -72,6 +72,16 @@ define([
     function translateFilterToBasicMap(filter){
         var propertyValueMap = {};
         var downConversion = false;
+
+        if (filter.filters && isAnyDate(filter)) {
```
rzwiefel: What is filter.filters? if it's not a Boolean, I would not recommend using implicit truthiness for a conditional

josephthweatt: `filter.filters` is an array of filters within some (but not all filters). That expression evaluates to false if `filter.filters` is null and is true if it is anything else. If you look through the rest of the code, you'll see this kind of statement getting used a lot. I think this is an acceptable JS format, but would you know of a better way to write this?

rzwiefel: You should try if at all possible to be explicit and avoid coercion because it's a) harder to read and understand b) more likely to break or do something you don't want it to do (because of coercion)
If you are wishing to check that it is null you should do `filter.filters === null`

clockard:
Ran manual tests successfully. 👍

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
```diff
@@ -72,9 +72,19 @@ define([
     function translateFilterToBasicMap(filter){
         var propertyValueMap = {};
         var downConversion = false;
-        if (filter.filters){
+
+        if (filter.filters !== null && isAnyDate(filter)) {
```
emmberk: ✏️ Could this `if (isAnyDate(filter))` be added in the body of `if (filter.filters !== null)` on line 85 instead?

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
```diff
@@ -518,6 +523,17 @@ define([
                 filters.push(typeFilter)
             }
 
+            var text = this.basicText.currentView.model.getValue()[0];
+
+            if (filters.length == 0) {
+                text = '*'
```
rzwiefel: If there are no filters yet but someone had typed something into the search box (eg, `foo`) wouldn't this logic see that there were no filters yet and then overwrite the actual search term with '*'?

andrewkfiedler:
I'll try and take a look over the weekend.

vinamartin:
I've just pulled this down and tested it and I'm seeing some issues I am currently debugging through. I will upload a patch to this issue once I've figured it out.

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
vinamartin: ❗️ `filter.filters === null` will not actually achieve the desired effect here, it's not that we're checking that filters is null, we're trying to see if filters _exists_, which is what that was doing previously. 
Seems like we should be checking if it is undefined or @djblue suggested something like : `Array.isArray(filter.filters) && filter.length > 0`

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
vinamartin: Yeah it totally does. I'll reorder this to not do that.

vinamartin:
Here's a link to a patch I made with IntelliJ (I find using IntelliJ's "Apply Patch" the best for these):
https://gist.github.com/vinamartin/66a0ec78ce60ad74701a5235526a1ed1

I've reverted back the `filter.filters` null checks for the time being but I'd like to see more feedback on this. My comment up above got squashed but the reason the null check won't work is because it's not that we're checking that filters is null, we're trying to see if filters _exists_, which is what that was doing previously. Seems like we should be checking if it is undefined or @djblue suggested something like : `Array.isArray(filter.filters) && filter.length > 0`. I'm still working through what would be the ideal solution but for now I've reset it to the thing that I know works.

I've changed the check in `constructFilter` to look like this : 
```       
            var text = this.basicText.currentView.model.getValue()[0];
            text = text === "" ? '*' : text;

            if (filters.length == 0 || text !== '*') {
                var matchCase = this.basicTextMatch.currentView.model.getValue()[0];
                filters.unshift(CQLUtils.generateFilter(matchCase, 'anyText', text));
            }
```
In other words, if the text is blank make it wildcard or if it exists grab the text. Then if there are no filters OR the text is not wildcard then add the text filter. 

Finally, I’d like to propose a different set of test steps here since there actually 4 cases:

Create 4 queries:
1. No text is added to the text box and no other query options are chosen, verify that wildcard is added after clicking save
2. Text is added to the text box and no other query options are chosen, verify that the custom text is saved
3. No text is added to the text box and a temporal filter is added, verify the text is blank but the temporal part of the query is saved
4. Text is added to the text box and a temporal filter is added, verify the custom text and the temporal part of the query is saved

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
```diff
@@ -72,9 +72,19 @@ define([
...
+        if (filter.filters !== null && isAnyDate(filter)) {
+           propertyValueMap['anyDate'] = propertyValueMap['anyDate'] || [];
+           if (propertyValueMap['anyDate'].filter(function(existingFilter){
+                   return existingFilter.type === filter.filters[0].type;
+               }).length === 0) {
+               propertyValueMap['anyDate'].push(filter.filters[0]);
+           }
+        }
+
+        if (filter.filters !== null){
```
coyotesqrl: ✏️ Your test on line 76 should already have ensured this.

catalog/ui/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.js
```diff
@@ -518,6 +523,17 @@ define([
                 filters.push(typeFilter)
             }
 
+            var text = this.basicText.currentView.model.getValue()[0];
+
+            if (filters.length == 0) {
```
rzwiefel: ❗️ This should likely be `===`

vinamartin:
build now

clockard:
Re-verified manual tests and performed additional requested tests. LGTM 👍

PR #2190, opened 2017-08-03
DDF-2863 Added optional flag to the SecurityFilters list in the DelegateServletFilter

emmberk:
#### What does this PR do?
This PR fixes the issue that restarting DDF caused some servlets to be missing the `SecurityFilter`s added in https://github.com/codice/ddf/commit/8c330bab9cf7c5e247b020beba115b60b6f66129.
#### Who is reviewing it? 
@mcalcote @Lambeaux @vinamartin 
@codice/security 
@clockard
@coyotesqrl

1. Install DDF. You shouldn't see any log messages like, 
    ```
    Bundle platform-filter-delegate/2.11.0.SNAPSHOT is waiting for dependencies...
    ```
    or
    ```
    Could not inject DelegateServletFilter into xxxxx because the servlet was already initialized.
    ```
2. Restart DDF. Again, you shouldn't see any of those log messages.
3. Confirm that ingesting from Intrigue works.
[DDF-2863](https://codice.atlassian.net/browse/DDF-2863)
#### Review Comment Legend:

### DIFF
--- a/platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
+++ b/platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
@@ -158,17 +158,19 @@ public class DelegateServletFilter implements Filter {
 
     public void removeSecurityFilter(
             final ServiceReference<SecurityFilter> securityFilterServiceReference) {
-        final BundleContext bundleContext = getContext();
+        if (securityFilterServiceReference != null) {
+            final BundleContext bundleContext = getContext();
 
-        if (bundleContext != null) {
-            // unmark the SecurityFilter as initialized so that it can be re-initialized if the SecurityFilter is registered again
-            keysOfInitializedSecurityFilters.remove(getFilterKey(securityFilterServiceReference,
-                    bundleContext));
-            bundleContext.getService(securityFilterServiceReference)
-                    .destroy();
-        } else {
-            LOGGER.warn(
-                    "Unable to remove SecurityFilter {}. Try restarting the system or turning up logging to monitor current SecurityFilters.");
+            if (bundleContext != null) {
+                // unmark the SecurityFilter as initialized so that it can be re-initialized if the SecurityFilter is registered again
+                keysOfInitializedSecurityFilters.remove(getFilterKey(securityFilterServiceReference,
+                        bundleContext));
+                bundleContext.getService(securityFilterServiceReference)
+                        .destroy();
+            } else {
+                LOGGER.warn(
+                        "Unable to remove SecurityFilter. Try restarting the system or turning up logging to monitor current SecurityFilters.");
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


platform/platform-filter-delegate/src/test/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilterTest.java
```diff
@@ -329,6 +327,11 @@ public void testRemoveSecurityFilter() throws IOException, ServletException {
     }
 
     @Test
+    public void testRemoveNullSecurityFilter() throws IOException, ServletException {
+        delegateServletFilter.removeSecurityFilter(null);
```
bdeining: I feel like something should be asserted in this test

platform/platform-filter-delegate/src/main/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilter.java
```diff
@@ -158,17 +158,19 @@ private void initializeSecurityFilter(BundleContext bundleContext,
...
+
+            if (bundleContext != null) {
+                // unmark the SecurityFilter as initialized so that it can be re-initialized if the SecurityFilter is registered again
+                keysOfInitializedSecurityFilters.remove(getFilterKey(securityFilterServiceReference,
+                        bundleContext));
+                bundleContext.getService(securityFilterServiceReference)
+                        .destroy();
+            } else {
+                LOGGER.warn(
+                        "Unable to remove SecurityFilter {}. Try restarting the system or turning up logging to monitor current SecurityFilters.");
```
bdeining: I think this is missing a parameter

platform/platform-filter-delegate/src/test/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilterTest.java
emmberk: I just want to assert that no exception is thrown. Is there a way to be explicit?

bdeining: Could you verify that `keysOfInitializedSecurityFilters.remove()` or `bundleContext.getService()` isn't called in the null case?

adimka: In Spock there is! 🖖 
As far as I'm aware, the test will fail if an exception is thrown, but there's not an explicit way to specify this in JUnit.

emmberk: I don't really care if those methods are called or not, as long as no exception is thrown.

bdeining: If you verify one of those methods isn't called, then it would show that the added logic is working as intended while adding an assertion to the unit test.  My comment is nit picky though

emmberk:
build now

clockard:
build now

clockard:
Manually tested and verified. LGTM 👍

clockard:
build now

platform/platform-filter-delegate/src/test/java/org/codice/ddf/platform/filter/delegate/DelegateServletFilterTest.java
coyotesqrl: In Junit tests, I usually just put a line comment at the end where my assertion would go had I been writing a Spock test.

PR #2193, opened 2017-08-04
DDF 3199 Update RegistryStore to ensure availability only when connected to a registry

clockard:
#### What does this PR do?
Update RegistryStore to ensure availability only when connected to a registry.
Also updated SourcePollerRunner.unbind() method to protect against NPEs.

#### Who is reviewing it? 
@mcalcote @emmberk @vinamartin 

@bdeining
@figliold
@shaundmorris

Install two DDF and install the registry app. For DDF A create a secondary node and then delete the identity node using the command `catalog:remove -p --cql "title like 'DDF-A'"` where DDF-A is the name of DDF A's identity node. On DDF B create a remote registry connection to DDF A and verify that the remote registry never shows up as available. Using bundle start/stop from the console, stop and restart the bundle registry-federation-admin-service-impl on DDF A. This should recreate the identity node. DDF B remote registry connection should show available within 5 minutes.
[DDF-3199](https://codice.atlassian.net/browse/DDF-3199)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
@@ -159,7 +159,10 @@ public class SourcePollerRunner implements Runnable {
         LOGGER.debug("Unbinding source [{}]", source);
         if (source != null) {
             sources.remove(source);
-            sourceStatusThreadLocks.remove(cachedSources.remove(getSourceKey(source)));
+            CachedSource cachedSource = cachedSources.remove(getSourceKey(source));
+            if (cachedSource != null) {
+                sourceStatusThreadLocks.remove(cachedSource);
+            }
         }
     }
 
--- a/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
+++ b/catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
@@ -28,6 +28,7 @@ import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang.StringUtils;
 import org.codice.ddf.cxf.SecureCxfClientFactory;
 import org.codice.ddf.parser.ParserException;
 import org.codice.ddf.registry.api.internal.RegistryStore;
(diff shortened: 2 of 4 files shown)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -159,7 +159,14 @@ public void unbind(Source source) {
         LOGGER.debug("Unbinding source [{}]", source);
         if (source != null) {
             sources.remove(source);
-            sourceStatusThreadLocks.remove(cachedSources.remove(getSourceKey(source)));
+            SourceKey key = getSourceKey(source);
+            CachedSource cachedSource = null;
```
shaundmorris: ❓  this is nice is it worth logging at trace level if the code was unable to unbind a Source?

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
```diff
@@ -473,4 +460,36 @@ public String getRegistryId() {
...
+            newAvailability = (getCapabilities() != null);
+            if (oldAvailability != newAvailability) {
+                // If the source becomes available, configure it.
+                if (newAvailability) {
+                    configureCswSource();
+                    try {
+                        // Make sure the endpoint supports registry operations
+                        newAvailability = registryInfoQuery();
+                        if(!newAvailability){
+                            LOGGER.info("Unable to retrieve identity node of remote registry {}",
```
shaundmorris: ❓  since these are at `info` level, how often do you think these log statements will be hit at runtime? Is there something an administrator can do to help resolve this?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
```diff
@@ -159,7 +159,14 @@ public void unbind(Source source) {
...
+            if(key != null){
```
shaundmorris: ✏️  this always evaluates to `true` since getSourceKey() returns a new SourceKey. You could remove this condition.

shaundmorris:
❓ should there be new unit tests to hit the availability threads in`RegistryAvailabilityCommand`

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
clockard: Good catch.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/SourcePollerRunner.java
clockard: I don't know that there is much value in logging here. If there is a problem here then it has already been removed.

catalog/spatial/registry/registry-api-impl/src/main/java/org/codice/ddf/registry/api/impl/RegistryStoreImpl.java
clockard: They should hopefully never happen but if they do it would be very infrequent. I will update the message as well to direct an admin to check configurations.

clockard:
build now

kcrowe01:
Manually verified ✅

PR #2195, opened 2017-08-07
DDF-3182 Added paging to search endpoint and config for max page size

mcalcote:
#### What does this PR do?
The search endpoint has been changed to enforce its own maximum page size, the user is able to configure this value. It has also been updated to make sure it properly pages if the number of results returned by the catalog framework doesn't match its own maximum page size.

#### Who is reviewing it? 
@mweser @emmberk @josephthweatt @vinamartin @alexaabrd 
@clockard @lessarderic 

Install a DDF, then configure the Standard Search Endpoint max results to be lower than the number of results in the catalog. Confirm the configured max is returned using the standard search ui.
Configure the max standard search results to be higher than the catalog.maxPageSize, but both lower than the number of products in the catalog. Confirm the max search results are returned, which would require paging.

[DDF-3182](https://codice.atlassian.net/browse/DDF-3182)

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
+++ b/catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
@@ -39,7 +39,6 @@ import org.opengis.filter.sort.SortOrder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.Function;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 
@@ -52,6 +51,7 @@ import ddf.catalog.operation.Query;
 import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.QueryResponse;
 import ddf.catalog.operation.impl.ProcessingDetailsImpl;
+import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
 import ddf.catalog.operation.impl.QueryResponseImpl;
 import ddf.catalog.source.SourceUnavailableException;
@@ -190,16 +190,12 @@ public abstract class QueryRunnable implements Runnable {
                 .getPageSize() > 0 ?
                 request.getQuery()
                         .getPageSize() :
-                Integer.MAX_VALUE;
+                searchController.getMaxQueryResults();
     }
 
     protected void addResults(Collection<Result> responseResults) {
-        results.putAll(Maps.uniqueIndex(responseResults, new Function<Result, String>() {
-            @Override
-            public String apply(Result result) {
-                return getResultKey(result.getMetacard());
-            }
-        }));
+        results.putAll(Maps.uniqueIndex(responseResults,
+                result -> getResultKey(result != null ? result.getMetacard() : null)));
     }
 
     protected String getResultKey(Metacard metacard) {
@@ -257,6 +253,7 @@ public abstract class QueryRunnable implements Runnable {
                 LOGGER.debug("Sending query: {}", query);
                 response = searchController.getFramework()
                         .query(request);
+                collectResponses(request, response);
             }
         } catch (UnsupportedQueryException | FederationException e) {
             LOGGER.info("Error executing query. {}. Set log level to DEBUG for more information",
@@ -287,6 +284,32 @@ public abstract class QueryRunnable implements Runnable {
                 .put("elapsed", estimatedTime);
 
         return response;
+    }
+
+    private void collectResponses(QueryRequest request, QueryResponse response)
+            throws UnsupportedQueryException, SourceUnavailableException, FederationException {
+        int resultSize = response.getResults()
+                .size();
+        int adjustedStartIndex = request.getQuery()
+                .getStartIndex();
+        while (resultSize != 0 && resultSize < request.getQuery()
+                .getPageSize()) {
+            adjustedStartIndex += resultSize;
+            QueryImpl queryCopy = new QueryImpl(request.getQuery());
+            queryCopy.setStartIndex(adjustedStartIndex);
+            QueryRequestImpl requestCopy = new QueryRequestImpl(queryCopy);
+            requestCopy.setProperties(request.getProperties());
+            QueryResponse singleResponse = searchController.getFramework()
+                    .query(requestCopy);
+            response.getResults()
+                    .addAll(singleResponse.getResults());
+            response.getProcessingDetails()
+                    .addAll(singleResponse.getProcessingDetails());
+            response.getProcessingErrors()
+                    .addAll(singleResponse.getProcessingErrors());
+            resultSize = singleResponse.getResults()
+                    .size();
+        }
     }
 
     private QueryResponse getEmptyResponse(String sourceId) {
--- a/catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
+++ b/catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
@@ -14,6 +14,7 @@
 package org.codice.ddf.ui.searchui.query.service;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -62,7 +63,7 @@ import ddf.security.Subject;
 @Service
 public class SearchService {
 
-    public static final String LOCAL_SOURCE = "local";
+    private static final String LOCAL_SOURCE = "local";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SearchService.class);
 
@@ -101,10 +102,8 @@ public class SearchService {
     /**
      * Creates a new SearchService
      *
-     * @param filterBuilder
-     *            - FilterBuilder to use for queries
-     * @param searchController
-     *            - SearchController to handle async queries
+     * @param filterBuilder    FilterBuilder to use for queries
+     * @param searchController SearchController to handle async queries
      */
     public SearchService(FilterBuilder filterBuilder, SearchController searchController) {
         this.filterBuilder = filterBuilder;
@@ -114,10 +113,8 @@ public class SearchService {
     /**
      * Service method called by Cometd when something arrives on the service channel
      *
-     * @param remote
-     *            - Client session
-     * @param message
-     *            - JSON message
+     * @param remote  Client session
+     * @param message JSON message
      */
     @Listener("/service/query")
     public void processQuery(final ServerSession remote, Message message) {
@@ -128,11 +125,7 @@ public class SearchService {
 
         if (queryMessage != null && queryMessage.containsKey(Search.ID)) {
             bayeux.createChannelIfAbsent("/" + queryMessage.get(Search.ID),
-                    new ConfigurableServerChannel.Initializer() {
-                        public void configureChannel(ConfigurableServerChannel channel) {
-                            channel.setPersistent(true);
-                        }
-                    });
+                    (ConfigurableServerChannel.Initializer) channel -> channel.setPersistent(true));
 
             BayeuxContext context = bayeux.getContext();
             Subject subject = null;
@@ -176,8 +169,7 @@ public class SearchService {
     /**
      * Creates the query requests for each source and hands off the query to the Search Controller
      *
-     * @param queryMessage
-     *            - JSON message received from cometd
+     * @param queryMessage JSON message received from cometd
      */
     public void executeQuery(Map<String, Object> queryMessage, Subject subject) {
         String sources = castObject(String.class, queryMessage.get(SOURCES));
@@ -216,12 +208,11 @@ public class SearchService {
         Set<String> sourceIds;
         if (StringUtils.equalsIgnoreCase(sources, LOCAL_SOURCE)) {
             LOGGER.debug("Received local query");
-            sourceIds = new HashSet<String>(Arrays.asList(searchController.getFramework()
-                    .getId()));
+            sourceIds = Collections.singleton(searchController.getFramework()
+                    .getId());
         } else if (!(StringUtils.isEmpty(sources))) {
             LOGGER.debug("Received source names from client: {}", sources);
-            sourceIds =
-                    new HashSet<String>(Arrays.asList(StringUtils.stripAll(sources.split(","))));
+            sourceIds = new HashSet<>(Arrays.asList(StringUtils.stripAll(sources.split(","))));
         } else {
             LOGGER.debug("Received enterprise query");
             SourceInfoResponse sourceInfo = null;
@@ -234,7 +225,7 @@ public class SearchService {
             }
 
             if (sourceInfo != null) {
-                sourceIds = new HashSet<String>();
+                sourceIds = new HashSet<>();
                 for (SourceDescriptor source : sourceInfo.getSourceInfo()) {
                     if (source.isAvailable()) {
                         sourceIds.add(source.getSourceId());
@@ -251,16 +242,11 @@ public class SearchService {
     /**
      * Creates a new query from the incoming parameters
      *
-     * @param filter
-     *            - Filter to query
-     * @param startIndexLng
-     *            - Start index for the query
-     * @param countLng
-     *            - number of results for the query
-     * @param sortStr
-     *            - How to sort the query results
-     * @param maxTimeoutLng
-     *            - timeout value on the query execution
+     * @param filter        Filter to query
+     * @param startIndexLng Start index for the query
+     * @param countLng      number of results for the query
+     * @param sortStr       How to sort the query results
+     * @param maxTimeoutLng timeout value on the query execution
      * @return - the new query
      */
     private Query createQuery(Filter filter, Long startIndexLng, Long countLng, String sortStr,
@@ -269,7 +255,18 @@ public class SearchService {
         String sortField = Result.TEMPORAL;
         String sortOrder = DEFAULT_SORT_ORDER;
         Long startIndex = startIndexLng == null ? Long.valueOf(DEFAULT_START_INDEX) : startIndexLng;
-        Long count = countLng == null ? Long.valueOf(DEFAULT_COUNT) : countLng;
+        int count;
+        if (countLng == null) {
+            count = DEFAULT_COUNT;
+        } else if (countLng > searchController.getMaxQueryResults()) {
+            count = searchController.getMaxQueryResults();
+            LOGGER.debug(
+                    "Number of query results requested ({}) is greater than the max page count ({}). Using max count instead. ",
+                    countLng,
+                    searchController.getMaxQueryResults());
+        } else {
+            count = countLng.intValue();
+        }
         long maxTimeout = maxTimeoutLng == null ? DEFAULT_TIMEOUT : maxTimeoutLng;
 
         // Updated to use the passed in index if valid (=> 1)
@@ -305,11 +302,6 @@ public class SearchService {
                     .like()
                     .text(FilterDelegate.WILDCARD_CHAR);
         }
-        return new QueryImpl(filter,
-                startIndex.intValue(),
-                count.intValue(),
-                sort,
-                true,
-                maxTimeout);
+        return new QueryImpl(filter, startIndex.intValue(), count, sort, true, maxTimeout);
     }
 }
--- a/catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
+++ b/catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
@@ -29,26 +29,30 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
-import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
 
 import org.codice.ddf.ui.searchui.query.actions.ActionRegistryImpl;
+import org.codice.ddf.ui.searchui.query.controller.search.SourceQueryRunnable;
 import org.codice.ddf.ui.searchui.query.model.Search;
 import org.codice.ddf.ui.searchui.query.model.SearchRequest;
 import org.cometd.bayeux.server.BayeuxServer;
@@ -74,9 +78,7 @@ import com.google.common.util.concurrent.Futures;
 
 import ddf.catalog.CatalogFramework;
 import ddf.catalog.data.Result;
-import ddf.catalog.data.impl.BasicTypes;
 import ddf.catalog.data.impl.MetacardImpl;
-import ddf.catalog.federation.FederationException;
 import ddf.catalog.filter.impl.SortByImpl;
 import ddf.catalog.filter.proxy.adapter.GeotoolsFilterAdapterImpl;
 import ddf.catalog.operation.Query;
@@ -86,6 +88,7 @@ import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryResponseImpl;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
+import ddf.security.Subject;
 
 /**
  * Test cases for {@link org.codice.ddf.ui.searchui.query.controller.SearchController}
@@ -114,6 +117,10 @@ public class SearchControllerTest {
 
     private static final String[] SOURCE_ID_LIST = {"any_source_id"};
 
+    private ArgumentCaptor<QueryResponse> queryResponseArgumentCaptor;
+
+    private ArgumentCaptor<QueryRequest> queryRequestArgumentCaptor;
+
     private SearchController searchController;
 
     private CatalogFramework framework;
@@ -138,9 +145,14 @@ public class SearchControllerTest {
     @Mock
     private Query mockQuery;
 
+    @Mock
+    private Search search;
+
     @Before
     public void setUp() throws Exception {
         framework = createFramework();
+        queryResponseArgumentCaptor = ArgumentCaptor.forClass(QueryResponse.class);
+        queryRequestArgumentCaptor = ArgumentCaptor.forClass(QueryRequest.class);
 
         searchController = new SearchController(framework,
                 new ActionRegistryImpl(Collections.emptyList(), Collections.emptyList()),
@@ -155,13 +167,11 @@ public class SearchControllerTest {
         when(mockSearchRequest.getQuery()).thenReturn(mockQuery);
         when(mockQuery.getSortBy()).thenReturn(SortBy.NATURAL_ORDER);
         when(mockBayeuxServer.getChannel(anyString())).thenReturn(mockServerChannel);
-        when(mockExecutor.submit(any(Runnable.class))).thenAnswer(
-                (args) -> {
-                    Runnable runnable = (Runnable) args.getArguments()[0];
-                    runnable.run();
-                    return mockFuture;
-                }
-        );
+        when(mockExecutor.submit(any(Runnable.class))).thenAnswer((args) -> {
+            Runnable runnable = (Runnable) args.getArguments()[0];
+            runnable.run();
+            return mockFuture;
+        });
     }
 
     @Test
@@ -238,6 +248,7 @@ public class SearchControllerTest {
         Set<String> srcIds = new HashSet<>();
         srcIds.add(ID);
 
+        setupModifiedFramework(2L, 2, 2, 0);
         BayeuxServer bayeuxServer = mock(BayeuxServer.class);
         ServerChannel channel = mock(ServerChannel.class);
         ArgumentCaptor<ServerMessage.Mutable> reply =
@@ -325,9 +336,9 @@ public class SearchControllerTest {
         verify(channel, times(1)).publish(any(), any());
     }
 
-    private List<String> cacheQuery(Set<String> srcIds, int queryRequestCount)
-            throws CQLException, UnsupportedQueryException, SourceUnavailableException,
-            FederationException {
+    private List<String> cacheQuery(Set<String> srcIds, int queryRequestCount) throws Exception {
+
+        setupModifiedFramework(2L, 0);
         SearchRequest request = new SearchRequest(srcIds,
                 getQueryRequest("anyText LIKE '*'"),
                 "queryId");
@@ -384,10 +395,160 @@ public class SearchControllerTest {
         searchController.executeQuery(request, mockServerSession, null);
 
         // Verify
-        verify(framework).query(queryRequestCaptor.capture());
+        verify(framework, times(2)).query(queryRequestCaptor.capture());
         assertThat(queryRequestCaptor.getValue()
                 .getProperties()
                 .size(), is(0));
+    }
+
+    @Test
+    public void testOffsetStartIndexMultiPage() throws Exception {
+        final int START_INDEX = 13;
+        final long TOTAL_CATALOG_HITS = 20L;
+        final int EXPECTED_RESULT = 8;
+        final int FIRST_PAGE_RESULT = 5;
+        final int SECOND_PAGE_RESULT = 3;
+        final int EMPTY_PAGE_RESULT = 0;
+        final String EXPECTED_HITS_MESSAGE =
+                "The framework should have only return " + TOTAL_CATALOG_HITS + " hits";
+        final String EXPECTED_RESULTS_MESSAGE =
+                "The max query results was set to expect " + EXPECTED_RESULT + " results";
+
+        when(mockQuery.getStartIndex()).thenReturn(START_INDEX);
+        setupModifiedFramework(TOTAL_CATALOG_HITS,
+                FIRST_PAGE_RESULT,
+                SECOND_PAGE_RESULT,
+                EMPTY_PAGE_RESULT);
+        SourceQueryRunnable sourceQueryRunnable = setupSourceQueryRunnable(search, EXPECTED_RESULT);
+        sourceQueryRunnable.run();
+
+        verify(framework, times(3)).query(queryRequestArgumentCaptor.capture());
+        verify(search).update(any(), queryResponseArgumentCaptor.capture());
+        List<Integer> startIndices = queryRequestArgumentCaptor.getAllValues()
+                .stream()
+                .map(qR -> qR.getQuery()
+                        .getStartIndex())
+                .collect(Collectors.toList());
+        List<Integer> expectedIndices = Arrays.asList(13, 18, 21);
+        assertThat(startIndices, is(expectedIndices));
+        assertThat(EXPECTED_HITS_MESSAGE,
+                queryResponseArgumentCaptor.getValue()
+                        .getHits(),
+                is(TOTAL_CATALOG_HITS));
+        assertThat(EXPECTED_RESULTS_MESSAGE,
+                queryResponseArgumentCaptor.getValue()
+                        .getResults()
+                        .size(),
+                is(EXPECTED_RESULT));
+    }
+
+    @Test
+    public void testMatchingPageSize() throws Exception {
+        final long TOTAL_CATALOG_HITS = 60L;
+        final int EXPECTED_RESULT = 60;
+        final int FIRST_PAGE_RESULT = 30;
+        final int SECOND_PAGE_RESULT = 30;
+        final int EMPTY_PAGE_RESULT = 0;
+        final String EXPECTED_HITS_MESSAGE =
+                "The framework should have only return " + TOTAL_CATALOG_HITS + " hits";
+        final String EXPECTED_RESULTS_MESSAGE =
+                "The max query results was set to expect " + EXPECTED_RESULT + " results";
+
+        setupModifiedFramework(TOTAL_CATALOG_HITS,
+                FIRST_PAGE_RESULT,
+                SECOND_PAGE_RESULT,
+                EMPTY_PAGE_RESULT);
+        SourceQueryRunnable sourceQueryRunnable = setupSourceQueryRunnable(search, EXPECTED_RESULT);
+        sourceQueryRunnable.run();
+
+        verify(framework, times(3)).query(queryRequestArgumentCaptor.capture());
+        verify(search).update(any(), queryResponseArgumentCaptor.capture());
+        List<Integer> startIndices = queryRequestArgumentCaptor.getAllValues()
+                .stream()
+                .map(qR -> qR.getQuery()
+                        .getStartIndex())
+                .collect(Collectors.toList());
+        List<Integer> expectedIndices = Arrays.asList(0, 30, 60);
+        assertThat(startIndices, is(expectedIndices));
+        assertThat(EXPECTED_HITS_MESSAGE,
+                queryResponseArgumentCaptor.getValue()
+                        .getHits(),
+                is(TOTAL_CATALOG_HITS));
+        assertThat(EXPECTED_RESULTS_MESSAGE,
+                queryResponseArgumentCaptor.getValue()
+                        .getResults()
+                        .size(),
+                is(EXPECTED_RESULT));
+    }
+
+    @Test
+    public void testMaxQueryPaging() throws Exception {
+        final long TOTAL_CATALOG_HITS = 1100L;
+        final int EXPECTED_MAX_RESULT = 1000;
+        final int FIRST_PAGE_RESULT = 400;
+        final int SECOND_PAGE_RESULT = 400;
+        final int THIRD_PAGE_RESULT = 300;
+        final int EMPTY_PAGE_RESULT = 0;
+        final String EXPECTED_HITS_MESSAGE =
+                "The framework should have only return " + TOTAL_CATALOG_HITS + " hits";
+        final String EXPECTED_RESULTS_MESSAGE =
+                "The max query results was set to expect " + EXPECTED_MAX_RESULT + " results";
+
+        setupModifiedFramework(TOTAL_CATALOG_HITS,
+                FIRST_PAGE_RESULT,
+                SECOND_PAGE_RESULT,
+                THIRD_PAGE_RESULT,
+                EMPTY_PAGE_RESULT);
+        SourceQueryRunnable sourceQueryRunnable = setupSourceQueryRunnable(search,
+                EXPECTED_MAX_RESULT);
+        sourceQueryRunnable.run();
+
+        verify(framework, times(4)).query(queryRequestArgumentCaptor.capture());
+        verify(search).update(any(), queryResponseArgumentCaptor.capture());
+        List<Integer> startIndices = queryRequestArgumentCaptor.getAllValues()
+                .stream()
+                .map(qR -> qR.getQuery()
+                        .getStartIndex())
+                .collect(Collectors.toList());
+        List<Integer> expectedIndices = Arrays.asList(0, 400, 800, 1100);
+        assertThat(startIndices, is(expectedIndices));
+        assertThat(EXPECTED_HITS_MESSAGE,
+                queryResponseArgumentCaptor.getValue()
+                        .getHits(),
+                is(TOTAL_CATALOG_HITS));
+        assertThat(EXPECTED_RESULTS_MESSAGE,
+                queryResponseArgumentCaptor.getValue()
+                        .getResults()
+                        .size(),
+                is(EXPECTED_MAX_RESULT));
+    }
+
+    private SourceQueryRunnable setupSourceQueryRunnable(Search search, int pageSize) {
+        when(mockQuery.getPageSize()).thenReturn(pageSize);
+        Subject subject = mock(Subject.class);
+        BayeuxServer bayeuxServer = mock(BayeuxServer.class);
+        ServerChannel channel = mock(ServerChannel.class);
+        when(bayeuxServer.getChannel(any(String.class))).thenReturn(channel);
+        searchController.setBayeuxServer(bayeuxServer);
+        when(mockSearchRequest.getQuery()).thenReturn(mockQuery);
+
+        return new SourceQueryRunnable(searchController,
+                "sampleSourceId",
+                mockSearchRequest,
+                subject,
+                new HashMap<>(),
+                search,
+                mockServerSession,
+                mockFuture,
+                mockFuture);
+    }
+
+    private void setupModifiedFramework(long totalHits, int... catalogResults) throws Exception {
+        framework = createFramework(totalHits, catalogResults);
+        searchController = new SearchController(framework,
+                new ActionRegistryImpl(Collections.emptyList(), Collections.emptyList()),
+                new GeotoolsFilterAdapterImpl(),
+                new SequentialExecutorService());
     }
 
     private void assertReplies(List<Mutable> replies) {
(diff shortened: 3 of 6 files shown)


catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -269,7 +261,20 @@ private Query createQuery(Filter filter, Long startIndexLng, Long countLng, Stri
         String sortField = Result.TEMPORAL;
         String sortOrder = DEFAULT_SORT_ORDER;
         Long startIndex = startIndexLng == null ? Long.valueOf(DEFAULT_START_INDEX) : startIndexLng;
-        Long count = countLng == null ? Long.valueOf(DEFAULT_COUNT) : countLng;
+        Long count;
+        if (countLng == null) {
+            count = DEFAULT_COUNT;
+        } else if (countLng > searchController.getMaxQueryResults()) {
+            count = (long) searchController.getMaxQueryResults();
+            LOGGER.debug(
```
clockard: This might be good information to put at the info level.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -84,7 +85,7 @@
 
     private static final long DEFAULT_TIMEOUT = 300000;
 
-    private static final int DEFAULT_COUNT = 10;
+    private static final long DEFAULT_COUNT = 10L;
```
josephthweatt: `DEFAULT_COUNT` is only being used once, to set local long value `count` in **CreateQuery()**. `count` is then used to create a **QueryImpl** object, which is getting converted back into an int (315). Maybe we should have `DEFAULT_COUNT` and `count` as integers, and on line 276 we just convert `countLng` to an int.

catalog/ui/search-ui/search-endpoint/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -23,6 +23,9 @@
         <AD description="Disables relevance and distance normalization." name="Disable Normalization"
             id="normalizationDisabled" required="false" type="Boolean" default="false"/>
 
+        <AD description="Controls the maximum number of results returned from a query." name="Maximum Query Results"
```
emmberk: ✏️  Maybe include what maximum results means. Like, what happens when I have more results than this maximum? Why set a maximum in the first place?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mcalcote: 👍 will update

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,96 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
         assertReplies(replies);
     }
 
+    @Test
+    public void testReducedMaxQueryResults() throws Exception {
+
```
josephthweatt: Extra line here should be deleted  to be in the same format of other methods

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,96 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
...
+        framework = createFramework(20);
```
emmberk: ✏️  You could make 20 and 7 final variables here and use them in the setup and in the asserts to make it clore clear why you're expecting 20L and 7.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,96 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
...
+                mockFuture,
+                mockFuture);
+
+        sourceQueryRunnable.run();
+
+        verify(search).update(any(), queryResponseArgumentCaptor.capture());
+
+        assertThat(queryResponseArgumentCaptor.getAllValues()
+                .size(), is(1));
+        assertThat(queryResponseArgumentCaptor.getValue()
```
emmberk: ✏️  These asserts could use a message. Why are you expecting 20L and 15?

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Good call, will remove

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,96 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
...
+                .getHits(), is(20L));
+        assertThat(queryResponseArgumentCaptor.getValue()
+                .getResults()
+                .size(), is(15));
+
+    }
+
```
josephthweatt: **testReducedMaxQueryResults** and **testPagingFromCatalog** differ by only 3 lines (249/293, 267/311, 288/332). You could probably make some methods that simplify the tests and reduce the lines of code being added.

distribution/docs/src/main/jdocs/content/_tables/org.codice.ddf.ui.search.standard.endpoint-table-contents.adoc
```diff
@@ -26,5 +26,12 @@
 |false
 |false
 
+|Set Max Query Results
+|setMaxQueryResults
```
vinamartin: ✏️  Looks like the name of this is actually `Maximum Query Results`/`maxQueryResults` (without the set part). Change this to reflect what is in the blueprint for consistency.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mcalcote: Updated

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Will add a method or two to reduce duplication.

catalog/ui/search-ui/search-endpoint/src/main/resources/OSGI-INF/metatype/metatype.xml
mcalcote: Tried to explain a bit more. Please review the new text and let me know what else should be altered.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Will leave comment, but I don't want to alter the setup when they are only used for a single test.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Will add messages. 👍

distribution/docs/src/main/jdocs/content/_tables/org.codice.ddf.ui.search.standard.endpoint-table-contents.adoc
mcalcote: Nice catch 👍

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -216,8 +212,9 @@ public void executeQuery(Map<String, Object> queryMessage, Subject subject) {
         Set<String> sourceIds;
         if (StringUtils.equalsIgnoreCase(sources, LOCAL_SOURCE)) {
             LOGGER.debug("Received local query");
-            sourceIds = new HashSet<String>(Arrays.asList(searchController.getFramework()
-                    .getId()));
+            sourceIds =
+                    new HashSet<String>(Collections.singletonList(searchController.getFramework()
```
mweser: ❓ What drove the change here from Arrays to Collections?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -269,7 +261,20 @@ private Query createQuery(Filter filter, Long startIndexLng, Long countLng, Stri
         String sortField = Result.TEMPORAL;
         String sortOrder = DEFAULT_SORT_ORDER;
         Long startIndex = startIndexLng == null ? Long.valueOf(DEFAULT_START_INDEX) : startIndexLng;
-        Long count = countLng == null ? Long.valueOf(DEFAULT_COUNT) : countLng;
+        int count;
+        if (countLng == null) {
```
mweser: ❓ Is this change here mostly to clarify the ternary operator functionality?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
bdeining: If there is only one item in the collection, the guidance is to use `Collections.singletonList` instead.  Intellij will actually mark it as a warning.
https://stackoverflow.com/questions/26027396/arrays-aslist-vs-collections-singletonlist

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -409,91 +489,126 @@ private void assertReplies(List<Mutable> replies) {
...
         }
         return framework;
     }
 
-    /**
-     * Helper methods for generating exception stacks
-     */
-    private Exception generateExceptionStack(int targetSize) {
-        return generateExceptionStack(targetSize, 0, null);
+    private Result generateResult(int value, long count) {
```
mweser: ✏️ Changing the name to "generateResultMock" would increase the readability of the code when this method is invoked. In addition, changing "count" to something more descriptive would help as well.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -409,91 +489,126 @@ private void assertReplies(List<Mutable> replies) {
...
+            when(framework.query(any(QueryRequest.class))).thenReturn(firstResponse,
+                    secondResponse,
+                    emptyResponse);
+        } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            LOGGER.debug("Error querying framework", e);
+        }
+        return framework;
+    }
+
+    private CatalogFramework createFramework(long count) {
```
mweser: ✏️ Specifying this creates a framework _mock_ would help with readability, as well as making "count" more specific. As it stands, it's rather vague what "count" is the count of. It should be immediately apparent by its name that it's the "count of results" you wish to return from a call to the catalog framework.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -409,91 +489,126 @@ private void assertReplies(List<Mutable> replies) {
...
 
         QueryResponse response = new QueryResponseImpl(mock(QueryRequest.class),
-                new ArrayList<Result>(),
-                COUNT);
+                new ArrayList<>(),
+                count);
         response.getResults()
                 .addAll(results);
 
+        QueryResponse emptyResponse = new QueryResponseImpl(mock(QueryRequest.class),
```
mweser: ❓ What value does having an actual QueryResponse bring versus simply a mock of a QueryResponse?

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -409,91 +489,126 @@ private void assertReplies(List<Mutable> replies) {
...
+        for (int i = 0; i < count * pagesExpected; i++) {
+            Result result = generateResult(i, count);
+            results.add(result);
+        }
+
+        QueryResponse firstResponse = new QueryResponseImpl(mock(QueryRequest.class),
+                new ArrayList<>(),
+                count);
+        for (int i = 0; i < count; i++) {
+            firstResponse.getResults()
```
mweser: ✏️  I'd have you take a look at https://github.com/codice/ddf/blob/b132cafeb6b770730c0c243dcd988f57b0580b36/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java#L123 to see how to more cleanly set up some of these responses so that your code is a bit more readable and functional.

mweser: https://github.com/codice/ddf/blob/b132cafeb6b770730c0c243dcd988f57b0580b36/catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/RemoveAllCommandTest.java#L211

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -384,7 +464,7 @@ public void testExecuteQueryCacheDisabled() throws Exception {
         searchController.executeQuery(request, mockServerSession, null);
 
         // Verify
-        verify(framework).query(queryRequestCaptor.capture());
+        verify(framework, times(2)).query(queryRequestCaptor.capture());
```
mweser: ✏️  It would help readability to extract the value 2 and set as a final constant at the top of the class.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -273,21 +353,21 @@ private Query getQueryRequest(String cql) throws CQLException {
...
     public void testExecuteQueryCacheEnabledWithMultipleSources() throws Exception {
         Set<String> srcIds = new HashSet<>(2);
         srcIds.add("id1");
         srcIds.add("id2");
-        List<String> modes = cacheQuery(srcIds, 3);
+        List<String> modes = cacheQuery(srcIds, 4);
 
-        assertThat(modes.size(), is(3));
-        assertThat(modes, hasItems("cache", "update", "update"));
+        assertThat(modes.size(), is(4));
```
mweser: ✏️  It would help readability to extract the value 4 and set as a final constant at the top of the class. That would also help with your cacheQuery() call readability.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -273,21 +353,21 @@ private Query getQueryRequest(String cql) throws CQLException {
     public void testExecuteQueryCacheEnabledWithSingleSource() throws Exception {
         Set<String> srcIds = new HashSet<>(1);
         srcIds.add("id");
-        List<String> modes = cacheQuery(srcIds, 2);
+        List<String> modes = cacheQuery(srcIds, 3);
 
-        assertThat(modes.size(), is(2));
-        assertThat(modes, hasItems("cache", "update"));
+        assertThat(modes.size(), is(3));
```
mweser: ✏️  It would help readability to extract the value 3 and set as a final constant at the top of the class.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,101 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
...
+
+        Subject subject = mock(Subject.class);
+        Search search = mock(Search.class);
+        ArgumentCaptor<QueryResponse> queryResponseArgumentCaptor = ArgumentCaptor.forClass(
+                QueryResponse.class);
+        searchController = preparePagingTestController(framework);
+
+        //this is the maximum amount of results a query should return
+        searchController.setMaxQueryResults(MAX_QUERY_RESULTS);
+        SourceQueryRunnable sourceQueryRunnable = new SourceQueryRunnable(searchController,
```
mweser: ✏️  This middle section could benefit from extracting some functionality to external methods to make more readable.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,101 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
...
+                subject,
+                new HashMap<>(),
+                search,
+                mockServerSession,
+                mockFuture,
+                mockFuture);
+        sourceQueryRunnable.run();
+
+        verify(search).update(any(), queryResponseArgumentCaptor.capture());
+        assertThat("The framework should have only return " + (MAX_FRAMEWORK_PAGE * PAGES_EXPECTED)
```
mweser: ✏️  It would be more readable to extract the string message to a local variable, as well as extracting the call to the queryResponseArgumentCaptor to a local variable. We typically don't like having a lot of nested functionality inside assert statements, as that muddies the readability.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,101 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
...
+                        .getResults()
+                        .size(),
+                is(MAX_QUERY_RESULTS));
+    }
+
+    @Test
+    public void testPagingFromCatalog() throws Exception {
+        final int MAX_QUERY_RESULTS = 15;
+        final int MAX_FRAMEWORK_PAGE = 10;
+        final int PAGES_EXPECTED = 2;
```
mweser: 😄  I like how you extracted these values. I'd like to see more of this, as it makes it clear to someone with little context what the important inputs to the test are.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -258,6 +243,101 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
         assertReplies(replies);
     }
 
+    @Test
+    public void testReducedMaxQueryResults() throws Exception {
+        final int MAX_QUERY_RESULTS = 7;
```
mweser: ✏️  See the comments on the next test as they apply here as well.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -140,7 +139,7 @@
 
     @Before
     public void setUp() throws Exception {
-        framework = createFramework();
+        framework = createFramework(2);
```
mweser: ❗️ I notice this same method is called in your individual tests, so is should this remain? Also the method name is rather vague, leading a reader to wonder what the 2 is there for. Specifying that this sets up a mock would be very helpful.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mcalcote: New flow control was needed to account for the possibility a non-null countLng was greater than the maxResultQuery size. I tried to keep the ternary, but it would have resulted in nested operations.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: For a majority of tests, this case is used, so I only called createFramework when this needed an alternate value.  I can add a comment for clarity in those cases and overload the method so this remains 'createFramework()'.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: 👍 @emmberk suggested this and deserves the kudos

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: 🤝

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mweser: Well I appreciate the change then 👍

catalog/ui/search-ui/search-endpoint/src/main/resources/OSGI-INF/metatype/metatype.xml
emmberk: Looks great, except that the documentation is missing the extra explanation. Thanks!

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
emmberk: I meant setup in the unit test method itself. Looks like you went ahead and did that already with the `final int MAX_QUERY_RESULTS = 7;` and others. Thanks!

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Are you referring to the creation/instantiation of the Query Runnable? The formatter makes it seem like more than it is, we really only have 3 lines of code here. Set max query results, create the runnable, run the runnable.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: I feel like that would add a lot of lines of code to create a local variable to hold each portion of this assertion... The line was more concise before extracting all the values out. Maybe too much extraction and instantiation also hurts readability? 

I will attempt to clean this up a bit.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: This class is already 800+ lines, I would really like to avoid creating these variables unless they really helps illuminate some facet of the test. Thankfully with IntelliJ, the method call shows the signature's name for each variable and helps the readability inside the IDE.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Same comment as above.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Same comment.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: That is an elegant solution. Perhaps it should be moved to a location so that other paging improvements could use it in their tests...

I will adapt the logic in this class a bit to use some of it, but will try to keep it under the number of lines for my "dirty" implementation.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Updated names to make their use more clear.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: I agree the naming here is very weak. Will update

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mweser: They do illuminate facets of the test, and the size of the file as a whole should be irrelevant. Also, the IDE being used with this code should not be assumed as the code should be as environment-agnostic as possible.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mweser: @lessarderic What might your take on this be? Might this class be broken up a bit given how large it has become? A couple classes to test different aspects of SearchController?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
```diff
@@ -255,22 +254,13 @@ protected QueryResponse queryCatalog(String sourceId, SearchRequest searchReques
                 }
 
                 LOGGER.debug("Sending query: {}", query);
-                response = searchController.getFramework()
-                        .query(request);
+                ResultIterable results = new ResultIterable(searchController.getFramework(),
```
lessarderic: ❗️ We should keep using `searchController.getFramework().query(request)` here instead of switching to the new `ResultIterable`, otherwise we lose all the processing details returned by the `QueryResponse` that we used to handle in the catch blocks.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
```diff
@@ -255,22 +254,13 @@ protected QueryResponse queryCatalog(String sourceId, SearchRequest searchReques
...
+                        request);
+                List<Result> resultsList = new ArrayList<>();
+                results.iterator()
+                        .forEachRemaining(resultsList::add);
+                response = new QueryResponseImpl(request, resultsList, (long) resultsList.size());
             }
```
lessarderic: ❗️ We need to keep handling the processing details contained in the exceptions.

catalog/ui/search-ui/search-endpoint/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -20,9 +20,17 @@
         <AD description="Disables use of cache." name="Disable Cache" id="cacheDisabled"
             required="false" type="Boolean" default="false"/>
 
-        <AD description="Disables relevance and distance normalization." name="Disable Normalization"
+        <AD description="Disables relevance and distance normalization."
+            name="Disable Normalization"
             id="normalizationDisabled" required="false" type="Boolean" default="false"/>
 
+        <AD description="Controls the maximum number of results returned from a single query. This
```
lessarderic: ❓ Do we really need to provide a configuration for this in the Admin UI or can we pick a number and let the endpoint clients (CometD, UI, etc.) do the paging instead if needed?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -101,10 +102,8 @@
     /**
      * Creates a new SearchService
      *
-     * @param filterBuilder
-     *            - FilterBuilder to use for queries
-     * @param searchController
-     *            - SearchController to handle async queries
+     * @param filterBuilder    - FilterBuilder to use for queries
```
lessarderic: ✏️  Since we're touching those comments, can we remove the `-`? They aren't needed.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
lessarderic: How about `sourceIds = Collections.singleton(searchController.getFramework().getId());`?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -269,7 +261,20 @@ private Query createQuery(Filter filter, Long startIndexLng, Long countLng, Stri
...
         String sortOrder = DEFAULT_SORT_ORDER;
         Long startIndex = startIndexLng == null ? Long.valueOf(DEFAULT_START_INDEX) : startIndexLng;
-        Long count = countLng == null ? Long.valueOf(DEFAULT_COUNT) : countLng;
+        int count;
+        if (countLng == null) {
+            count = DEFAULT_COUNT;
+        } else if (countLng > searchController.getMaxQueryResults()) {
+            count = searchController.getMaxQueryResults();
+            LOGGER.info(
+                    "Number of query results requested ({}) is greater than the max page count({}). Using max count instead, "
```
lessarderic: ✏️  Missing space between `page count` and `({})`.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
```diff
@@ -255,22 +254,13 @@ protected QueryResponse queryCatalog(String sourceId, SearchRequest searchReques
                 }
 
                 LOGGER.debug("Sending query: {}", query);
-                response = searchController.getFramework()
-                        .query(request);
+                ResultIterable results = new ResultIterable(searchController.getFramework(),
+                        request);
+                List<Result> resultsList = new ArrayList<>();
+                results.iterator()
```
lessarderic: ❗️ As an FYI, the `ResultIterable` iterator will _not_ stop once max query result has been reached but will continue until _all_ results have been retrieved. The `ResultListIterable` could have been used instead, but it doesn't matter since we can't use those here anyways.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
```diff
@@ -269,7 +261,20 @@ private Query createQuery(Filter filter, Long startIndexLng, Long countLng, Stri
         String sortField = Result.TEMPORAL;
         String sortOrder = DEFAULT_SORT_ORDER;
         Long startIndex = startIndexLng == null ? Long.valueOf(DEFAULT_START_INDEX) : startIndexLng;
-        Long count = countLng == null ? Long.valueOf(DEFAULT_COUNT) : countLng;
+        int count;
+        if (countLng == null) {
+            count = DEFAULT_COUNT;
+        } else if (countLng > searchController.getMaxQueryResults()) {
+            count = searchController.getMaxQueryResults();
+            LOGGER.info(
```
lessarderic: ❓ Since we could we end up flooding the log file if a client keeps using a count that's too high, are we sure we want to log this at info?

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
rzwiefel: @mweser IntelliJ will now tell you what an ambiguous parameter name is -- so not knowing what the 2 is there for is just a shortcoming of github reviews 😄

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
rzwiefel: Why not just `mockResult`? 🙂

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mweser: @rzwiefel I know and love that feature in IntelliJ, but it is bad practice to assume the development environment when this is an open-source project and the code itself should be universally readable to all. The feature is simply for ease of development, but many people still view the code in raw form in the GitHub text editor, and we should always assume that there are no environment features to help with viewing raw text.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
rzwiefel: As soon as https://github.com/codice/ddf/pull/2207 is merged, there will be a way to cap the maximum number of results returned by the ResultIterable.

catalog/ui/search-ui/search-endpoint/src/main/resources/OSGI-INF/metatype/metatype.xml
mcalcote: It isn't needed and I was on the fence about adding it to be honest. Will remove.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
mcalcote: Removed ResultIterable reference.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
mcalcote: Removed reference

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
mcalcote: Added back in 👍

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mcalcote: removed

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mcalcote: Since the value will not longer be configurable by the user, and this info will be less relevant as a result, I will set it back to debug.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/service/SearchService.java
mcalcote: 👍 Nice catch

mcalcote:
build now

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
```diff
@@ -190,16 +190,12 @@ protected int getMaxResults(SearchRequest request) {
...
 
     protected void addResults(Collection<Result> responseResults) {
-        results.putAll(Maps.uniqueIndex(responseResults, new Function<Result, String>() {
-            @Override
-            public String apply(Result result) {
-                return getResultKey(result.getMetacard());
-            }
-        }));
+        results.putAll(Maps.uniqueIndex(responseResults,
+                result -> getResultKey(result != null ? result.getMetacard() : null)));
```
mweser: 👍  Great use of lambda expression to make this easier to read.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -238,6 +240,7 @@ public void testMetacardTypeValuesCacheEnabled() throws Exception {
         Set<String> srcIds = new HashSet<>();
         srcIds.add(ID);
 
+        setupModifiedFramework(2L, 2, 2);
```
mweser: 👍  I really like this abstraction.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -390,6 +395,64 @@ public void testExecuteQueryCacheDisabled() throws Exception {
                 .size(), is(0));
     }
 
+    @Test
+    public void testMaxQueryPaging() throws Exception {
+        final long TOTAL_CATALOG_HITS = 1100L;
```
mweser: 👍  Great use of verbose inputs to test

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -390,6 +395,64 @@ public void testExecuteQueryCacheDisabled() throws Exception {
...
+        final int EXPECTED_MAX_RESULT = 1000;
+        final int SECOND_PAGE_RESULT = 500;
+        final int FIRST_PAGE_RESULT = 600;
+        final int EMPTY_PAGE = 0;
+        ArgumentCaptor<QueryResponse> queryResponseArgumentCaptor = ArgumentCaptor.forClass(
+                QueryResponse.class);
+        Search search = mock(Search.class);
+        when(mockQuery.getPageSize()).thenReturn(EXPECTED_MAX_RESULT);
+
+        setupModifiedFramework(TOTAL_CATALOG_HITS,
```
mweser: 👍

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -390,6 +395,64 @@ public void testExecuteQueryCacheDisabled() throws Exception {
...
+                FIRST_PAGE_RESULT,
+                SECOND_PAGE_RESULT,
+                EMPTY_PAGE);
+        SourceQueryRunnable sourceQueryRunnable = setupSourceQueryRunnable(search);
+        sourceQueryRunnable.run();
+
+        verify(search).update(any(), queryResponseArgumentCaptor.capture());
+        assertThat("The framework should have only return " + TOTAL_CATALOG_HITS + " hits",
```
mweser: ✏️  Strings for assert reasons could be cleaned up a bit and constructed up where the input variables are declared. "expected" strings declared with variables.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -390,6 +395,64 @@ public void testExecuteQueryCacheDisabled() throws Exception {
...
+                mockSearchRequest,
+                subject,
+                new HashMap<>(),
+                search,
+                mockServerSession,
+                mockFuture,
+                mockFuture);
+    }
+
+    private void setupModifiedFramework(long totalHits, int... catalogResults) {
```
mweser: 👍  Great abstraction. Goes a long way to help with readability throughout test class.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -442,58 +505,54 @@ private void assertReplies(List<Mutable> replies) {
...
         }
         return framework;
     }
 
-    /**
-     * Helper methods for generating exception stacks
-     */
-    private Exception generateExceptionStack(int targetSize) {
-        return generateExceptionStack(targetSize, 0, null);
+    private QueryResponse getNextResponse(List<QueryResponse> responseList) {
```
mweser: ✏️  Might consider changing "get" to "pop" to indicate the side-effect of returning top value as well as removing it. A get is more passive and doesn't remove any elements, which may be what future developers might assume when using this method.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -442,58 +505,54 @@ private void assertReplies(List<Mutable> replies) {
...
+        if (responseList.size() > 1) {
+            return responseList.remove(1);
+        } else {
+            return new QueryResponseImpl(mock(QueryRequest.class), new ArrayList<>(), 0);
+        }
+    }
+
+    private List<Result> generateResults(int resultsPerPage) {
+        return Stream.generate(() -> generateResult(UUID.randomUUID()
```
mweser: 👍  Beautiful

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -442,58 +505,54 @@ private void assertReplies(List<Mutable> replies) {
...
+                .toString()))
+                .limit(resultsPerPage)
+                .collect(Collectors.toList());
+    }
+
+    private Result generateResult(String id) {
```
mweser: 👍  Great abstraction

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Good call, that is more clear.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Fixed

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -442,58 +510,54 @@ private void assertReplies(List<Mutable> replies) {
...
-                COUNT);
-        response.getResults()
-                .addAll(results);
+        List<QueryResponse> responseList = new ArrayList<>();
+        for (int resultsPerPage : pagingResults) {
+            List<Result> results = generateResults(resultsPerPage);
+            QueryResponse response = new QueryResponseImpl(mock(QueryRequest.class),
+                    new ArrayList<>(),
+                    totalHits);
+            response.getResults()
```
lessarderic: ❓ Could we initialize the `results` first and pass that in to the `QueryResponseImpl` constructor instead of giving an empty `ArrayList` and then populating it? Might make the code a little more straightforward.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -442,58 +510,54 @@ private void assertReplies(List<Mutable> replies) {
...
         try {
-            when(framework.query(any(QueryRequest.class))).thenReturn(response);
-        } catch (UnsupportedQueryException e) {
-            LOGGER.debug("Error querying framework", e);
-        } catch (SourceUnavailableException e) {
-            LOGGER.debug("Error querying framework", e);
-        } catch (FederationException e) {
+            when(framework.query(any(QueryRequest.class))).thenReturn(responseList.get(0),
+                    popNextResponse(responseList));
+        } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
```
lessarderic: ❗️Since this is a test class, we shouldn't worry about catching exceptions here and just let the exceptions (which will never happen) bubble up. Adding `throws Exception` to the method should be good enough here.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -442,58 +510,54 @@ private void assertReplies(List<Mutable> replies) {
...
 
         try {
-            when(framework.query(any(QueryRequest.class))).thenReturn(response);
-        } catch (UnsupportedQueryException e) {
-            LOGGER.debug("Error querying framework", e);
-        } catch (SourceUnavailableException e) {
-            LOGGER.debug("Error querying framework", e);
-        } catch (FederationException e) {
+            when(framework.query(any(QueryRequest.class))).thenReturn(responseList.get(0),
+                    popNextResponse(responseList));
```
lessarderic: ❓Just to make sure here. `popNextResponse()` will be called when the mock is setup, not when the test is run and `framework.query()` is called. Is that what we want?

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
```diff
@@ -390,6 +395,69 @@ public void testExecuteQueryCacheDisabled() throws Exception {
                 .size(), is(0));
     }
 
+    @Test
+    public void testMaxQueryPaging() throws Exception {
```
lessarderic: ❓ Shouldn't we have more tests here, i.e., one when we reach the total number of hits, one when we reach the number requested and one we reach the hard-coded limit (`1,000`)?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
```diff
@@ -289,6 +286,31 @@ protected QueryResponse queryCatalog(String sourceId, SearchRequest searchReques
         return response;
     }
 
+    private void collectResponses(QueryRequest request, QueryResponse response)
+            throws UnsupportedQueryException, SourceUnavailableException, FederationException {
+        int intialResultsSize = response.getResults()
+                .size();
+        while(intialResultsSize < response.getHits() && intialResultsSize < request.getQuery()
```
mcalcote: I need a space after `while`!

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: We reach the total number of hits for many of the pre-existing tests. We reach the hard-coded limit in this test (1100 total hits with simulated max catalog result of 600 and 1000 results returned over two pages). These two conditions cover not using paging and paging. I can add more tests, but I might be missing what you mean by the number requested.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Good call, easy change :+1:

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
mcalcote: Nope, it should have been an array, not a list. Fixed and updated the test to have more pages.

catalog/ui/search-ui/search-endpoint/src/test/java/org/codice/ddf/ui/searchui/query/controller/SearchControllerTest.java
lessarderic: So I guess my point was that we're trying to test all the different paths for the `while` loop here: https://github.com/codice/ddf/pull/2195/files#diff-675e304e397476c2b776d7b512286af6R293.

Here's what I had in mind, but there could be more:
1. The first query returns everything requested, e.g., page size = 100, total hits = 100 and result size = 100. In other words, we don't even go inside the while loop.
2. We need to query the Catalog multiple times until we get to the page size requested, which is less than the total hits, e.g., page size = 100, total hits = 300 and result sizes = 50, 50. In this case, we go in the while loop once and leave because `intialResultsSize < request.getQuery().getPageSize()`
3. The test you currently have where we need to query the Catalog until we hit the total hits of 1,000, which covers the case where we exit the while loop because `intialResultsSize < response.getHits()`

We may also want to have some tests where the start index for the query is not 1 with the above variations to make sure we stop when expected. In other words, to test that lines https://github.com/codice/ddf/pull/2195/files#diff-675e304e397476c2b776d7b512286af6R295 and https://github.com/codice/ddf/pull/2195/files#diff-675e304e397476c2b776d7b512286af6R298 work as expected.

Hope this makes sense.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/search/QueryRunnable.java
```diff
@@ -289,6 +286,31 @@ protected QueryResponse queryCatalog(String sourceId, SearchRequest searchReques
         return response;
     }
 
+    private void collectResponses(QueryRequest request, QueryResponse response)
+            throws UnsupportedQueryException, SourceUnavailableException, FederationException {
+        int intialResultsSize = response.getResults()
+                .size();
+        while (intialResultsSize < response.getHits() && intialResultsSize < request.getQuery()
```
rzwiefel: ❗️ getHits() is not guaranteed, and can possibly return `-1`

mcalcote: ...sigh, this means there is no way I can prevent an extra query if the results returned are less than the query page size

mcalcote:
build now

clockard:
build now

clockard:
build now

PR #2204, opened 2017-08-09
DDF-3207 Updates CswEndpoint to support paging

coyotesqrl:
#### What does this PR do?
Uses paging internally for get, update, and delete operations to avoid failing or producing incorrect and unexpected results.

#### Who is reviewing it? 
@rzwiefel @AzGoalie @kcwire 

@codice/io

@clockard
@lessarderic

Full build and test. Then a larger number of records should be ingested - at minimum, more than 20 - and then queried, updated, and deleted through CSW queries. All records should be correctly processed.

N/A

[DDF-3207](https://codice.atlassian.net/browse/DDF-3207)

N/A

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
@@ -28,8 +28,11 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import javax.activation.MimeType;
 import javax.activation.MimeTypeParseException;
@@ -76,6 +79,7 @@ import org.w3c.dom.Element;
 import org.xml.sax.SAXException;
 
 import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
 import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.io.ParseException;
@@ -93,10 +97,12 @@ import ddf.catalog.operation.DeleteResponse;
 import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.QueryResponse;
 import ddf.catalog.operation.ResourceResponse;
+import ddf.catalog.operation.SourceResponse;
 import ddf.catalog.operation.UpdateRequest;
 import ddf.catalog.operation.UpdateResponse;
 import ddf.catalog.operation.impl.CreateRequestImpl;
 import ddf.catalog.operation.impl.DeleteRequestImpl;
+import ddf.catalog.operation.impl.QueryResponseImpl;
 import ddf.catalog.operation.impl.ResourceRequestById;
 import ddf.catalog.operation.impl.UpdateRequestImpl;
 import ddf.catalog.resource.Resource;
@@ -106,6 +112,9 @@ import ddf.catalog.resource.impl.ResourceImpl;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
+import ddf.catalog.util.impl.CatalogQueryException;
+import ddf.catalog.util.impl.QueryFunction;
+import ddf.catalog.util.impl.ResultIterable;
 import net.opengis.cat.csw.v_2_0_2.BriefRecordType;
 import net.opengis.cat.csw.v_2_0_2.CapabilitiesType;
 import net.opengis.cat.csw.v_2_0_2.DescribeRecordResponseType;
@@ -200,6 +209,8 @@ public class CswEndpoint implements Csw {
                     SERVICE_PROVIDER,
                     OPERATIONS_METADATA,
                     FILTER_CAPABILITIES));
+
+    static final int DEFAULT_BATCH = 500;
 
     private static final List<String> ELEMENT_NAMES = Arrays.asList("brief", "summary", "full");
 
@@ -563,7 +574,7 @@ public class CswEndpoint implements Csw {
         for (UpdateAction updateAction : request.getUpdateActions()) {
             try {
                 numUpdated += updateRecords(updateAction);
-            } catch (CswException | FederationException | IngestException | SourceUnavailableException | UnsupportedQueryException e) {
+            } catch (CswException | FederationException | IngestException | SourceUnavailableException | UnsupportedQueryException | CatalogQueryException e) {
                 LOGGER.debug("Unable to update record(s)", e);
                 throw new CswException("Unable to update record(s).",
                         CswConstants.TRANSACTION_FAILED,
@@ -578,7 +589,7 @@ public class CswEndpoint implements Csw {
         for (DeleteAction deleteAction : request.getDeleteActions()) {
             try {
                 numDeleted += deleteRecords(deleteAction);
-            } catch (CswException | FederationException | IngestException | SourceUnavailableException | UnsupportedQueryException e) {
+            } catch (CswException | FederationException | IngestException | SourceUnavailableException | UnsupportedQueryException | CatalogQueryException e) {
                 LOGGER.debug("Unable to delete record(s)", e);
                 throw new CswException("Unable to delete record(s).",
                         CswConstants.TRANSACTION_FAILED,
@@ -659,29 +670,34 @@ public class CswEndpoint implements Csw {
         queryRequest = queryFactory.updateQueryRequestTags(queryRequest,
                 schemaTransformerManager.getTransformerSchemaForId(deleteAction.getTypeName()));
 
-        QueryResponse response = framework.query(queryRequest);
+        Iterable<List<Result>> resultLists = Iterables.partition(ResultIterable.resultIterable(
+                framework,
+                queryRequest), DEFAULT_BATCH);
+        int batchCount = 1;
+        int deletedCount = 0;
+        for (List<Result> results : resultLists) {
+            Set<String> idsToDelete = results.stream()
+                    .filter(Objects::nonNull)
+                    .map(Result::getMetacard)
+                    .filter(Objects::nonNull)
+                    .map(Metacard::getId)
+                    .collect(Collectors.toSet());
 
-        List<String> ids = new ArrayList<>();
+            if (!idsToDelete.isEmpty()) {
+                DeleteRequestImpl deleteRequest =
+                        new DeleteRequestImpl(idsToDelete.toArray(new String[0]));
 
-        for (Result result : response.getResults()) {
-            if (result != null && result.getMetacard() != null) {
-                ids.add(result.getMetacard()
-                        .getId());
+                LOGGER.debug("Attempting to delete {} metacards from batch {}.",
+                        idsToDelete.size(),
+                        batchCount++);
+                DeleteResponse deleteResponse = framework.delete(deleteRequest);
+
+                deletedCount += deleteResponse.getDeletedMetacards()
+                        .size();
             }
         }
 
-        if (ids.size() > 0) {
-            DeleteRequestImpl deleteRequest =
-                    new DeleteRequestImpl(ids.toArray(new String[ids.size()]));
-
-            LOGGER.debug("Attempting to delete {} metacards. ", ids.size());
-            DeleteResponse deleteResponse = framework.delete(deleteRequest);
-
-            return deleteResponse.getDeletedMetacards()
-                    .size();
-        }
-
-        return 0;
+        return deletedCount;
     }
 
     private int updateRecords(UpdateAction updateAction)
@@ -711,43 +727,56 @@ public class CswEndpoint implements Csw {
             queryRequest = queryFactory.updateQueryRequestTags(queryRequest,
                     schemaTransformerManager.getTransformerSchemaForId(updateAction.getTypeName()));
 
-            QueryResponse response = framework.query(queryRequest);
+            Map<String, Serializable> recordProperties = updateAction.getRecordProperties();
+            Iterable<List<Result>> resultList = Iterables.partition(ResultIterable.resultIterable(
+                    framework,
+                    queryRequest), DEFAULT_BATCH);
+            int batchCount = 1;
+            int updatedCount = 0;
 
-            if (response.getHits() > 0) {
-                Map<String, Serializable> recordProperties = updateAction.getRecordProperties();
-
-                List<String> updatedMetacardIdsList = new ArrayList<>();
-                List<Metacard> updatedMetacards = new ArrayList<>();
-
-                for (Result result : response.getResults()) {
-                    Metacard metacard = result.getMetacard();
-
-                    if (metacard != null) {
-                        for (Entry<String, Serializable> recordProperty : recordProperties.entrySet()) {
-                            Attribute attribute = new AttributeImpl(recordProperty.getKey(),
-                                    recordProperty.getValue());
-                            metacard.setAttribute(attribute);
-                        }
-                        updatedMetacardIdsList.add(metacard.getId());
-                        updatedMetacards.add(metacard);
-                    }
-                }
-
-                if (updatedMetacardIdsList.size() > 0) {
-                    String[] updatedMetacardIds =
-                            updatedMetacardIdsList.toArray(new String[updatedMetacardIdsList.size()]);
-                    UpdateRequest updateRequest = new UpdateRequestImpl(updatedMetacardIds,
-                            updatedMetacards);
-
-                    LOGGER.debug("Attempting to update {} metacards.",
-                            updatedMetacardIdsList.size());
-                    UpdateResponse updateResponse = framework.update(updateRequest);
-                    return updateResponse.getUpdatedMetacards()
-                            .size();
-                }
+            for (List<Result> results : resultList) {
+                updatedCount += updateResultList(recordProperties, batchCount++, results);
             }
+
+            return updatedCount;
         }
         return 0;
+    }
+
+    private int updateResultList(Map<String, Serializable> recordProperties, int batchCount,
+            List<Result> resultList) throws IngestException, SourceUnavailableException {
+        List<String> updatedMetacardIdsList = new ArrayList<>();
+        List<Metacard> updatedMetacards = new ArrayList<>();
+
+        int updatedCount = 0;
+        for (Result result : resultList) {
+            Metacard metacard = result.getMetacard();
+
+            if (metacard != null) {
+                for (Entry<String, Serializable> recordProperty : recordProperties.entrySet()) {
+                    Attribute attribute = new AttributeImpl(recordProperty.getKey(),
+                            recordProperty.getValue());
+                    metacard.setAttribute(attribute);
+                }
+                updatedMetacardIdsList.add(metacard.getId());
+                updatedMetacards.add(metacard);
+            }
+        }
+
+        if (updatedMetacardIdsList.size() > 0) {
+            String[] updatedMetacardIds = updatedMetacardIdsList.toArray(new String[0]);
+            UpdateRequest updateRequest = new UpdateRequestImpl(updatedMetacardIds,
+                    updatedMetacards);
+
+            LOGGER.debug("Attempting to update {} metacards in batch {}.",
+                    updatedMetacardIdsList.size(),
+                    batchCount);
+            UpdateResponse updateResponse = framework.update(updateRequest);
+            updatedCount = updateResponse.getUpdatedMetacards()
+                    .size();
+        }
+
+        return updatedCount;
     }
 
     @GET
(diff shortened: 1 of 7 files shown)


catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -228,6 +235,8 @@
 
     private static final XMLUtils XML_UTILS = XMLUtils.getInstance();
 
+    private static final int DEFAULT_PAGE_SIZE = 20;
```
kcwire: Why 20? Do we have any metrics to attempt to optimize this to some degree? 20 seems small to me but I have no physical evidence to back that up.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -929,10 +946,18 @@ private CswRecordCollection queryCsw(GetRecordsType request) throws CswException
             try {
                 queryRequest = queryFactory.updateQueryRequestTags(queryRequest,
                         request.getOutputSchema());
-                LOGGER.debug("Attempting to execute query: {}", queryRequest);
-                QueryResponse queryResponse = framework.query(queryRequest);
+                queryRequest = setQueryPageSize(queryRequest);
```
kcwire: I believe we have lost the requests original `maxRecords` at this point - which means we will return all results possible even if they have only requested 5 (or any subset).

distribution/test/itests/test-itests-common/src/main/resources/TestSpatial/csw/request/query/CswPagingTestLikeQuery.xml
```diff
@@ -0,0 +1,31 @@
...
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/ -->
+<csw:GetRecords
+        xmlns:ogc="http://www.opengis.net/ogc"
+        xmlns:csw="http://www.opengis.net/cat/csw/2.0.2"
+        resultType="results"
+        outputFormat="application/xml" outputSchema="http://www.opengis.net/cat/csw/2.0.2"
+        startPosition="1" maxRecords="20" service="CSW" version="2.0.2">
```
kcwire: Here we are specifying we only want 20 results - but in the test we verify all 30 are returned - I believe this confirms the `maxRecords` is incorrectly being ignored.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: You're correct. This is being discussed out-of-band and will be addressed either in this, or a separate issue.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: I matched the default used by the paging Iterables. I do not know if there was extensive (or any) empirical study to support that page size.

distribution/test/itests/test-itests-common/src/main/resources/TestSpatial/csw/request/query/CswPagingTestLikeQuery.xml
coyotesqrl: Agreed. Also once fixed, more complete tests will ask for `maxRecords` < `pageSize`, `maxRecords` == `pageSize`, and `maxRecords` > `pageSize` results to confirm all three cases.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -661,30 +670,34 @@ private int deleteRecords(DeleteAction deleteAction)
...
-
-        QueryResponse response = framework.query(queryRequest);
-
-        List<String> ids = new ArrayList<>();
-
-        for (Result result : response.getResults()) {
-            if (result != null && result.getMetacard() != null) {
-                ids.add(result.getMetacard()
-                        .getId());
+        queryRequest = setQueryPageSize(queryRequest);
```
lessarderic: ❓ Deleting metacards 20 at a time seems a little sub-optimal. Should we either increase `DEFAULT_PAGE_SIZE` or provide the page size parameter to the `setQueryPageSIze()` and give a higher number here?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -929,10 +946,18 @@ private CswRecordCollection queryCsw(GetRecordsType request) throws CswException
...
+                ResultIterable results = new ResultIterable(framework, queryRequest);
+                List<Result> resultList = results.stream()
+                        .collect(Collectors.toList());
+
+                QueryResponse queryResponse = new QueryResponseImpl(queryRequest,
+                        resultList,
+                        resultList.size());
                 response.setSourceResponse(queryResponse);
-            } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            } catch (UnsupportedQueryException e) {
```
lessarderic: ❓ Shouldn't we catch `QueryResultException` here and pass the cause as a parameter to the `CswException`?

coyotesqrl:
build now

coyotesqrl:
This PR has been updated to support the changes merged today for `ResultIterable`s. It should be ready for final review.

coyotesqrl:
build now

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -713,43 +725,54 @@ private int updateRecords(UpdateAction updateAction)
             queryRequest = queryFactory.updateQueryRequestTags(queryRequest,
                     schemaTransformerManager.getTransformerSchemaForId(updateAction.getTypeName()));
 
-            QueryResponse response = framework.query(queryRequest);
+            Map<String, Serializable> recordProperties = updateAction.getRecordProperties();
+            Iterable<List<Result>> resultList = Iterables.partition(ResultIterable.resultIterable(
+                    framework,
+                    queryRequest), 64);
+            AtomicInteger batch = new AtomicInteger(1);
```
lessarderic: ✏️  Should we rename `batch` to `batchCount` or `batchNumber` to make it clear that this is _not_ the batch of results?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -713,43 +725,54 @@ private int updateRecords(UpdateAction updateAction)
...
-                List<String> updatedMetacardIdsList = new ArrayList<>();
-                List<Metacard> updatedMetacards = new ArrayList<>();
+            return updatedCount.get();
+        }
+        return 0;
+    }
 
-                for (Result result : response.getResults()) {
-                    Metacard metacard = result.getMetacard();
+    private void updateResultList(Map<String, Serializable> recordProperties, AtomicInteger batch,
```
lessarderic: ✏️  `batchCount` instead of `batch` if we keep the in-out parameters?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -929,10 +953,32 @@ private CswRecordCollection queryCsw(GetRecordsType request) throws CswException
...
+                                .intValue());
+                List<Result> resultList = results.stream()
+                        .collect(Collectors.toList());
+
+                long totalHits = hitCount.get();
+                totalHits = totalHits != 0 ? totalHits : resultList.size();
+
+                QueryResponse queryResponse = new QueryResponseImpl(queryRequest,
+                        resultList,
+                        totalHits);
```
lessarderic: ❓ Any reason we're using `totalHits` instead of just always use `resultList.size()`, which is always guaranteed to be set and accurate?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -929,10 +953,32 @@ private CswRecordCollection queryCsw(GetRecordsType request) throws CswException
...
+
                 response.setSourceResponse(queryResponse);
-            } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+            } catch (UnsupportedQueryException e) {
```
lessarderic: ❓ Shouldn't we catch `CatalogQueryException` here as well?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -713,43 +725,54 @@ private int updateRecords(UpdateAction updateAction)
             queryRequest = queryFactory.updateQueryRequestTags(queryRequest,
                     schemaTransformerManager.getTransformerSchemaForId(updateAction.getTypeName()));
 
-            QueryResponse response = framework.query(queryRequest);
+            Map<String, Serializable> recordProperties = updateAction.getRecordProperties();
+            Iterable<List<Result>> resultList = Iterables.partition(ResultIterable.resultIterable(
+                    framework,
+                    queryRequest), 64);
```
rzwiefel: @lessarderic can a system hold 1000 records per user per cswsource in memory at once?
Though I do agree that since the front end does 250 per source that might make sense to align this with that

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -662,29 +669,34 @@ private int deleteRecords(DeleteAction deleteAction)
...
 
-        QueryResponse response = framework.query(queryRequest);
-
-        List<String> ids = new ArrayList<>();
-
-        for (Result result : response.getResults()) {
-            if (result != null && result.getMetacard() != null) {
-                ids.add(result.getMetacard()
-                        .getId());
+        Iterable<List<Result>> resultLists = Iterables.partition(ResultIterable.resultIterable(
```
rzwiefel: static import `resultIterable`

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -662,29 +669,34 @@ private int deleteRecords(DeleteAction deleteAction)
...
+                framework,
+                queryRequest), DEFAULT_BATCH);
+        AtomicInteger batch = new AtomicInteger(1);
```
coyotesqrl: Mostly just consistency. I need the in-out objects in the `updateRecords()` method that is similar to this. I can change them here if it bothers you.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: I'm guessing on an appropriate size that will definitely not trigger an OOM condition, but I agree that it's probably lower than strictly necessary. @kcwire?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: Yeah. I can do that. I'll update both sets of Atomics.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/TestCswEndpoint.java
```diff
@@ -2039,6 +2048,46 @@ private void setUpMocksForProductRetrieval(boolean includeMimeType)
...
+            MetacardImpl metacard = new MetacardImpl();
+            metacard.setId(i + "");
+            results.add(new ResultImpl(metacard));
+        }
+
+        List<QueryResponse> queryResponses = new ArrayList<>();
+        while (!results.isEmpty()) {
+            List<Result> batchList = new ArrayList<>();
+            for (int i = 0; i < batchSize; i++) {
+                Result result = results.poll();
```
rzwiefel: I'm not sure how I feel about trying to recreate catalog framework logic instead of just mocking the responses verbatim. Something feels wrong with duplicating its' logic.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
lessarderic: I'm not a fan of in-out parameters in general and `Atomic` objects make it feel like the code is threaded somehow. But it's not that big a deal. Up to you.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: I'm sad how long I had to look at this until I remembered the whole reason I added the `QueryFunction` and tracked the `hitCount` with the atomic.

The size of the `resultList` can be limited to `request.getMaxRecords().intValue()` which would lead to an incorrect response for hits.

rzwiefel: Maybe we should leave a comment for future readers! 😄

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: YES. Good catch.

coyotesqrl:
build now

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
coyotesqrl: :man_shrugging: I'm pretty indifferent and inconsistent w/r/t static imports.

coyotesqrl:
@lessarderic For your comment about the page size being set to `-1` by `CswQueryFactory`, it is ignored in the `ResultIterable` here: https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java#L237

Result iteration handles paging internally so the semantics of `-1` aren't used at all.

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
lessarderic: Ah! Make sense now. @rzwiefel is right. This would be a great time to add an inline comment to explain _why_ we're using `totalHits` instead of `resultList.size()`.

coyotesqrl: Comment added.

coyotesqrl:
build now

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestSpatial.java
```diff
@@ -446,21 +488,77 @@ private boolean hasExpectedMetacardsReturned(String queryResult,
...
+        StrSubstitutor strSubstitutor = new StrSubstitutor(ImmutableMap.of("maxRecords",
+                "" + maxRecords));
+
+        strSubstitutor.setVariablePrefix(RESOURCE_VARIABLE_DELIMETER);
+        strSubstitutor.setVariableSuffix(RESOURCE_VARIABLE_DELIMETER);
+        return strSubstitutor.replace(rawCswQuery);
+    }
+
+    private Set<String> ingestPagingRecords(int number) {
+        StringBuilder sb = new StringBuilder();
```
lessarderic: ❓ Any reason we're building the CSW record inline instead of using a resource file and [`getFileContent()`](https://github.com/codice/ddf/blob/master/distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java#L694)? I thought we were trying to avoid manually building XML documents in itests.

coyotesqrl: It was a bit simpler than using a resource with three replacement params, but I could certainly change it.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/TestCswEndpoint.java
```diff
@@ -184,20 +193,20 @@
...
 
-    private static ArgumentCaptor<QueryRequest> argument;
-
     private static final long RESULT_COUNT = 10;
 
     private static final long TOTAL_COUNT = 10;
 
     private Validator validator = mock(Validator.class);
 
+    private List<QueryResponse> qrBatch;
```
coyotesqrl: I can spell it out and fix that capitalization.

coyotesqrl:
I'm not sure of the initial motivation for setting the pageSize to -1 on that factory method. It's been like that for a year and a half and I worry that changing it now would have unintended side effects. I would add a doc explaining why/what it is doing, but since I'm not sure of the motivations, I'm not sure I can even explain or justify it correctly. 

@ryeats do you recall anything about that change? It appears to have been made for DDF-1858.

coyotesqrl:
@lessarderic Added a new test resource - `CswPagingRecord.xml` - to remove the inline creation of the records for itest ingest. Also renamed variable/method as requested in unit test.

coyotesqrl:
build now

ryeats:
In DDF-1858 I just refactored the Query creation logic out of CswEndpoint and into its own class so I could reuse it.  I see that deleteRecords and updateRecords had pageSize set to -1 before, It looks like that should not have been pull out of those methods and into the QueryFactory ...

coyotesqrl:
build now

lessarderic:
So @ryeats, are you ok with @coyotesqrl removing that `setPageSize(-1)` from there then?

coyotesqrl:
build now

PR #2206, opened 2017-08-09
DDF-3205 Expose Solr faceting capabilities to the CatalogFramework

adimka:
#### What does this PR do?
Creates a way for the DDF system to create faceted queries in Solr, and receive back faceting information with query results.

#### Who is reviewing it? 
@albany551 
@brianfelix 
@mackncheesiest
@glenhein 
@bdeining 
@troymohl 
@mojogitoverhere
@rzwiefel 

@codice/core-apis 
@codice/solr 

@millerw8
@pklinef

Full build with unit tests. There's no way with just these changes to create a faceted query in the system, but we should verify that the changes to `SolrMetacardClientImpl` aren't breaking changes by ingesting some data and querying (via the UI is fine).

[DDF-3205](https://codice.atlassian.net/browse/DDF-3205)

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Review Comment Legend:


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetProperties.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Set;
+
+public class FacetProperties implements Serializable {
+
+    public enum SortFacetsBy {INDEX, COUNT}
+
+    private static final int DEFAULT_FACET_LIMIT = 100;
+    private static final int DEFAULT_MIN_FACET_COUNT = 0;
+
+    private Set<String> facetFields;
```
bdeining: I think the formatting is off on this class

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedFieldResult.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param valueCounts A list of the number of occurrences for each facet value
+     */
+    public FacetedFieldResult(String fieldName, List<String> fieldValues, List<Long> valueCounts) {
+        this.fieldName = fieldName;
+        facetValues = new ArrayList<>();
+
+        Iterator<String> valueItr = fieldValues.iterator();
+        Iterator<Long> countItr = valueCounts.iterator();
+
+        while (valueItr.hasNext() && countItr.hasNext()) {
```
bdeining: Should we actually check if the `fieldValues` and `valueCounts` are the same size?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +117,39 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
 
         SolrQuery query = getSolrQuery(request, filterDelegateFactory.newInstance(resolver));
 
+        boolean isFacetedQuery = false;
+        FacetProperties textFacetsProp = (FacetProperties) request.getPropertyValue(FACET_FIELDS_KEY);
```
bdeining: Should we check if the result of `getPropertyValue` is actually a `FacetProperties` before casting it?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedFieldResult.java
adimka: When I chose to do it this way, I considered what to do in the case that they weren't the same size. Failing to create the object seemed extreme, so doing as close to what is expected as possible seemed the right choice. I think we could get some value out of logging to debug in that case though.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
adimka: Probably for safety. This should really never be the case, but better safe than sorry.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +117,40 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
 
         SolrQuery query = getSolrQuery(request, filterDelegateFactory.newInstance(resolver));
 
+        boolean isFacetedQuery = false;
```
mackncheesiest: Might consider adding a test that mocks out a Solr response that includes faceting to make sure all this new logic is good.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedFieldResult.java
bdeining: I agree that failing completely is extreme; I agree with a logging statement

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/faceting/FacetedFieldResultSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package ddf.catalog.operation.faceting
+
+import spock.lang.Specification
+
+class FacetedFieldResultSpec extends Specification {
+    static testSameLenValues = ["keyboards", "cutting boards", "school boards", "skateboards"]
+    static testShorterValues = ["apples", "oranges"]
+    static testLongerValues = ["barbecued", "boiled", "broiled", "sauted", "kabob'd", "creole",
```
mackncheesiest: 🦐  🍤

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +117,40 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
...
+        Serializable textFacetPropRaw = request.getPropertyValue(EXPERIMENTAL_FACET_FIELDS_KEY);
+
+        if(textFacetPropRaw != null && textFacetPropRaw instanceof FacetProperties) {
+            FacetProperties textFacetProp = (FacetProperties) textFacetPropRaw;
+            isFacetedQuery = true;
+
+            textFacetProp.getFacetFields().stream()
+                    .map(facet -> !facet.endsWith("_txt") ? facet + "_txt" : facet)
```
glenhein: can you make `_txt` a constant?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetProperties.java
```diff
@@ -0,0 +1,87 @@
...
+    public FacetProperties(Set<String> facetProperties, SortFacetsBy sortKey, int facetLimit,
+            int minFacetCount) {
+        this.facetFields = facetProperties;
+        this.sortKey = sortKey;
+        this.facetLimit = facetLimit;
+        this.minFacetCount = minFacetCount;
+    }
+
+    public Set<String> getFacetFields() {
+        return facetFields;
```
adimka: Probably a good idea, I don't think we want people editing the Set through this.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetValueCount.java
```diff
@@ -0,0 +1,32 @@
...
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.operation.faceting;
+
+public class FacetValueCount {
+    private String value;
+    private long count;
+
+    FacetValueCount(String value, long count) {
```
adimka: After some tinkering, it seems that auto-boxing makes the clunkiness I was afraid of using `long` a non-factor. Changed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedFieldResult.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param fieldName The field name for which faceting data is reported
+     * @param fieldValues A list of the discovered facet values
+     * @param valueCounts A list of the number of occurrences for each facet value
+     */
+    public FacetedFieldResult(String fieldName, List<String> fieldValues, List<Long> valueCounts) {
+        this.fieldName = fieldName;
+        facetValues = new ArrayList<>();
+
+       if (fieldValues.size() != valueCounts.size()) {
```
adimka: I'll add a second constructor for this. I want it to be as simple as possible for whoever has the data they're inputting to convert to our format. If they have Lists, I don't want them to have to convert to Pairs - and vice versa. We can do that work for them. 👍

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
mackncheesiest: (By which I mean having tests in place to catch future regressions is always good)

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetValueCount.java
adimka: Actually, upon looking again, Solr takes in an `int` for this and returns a `long` there's no way to guarantee we don't lose data if we go from `long` to `int` when passing this in, so I think we'll need to use the same mismatched types they do. I don't care for this coupling, but I'm not sure there's a way around it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedFieldResult.java
```diff
@@ -0,0 +1,67 @@
...
+package ddf.catalog.operation.faceting;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FacetedFieldResult {
```
pklinef: ❗️ This should be marked as experimental.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedFieldResult.java
```diff
@@ -0,0 +1,67 @@
...
+     * query. A FacetedFieldResult is representative of a single attribute's faceting results,
+     * and zero to many FacetedFieldResults may make up a complete faceted query result.
+     * This constructor zips together the fieldValues and valueCounts provided, and these list
+     * should correspond and be of the same length if sane results are desired.
+     *
+     * @param fieldName The field name for which faceting data is reported
+     * @param fieldValues A list of the discovered facet values
+     * @param valueCounts A list of the number of occurrences for each facet value
+     */
+    public FacetedFieldResult(String fieldName, List<String> fieldValues, List<Long> valueCounts) {
```
pklinef: ❗️ Fields is a Solr concept.  Attribute should be used in a catalog framework context.

catalog/core/catalog-core-api/src/main/java/ddf/catalog/Constants.java
```diff
@@ -168,4 +168,8 @@
     public static final String ATTRIBUTE_UPDATE_MAP_KEY = "attributeUpdateMap";
 
     public static final String STORE_REFERENCE_KEY = "storeReference";
+
+    public static final String EXPERIMENTAL_FACET_FIELDS_KEY = "facet-fields";
```
pklinef: ❗️ Should be attributes instead of fields here too.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +117,40 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
 
         SolrQuery query = getSolrQuery(request, filterDelegateFactory.newInstance(resolver));
 
+        boolean isFacetedQuery = false;
+        Serializable textFacetPropRaw = request.getPropertyValue(EXPERIMENTAL_FACET_FIELDS_KEY);
+
+        if(textFacetPropRaw != null && textFacetPropRaw instanceof FacetProperties) {
```
pklinef: ✏️  Make sure to run the code formatter.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
pklinef: It already is a constant.

https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SchemaFields.java#L44

pklinef: ❗️ You should use the metacard type and `DynamicSchemaResolver` to figure out the attribute type and its format.  You cannot assume that the given attribute to facet will be a String attribute.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetValueCount.java
pklinef: We could make it long for both cases but warn about data lose in the Solr provider if they set the limit using something greater than an int.  I could see Solr changing this in the future to support long in Solrj.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/faceting/FacetedFieldResultSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+
+package ddf.catalog.operation.faceting
+
+import spock.lang.Specification
+
+class FacetedFieldResultSpec extends Specification {
```
pklinef: ❗️ Add facet tests to `SolrProviderTest` as well to verify proper integration and regression testing with Solrj and Solr.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetValueCount.java
adimka: The incoming value is the minimum count required in order for the faceting information to be reported, which is different from the outgoing value - the number of results reported.
An `int` seems appropriate.

adimka:
build now

adimka:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedAttributeResult.java
```diff
@@ -0,0 +1,83 @@
...
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.operation.faceting;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import org.apache.commons.lang3.tuple.Pair;
```
rzwiefel: Btw using a third party library (like commons-lang3) inside of this jar (catalog-core-api-impl) will require each consuming jar to embed commons-lang3 unless we have it exported already in our container. (I thought we only had lang2 exported but I could be mistaken).

bdeining: To my knowledge, it's `commons-lang3`

rzwiefel: Any reason we don't just use a data container class (eg, `ValueCount` or something) instead of using the Pair ?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedAttributeResult.java
```diff
@@ -0,0 +1,83 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class FacetedAttributeResult {
```
rzwiefel: ❗️ Missing class javadoc
(are we marking these experimental as well?)

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetValueCount.java
```diff
@@ -0,0 +1,32 @@
...
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.catalog.operation.faceting;
+
+public class FacetValueCount {
```
rzwiefel: ❗️ class javadoc missing
Also, experimental?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +118,43 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
 
         SolrQuery query = getSolrQuery(request, filterDelegateFactory.newInstance(resolver));
 
+        boolean isFacetedQuery = false;
+        Serializable textFacetPropRaw = request.getPropertyValue(EXPERIMENTAL_FACET_PROPERTIES_KEY);
+
+        if (textFacetPropRaw != null && textFacetPropRaw instanceof FacetProperties) {
+            FacetProperties textFacetProp = (FacetProperties) textFacetPropRaw;
+            isFacetedQuery = true;
```
rzwiefel: Might consider adding a log statement here? Something really low like trace, and make sure to put a isTraceEnabled check on it so we don't performance suffer from it in prod

```java
if (LOGGER.isTraceEnabled()) {
    LOGGER.trace("Enabling faceted query for request [{}] on field {}", request, textFacetProp);
}

```

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedAttributeResult.java
rzwiefel: Just making sure we're fine tying this jar to an external dependency that is not shadowed in 👍

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +118,43 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
...
+
+            textFacetProp.getFacetAttributes()
+                    .stream()
+                    .map(this::addAttributeTypeSuffix)
+                    .filter(attr -> attr.contains(String.valueOf(FIRST_CHAR_OF_SUFFIX)))
+                    .forEach(query::addFacetField);
+
+            query.setFacetSort(textFacetProp.getSortKey().name());
+            query.setFacetLimit(textFacetProp.getFacetLimit());
+            query.setFacetMinCount(textFacetProp.getMinFacetCount());
```
rzwiefel: :question: Do we not need to set FacetMaxCount?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -110,14 +118,43 @@ public SourceResponse query(QueryRequest request) throws UnsupportedQueryExcepti
...
         SolrQuery query = getSolrQuery(request, filterDelegateFactory.newInstance(resolver));
 
+        boolean isFacetedQuery = false;
+        Serializable textFacetPropRaw = request.getPropertyValue(EXPERIMENTAL_FACET_PROPERTIES_KEY);
+
+        if (textFacetPropRaw != null && textFacetPropRaw instanceof FacetProperties) {
+            FacetProperties textFacetProp = (FacetProperties) textFacetPropRaw;
+            isFacetedQuery = true;
+
+            textFacetProp.getFacetAttributes()
```
rzwiefel: :question: Do we have any guarantees that `textFacetProp` is not null before being used here?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedAttributeResult.java
adimka: I didn't want you to have to massage your result data coming from the database implementation if I could avoid it. My goal using Pair here is to be as generic as possible and do most of the work. If Java had a base tuple this could be avoided.
I'll check on the export.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedQueryRequest.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param isEnterprise Specifies if this FacetedQueryRequest is an enterprise query
+     * @param sourceIds A list of sources to query
+     * @param properties Properties supplied to this query for auth, transactions, etc
+     * @param facetProperties A structure describing the desired faceting constraints
+     */
+    public FacetedQueryRequest(Query query, boolean isEnterprise, Collection<String> sourceIds,
+            Map<String, Serializable> properties, FacetProperties facetProperties) {
+        super(query, isEnterprise, sourceIds, properties);
+
+        this.properties = new HashMap<>(this.properties);
```
rzwiefel: :pencil: How about copying the properties before passing into the `super` above? eg line 61 : `super(query, isEnterprise, sourceIds, new HashMap<>(properties);`. then you can get rid of this line completely since the super ctor autmatically sets the properties for you.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedAttributeResult.java
adimka: I'll add the Javadoc.
I had a discussion about this with Phil. Since we're not adding this to 'catalog-core-api` for the time being, none of it needs to be marked experimental. I figured we could mark the primary "interface-ish" classes as Experimental so no one goes off using on them without expecting change.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetValueCount.java
adimka: Same as above, I'll add javadoc

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedAttributeResult.java
rzwiefel: Cool 👍

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
adimka: This value is the minimum number of occurrences a facet value needs in order to be reported.
`query.setFacetMinCount(x)` says "If it doesn't show up at least x times, don't even tell me about it". A max really has no meaning in this context.
The facet limit is the maximum number of different facet values to return.
`query.setFacetLimit(x)` says "Give me the top x values found in this field" which might be what you're thinking here, in which case, we're doing it.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
adimka: Nope. Just fixed in `FacetProperties` constructor. Good catch.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
adimka: Nope. Just fixed in FacetProperties constructor. Good catch.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedQueryRequest.java
adimka: Hmm, looks like doing this makes FindBugs unhappy.
I'll leave it as is.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
rzwiefel: Ah that make sense. Good to know :+1:

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/faceting/FacetedQueryRequest.java
rzwiefel: What does findbugs say about it??

adimka:
build now

bdeining:
build now

adimka:
build now

adimka:
build now

adimka:
build now

adimka:
build now

adimka:
build now

PR #2207, opened 2017-08-09
DDF-3034 Added a max result count && condensed ResultListIterable

rzwiefel:
#### What does this PR do?
- Adds a max result count limit onto the ResultIterable class so that callers may specify the total number of results they wish to get back over the set of all results pertaining to their query
- refactored the resultl into resultiterable
#### Who is reviewing it? 
anyone

@lessarderic 
@coyotesqrl

https://codice.atlassian.net/browse/DDF-3034

#### Review Comment Legend:


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -76,21 +112,38 @@ public ResultIterable(CatalogFramework catalogFramework, QueryRequest queryReque
      *                      results.
      * @param queryRequest  request used to retrieve the results.
      */
-    public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest) {
+    public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest,
+            int maxResultCount) {
         notNull(queryFunction, "Query function cannot be null");
         notNull(queryRequest, "Query request cannot be null");
```
rzwiefel: yes

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -76,21 +112,38 @@ public ResultIterable(CatalogFramework catalogFramework, QueryRequest queryReque
      *                      results.
      * @param queryRequest  request used to retrieve the results.
      */
-    public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest) {
+    public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest,
```
rzwiefel: yes

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -61,11 +68,40 @@
...
+     */
+    public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest) {
+        this(queryFunction, queryRequest, 0);
+    }
+
+    /**
+     * Creates an iterable that will call the {@link CatalogFramework} to retrieve the results
+     * that match the {@link QueryRequest} provided.
+     *
+     * @param catalogFramework reference to the {@link CatalogFramework} to call to retrieve the
```
rzwiefel: yes

lessarderic: Still need to mention that `0` will indicate no maximum. Should we consider having a constant for it, e.g., `NO_MAX_RESULT_COUNT`?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -77,20 +80,75 @@ public ResultIterable(CatalogFramework catalogFramework, QueryRequest queryReque
...
+    }
+
+    /**
+     * Creates an iterable that will call a {@link QueryFunction} to retrieve the results
+     * that match the {@link QueryRequest} provided.
+     *
+     * @param queryFunction  reference to the {@link QueryFunction} to call to retrieve the
+     *                       results.
+     * @param queryRequest   request used to retrieve the results.
+     * @param maxResultCount maximum number of results in total to query for
```
lessarderic: Same comment as above.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -77,20 +80,75 @@ public ResultIterable(CatalogFramework catalogFramework, QueryRequest queryReque
...
+     * @param catalogFramework reference to the {@link CatalogFramework} to call to retrieve the
+     *                         results.
+     * @param queryRequest     request used to retrieve the results.
+     * @param maxResultCount   maximum number of results in total to query for
+     */
+    public ResultIterable(CatalogFramework catalogFramework, QueryRequest queryRequest,
+            int maxResultCount) {
+        notNull(catalogFramework, "CatalogFramework cannot be null");
+        notNull(queryRequest, "Query request cannot be null");
+        Validate.isTrue(maxResultCount >= 0,
```
lessarderic: ✏️  static import `Validate.isTrue()`?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -77,20 +80,75 @@ public ResultIterable(CatalogFramework catalogFramework, QueryRequest queryReque
...
 
     @Override
     public Iterator<Result> iterator() {
+        if (maxResultCount > 0) {
+            return limit(new QueryResultIterator(queryFunction, queryRequest), maxResultCount);
+        }
         return new QueryResultIterator(queryFunction, queryRequest);
     }
 
+    /**
```
lessarderic: ❓ If the purpose of this method was to replace `ResultListIterable`, it should return an `Iterable` so clients can use `spliterator()` and `forEach()` on it, but I'm not sure returning an `Iterable` from an `Iterable` would make a lot of sense. Keeping them separate still seems to make the most sense to me, but I'm curious to see what the other reviewers think.

lessarderic: ❗️ If we decide to keep this method, we should provide a description for this method.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -190,4 +248,26 @@ private void copyQueryRequestAndQuery(QueryRequest queryRequest) {
...
+        }
+
+        @Override
+        public boolean hasNext() {
+            return iterator.hasNext();
+        }
+
+        @Override
+        public List<Result> next() {
+            return stream(iterator).limit(pageSize)
```
rzwiefel: ❗️ Need to double check that streaming over the iterator will not consume more than the `limit` in any scenario. And streaming multiple times over the same iterator causes no ill effects.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -190,4 +248,26 @@ private void copyQueryRequestAndQuery(QueryRequest queryRequest) {
                     queryRequest.getProperties());
         }
     }
+
+    private static class ResultListIterator implements Iterator<List<Result>> {
```
lessarderic: ✏️  If we keep this here, should we keep the iterator inner-class names consistent, i.e., name this one `QueryResultListIterator` or name the other one `ResultIterator`?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -147,7 +149,7 @@ class QueryResultPaginatorSpec extends Specification {
         1.upto(nextCalls, { results << resultIterator.next() })
 
         then:
-        1.upto(nextCalls, { results.get(it - 1).distanceInMeters == (Double) it })
+        1.upto(nextCalls, { assert results.get(it - 1).distanceInMeters == (Double) it })
```
lessarderic: Thanks for fixing this @rzwiefel !

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -195,31 +197,98 @@ class QueryResultPaginatorSpec extends Specification {
         1.upto(resultSize + 1, { results << resultIterator.next() })
 
         then:
-        1.upto(resultSize + 1, { results.get(it - 1).distanceInMeters == (Double) it })
+        1.upto(resultSize + 1, { assert results.get(it - 1).distanceInMeters == (Double) it })
+    }
+
+    def "Properly pages when default page size is used with no maxResults"() {
```
rzwiefel: this unit test can be removed since it is mostly a duplicate of the more thorough one below

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: Oh yes I could make it return an iterable instead of an iterator. That would make sense if someone is using it in a for-each. 

Though a counter point, If a client needs a spliterator would they not simply use the spliterator over the results? Wouldn't a spliterator over Lists of results be slightly redundant?

In my eyes it just seems that a large majority of the code is needlessly copied directly. As seen by this implementation there is hardly any real logic needed to produce the functionality that is desired. I'm not sure what we gain by duplicating a large amount of code and keeping it in a separate class. In order to apply the maxResultCount change in it's current state I would essentially need to duplicate my logic in both classes. However in the new condensed version of the ResultListIterator it gets this for almost no effort.

 I know that ResultIterable has already been used extensively. It seems that the most common case is simply for this plain iterator over results. What factored in for the need of creating the List version of this iterable initially? 

In response to the Iterable returning an Iterable, Do we gain much by having the ResultListIterable as a top level class as opposed to just being able to access it via a method (`.listIterator`)-- Especially if it is not frequently used? Forcing to people do have an extra call on the class for a .listIterator won't make me lose much sleep at night, especially with the simplicity it introduces. If the need for it as a top level class increases , we could always split it out and duplicate it again

coyotesqrl: I'm with @rzwiefel on this. Other than two javadoc links in `CatalogFramework`, the first usage of `ResultListIterable` is in my PR #2204 to do batch delete/update in the `CswEndpoint`. So in terms of usage frequency and utility, it's relatively low (which means removing it won't impact much code as well). The fact that the two classes have a lot of similarity in their contents vs. this compact implementation also makes me believe this is the better solution.

If the low usage rate for the `ResultListIterable` continues as expected, I can certainly foresee a day when someone would make an enhancement to `ResultIterable` and neglect to duplicate the logic into its sister class. And even if no one ever forgets to duplicate logic as needed, the fact that it is required says to me that we should collapse/refactor for simplicity.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: Fixed. Renamed to `ResultIterator` to be consistent with the outer class name.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: I would rather that someone use the simplified constructor that has no maxResultCount if they do not wish to impose one. 

I have refactored the constructors and javadoc to make this clear.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -48,53 +52,117 @@
...
     public Iterator<Result> iterator() {
-        return new QueryResultIterator(queryFunction, queryRequest);
+        if (maxResultCount > 0) {
+            return limit(new ResultIterator(queryFunction, queryRequest), maxResultCount);
+        }
+        return new ResultIterator(queryFunction, queryRequest);
+    }
+
+    /**
+     * @param pageSize the maximum page size for each list segment
```
lessarderic: ❗️ I think I know what's bugging me with having this method here instead of having a top level class. Even though it reduces the number of top-level classes, it feels like it goes against the Single Responsibility Pattern and will make it more difficult for people that don't know the code where to find this paging iterator. If I'm looking for something to return me batches of `Result` objects, I'm not sure I would think to look for a method on this class. However, if I search for `ResultIterator`, both classes would pop-up.

I asked @paouelle's opinion and he also felt that this method didn't belong here. His suggestion was instead to make the `ResultListIterable` more reusable and treat it like a utility class that returns batches of objects of a generic type.

I took a quick look at Guava's `Iterables.partition()` and it looks like we should be able to use that instead of writing our own batch iterator class. Based on their documentation (https://github.com/google/guava/wiki/CollectionUtilitiesExplained#iterables), most `Iterables` are lazy and will not advance until needed, which means that if we were to give our `ResultIterable` to `partition()`, it should only fetch the next batch of results when `next()` is called on the returned `Iterator`. Which is confirmed by looking at the implementation: https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Iterators.java#L620-L649.

So my recommendation would be to get rid of all the `ResultListIterable` and all the code related to batching and add a note telling people to use a utility such as `Iterables.partition()` to do batching.

rzwiefel: Good idea, done!

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -48,53 +52,117 @@
...
-     * that match the {@link QueryRequest} provided.
+     * that match the {@link QueryRequest} provided. There will be no limit to the number of
+     * results returned.
      *
      * @param queryFunction reference to the {@link QueryFunction} to call to retrieve the
      *                      results.
      * @param queryRequest  request used to retrieve the results.
      */
-    public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest) {
+    public static ResultIterable resultIterable(QueryFunction queryFunction,
```
coyotesqrl: @rzwiefel and I had discussed this last week and I suggested that the factory methods would probably be a bit cleaner than multiple ctors with an init method. It allows us to keep the `final` fields (not the biggest deal in the world, of course) and sets us up for any future customizations either through more factory methods or by adding a Builder. It was a merely a preference of mine and certainly isn't necessary; if the impact to other work is going to be too large we definitely don't have to make it.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: Also 0 is no longer an allowed value (externally). All callers must now pass a positive integer OR use the one that does not take a cap. This way the 0 remains an internal detail

rzwiefel:
@lessarderic @coyotesqrl 
Updated.
 - Refactored constructors to static factory methods
 - Removed all 'List' classes in favor of using Guavas `Iterables.partition`
 - Resolved code style comments

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
coyotesqrl: Oh...also, it lets us hide the use of `0` internally.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
lessarderic: ❓ Do we still want to remove this test?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -188,38 +192,105 @@ class QueryResultPaginatorSpec extends Specification {
...
+        setup:
+        def actualResults = (1..25).collect { new ResultImpl() }
+        def totalResults = actualResults.size()
+
+        // pageSize = 20
+        expectedQueriesMade * catalogFramework.query(_ as QueryRequest) >>
+                { QueryRequest queryRequest -> buildQueryResponse(actualResults[0..19]) } >>
+                { QueryRequest queryRequest -> buildQueryResponse(actualResults[20..24]) } >>
+                { QueryRequest queryRequest -> buildQueryResponse([]) }
+
```
lessarderic: ✏️  Why the extra blanks lines?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -188,38 +192,105 @@ class QueryResultPaginatorSpec extends Specification {
...
+        QueryRequest queryRequestMock = createQueryRequestMock(queryMock)
+
+        ResultIterable resultIterable = resultIterable(catalogFramework, queryRequestMock)
+
+        when:
+        def results = resultIterable.stream().collect(Collectors.toList())
+
+
+        then:
+        results.size() == 25
```
lessarderic: ❓ If we keep this test, would it be safer and more consistent to verify that all the `distanceInMeters` values are as expected?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -188,38 +192,105 @@ class QueryResultPaginatorSpec extends Specification {
...
+                .collect(Collectors.toList())
+
+        then:
+        def resultCount = Math.min(expectedResultCount, totalResults)
+        // Either expect the maxResults it was capped to, or the actual result count if it was less
+        queryResults.size() == resultCount
+        queryResults == actualResults.subList(0, resultCount)
+
+        where:
+        expectedResultCount | expectedQueriesMade
```
lessarderic: ✏️  Rename `expectedResultCount` to `maxResultCount`?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -188,38 +192,105 @@ class QueryResultPaginatorSpec extends Specification {
...
+
+        when:
+        def queryResults = resultIterable.stream()
+                .collect(Collectors.toList())
+
+        then:
+        def resultCount = Math.min(expectedResultCount, totalResults)
+        // Either expect the maxResults it was capped to, or the actual result count if it was less
+        queryResults.size() == resultCount
+        queryResults == actualResults.subList(0, resultCount)
```
lessarderic: ❓ Is this equivalent to comparing the `distanceInMeters` like we do in the other tests? If so, maybe we should use this approach instead of the one I've used for the other tests. I like this one better :-)

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: Fixed. *We need an external formatter..*

coyotesqrl:
You should update the PR title/description to reflect that this is no longer a WIP.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: I saved the actual metacards that were created so I check equality based on it returning the same metacards in the mock. I'll update the other tests to use this as well as you mentioned below https://github.com/codice/ddf/pull/2207#discussion_r133031684)

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: Updated

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: done

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -155,18 +212,22 @@ private void fetchNextResults() {
...
-                    results = RESULTS_EXHAUSTED;
+                    finished = true;
                     return;
                 }
 
                 results = response.getResults()
                         .iterator();
                 currentIndex += response.getResults()
                         .size();
+
```
lessarderic: ❗️ This will break when `getHits()` returns `-1`.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -109,7 +162,11 @@ public ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest) {
 
         private Iterator<Result> results = Collections.emptyIterator();
 
-        QueryResultIterator(QueryFunction queryFunction, QueryRequest queryRequest) {
+        SourceResponse response;
```
lessarderic: ❗️ Not sure why we made this a member variable.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
lessarderic: ❓ Since we don't expose the total number of hits anywhere, and since `getHits()` could return `-1` (or could even vary between requests), should we even worry about this condition and simply keep querying until the result list is empty?

rzwiefel: We can ensure `getHits() != -1 && currentIndex > response.getHits()` and if we're worried about it changing then we can store the first `getHits` we receive. Would that alleviate this issue?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -128,104 +132,157 @@ class QueryResultPaginatorSpec extends Specification {
...
+    }
 
-        where:
-        nextCalls | pageSize | resultSize | expectedQueries
-        1         | 1        | 1          | 1
-        2         | 1        | 2          | 1
-        3         | 1        | 1          | 3
-        2         | 2        | 1          | 2
-        2         | 0        | 1          | 2
+    QueryResponse buildDynamicQueryResponse(List<Result> actualResults, Queue<Result> resultQueue, QueryRequest) {
```
lessarderic: ❓ Move after all the test methods?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
lessarderic: ❓How about `(response.getHits() => 0) && (currentIndex > response.getHits())`? Also, should we always set the `requestsTotalResultsCount` flag here https://github.com/codice/ddf/pull/2207/files#diff-9ee238e3d9003aab5d04411cde4ca2e6R245 to `true`?

lessarderic: ❗️We should add a unit test for the case where `getHits() < 0`.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: Oh this was just leftover cruft, need to delete

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: screenshot 2017-08-15 07 02 33
I changed that yesterday, just must have missed it in the commit

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: leftover cruft from something I tried.
removed.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: added a test.

rzwiefel:
@lessarderic updated

rzwiefel:
build now

rzwiefel:
build now

rzwiefel:
Build now

coyotesqrl:
build now

rzwiefel:
build now

rzwiefel:
build now

rzwiefel:
build now

PR #2209, opened 2017-08-10
DDF-3209 Create a plugin that can be configured to expand metacard attributes on ingest

vinamartin:
#### What does this PR do?
Creates a plugin that can be configured to expand metacard attributes on ingest. 
If metacards are ingested with attributes that are incorrect or abbreviated, this plugin can map them to the correct or desired value. 

#### Who is reviewing it? 
@josephthweatt (Author)
@emmberk @emanns95 @brendan-hofmann @AzGoalie @mcalcote @clockard 

@coyotesqrl 
@lessarderic

Update the ddf-metacard-attribute-ruleset.cfg with the following form:
`<attribute name>:<original value>:<expanded value>`
Ingest a metacard and inspect its attributes using Intrigue, select an attribute to override and add its value to that cfg file. Then delete the metacard and re-ingest the metacard and verify in Intrigue that the metacard attribute has been expanded. 

[DDF-3209](https://codice.atlassian.net/browse/DDF-3209)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:


catalog/plugin/catalog-plugin-configuration-mapping/src/test/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPluginTest.java
```diff
@@ -0,0 +1,184 @@
...
+        Test that if there exists an expansion for a metacard attribute that it is expanded.
+    */
+    @Test
+    public void testProcessMetacardMultipleAttributesWithExpansion() throws Exception {
+
+        CreateRequest input = new CreateRequestImpl(Arrays.asList(metacard1));
+
+        CreateRequest output = configurationMappingPlugin.process(input);
+
+        assert (output.getMetacards()
```
bdeining: why `assert`?  These really ought to be `assertThat(actual, is(expected));`

catalog/plugin/catalog-plugin-configuration-mapping/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,29 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
+        xsi:schemaLocation="http://www.osgi.org/xmlns/blueprint/v1.0.0 http://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd">
+
+    <reference id="expansionService" interface="ddf.security.expansion.Expansion"
+               filter="(mapping=security.metacard.attribute.mapping)"/>
+
+    <bean id="configurationMappingPlugin"
+          class="org.codice.ddf.catalog.plugin.configuration.mapping.ConfigurationMappingPlugin">
+        <property name="expansionService" ref="expansionService"/>
```
bdeining: Should there be an associated metatype for this plugin so that it can be configured through the Admin UI?

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public DeleteRequest process(DeleteRequest input)
+            throws PluginExecutionException, StopProcessingException {
+        return input;
+    }
+
+    public void setExpansionService(Expansion expansionService) {
+        this.expansionService = expansionService;
```
vinamartin: No reason.

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public UpdateRequest process(UpdateRequest input)
+            throws PluginExecutionException, StopProcessingException {
+        List<Metacard> metacards = new ArrayList<>();
+
+        for (Map.Entry<Serializable, Metacard> metacardEntry : input.getUpdates()) {
+            metacards.add(metacardEntry.getValue());
+        }
+
+        expandAttributes(metacards);
```
rzwiefel: up to you, but if you wish to do this functionally you could write it as, 
```java
input.getUpdates()
    .stream()
    .map(Map.Entry::getValue)
```

catalog/plugin/catalog-plugin-configuration-mapping/src/main/resources/OSGI-INF/blueprint/blueprint.xml
vinamartin: I've been going back and forth on this one. 
We were trying to use expansion service as it existed and since it already had the config file we thought we should use that. It builds the map for us which was helpful. 
If we provide a metatype, we're going to have to be able to restrict the input to some format we can understand and then parse it ourselves (which to be fair, is not the worst considering we still have to look through the map and all that). 
I could see it working either way.

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
josephthweatt: @vinamartin did we want to create the list functionally?

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public void setExpansionService(Expansion expansionService) {
+        this.expansionService = expansionService;
+    }
+
+    private void expandAttributes(List<Metacard> metacards) {
+        for (Map.Entry<String, List<String[]>> expansion : expansionService.getExpansionMap()
+                .entrySet()) {
+            String attribute = expansion.getKey();
+            Map<String, String> expansionMap = toMap(expansion.getValue());
```
rzwiefel: I'm definitely not a fan of the list of string[] stuff. 

Was it not possible here to use the original map to do the mapping transformation, or did you encounter problems?

Maybe even instead of having the service make available the data and pass to the callers, how about having the service take in the metacard and do the transformations itself, that way it doesn't force the callers to process them?

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
vinamartin: Fine by me. 👍

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
```diff
@@ -0,0 +1,94 @@
...
+
+            for (Metacard metacard : metacards) {
+                Attribute metacardAttribute = metacard.getAttribute(attribute);
+
+                if (metacardAttribute != null
+                        && expansionMap.containsKey(metacardAttribute.getValue())) {
```
emanns95: What about multi-valued attributes? It looks like here if we have multiple entries and the first one (which is what `getValue` will return) matches the replacement, we're replacing all of them. So we could miss replacements or lose other data

josephthweatt:
@vinamartin we need to rename this PR to say DDF-3209, its 3029 rn

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
```diff
@@ -60,35 +67,77 @@ public DeleteRequest process(DeleteRequest input)
...
+                    attributeRuleSet.keySet()));
+
+            for (String attributeName : commonAttributes) {
+                Attribute metacardAttribute = metacard.getAttribute(attributeName);
+
+                if (metacardAttribute != null) {
+
+                    // expandTo holds the value that the ruleset
+                    String expandTo = getCorrespondingValue(attributeName,
+                            metacardAttribute.getValue());
```
emanns95: Unless I'm missing something, I think we need to loop through all the possible values in a metacard attribute since it can be a collection.

Say you want to expand `contact.name b -> bar`.
And the metacard attribute for `contact.name` contains a collection `a, b, c`. 

It will only check if `a` is in the `attributeRuleSet`, not expanding `b` to `bar`. This is per the `Attribute` interface javadoc, which states for `getValue` that `this method must return the first, if any, value`. If we instead use `getValues`, we can `getCorrespondingValue` on each of those to make sure we check everything and reach `b`. There should probably also be a test for this case.

vinamartin: Fair, I think one thing we didn't realize before is that the interface actually already supports this and we don't have to iterate over it ourselves. I'm currently testing it now.

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
vinamartin: I think the reason we were doing this is due to some confusion with the javadoc for the expansion interface. There was no reason we actually ever needed to do this. 
I'm refactoring it to do it the proper way.

brendan-hofmann: Perhaps we should clarify the javadoc on the expansion service while we're here? 😃

catalog/plugin/catalog-plugin-configuration-mapping/src/main/resources/OSGI-INF/blueprint/blueprint.xml
brendan-hofmann: Documenting the expected input format should be sufficient, we don't extensively validate that in other configurable areas.

A downside of using the expansion service is that those values are used elsewhere as well. It may not always be desirable or intended to declare a value as acceptable just so that it can be rewritten into the value that is actually desired. If the expansion service's internal logic has value, perhaps it should be refactored into a library that both it and this plugin can use?

catalog/plugin/catalog-plugin-configuration-mapping/src/main/java/org/codice/ddf/catalog/plugin/configuration/mapping/ConfigurationMappingPlugin.java
```diff
@@ -0,0 +1,143 @@
...
+
+            Set<String> commonAttributes = new HashSet<>(CollectionUtils.intersection(attributeSet,
+                    attributeRuleSet.keySet()));
+
+            for (String attributeName : commonAttributes) {
+                Attribute metacardAttribute = metacard.getAttribute(attributeName);
+
+                if (metacardAttribute != null) {
+                    HashSet attributeValues = new HashSet<String>();
+                    attributeValues.addAll(metacardAttribute.getValues());
```
emanns95: I feel like we should still add a test to verify that given a multi-valued attribute, only the value matching the expansion gets replaced. It looks good to me though 👍

vinamartin: Yup I agree, adding that test now

catalog/plugin/catalog-plugin-configuration-mapping/src/main/resources/OSGI-INF/blueprint/blueprint.xml
vinamartin: It looks like if I wanted to create my config file for this I could do something similar to here: https://github.com/codice/ddf/blob/master/platform/security/expansion/security-expansion-metacardattr-map/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L24 Thoughts?

brendan-hofmann: Pro: avoids potentially complex configuration in our admin UI.
Con: requires admins to create/modify an external file to configure the behavior.

It looks like we'd be more internally consistent (and it's easier) to implement this in a similar way, with an eye towards code reuse. 👍

vinamartin: 👍  Thankfully this is easy enough to leverage in our own way without having to write our own implementation.

vinamartin:
Closing out this PR in favor of a better solution

PR #2213, opened 2017-08-13
DDF-2969 Updating Jenkinsfile for various build improvements

mcalcote:
#### What does this PR do?
Specifying JAVA_HOME for windows builds, adjusting cron statement to avoid overlapping deployments, adding retry to all scm operations.

#### Who is reviewing it? 
@oconnormi @LinkMJB 

@codice/build 

@clockard @shaundmorris

Run the pipeline build in a test jenkins environment and confirm the changes compile/operate as expected. Inspect the java_home change and make sure the windows path is updated correctly.

See recent jenkins failures for origin of these changes.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,7 +6,7 @@ pipeline {
         buildDiscarder(logRotator(numToKeepStr:'25'))
     }
     triggers {
-        cron('H H(20-23) * * *')
+        cron('H H(17-19) * * *')
     }
     environment {
         DOCS = 'distribution/docs'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -6,11 +6,12 @@ pipeline {
         buildDiscarder(logRotator(numToKeepStr:'25'))
     }
     triggers {
-        cron('H H(20-23) * * *')
+        cron('H H(17-19) * * *')
```
shaundmorris: ❓ Why is this an improvement?

Jenkinsfile
```diff
@@ -6,11 +6,12 @@ pipeline {
...
     }
     environment {
         DOCS = 'distribution/docs'
         ITESTS= 'distribution/test/itests/test-itests-ddf'
+        JDK_LOC = 'C:/Program Files/Java/jdk1.8.0_131'
```
shaundmorris: ❓ Is the JDK something we should plan to update frequently?

mcalcote: I figured it was marginally volatile so I chose to place it here even though it is only used once below. I don't imagine we would update it frequently to be honest.

Jenkinsfile
mcalcote: I adjusted this trigger to avoid overlapping this job with others pipelines (Alliance). This may have been a contributor to some of the mismatched/missing artifact issues that caused build failures.

mcalcote:
@clockard - Recent builds have had windows nodes fail with a missing JAVA_HOME environment variable. The cause may be due to a regression in maven version to address another bug, but the goal here is to explicitly set this and avoid the same failures.

Jenkinsfile
oconnormi: We should be able to just use the `jdk` property in the `withMaven` step (this is currently being verified in another pipeline at the moment, but i'm pretty sure it already got us past the `JAVA_HOME` issue). This will enable us to either make a project follow the latest jdk8 version/update or pin the project to a specific version if necessary

mcalcote:
build now

mcalcote:
I did test it on the production jenkins and it did not suffer the now infamous `Error: JAVA_HOME not found in your environment. `

PR #2219, opened 2017-08-15
External SAML2 IdP ( Keycloak )

gradygit:
#### What does this PR do?
Provides minor modifications to the DDF IdP client to allow it to connect to an external IdP using SAML2.   These changes were made in order to allow [Keycloak](http://www.keycloak.org/) to act as the IdP.

There are two commits.

Add a failover mechanism when parsing SAML Assertion namespaces which are invalid/incomplete when they are extracted from a larger SAMLResponse.  

As an example, consider this SAMLResponse:
```
<samlp:Response
	xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
	xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" Destination="https://{my_servers_FQHN}/services/saml/sso" ID="ID_d35a0cc9-63c2-4427-bf99-c634ced2026e" IssueInstant="2017-03-15T16:23:54.827Z" Version="2.0">
	<saml:Issuer>http://{my_servers_FQHN}:3389/auth/realms/{somerealm}</saml:Issuer>
	<dsig:Signature
		xmlns:dsig="http://www.w3.org/2000/09/xmldsig#">
...
	</dsig:Signature>
	<samlp:Status>
		<samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
	</samlp:Status>
	<saml:Assertion
		xmlns="urn:oasis:names:tc:SAML:2.0:assertion" ID="ID_18545e6c-7062-414f-887b-21fd67b22344" IssueInstant="2017-03-15T16:23:54.826Z" Version="2.0">
....
	</saml:Assertion>
</samlp:Response>
```

When the assertion is extracted in AssertionConsumerService.login  using DOM2Writer.nodeToString, the xml that is extracted looks like this:
```
<saml:Assertion
		xmlns="urn:oasis:names:tc:SAML:2.0:assertion" ID="ID_18545e6c-7062-414f-887b-21fd67b22344" IssueInstant="2017-03-15T16:23:54.826Z" Version="2.0">
...
</saml:Assertion>
```
This is invalid XML, but still passes the:
`if (tokenString.contains(SAML_NAMESPACE)) {`

test in SAMLAssertionHandler.java - ultimately causing a failure because the code proceeds without a SAML token.

The second commit allows administration configuration of whether or not the AuthnRequest generated by DDF's IDP client will contain assertionConsumerServiceURL.  Setting assertionConsumerServiceURL is optional per the SAML2.0 spec and causes an issue when DDF is deployed in a containerized environment.  Namely because the internal container hostname/port that is running DDF is not necessarily the same as what is publicly exposed on the docker-machine via Docker Run.    Having AssertionConsumerServiceURL in the AuthnRequest overrides the Assertion Consumer Service value that has been manually defined in the Keycloak Client configuration.

This commit package defaults the configuration to continue to generate assertionConsumerServiceURL as per how it currently operates with the internal DDF IDP.
 
#### Who is reviewing it? 
@brendan-hofmann 
@stustison
@brendan-hofmann 
@stustison
Preconditions:
Keycloak has been deployed and a security realm has been created.  Can use the official docker image if desired.
DDF has been built using these changes and deployed.

Steps:

1. Install security-idp from the DDF admin console
2. Set up a Keycloak client using DDF's metadata as input ( /services/saml/sso/metadata ) 
3. Configure DDF client-idp with the EntityDescriptor metadata from the keycloak client
4. Configure DDF client-idp to 'uncheck' the box for Set Assertion Consumer Service URL
5. Configure DDF web context policy manager to protect desired endpoint with IDP.
6. Import the Keycloak Realm certificate into DDF's keystore
7. Restart DDF
8. Navigate to the endpoint that was protected by IDP above.  Redirection to Keycloak will occur for sign in.  On successful auth/auth, there will be a redirection back to {ddf}/services/saml/sso and then finally a redirection back to the original requested resource.

Notes: When creating the client in keycloak, use the role list mapper to set the role in http://schemas.xmlsoap.org/ws/2005/05/identity/claims/role

#### Checklist:
- [ ] Documentation Updated
- [N/A] Update / Add Unit Tests
#### Review Comment Legend:


platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
```diff
@@ -494,7 +494,12 @@ private String createAndSignAuthnRequest(boolean isPost, boolean wantSigned)
         issuer.setValue(spIssuerId);
         authnRequest.setIssuer(issuer);
 
-        authnRequest.setAssertionConsumerServiceURL(spAssertionConsumerServiceUrl);
+        if (idpMetadata.getAssertionConsumerServiceURLCheck()){
+            LOGGER.info("Setting optional AssertionConsumerServiceURL in AuthnRequest");
```
stustison: Can you set these to the trace level to fit with our logging guidelines.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
```diff
@@ -494,7 +494,12 @@ private String createAndSignAuthnRequest(boolean isPost, boolean wantSigned)
         issuer.setValue(spIssuerId);
         authnRequest.setIssuer(issuer);
 
-        authnRequest.setAssertionConsumerServiceURL(spAssertionConsumerServiceUrl);
+        if (idpMetadata.getAssertionConsumerServiceURLCheck()){
```
stustison: This is fine for 2.10.x, but this is fixed in 2.11.0 where we now have a different host/port for exposing services from within a container and shouldn't be needed when we port this upstream.

stustison:
build now

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/IdpHandler.java
rzwiefel: The guidelines can be found here, for future reference: https://codice.atlassian.net/wiki/spaces/DDF/pages/7045131/Log+Levels

stustison:
I didn't notice it before, but could you please create a ticket https://codice.atlassian.net/projects/DDF/issues and update your commits with the ticket number. I think we'll be good to merge after that.

mrmateo:
@stustison  - Was looking at this the other day and I think it would be good to get in before 2.11.0 is released.

@gradygit  did you need any help updating the commits like what @stustison  mentioned in his last comment?

stustison:
@mrmateo I agree 100%. I'd like to get this in for 2.11.0. It is unlikely that we'll be releasing another 2.10.x release anyways with 2.11.0 on the way.

stustison:
@clockard make sure we add this to the 2.11 tickets

mrmateo:
@stustison - I am playing around with this on the latest master baseline to see if it can work. Not fully working yet, but making some decent progress. Only changes made so far outside of the ones in this ticket were to revert the changes to post-binding.html done in this commit - https://github.com/codice/ddf/commit/f8b4b4d2f247987fc688fea538abe5fca83ac68d due to that commit update the file to make it invalid XHTML (the file is marked as XHTML which requires valid XML marking...eliminating the end tags causes errors to appear in browsers). 

Other than that just running in to an issue where the SAML assertion isn't getting its namespaces copied over correctly from the parent -- i.e. the main Response in the log shows the correct namespace for the assertion:
`
<saml:Assertion xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" ID="ID_7d88b9e7-df5a-40a4-a7db-4d5218e126ee" IssueInstant="2017-09-20T17:06:35.187Z" Version="2.0">
`
but when it is pulled out and added to the header, the namespace doesn't match up any more:
`
<saml:Assertion ID="ID_7d88b9e7-df5a-40a4-a7db-4d5218e126ee" IssueInstant="2017-09-20T17:06:35.187Z" Version="2.0" xmlns="urn:oasis:names:tc:SAML:2.0:assertion">
`
which causes a XML parsing exception in the login method.

I didn't see any differences in that part of the code (uses the same OpenSAML library) so right now am looking to see if it's a difference in the response from Keycloak versions. Have you seen that before at all in any testing you've been doing?

stustison:
@mrmateo No, I haven't seen that particular issue with the namespaces. I haven't had a chance to try using Keycloak as an IdP yet. I also did not notice that the commit you mentioned made that invalid XML, good catch! Sounds like we'll need another PR to fix that up.

coyotesqrl:
Closing this ticket targeting 2.10.x as we will likely not be releasing another version of that release line. The work has been included in the master branch for release in 2.11.0 with PR #2353.

PR #2222, opened 2017-08-16
DDF-3216 add the ability to attach notes to workspace queries

glenhein:
ATTN: DO NOT MERGE UNTIL 2.11 IS FORKED FROM MASTER. WAIT FOR 2.12-SNAPSHOT!

#### What does this PR do?
Add the ability to attach notes to workspace queries.
#### Who is reviewing it? 
@bdeining 
@troymohl 
@andrewkfiedler 

@andrewkfiedler
@bdeining
1) build and install
2) Create a workspace query. After the initial query has been created, there will be a new tab on the query called 'Notes'. You can add, edit and delete notes. You must be logged in as a non-guest user.
[DDF-3216](https://codice.atlassian.net/browse/DDF-3216)
screen shot 2017-08-16 at 9 07 46 am

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -40,6 +40,7 @@ import ddf.catalog.core.versioning.DeletedMetacard;
 import ddf.catalog.core.versioning.MetacardVersion;
 import ddf.catalog.core.versioning.MetacardVersion.Action;
 import ddf.catalog.core.versioning.impl.MetacardVersionImpl;
+import ddf.catalog.data.Attribute;
 import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.AttributeRegistry;
 import ddf.catalog.data.AttributeType;
@@ -49,6 +50,8 @@ import ddf.catalog.data.MetacardType;
 import ddf.catalog.data.Result;
 import ddf.catalog.data.impl.AttributeImpl;
 import ddf.catalog.data.impl.ResultImpl;
+import ddf.catalog.data.impl.types.SecurityAttributes;
+import ddf.catalog.data.types.Core;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.filter.FilterBuilder;
 import ddf.catalog.operation.DeleteResponse;
@@ -71,6 +74,7 @@ import ddf.catalog.transform.QueryResponseTransformer;
 import ddf.catalog.util.impl.ResultIterable;
 import ddf.security.Subject;
 import ddf.security.SubjectUtils;
+import ddf.security.common.audit.SecurityLogger;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -110,6 +114,9 @@ import org.codice.ddf.catalog.ui.metacard.associations.Associated;
 import org.codice.ddf.catalog.ui.metacard.edit.AttributeChange;
 import org.codice.ddf.catalog.ui.metacard.edit.MetacardChanges;
 import org.codice.ddf.catalog.ui.metacard.history.HistoryResponse;
+import org.codice.ddf.catalog.ui.metacard.notes.NoteConstants;
+import org.codice.ddf.catalog.ui.metacard.notes.NoteMetacard;
+import org.codice.ddf.catalog.ui.metacard.notes.NoteUtil;
 import org.codice.ddf.catalog.ui.metacard.transform.CsvTransform;
 import org.codice.ddf.catalog.ui.metacard.validation.Validator;
 import org.codice.ddf.catalog.ui.metacard.workspace.WorkspaceAttributes;
@@ -137,6 +144,12 @@ public class MetacardApplication implements SparkApplication {
 
   private static final Security SECURITY = Security.getInstance();
 
+  private static final String ERROR_RESPONSE_TYPE = "error";
+
+  private static final String SUCCESS_RESPONSE_TYPE = "success";
+
+  private static final MetacardType SECURITY_ATTRIBUTES = new SecurityAttributes();
+  private static int pageSize = 250;
   private final ObjectMapper mapper =
       JsonFactory.create(
           new JsonParserFactory().usePropertyOnly(),
@@ -145,32 +158,22 @@ public class MetacardApplication implements SparkApplication {
               .includeNulls()
               .includeDefaultValues()
               .setJsonFormatForDates(false));
-
   private final CatalogFramework catalogFramework;
-
   private final FilterBuilder filterBuilder;
-
   private final EndpointUtil util;
-
   private final Validator validator;
-
   private final WorkspaceTransformer transformer;
-
   private final ExperimentalEnumerationExtractor enumExtractor;
-
   private final SubscriptionsPersistentStore subscriptions;
-
   private final List<MetacardType> types;
-
   private final Associated associated;
-
-  private static int pageSize = 250;
-
   private final QueryResponseTransformer csvQueryResponseTransformer;
 
   private final AttributeRegistry attributeRegistry;
 
   private final ConfigurationApplication configuration;
+
+  private final NoteUtil noteUtil;
 
   public MetacardApplication(
       CatalogFramework catalogFramework,
@@ -184,7 +187,8 @@ public class MetacardApplication implements SparkApplication {
       Associated associated,
       QueryResponseTransformer csvQueryResponseTransformer,
       AttributeRegistry attributeRegistry,
-      ConfigurationApplication configuration) {
+      ConfigurationApplication configuration,
+      NoteUtil noteUtil) {
     this.catalogFramework = catalogFramework;
     this.filterBuilder = filterBuilder;
     this.util = endpointUtil;
@@ -197,6 +201,7 @@ public class MetacardApplication implements SparkApplication {
     this.csvQueryResponseTransformer = csvQueryResponseTransformer;
     this.attributeRegistry = attributeRegistry;
     this.configuration = configuration;
+    this.noteUtil = noteUtil;
   }
 
   private String getSubjectEmail() {
@@ -584,6 +589,137 @@ public class MetacardApplication implements SparkApplication {
             }
           }
           return "";
+        });
+
+    post(
+        "/annotations",
+        (req, res) -> {
+          Map<String, Object> incoming =
+              JsonFactory.create().parser().parseMap(util.safeGetBody(req));
+          String workspaceId = incoming.get("workspace").toString();
+          String queryId = incoming.get("parent").toString();
+          String annotation = incoming.get("note").toString();
+          String user = getSubjectEmail();
+          if (user == null) {
+            res.status(401);
+            return util.getResponseWrapper(
+                ERROR_RESPONSE_TYPE,
+                "You are not authorized to create notes! A user email is required. "
+                    + "Please ensure you are logged in and/or have a valid email registered in the system.");
+          }
+          if (StringUtils.isBlank(annotation)) {
+            res.status(400);
+            return util.getResponseWrapper(ERROR_RESPONSE_TYPE, "No annotation!");
+          }
+          NoteMetacard noteMetacard = new NoteMetacard(queryId, user, annotation);
+
+          Metacard workspaceMetacard = util.findWorkspace(workspaceId);
+
+          if (workspaceMetacard == null) {
+            res.status(404);
+            return util.getResponseWrapper(
+                ERROR_RESPONSE_TYPE, "Cannot find the workspace metacard!");
+          }
+
+          util.copyAttributes(workspaceMetacard, SECURITY_ATTRIBUTES, noteMetacard);
+
+          Metacard note = saveMetacard(noteMetacard);
+
+          SecurityLogger.auditWarn(
+              "Attaching an annotation to a resource: resource={} annotation={}",
+              SecurityUtils.getSubject(),
+              workspaceId,
+              noteMetacard.getId());
+
+          Map<String, String> responseNote = noteUtil.getResponseNote(note);
+          if (responseNote == null) {
+            res.status(500);
+            return util.getResponseWrapper(
+                ERROR_RESPONSE_TYPE, "Cannot serialize note metacard to json!");
+          }
+          return util.getResponseWrapper(SUCCESS_RESPONSE_TYPE, util.getJson(responseNote));
+        });
+
+    get(
+        "/annotations/:queryid",
+        (req, res) -> {
+          String queryId = req.params(":queryid");
+
+          List<Metacard> retrievedMetacards =
+              noteUtil.getAssociatedMetacardsByTwoAttributes(
+                  NoteConstants.PARENT_ID, Core.METACARD_TAGS, queryId, "note");
+          ArrayList<String> getResponse = new ArrayList<>();
+          retrievedMetacards.sort(Comparator.comparing(Metacard::getCreatedDate));
+          for (Metacard metacard : retrievedMetacards) {
+            Map<String, String> responseNote = noteUtil.getResponseNote(metacard);
+            if (responseNote != null) {
+              getResponse.add(util.getJson(responseNote));
+            }
+          }
+          return util.getResponseWrapper(SUCCESS_RESPONSE_TYPE, getResponse.toString());
+        });
+
+    put(
+        "/annotations/:id",
+        APPLICATION_JSON,
+        (req, res) -> {
+          Map<String, Object> incoming =
+              JsonFactory.create().parser().parseMap(util.safeGetBody(req));
+          String noteMetacardId = req.params(":id");
+          String note = incoming.get("note").toString();
+          Metacard metacard;
+          try {
+            metacard = util.getMetacard(noteMetacardId);
+          } catch (NotFoundException e) {
+            LOGGER.debug("Note metacard was not found for updating. id={}", noteMetacardId);
+            res.status(404);
+            return util.getResponseWrapper(ERROR_RESPONSE_TYPE, "Note metacard was not found!");
+          }
+
+          Attribute attribute = metacard.getAttribute(Core.METACARD_OWNER);
+          if (attribute != null
+              && attribute.getValue() != null
+              && !attribute.getValue().equals(getSubjectEmail())) {
+            res.status(401);
+            return util.getResponseWrapper(
+                ERROR_RESPONSE_TYPE, "Owner of note metacard is invalid!");
+          }
+          metacard.setAttribute(new AttributeImpl(NoteConstants.COMMENT, note));
+          metacard = updateMetacard(metacard.getId(), metacard);
+          Map<String, String> responseNote = noteUtil.getResponseNote(metacard);
+          return util.getResponseWrapper(SUCCESS_RESPONSE_TYPE, util.getJson(responseNote));
+        });
+
+    delete(
+        "/annotations/:id",
+        (req, res) -> {
+          String noteToDeleteMetacardId = req.params(":id");
+          Metacard metacard;
+          try {
+            metacard = util.getMetacard(noteToDeleteMetacardId);
+          } catch (NotFoundException e) {
+            LOGGER.debug("Note metacard was not found for deleting. id={}", noteToDeleteMetacardId);
+            res.status(404);
+            return util.getResponseWrapper(ERROR_RESPONSE_TYPE, "Note metacard was not found!");
+          }
+          Attribute attribute = metacard.getAttribute(Core.METACARD_OWNER);
+          if (attribute != null
+              && attribute.getValue() != null
+              && !attribute.getValue().equals(getSubjectEmail())) {
+            res.status(401);
+            return util.getResponseWrapper(
+                ERROR_RESPONSE_TYPE, "Owner of note metacard is invalid!");
+          }
+          DeleteResponse deleteResponse =
+              catalogFramework.delete(new DeleteRequestImpl(noteToDeleteMetacardId));
+          if (deleteResponse.getDeletedMetacards() != null
+              && !deleteResponse.getDeletedMetacards().isEmpty()) {
+            Map<String, String> responseNote =
+                noteUtil.getResponseNote(deleteResponse.getDeletedMetacards().get(0));
+            return util.getResponseWrapper(SUCCESS_RESPONSE_TYPE, util.getJson(responseNote));
+          }
+          res.status(500);
+          return util.getResponseWrapper(ERROR_RESPONSE_TYPE, "Could not delete note metacard!");
         });
 
     after(
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.metacard.notes;
+
+import ddf.catalog.data.AttributeDescriptor;
+import ddf.catalog.data.MetacardType;
+import ddf.catalog.data.impl.AttributeDescriptorImpl;
+import ddf.catalog.data.impl.BasicTypes;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+public class NoteAttributes implements MetacardType {
+
+  private static final Set<AttributeDescriptor> DESCRIPTORS;
+
+  private static final String NAME = "note";
+
+  static {
+    Set<AttributeDescriptor> descriptors = new HashSet<>();
+    descriptors.add(
+        new AttributeDescriptorImpl(
+            NoteConstants.PARENT_ID,
+            true /* indexed */,
+            true /* stored */,
+            true /* tokenized */,
+            false /* multivalued */,
+            BasicTypes.STRING_TYPE));
+    descriptors.add(
+        new AttributeDescriptorImpl(
+            NoteConstants.COMMENT,
+            true /* indexed */,
+            true /* stored */,
+            true /* tokenized */,
+            true /* multivalued */,
+            BasicTypes.STRING_TYPE));
+    DESCRIPTORS = Collections.unmodifiableSet(descriptors);
+  }
+
+  @Override
+  public Set<AttributeDescriptor> getAttributeDescriptors() {
+    return DESCRIPTORS;
+  }
+
+  @Override
+  public AttributeDescriptor getAttributeDescriptor(String name) {
+    for (AttributeDescriptor attributeDescriptor : DESCRIPTORS) {
+      if (attributeDescriptor.getName().equals(name)) {
+        return attributeDescriptor;
+      }
+    }
+    return null;
+  }
+
+  @Override
+  public String getName() {
+    return NAME;
+  }
+}
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteMetacard.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteMetacard.java
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.metacard.notes;
+
+import ddf.catalog.data.MetacardType;
+import ddf.catalog.data.impl.MetacardImpl;
+import ddf.catalog.data.types.Core;
+
+public class NoteMetacard extends MetacardImpl {
+
+  private static final MetacardType TYPE = new NoteMetacardType();
+
+  /**
+   * @param id The parent ID of the note
+   * @param user the owner of the note
+   * @param comment The comment itself
+   */
+  public NoteMetacard(final String id, final String user, final String comment) {
+    super(TYPE);
+    this.setAttribute(NoteConstants.PARENT_ID, id);
+    this.setAttribute(Core.METACARD_OWNER, user);
+    this.setAttribute(NoteConstants.COMMENT, comment);
+    this.setAttribute(Core.METACARD_TAGS, "note");
+    this.setAttribute(Core.DESCRIPTION, "Note for a Metacard with ID " + id);
+  }
+}
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
@@ -1,0 +1,136 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.metacard.notes;
+
+import static ddf.catalog.util.impl.ResultIterable.resultIterable;
+
+import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Metacard;
+import ddf.catalog.data.Result;
+import ddf.catalog.data.types.Core;
+import ddf.catalog.federation.FederationException;
+import ddf.catalog.filter.FilterBuilder;
+import ddf.catalog.operation.impl.QueryImpl;
+import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.source.SourceUnavailableException;
+import ddf.catalog.source.UnsupportedQueryException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import org.opengis.filter.Filter;
+import org.opengis.filter.sort.SortBy;
+
+public class NoteUtil {
+
+  private static final int PAGE_SIZE = 250;
+
+  private final FilterBuilder filterBuilder;
+
+  private final CatalogFramework catalogFramework;
+
+  public NoteUtil(FilterBuilder filterBuilder, CatalogFramework catalogFramework) {
+    this.filterBuilder = filterBuilder;
+    this.catalogFramework = catalogFramework;
+  }
+
+  /**
+   * Extracts the data required by the user from the note {@link Metacard} object. If any required
+   * fields are null the return object will also be null.
+   *
+   * @param metacard - the metacard object to abstract response data from
+   * @return map of the response note fields or null
+   */
+  @Nullable
+  public Map<String, String> getResponseNote(Metacard metacard) {
+    Map<String, String> requiredAttributes = new HashMap<>();
+    if (metacard.getId() != null) {
+      requiredAttributes.put("id", metacard.getId());
+    } else {
+      return null;
+    }
+    if (metacard.getAttribute(NoteConstants.PARENT_ID) != null) {
+      requiredAttributes.put(
+          "parent", (String) metacard.getAttribute(NoteConstants.PARENT_ID).getValue());
+    } else {
+      return null;
+    }
+    if (metacard.getCreatedDate() != null) {
+      requiredAttributes.put("created", metacard.getCreatedDate().toString());
+    }
+    if (metacard.getModifiedDate() != null) {
+      requiredAttributes.put("modified", metacard.getModifiedDate().toString());
+    }
+    if (metacard.getAttribute(NoteConstants.COMMENT) != null) {
+      requiredAttributes.put(
+          "note", (String) metacard.getAttribute(NoteConstants.COMMENT).getValue());
+    } else {
+      return null;
+    }
+    if (metacard.getAttribute(Core.METACARD_OWNER) != null) {
+      requiredAttributes.put(
+          "owner", (String) metacard.getAttribute(Core.METACARD_OWNER).getValue());
+    } else {
+      return null;
+    }
+    return requiredAttributes;
+  }
+
+  /**
+   * Returns all associated metacards given two relatable attributes.
+   *
+   * @param primaryFilterAttribute The primary attribute to search against
+   * @param secondaryFilterAttribute The secondary attribute to search against
+   * @param primaryAttributeValue A primary attribute value
+   * @param secondaryAttributeValue A secondary attribute value
+   * @return A set of metacards, filtered by the given attributes and it's values
+   * @throws UnsupportedQueryException Error
+   * @throws SourceUnavailableException Error
+   * @throws FederationException Error
+   */
+  public List<Metacard> getAssociatedMetacardsByTwoAttributes(
+      String primaryFilterAttribute,
+      String secondaryFilterAttribute,
+      String primaryAttributeValue,
+      String secondaryAttributeValue)
+      throws UnsupportedQueryException, SourceUnavailableException, FederationException {
+
+    Filter idFilter =
+        filterBuilder.attribute(primaryFilterAttribute).is().equalTo().text(primaryAttributeValue);
+    Filter tagsFilter =
+        filterBuilder
+            .attribute(secondaryFilterAttribute)
+            .is()
+            .equalTo()
+            .text(secondaryAttributeValue);
+    Filter filter = filterBuilder.allOf(idFilter, tagsFilter);
+
+    return resultIterable(
+            catalogFramework,
+            new QueryRequestImpl(
+                new QueryImpl(
+                    filter,
+                    1,
+                    PAGE_SIZE,
+                    SortBy.NATURAL_ORDER,
+                    true,
+                    TimeUnit.SECONDS.toMillis(10)),
+                false))
+        .stream()
+        .map(Result::getMetacard)
+        .collect(Collectors.toList());
+  }
+}
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
@@ -26,6 +26,7 @@ import ddf.catalog.data.InjectableAttribute;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.MetacardType;
 import ddf.catalog.data.Result;
+import ddf.catalog.data.impl.AttributeImpl;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.filter.FilterBuilder;
 import ddf.catalog.impl.filter.GeoToolsFunctionFactory;
@@ -81,6 +82,7 @@ import spark.Request;
 import spark.Response;
 
 public class EndpointUtil {
+
   private static final Logger LOGGER = LoggerFactory.getLogger(EndpointUtil.class);
 
   private final List<MetacardType> metacardTypes;
@@ -155,6 +157,13 @@ public class EndpointUtil {
     Result result = queryResponse.getResults().get(0);
 
     return result.getMetacard();
+  }
+
+  public String getResponseWrapper(String responseType, Object response) {
+    Map<String, Object> result = new HashMap<>();
+    result.put("responseType", responseType);
+    result.put("response", response);
+    return getJson(result);
   }
 
   public Map<String, Result> getMetacardsByFilter(String tagFilter)
@@ -525,5 +534,49 @@ public class EndpointUtil {
           incoming);
     }
     return current;
+  }
+
+  /**
+   * Copy the attributes from a metacard to another metacard.
+   *
+   * @param sourceMetacard the source metacard
+   * @param metacardType copy all attributes represented by this metacard type
+   * @param destinationMetacard the destination metacard
+   */
+  public void copyAttributes(
+      Metacard sourceMetacard, MetacardType metacardType, Metacard destinationMetacard) {
+    metacardType
+        .getAttributeDescriptors()
+        .stream()
+        .filter(descriptor -> sourceMetacard.getAttribute(descriptor.getName()) != null)
+        .map(descriptor -> copyAttribute(descriptor, sourceMetacard))
+        .forEach(destinationMetacard::setAttribute);
+  }
+
+  private Attribute copyAttribute(AttributeDescriptor attributeDescriptor, Metacard metacard) {
+    String name = attributeDescriptor.getName();
+    if (attributeDescriptor.isMultiValued()) {
+      List<Serializable> values = new ArrayList<>(metacard.getAttribute(name).getValues());
+      return new AttributeImpl(name, values);
+    }
+    return metacard.getAttribute(name);
+  }
+
+  /**
+   * Find the workspace metacard based on the workspace identifier. If the workspsace cannot be
+   * found, then return <code>null</code>.
+   *
+   * @param workspaceId the workspace identifier
+   * @return workspace metacard
+   */
+  public Metacard findWorkspace(String workspaceId) {
+    try {
+      return getMetacard(workspaceId);
+    } catch (NotFoundException
+        | UnsupportedQueryException
+        | SourceUnavailableException
+        | FederationException e) {
+      return null;
+    }
   }
 }
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/annotation/annotation.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/annotation/annotation.js
@@ -1,4 +1,3 @@
-{{!--
 /**
  * Copyright (c) Codice Foundation
  *
@@ -10,13 +9,18 @@
  * <http://www.gnu.org/licenses/lgpl.html>.
  *
  **/
- --}}
-<div class="if-editing">
-    <textarea class="ta-enabled" id="{{cid}}" value="{{value}}"
-              placeholder="{{property.placeholder}}" name="{{property.id}}">{{value}}</textarea>
-</div>
-<div class="if-viewing">
-    <textarea class="ta-disabled" disabled id="{{cid}}" value="{{value}}"
-              placeholder="{{property.placeholder}}" name="{{property.id}}">{{value}}</textarea>
-    <button class="ta-expand"><i class="fa fa-ellipsis-h"></i></button>
-</div>
+/*global require*/
+var Backbone = require('backbone');
+
+module.exports = Backbone.Model.extend({
+    defaults: function () {
+        return {
+            id: undefined,
+            parent: undefined,
+            created: undefined,
+            modified: undefined,
+            annotation: undefined,
+            owner: undefined
+        };
+    }
+});
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/input/textarea/input-textarea.less
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/input/textarea/input-textarea.less
@@ -1,4 +1,7 @@
 @{customElementNamespace}input.is-textarea {
+    .if-overflow {
+        display:none;
+    }
     .if-editing {
         position: relative;
     }
@@ -6,13 +9,48 @@
     textarea {
         line-height: normal;
         padding: 6px 10px 5px 10px;
-        width: 100% !important;
-        min-height: 100px;
-        max-height: 400px;
+      width: 100%;
+        max-height:400px;
+        min-height:100px;
         color: @text-color;
         background: @medium-lighten-inherited-background-color;
         .is-input;
+
     }
+
+    textarea:disabled {
+        background-color: #2e2e33;
+      position: relative;
+      z-index: @zIndexContent;
+        border: none;
+        height:75px;
+        max-height:75px;
+        min-height:75px;
+        user-select: none;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        flex-direction: unset;
+        resize:none;
+    }
+
+    .ta-expand {
+        display:block;
+        height:20px;
+        text-align:center;
+        width:100%;
+        position:relative;
+      bottom: 30px;
+      margin-bottom: -15px;
+      z-index: @zIndexContent;
+      background-color: rgba(46, 46, 51, 0.7);
+
+      i {
+        position: relative;
+        bottom: 0;
+      }
+
+    }
+
 }
 
 @{customElementNamespace}input.is-textarea.is-editing {
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/query-annotations/query-annotations.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/query-annotations/query-annotations.view.js
@@ -1,0 +1,225 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+/*global define, setTimeout*/
+define([
+    'jquery',
+    'backbone',
+    'marionette',
+    'underscore',
+    'properties',
+    'js/model/Metacard',
+    'wreqr',
+    './query-annotations.hbs',
+    'maptype',
+    'js/store',
+    'js/CustomElements',
+    'component/property/property.view',
+    'component/property/property',
+    'component/annotation/annotation.collection',
+    'component/annotation/annotation.collection.view',
+    'component/input/textarea/input-textarea.view',
+    'component/loading-companion/loading-companion.view',
+    'component/announcement'
+], function ($, Backbone, Marionette, _, properties, MetaCard, wreqr, template, maptype,
+             store, CustomElements, PropertyView, Property, AnnotationCollection, AnnotationCollectionView,
+             TextAreaView, LoadingCompanionView, announcement) {
+
+    return Marionette.LayoutView.extend({
+        setDefaultModel: function () {
+            this.model = this.selectionInterface.getSelectedResults().first();
+        },
+        template: template,
+        tagName: CustomElements.register('query-annotations'),
+        events: {
+            'click .add-annotation-btn ': 'handleCreate',
+            'click .refresh-btn ': 'handleRefresh'
+        },
+        childEvents: {
+            "annotation:add": "handleChildDelete"
+        },
+        regions: {
+            addAnnotationField: '.add-annotation-field',
+            annotationsList: '.annotations-list'
+        },
+        _annotationsCollection: undefined,
+        _annotations: undefined,
+        initialize: function (options) {
+            this.selectionInterface = options.selectionInterface || this.selectionInterface;
+            if (!options.model) {
+                this.setDefaultModel();
+            }
+        },
+        onBeforeShow: function(){
+            this.clearAnnotations();
+            this.showAddAnnotationView();
+            this.getAnnotationsForQuery();
+            this.showAnnotationsListView();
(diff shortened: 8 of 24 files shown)


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+            String parentID = incoming.get("parent")
+                    .toString();
+            String comment = incoming.get("note")
+                    .toString();
+            Boolean isVirtual = Boolean.parseBoolean(incoming.getOrDefault("isVirtual", "false")
+                    .toString());
+            String user = getSubjectEmail();
+            if (user == null) {
+                return util.getResponseWrapper("error",
+                        "You are not authorised to create notes! A user email is required. "
```
bdeining: Is there a reason to use the UK spelling of `authorized`?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                                + "Please ensure you are logged in and/or have a valid email registered in the system.");
+            }
+            //Test if parent metacard ID is valid and comment is not empty
```
bdeining: We have a service to validate the metacard ID that could possibly be leveraged: https://github.com/codice/ddf/blob/master/platform/util/platform-util-uuidgenerator/util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/UuidGeneratorImpl.java#L46.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+        get("/notes/:parentid", (req, res) -> {
+            LOGGER.trace("Hit notes GET method");
+            String parentID = req.params(":parentid");
+            String[] isVirtualParams = req.queryMap()
+                    .toMap()
+                    .getOrDefault("isVirtual", new String[] {"false"});
+            Boolean isVirtual =
+                    isVirtualParams.length > 0 && Boolean.parseBoolean(isVirtualParams[0]);
+            if (!isVirtual) {
+                //Test if parent metacard ID is valid
```
bdeining: Same comment on id validation

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                for (Metacard metacard : retrievedMetacards) {
+                    String jsonStringToAppend = util.getResponseNoteJson(metacard);
+                    if (jsonStringToAppend != null) {
+                        LOGGER.trace("jsonStringToAppend: " + jsonStringToAppend);
+                        getResponse.add(jsonStringToAppend);
+                    }
+                }
+                LOGGER.trace("getResponse: " + getResponse.toString());
+                LOGGER.trace(
+                        "output: " + util.getResponseWrapper("success", getResponse.toString()));
```
bdeining: Use parameterized logging.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                //Reorder in terms of date created using timsort
+                retrievedMetacards.sort(Comparator.comparing(Metacard::getCreatedDate));
+                for (Metacard metacard : retrievedMetacards) {
+                    String jsonStringToAppend = util.getResponseNoteJson(metacard);
+                    if (jsonStringToAppend != null) {
+                        LOGGER.trace("jsonStringToAppend: " + jsonStringToAppend);
+                        getResponse.add(jsonStringToAppend);
+                    }
+                }
+                LOGGER.trace("getResponse: " + getResponse.toString());
```
bdeining: Use parameterized logging.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                            parentID,
+                            "note");
+            if (!retrievedMetacards.isEmpty()) {
+                ArrayList<String> getResponse = new ArrayList<>();
+                //Reorder in terms of date created using timsort
+                retrievedMetacards.sort(Comparator.comparing(Metacard::getCreatedDate));
+                for (Metacard metacard : retrievedMetacards) {
+                    String jsonStringToAppend = util.getResponseNoteJson(metacard);
+                    if (jsonStringToAppend != null) {
+                        LOGGER.trace("jsonStringToAppend: " + jsonStringToAppend);
```
bdeining: Use parameterized logging.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+        delete("/notes/:noteid", (req, res) -> {
+            LOGGER.trace("Hit notes DELETE method");
+            String noteToDeleteMetacardId = req.params(":noteid");
+            Metacard metacard;
+            try {
+                metacard = util.getMetacard(noteToDeleteMetacardId);
+            } catch (NotFoundException e) {
+                throw new IllegalStateException("Note metacard ID is invalid!");
+            }
+            //TODO: The following check should be done as a PolicyExtension in the future.
```
bdeining: Is there a ticket for this TODO?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+            String noteMetacardId = req.params(":id");
+            String note = incoming.get("note")
+                    .toString();
+            Metacard metacard;
+            try {
+                metacard = util.getMetacard(noteMetacardId);
+            } catch (NotFoundException e) {
+                throw new IllegalStateException("Note metacard ID is invalid!");
+            }
+            //TODO: The following check should be done as a PolicyExtension in the future.
```
bdeining: Is there a ticket for this TODO?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -135,6 +155,91 @@ public Metacard getMetacard(String id)
...
+     * @return A set of metacards, filtered by the given attributes and it's values
+     * @throws UnsupportedQueryException  Error
+     * @throws SourceUnavailableException Error
+     * @throws FederationException        Error
+     */
+    public List<Metacard> getAssociatedMetacardsByTwoAttributes(String primaryFilterAttribute,
+            String secondaryFilterAttribute,
+            String primaryAttributeValue,
+            String secondaryAttributeValue)
+            throws UnsupportedQueryException, SourceUnavailableException, FederationException {
```
bdeining: I'm assuming these exceptions come from the query to the CatalogFramework.  Any reason not to catch them in the method?

catalog/ui/catalog-ui-search/src/main/webapp/component/input/textarea/input-textarea.less
```diff
@@ -1,18 +1,57 @@
...
+      margin-bottom: -15px;
+      z-index: @zIndexContent;
+      background-color: rgba(46, 46, 51, 0.7);
+
+      i {
+        position: relative;
+        bottom: 0;
+      }
+
+
```
bdeining: extra whitespace

catalog/ui/catalog-ui-search/src/main/webapp/component/query-status/query-status-body.less
```diff
@@ -1,2 +1,2 @@
-@{customElementNamespace}query-status-tbody {
+@{customElementNamespace}query-annotations-tbody {
```
bdeining: Why this change?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+            //Test if parent metacard ID is valid and comment is not empty
+            if (!isVirtual) {
+                try {
+                    util.getMetacard(parentID, true);
+                } catch (NotFoundException e) {
+                    LOGGER.debug("Parent metacard ID is invalid!");
+                    throw new IllegalStateException("Parent metacard ID is invalid!");
+                }
+            }
+            if (comment.equals("")) {
```
rzwiefel: ✏️ probably should use apache commons isBlank

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                LOGGER.info(util.getResponseWrapper("error", "No comment!"));
+                return util.getResponseWrapper("error", "No comment!");
+            }
+            MetacardType noteType = null;
+            for (MetacardType type : types) {
+                if (type.getName()
+                        .equals("resource-note")) {
+                    noteType = type;
```
rzwiefel: ✏️  We probably don't want to iterate through the list of metacard types every time to get the same metacard type. Could we store this off in a field or something like that?

rzwiefel: Could we do this with a service property instead?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                    break;
+                }
+            }
+            if (noteType == null) {
+                LOGGER.debug("Metacard type 'note' has not been registered in system");
+                throw new IllegalStateException(
+                        "Metacard type 'note' has not been registered in system");
+            }
+            Metacard note = saveMetacard(new NoteMetacard(noteType, parentID, user, comment));
+            String getResponse = util.getResponseNoteJson(note);
```
rzwiefel: ✏️  If this is specific logic concerning the Note logic it should probably be extracted into its own class instead of being put into the utils class.

catalog/ui/catalog-ui-search/src/main/webapp/component/query-status/query-status-body.less
glenhein: hmm, i don't know how the even got changed, but I will revert it

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: that was how the code came from PTE. but I'll change it

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: In this case, valid means that the metacard exists. BTW: It's not actually used by my code because I set the isVirtual flag. But I'm expecting the PTE code on metacard-notes to get merged into ddf and I wanted to minimize the eventual merge conflicts.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: Same answer as on id validation.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: fixed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: fixed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: fixed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
bdeining: Hmm in that case, this comment is a little misleading as it claims to verify that the `metacard ID is valid`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: Throwing the CatalogFramework exceptions is consistent with the other methods in EndpointUtil that deal with CatalogFramework exceptions.

catalog/ui/catalog-ui-search/src/main/webapp/component/input/textarea/input-textarea.less
glenhein: fixed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -94,6 +103,12 @@ public EndpointUtil(List<MetacardType> metacardTypes, CatalogFramework catalogFr
         registerGeoToolsFunctionFactory();
     }
 
+    @SuppressWarnings("unchecked")
+    public static <T, R extends Comparable> Comparator<T> compareBy(Function<T, R> getField) {
```
rzwiefel: Funny that this got back in. This is unneeded because it is provided by `Comparator.comparing`

catalog/ui/catalog-ui-search/src/main/webapp/component/query-annotations/query-annotations.view.js
```diff
@@ -0,0 +1,225 @@
...
+        },
+        setupListeners: function() {
+            this.listenTo(this._notesCollection, "remove", function() {
+                this.checkHasNotes();
+            });
+        },
+        handleRefresh: function() {
+            this.getNotesForMetacard();
+            announcement.announce({
+                            title: 'Success!',
```
bdeining: is this formatting correct?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -135,6 +155,91 @@ public Metacard getMetacard(String id)
...
+        if (queryResponse.getResults().isEmpty()) {
+            return resultList;
+        }
+        List<Result> results = queryResponse.getResults();
+        results.forEach(resultObject -> resultList.add(resultObject.getMetacard()));
+        return resultList;
+    }
+
+    /**
+     * Extracts the data required by the user from the note metacard object.
```
bdeining: nit : `{@link Metacard}`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -135,6 +155,91 @@ public Metacard getMetacard(String id)
...
+     * If any required fields are null the return object will also be null.
+     *
+     * @param metacard - the metacard object to abstract response data from
```
bdeining: nit : {@link Metacard}

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -135,6 +155,91 @@ public Metacard getMetacard(String id)
...
+                .equalTo()
+                .text(primaryAttributeValue);
+        Filter tagsFilter = filterBuilder.attribute(secondaryFilterAttribute)
+                .is()
+                .equalTo()
+                .text(secondaryAttributeValue);
+        Filter filter = filterBuilder.allOf(idFilter, tagsFilter);
+        QueryResponse queryResponse = catalogFramework.query(new QueryRequestImpl(new QueryImpl(
+                filter), false));
+        if (queryResponse.getResults().isEmpty()) {
```
bdeining: is this conditional even needed, given the next few lines of code?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                    .parser()
+                    .parseMap(req.body());
+            String noteMetacardId = req.params(":id");
+            String note = incoming.get("note")
+                    .toString();
+            Metacard metacard;
+            try {
+                metacard = util.getMetacard(noteMetacardId);
+            } catch (NotFoundException e) {
+                throw new IllegalStateException("Note metacard ID is invalid!");
```
bdeining: Might be worth logging `noteMetacardId` here.  Is the ID really invalid here?  Or is it just the case that it wasn't found?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+            }
+            //TODO: The following check should be done as a PolicyExtension in the future.
+            //All policy related decisions should be made by the PDP instead of being throughout the system.
+            if (!metacard.getAttribute(Core.METACARD_OWNER)
+                    .getValue()
```
bdeining: Potential NPE here if the attribute does not exist on the metacard

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+            try {
+                metacard = util.getMetacard(noteToDeleteMetacardId);
+            } catch (NotFoundException e) {
+                throw new IllegalStateException("Note metacard ID is invalid!");
+            }
+            //TODO: The following check should be done as a PolicyExtension in the future.
+            //All policy related decisions should be made by the PDP instead of being throughout the system.
+            if (!metacard.getAttribute(Core.METACARD_OWNER)
+                    .getValue()
+                    .toString()
```
bdeining: Why `toString()`?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+            Metacard metacard;
+            try {
+                metacard = util.getMetacard(noteToDeleteMetacardId);
+            } catch (NotFoundException e) {
+                throw new IllegalStateException("Note metacard ID is invalid!");
+            }
+            //TODO: The following check should be done as a PolicyExtension in the future.
+            //All policy related decisions should be made by the PDP instead of being throughout the system.
+            if (!metacard.getAttribute(Core.METACARD_OWNER)
+                    .getValue()
```
bdeining: Potential NPE here if the attribute does not exist on the metacard

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/workspace/QueryMetacardTypeImpl.java
```diff
@@ -109,6 +111,14 @@
                 false   /* tokenized */,
                 false    /* multivalued */,
                 BasicTypes.STRING_TYPE));
+
+        DESCRIPTORS.add(new AttributeDescriptorImpl(QUERY_ANNOTATION,
```
bdeining: Where is this used?

glenhein: removed, it was left over from the original approach

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: added debug log statement

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: added javadoc link for Metacard

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: added javadoc link for Metacard

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: looks redundant, removed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: looks like this method was even being referenced, so I removed it

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: I tried make it a service, but apparently you can't create a service and reference that service in the same pom. But I did create a service so other can use the metacard type, and I just pass the actual metacard type into the ctor.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: moved

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                return util.getResponseWrapper("error",
+                        "You are not authorised to create notes! A user email is required. "
+                                + "Please ensure you are logged in and/or have a valid email registered in the system.");
+            }
+            //Test if parent metacard ID is valid and comment is not empty
+            if (!isVirtual) {
+                try {
+                    util.getMetacard(parentID, true);
+                } catch (NotFoundException e) {
+                    LOGGER.debug("Parent metacard ID is invalid!");
```
glenhein: added debug log

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +520,164 @@ public void init() {
...
+                    .toMap()
+                    .getOrDefault("isVirtual", new String[] {"false"});
+            Boolean isVirtual =
+                    isVirtualParams.length > 0 && Boolean.parseBoolean(isVirtualParams[0]);
+            if (!isVirtual) {
+                //Test if parent metacard ID is valid
+                try {
+                    util.getMetacard(parentID, true);
+                } catch (NotFoundException e) {
+                    LOGGER.debug("Parent metacard ID is invalid!");
```
glenhein: I think this line is OBE, but in the two places where a NFE is caught, I added a debug log message with the ID.

catalog/ui/catalog-ui-search/src/main/webapp/component/query-annotations/query-annotations.view.js
glenhein: My current version looks OK, but I'm not 100% about the JS formatting rules. I re-ran the code formatter, and it didn't change anything.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: Added null check for the attribute and for the attribute value

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: Added null check for the attribute and for the attribute value

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: removed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +526,112 @@ public void init() {
             return "";
         });
 
+        post("/annotations", (req, res) -> {
+            Map<String, Object> incoming = JsonFactory.create()
+                    .parser()
+                    .parseMap(req.body());
```
stustison: What happens if the note is larger than your heap size?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +526,112 @@ public void init() {
...
+            } else {
+                return util.getResponseWrapper("error",
+                        "Could not find any annotations for this result!");
+            }
+        });
+
+        put("/annotations/:id", APPLICATION_JSON, (req, res) -> {
+            Map<String, Object> incoming = JsonFactory.create()
+                    .parser()
+                    .parseMap(req.body());
```
stustison: Same comment as before. Check the size?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +526,112 @@ public void init() {
...
+            String noteMetacardId = req.params(":id");
+            String note = incoming.get("note")
```
stustison: This variable looks like it is never used.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -517,6 +526,112 @@ public void init() {
...
+            String user = getSubjectEmail();
+            if (user == null) {
+                return util.getResponseWrapper("error",
+                        "You are not authorized to create notes! A user email is required. "
+                                + "Please ensure you are logged in and/or have a valid email registered in the system.");
+            }
+            if (StringUtils.isBlank(annotation)) {
+                return util.getResponseWrapper("error", "No annotation!");
+            }
+            Metacard note = saveMetacard(new NoteMetacard(noteMetacardType,
```
stustison: Should we be auditing when a user attaches a note to a resource that they are not the POC for?

stustison:
Unless I'm reading this wrong, it also looks like we're not protecting the Note metacards with the same security attributes that the "parent" metacard is protected with. This could potentially disclose information about the "parent" metacard that shouldn't be released.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
bdeining: Ooooh, I think in general we haven't been protecting against that in any of these endpoints.  It looks like the whole class would need to be updated.
https://github.com/codice/ddf/pull/2222/files#diff-691f835bd92e972f33f434a9c0957affL224
https://github.com/codice/ddf/pull/2222/files#diff-691f835bd92e972f33f434a9c0957affL238
https://github.com/codice/ddf/pull/2222/files#diff-691f835bd92e972f33f434a9c0957affL256
...

rzwiefel: @stustison How big of upload do we want to allow?

rzwiefel: Sticking with 1MB for now (`1_048_592` bytes)

stustison: I think 1MB is fine. It should probably be configurable though (I know).

rzwiefel: @bdeining Thats what I'm doing :)

bdeining: Of course you are

rzwiefel: 😄

rzwiefel: https://codice.atlassian.net/browse/DDF-3229

bdeining:
How does this work with Shared Queries?  We should add testing with shared queries to the hero criteria.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
bdeining: https://github.com/codice/ddf/pull/2242

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
bdeining: https://github.com/codice/ddf/pull/2242

glenhein:
@bdeining @stustison  I worked out the code to copy the security attributes defined in SecurityAttributes (MetacardType) from the workspace metacard to the query annotation metacard. However, the workspace metacard doesn't have any of those attributes. It does come back from the framework with a field called "security", which contains a copy of the metacard.owner field. Also, I added SecurityAttributes to the query annotation metacard type. However, when the query annotation metacard comes back from the framework, it does have the "security" field, but it does not contain the metacard.owner as the workspace does, even the query annotation metacard does contain metacard.owner as a first class attribute.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
rzwiefel: Yup this is fixed now. After you rebase just make sure to replace any instances of `req.body()` with `util.safeGetBody(req)`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/Note.java
```diff
@@ -0,0 +1,31 @@
...
+
+import ddf.catalog.data.Metacard;
+
+public interface Note {
+
+    /**
+     * The metacard ID of the parent object.
+     * Uses original taxonomy to preserve backwards compatibility.
+     */
+    String PARENT_ID = Metacard.RELATED;
```
rzwiefel: The original taxonomy uses RELATED & DERIVED as a directional relation from parent to child. It might be best to create a different attribute for a reversed direction relation.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Set;
+
+import ddf.catalog.data.AttributeDescriptor;
+import ddf.catalog.data.MetacardType;
+import ddf.catalog.data.impl.AttributeDescriptorImpl;
+import ddf.catalog.data.impl.BasicTypes;
+
+public class NoteAttributes implements Note, MetacardType {
+
+    private static final Set<AttributeDescriptor> DESCRIPTORS;
```
rzwiefel: If you're creating your own MetacardType and not extending from MetacardTypeImpl, You should probably make this a HashMap of Descriptor.name -> Descriptor. This way the get Attribute Descriptors below does not need to do a linear search to get a descriptor.

Alternatively you could extend the existing MetacardTypeImpl if you don't wish to keep the copied the logic.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
```diff
@@ -0,0 +1,69 @@
...
+                true /* stored */,
+                true /* tokenized */,
+                true /* multivalued */,
+                BasicTypes.STRING_TYPE));
+        DESCRIPTORS = Collections.unmodifiableSet(descriptors);
+    }
+
+    @Override
+    public Set<AttributeDescriptor> getAttributeDescriptors() {
+        return DESCRIPTORS;
```
rzwiefel: ❗️ Should probably return a non modifiable view of the set instead of exposing it. (`Collections.unmodifiableSet(..)`)

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteMetacard.java
```diff
@@ -0,0 +1,45 @@
...
+     * Constructor, creates a Note Metacard.
+     * @param metacardType A metacardType NoteMetacardType
+     * @param id The parent ID of the note
+     * @param comment The comment itself
+     */
+    public NoteMetacard(final MetacardType metacardType,
+            final String id,
+            final String user,
+            final String comment) {
+        super(metacardType);
```
rzwiefel: ❓ Should we be statically setting the type of NoteType in this implementation instead of passing in a type to set every time? (The type should not be changing in this case, so we should be able to set it without passing it in)

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
```diff
@@ -0,0 +1,69 @@
...
+        } else {
+            return null;
+        }
+        if (metacard.getAttribute(Core.METACARD_OWNER) != null) {
+            requiredAttributes.put("owner",
+                    (String) metacard.getAttribute(Core.METACARD_OWNER).getValue());
+        } else {
+            return null;
+        }
+        return JsonFactory.create().writeValueAsString(requiredAttributes);
```
rzwiefel: ✏️  I would recommend doing the encoding to a specific format (in this case JSON) outside of this method so we do not couple the encoding format with the actual method logic. So instead just return the map and have it encoded higher up in the application logic

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -135,6 +146,46 @@ public Metacard getMetacard(String id)
...
+        Filter filter = filterBuilder.allOf(idFilter, tagsFilter);
+        QueryResponse queryResponse = catalogFramework.query(new QueryRequestImpl(new QueryImpl(
+                filter), false));
+        List<Result> results = queryResponse.getResults();
+        results.forEach(resultObject -> resultList.add(resultObject.getMetacard()));
+        return resultList;
+    }
+
+    public String getResponseWrapper(String responseType, Object response) {
+        JSONObject responseObj = new JSONObject();
```
rzwiefel: ❗️ We should not be using `org.json.JSONObject`here since we already use boon throughout this module.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -135,6 +146,46 @@ public Metacard getMetacard(String id)
...
+     * @param primaryFilterAttribute   The primary attribute to search against
+     * @param secondaryFilterAttribute The secondary attribute to search against
+     * @param primaryAttributeValue    A primary attribute value
+     * @param secondaryAttributeValue  A secondary attribute value
+     * @return A set of metacards, filtered by the given attributes and it's values
+     * @throws UnsupportedQueryException  Error
+     * @throws SourceUnavailableException Error
+     * @throws FederationException        Error
+     */
+    public List<Metacard> getAssociatedMetacardsByTwoAttributes(String primaryFilterAttribute,
```
rzwiefel: ❓ Is this is directly related to the Note logic? Unless this has applications outside of notes, it should be moved into the NoteUtil class instead of the more generic EndpointUtil.

glenhein: While this is used by notes, there's nothing about this method that makes it specific to notes/annotations. It just retrieves metacards based on two caller-defined attributes and values.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: switched to boon

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteMetacard.java
glenhein: changed to static, which meant creating a proper NoteMetacardType class and pulling out the corresponding definition from blueprint.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
glenhein: switched to MetacardTypeImpl

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/Note.java
glenhein: But it is related, and there's nothing in the javadocs for RELATED that limit its use.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: @bdeining  and I don't think there is a reliable way to detect that the user attaching the note is not the POC, so I've added a SecurityLogger.auditWarn every time an annotation is created.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/Note.java
rzwiefel: Correct, I'm just saying we get a lot more power in the future if we keep all the relations semantically equivalent.

catalog/ui/catalog-ui-search/pom.xml
```diff
@@ -73,6 +73,10 @@
             <version>${camel.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.json</groupId>
```
rzwiefel: Also get rid of this dependency and on line 322 if you replaced the usages with boon

glenhein: removed from pom

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -164,6 +177,46 @@ public Metacard getMetacard(String id)
...
+            String secondaryFilterAttribute, String primaryAttributeValue,
+            String secondaryAttributeValue)
+            throws UnsupportedQueryException, SourceUnavailableException, FederationException {
+        LinkedList<Metacard> resultList = new LinkedList<>();
```
mackncheesiest: ❓ Is there a specific reason for choosing a LinkedList here?

catalog/ui/catalog-ui-search/src/main/webapp/component/annotation/annotation.js
```diff
@@ -0,0 +1,26 @@
+/**
```
mackncheesiest: ✏️  It might be helpful for maintainability in the future if the `Note` vs `annotation` terminology is normalized between the Java and UI components

catalog/ui/catalog-ui-search/src/main/webapp/component/query-annotations/query-annotations.view.js
```diff
@@ -0,0 +1,225 @@
...
+            this.turnOnEditing();
+            this.setupListeners();
+        },
+        clearannotations: function () {
+            if (!this._annotationsCollection) {
+                this._annotationsCollection = new annotationCollection();
+            }
+            this._annotationsCollection.reset();
+        },
+        showAddannotationView: function() {
```
mackncheesiest: ✏️  camelCase involving the word `annotation` throughout this file is a bit funky

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -522,6 +531,127 @@ public void init() {
             return "";
         });
 
+        post("/annotations", (req, res) -> {
+            Map<String, Object> incoming = JsonFactory.create()
+                    .parser()
+                    .parseMap(req.body());
```
rzwiefel: ❗️  This needs to be `util.safeGetBody(req)`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -522,6 +531,127 @@ public void init() {
...
+            } else {
+                return util.getResponseWrapper("error",
+                        "Could not find any annotations for this result!");
+            }
+        });
+
+        put("/annotations/:id", APPLICATION_JSON, (req, res) -> {
+            Map<String, Object> incoming = JsonFactory.create()
+                    .parser()
+                    .parseMap(req.body());
```
rzwiefel: ❗️ This needs to be `util.safeGetBody(req)`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -539,4 +586,38 @@ public FilterBuilder getFilterBuilder() {
...
+     * @param metacardType        copy all attributes represented by this metacard type
+     * @param destinationMetacard the destination metacard
+     */
+    public void copyAttributes(Metacard sourceMetacard, MetacardType metacardType,
+            Metacard destinationMetacard) {
+        metacardType.getAttributeDescriptors()
+                .forEach(attributeDescriptor -> Optional.ofNullable(sourceMetacard.getAttribute(
+                        attributeDescriptor.getName()))
+                        .ifPresent(attribute -> {
+                            List<Serializable> values = new LinkedList<>(attribute.getValues());
```
rzwiefel: ❓ Is it safe to assume that every attribute will be multivalued? Or do we need to key off the attributeDescriptors.isMultivalued ?

rzwiefel: ❓ Also why a linked list in this particular case?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -539,4 +586,38 @@ public FilterBuilder getFilterBuilder() {
...
+     *
+     * @param sourceMetacard      the source metacard
+     * @param metacardType        copy all attributes represented by this metacard type
+     * @param destinationMetacard the destination metacard
+     */
+    public void copyAttributes(Metacard sourceMetacard, MetacardType metacardType,
+            Metacard destinationMetacard) {
+        metacardType.getAttributeDescriptors()
+                .forEach(attributeDescriptor -> Optional.ofNullable(sourceMetacard.getAttribute(
+                        attributeDescriptor.getName()))
```
rzwiefel: ✏️ Not critical, but If we know that `sourceMetacard` isn't null, we could make this
```java
Optional.of(attributeDescriptor.getName())
    .map(sourceMetacard::getAttribute)
```

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
rzwiefel: One solution to this problem could be:
```java
    public void copyAttributes(Metacard sourceMetacard, MetacardType metacardType,
            Metacard destinationMetacard) {
        metacardType.getAttributeDescriptors()
                .stream()
                .filter(name -> sourceMetacard.getAttribute(name) != null)
                .map(ad -> copyAttribute(ad, sourceMetacard))
                .forEach(destinationMetacard::setAttribute);
    }

    private Attribute copyAttribute(AttributeDescriptor attributeDescriptor, Metacard metacard) {
        String name = attributeDescriptor.getName();
        if (attributeDescriptor.isMultiValued()) {
            return new AttributeImpl(name, new ArrayList<>(metacard.getAttribute(name)));
        }
        return metacard.getAttribute(name);

    }

```

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
```diff
@@ -0,0 +1,69 @@
...
+    private NoteUtil() { }
+
+    /**
+     * Extracts the data required by the user from the note {@link Metacard} object.
+     * If any required fields are null the return object will also be null.
+     *
+     * @param metacard - the metacard object to abstract response data from
+     * @return map of the response note fields
+     **/
+    public static Map<String, String> getResponseNote(Metacard metacard) {
```
rzwiefel: ❗️ Should annotate this method with `@Nullable` and also clarify the `@return` javadoc to state it can return null
❓ Would `EndpointUtil::getMetacardMap` serve a similiar enough purpose to this method it could be reused?

In the future it might be nice to create a transformer for this like we use for the other metacards that go through the backend, but thats a little outside the scope of this.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -522,6 +531,127 @@ public void init() {
...
+            Metacard note = saveMetacard(noteMetacard);
+
+            SecurityLogger.auditWarn(String.format(
+                    "Attaching an annotation to a resource: resource=%s annotation=%s",
+                    workspaceId,
+                    noteMetacard.getId()), SecurityUtils.getSubject());
+
+            Map<String, String> responseNote = NoteUtil.getResponseNote(note);
+            if (responseNote == null) {
+                throw new IllegalStateException("An error occurred while storing the annotation");
```
rzwiefel: ❗️ We should have an `exception(IllegalStateException.class, (ex, req, res) -> {...});` block to handle these exceptions

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -522,6 +531,127 @@ public void init() {
...
+            }
+
+            util.copyAttributes(workspaceMetacard, SECURITY_ATTRIBUTES, noteMetacard);
+
+            Metacard note = saveMetacard(noteMetacard);
+
+            SecurityLogger.auditWarn(String.format(
+                    "Attaching an annotation to a resource: resource=%s annotation=%s",
+                    workspaceId,
+                    noteMetacard.getId()), SecurityUtils.getSubject());
```
rzwiefel: ❓ Maybe @stustison Could clarify but I thought we didn't need to use the versions of the method and pass in the subject explicitly because it gets passed in automatically now

✏️  If that is indeed the case then we could use parameterized logging instead of String.format.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
rzwiefel: ☝️ same question

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
rzwiefel: image

It's just right now its only for notes so i'd rather generify this if/when we need to instead of prematurely.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -164,6 +177,46 @@ public Metacard getMetacard(String id)
...
+                .is()
+                .equalTo()
+                .text(primaryAttributeValue);
+        Filter tagsFilter = filterBuilder.attribute(secondaryFilterAttribute)
+                .is()
+                .equalTo()
+                .text(secondaryAttributeValue);
+        Filter filter = filterBuilder.allOf(idFilter, tagsFilter);
+        QueryResponse queryResponse = catalogFramework.query(new QueryRequestImpl(new QueryImpl(
+                filter), false));
```
rzwiefel: ❗️  This is a bug currently because this will use the default page size and cap the results at 20 (confirmed on upstream forks) .We want to fix that and the timeout (default is infinite, so we can set to 5 minutes or the like)

The best Route of action will probably be to use `ResultIterator` and return back as many results as possible. After importing it the only change needed should be to just something like this
```java
Filter filter = ...
return resultIterable(catalogFramework, new QueryRequestImpl(new QueryImpl(
    filter,
    ... all the other args including timeout,)
    .stream()
```

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: I'll move it to NoteUtil

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: no other reason than this is how the original author wrote the code. But it looks like I can clean it up a bit

catalog/ui/catalog-ui-search/src/main/webapp/component/annotation/annotation.js
glenhein: Note is the terminology used for the back-end because it's generalized and notes could be attached to various metacard types. Annotation is used in the front-end as it pertains to annotating queries. However, if the team feels strongly about it, the terminology can be unified.

brianfelix:
Manually tested/verified. LGTM 👍

catalog/ui/catalog-ui-search/src/main/webapp/component/query-annotations/query-annotations.view.js
glenhein: cleaned up the naming

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: fixed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: fixed

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: Is it safe to assume isMultiValued is always set correctly? What if the field should be single valued but the source metacard has multiple values anyways? Since in this use case, I'm copying security fields, I'd hate to drop values even though they don't agree with the attribute schema.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
glenhein: I'll add the `@Nullable`, but I don't see how it could be merged with `getMetacardMap`.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: I converted all the IllegalStateExceptions to `return util.getResponseWrapper("error", ...)`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: i made the change, please double check it, the code is now in NoteUtil

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: looks like there is a version of auditWarn that takes a subject and params

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
rzwiefel: Yes. Multivalued should always be correct. If it isn't there are a lot bigger problems at hand.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -522,6 +535,128 @@ public void init() {
...
+
+            SecurityLogger.auditWarn(
+                    "Attaching an annotation to a resource: resource={} annotation={}",
+                    SecurityUtils.getSubject(),
+                    workspaceId,
+                    noteMetacard.getId());
+
+            Map<String, String> responseNote = noteUtil.getResponseNote(note);
+            if (responseNote == null) {
+                return util.getResponseWrapper("error", "Cannot serialize note metacard to json!");
```
rzwiefel: ❗️ We need to be setting the correct http status codes on the response when we return errors for many of the places in here.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
glenhein: ok, switched over to your solution

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
glenhein: ok, i set a code for every error response for annotations. Please review my choice for codes and let me know if there is a better choice for the code.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteUtil.java
```diff
@@ -0,0 +1,139 @@
...
+                .equalTo()
+                .text(secondaryAttributeValue);
+        Filter filter = filterBuilder.allOf(idFilter, tagsFilter);
+
+        return resultIterable(catalogFramework,
+                new QueryRequestImpl(new QueryImpl(filter,
+                        1,
+                        PAGE_SIZE,
+                        SortBy.NATURAL_ORDER,
+                        false,
```
rzwiefel: ❗️ I think we do want to request the total result count.

glenhein: changed to `true`

mackncheesiest:
build now

mackncheesiest:
The build succeeded but an exception occurred in `gitflow-incremental-builder` at the end?

bdeining:
build now

rzwiefel:
@glenhein Looks like there's conflicts that need to be resolved.

Also pulled it down to check something out, but i'm not even seeing the note tab appear. 
image

glenhein:
@rzwiefel  The notes are for queries, not content metacards.

mackncheesiest:
I actually get an error when I try to use the Annotations tab currently
image
image

mackncheesiest:
Tried again with new changes. 

While logged in as `localhost:localhost`, I was successfully able to create a blank workspace, add a new search, and create, update, and delete annotations on that search. I then created a new local workspace, made some changes in that, went back to the original workspace, and verified that my existing annotations were properly retrieved.

bdeining:
build now

bdeining:
build now

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import ddf.catalog.data.AttributeDescriptor;
+import ddf.catalog.data.impl.AttributeDescriptorImpl;
+import ddf.catalog.data.impl.BasicTypes;
+import ddf.catalog.data.impl.MetacardTypeImpl;
+
+public class NoteAttributes extends MetacardTypeImpl implements Note {
```
bdeining: What's the reasoning behind extending `MetacardTypeImpl`?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteMetacard.java
```diff
@@ -0,0 +1,38 @@
...
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.metacard.notes;
+
+import ddf.catalog.data.MetacardType;
+import ddf.catalog.data.impl.MetacardImpl;
+import ddf.catalog.data.types.Core;
+
+public class NoteMetacard extends MetacardImpl {
```
bdeining: What's the reasoning behind extending MetacardImpl?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/notes/NoteAttributes.java
```diff
@@ -0,0 +1,56 @@
...
+                true /* multivalued */,
+                BasicTypes.STRING_TYPE));
+        DESCRIPTORS = Collections.unmodifiableSet(descriptors);
+    }
+
+    public NoteAttributes() {
+        this(NAME, DESCRIPTORS);
+    }
+
+    public NoteAttributes(String name, Set<AttributeDescriptor> descriptors) {
```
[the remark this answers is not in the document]
glenhein: NoteAttributes has been changed to follow the model of other attribute groups, such as LocationAttributes.

jlcsmith:
@glenhein - please rebase and resolve conflicts

bdeining:
build now

bdeining:
build now

glenhein:
build now

glenhein:
build now

bdeining:
build now

PR #2228, opened 2017-08-17
DDF-3223 Adds limit to attribute override size

rzwiefel:
#### What does this PR do?
 - Maximum attribute data size is capped to 65,535 bytes

#### Who is reviewing it? 
@anyone
@codice/security
@clockard
@stustison

1. Ensure you cannot override an attribute with data greater than 65,535 in size.

https://codice.atlassian.net/browse/DDF-3223

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
+++ b/catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
@@ -13,6 +13,9 @@
  */
 package org.codice.ddf.endpoints.rest;
 
+import static ddf.catalog.data.AttributeType.AttributeFormat.BINARY;
+import static ddf.catalog.data.AttributeType.AttributeFormat.OBJECT;
+
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -20,8 +23,10 @@ import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLDecoder;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -32,7 +37,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 
 import javax.activation.MimeType;
 import javax.activation.MimeTypeParseException;
@@ -59,6 +63,7 @@ import javax.ws.rs.core.UriInfo;
 import org.apache.commons.codec.CharEncoding;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.input.BoundedInputStream;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.cxf.jaxrs.ext.multipart.Attachment;
@@ -121,7 +126,6 @@ import ddf.mime.MimeTypeMapper;
 import ddf.mime.MimeTypeResolutionException;
 import ddf.mime.MimeTypeResolver;
 import ddf.mime.MimeTypeToTransformerMapper;
-
 import net.minidev.json.JSONArray;
 import net.minidev.json.JSONObject;
 import net.minidev.json.JSONValue;
@@ -161,6 +165,8 @@ public class RESTEndpoint implements RESTService {
     private static final String DEFAULT_MIME_TYPE = "application/octet-stream";
 
     private static final String DEFAULT_FILE_NAME = "file";
+
+    public static final int MAX_INPUT_SIZE = 65_536;
 
     private UuidGenerator uuidGenerator;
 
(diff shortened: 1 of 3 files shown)


rzwiefel:
build now

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
```diff
@@ -914,65 +917,82 @@ CreateInfo parseAttachments(List<Attachment> contentParts, String transformerPar
...
+            default:
+                LOGGER.debug("Attribute format '{}' not supported", attributeFormat);
+                break;
             }
-
-        });
+        } catch (IOException e) {
+            LOGGER.debug("Unable to read attribute to override", e);
+        } finally {
+            IOUtils.closeQuietly(inputStream);
```
clockard: Is there a reason we are doing this instead of try with resource?

rzwiefel: I copied from the above unit test verbatim minus one or two lines :)

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
```diff
@@ -914,65 +917,82 @@ CreateInfo parseAttachments(List<Attachment> contentParts, String transformerPar
...
+        try {
+            // ensure the input stream is no more than MAX_INPUT_SIZE
+            InputStream boundedStream = new BoundedInputStream(inputStream, MAX_INPUT_SIZE + 1);
+            byte[] bytes = IOUtils.toByteArray(boundedStream);
+            if (bytes.length > MAX_INPUT_SIZE) {
+                LOGGER.debug("Attribute length is limited to {} bytes", MAX_INPUT_SIZE);
+                return;
+            }
+
+            if (attributeFormat == OBJECT) {
```
rzwiefel: This should be moved above consuming the input stream so we don't unnecessarily consume it.

rzwiefel:
build now

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
```diff
@@ -914,65 +917,79 @@ CreateInfo parseAttachments(List<Attachment> contentParts, String transformerPar
...
+        try (InputStream is = inputStream; InputStream boundedStream = new BoundedInputStream(is,
+                MAX_INPUT_SIZE + 1)) {
+            if (attributeFormat == OBJECT) {
+                LOGGER.debug("Object type not supported for override");
+                return;
+            }
+
+            byte[] bytes = IOUtils.toByteArray(boundedStream);
+            if (bytes.length > MAX_INPUT_SIZE) {
+                LOGGER.debug("Attribute length is limited to {} bytes", MAX_INPUT_SIZE);
```
ahoffer: If I understand BoundedInputStream, it will never return more than MAX_INPUT_SIZE. It reads the wrapped input stream right up to MAX_INPUT_SIZE and then pretends like EOF was encountered. Should the RESTEndpoint fail instead of accepting the 65 KB it read? If the data is BINARY, then the 65 KB probably isn't useful. Maybe it should accept 65 KB of type String but cause a failure for other types?

ahoffer:
Why just the REST endpoint? I feel like all the endpoints should  limit the amount of data read?

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
rzwiefel: @ahoffer It will read up to` MAX_INPUT_SIZE + 1`. Meaning we only read one unnecessary byte and we are able to determine it was too large and treat it as a failure case. We do not process a partial/truncated value (which is why we have a `return` inside this if statement when `bytes.length > MAX_INPUT_SIZE`

rzwiefel:
@ahoffer Correct. There are individual tickets scoped to each endpoint. For example Intrigue endpoints have already been fixed https://github.com/codice/ddf/pull/2242.

distribution/docs/src/main/resources/_contents/_tables/org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor-table-contents.adoc
```diff
@@ -40,7 +40,7 @@
 |Attribute Overrides
 |attributeOverrides
 |String
-|Optional: Metacard attribute overrides (Key-Value pairs) that can be set on the content monitor. If an attribute is specified here, it will overwrite the metacard's attribute that was created from the content directory. The format should be 'key=value'.
+|Optional: Metacard attribute overrides (Key-Value pairs) that can be set on the content monitor. If an attribute is specified here, it will overwrite the metacard's attribute that was created from the content directory. The format should be 'key=value'. The maximum allowed size of an attribute override is 65,535 bytes.
```
ricklarsen: We'll also need this change in the same file on the `src/main/jdocs` path as well so it doesn't get lost when we transition to the templated docs.

distribution/docs/src/main/jdocs/content/_tables/org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor-table-contents.adoc
```diff
@@ -0,0 +1,55 @@
+:title: Content Directory Monitor
+:id: org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor-table-contents
```
ricklarsen: you don't need the `table-contents` part.

rzwiefel:
build now

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
ahoffer: Missed that `return`. Thanks!

distribution/docs/src/main/resources/_contents/_tables/org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor-table-contents.adoc
rzwiefel: done

rzwiefel:
Manually heroed and verified attributes are not allowed to be greater than the specified size

PR #2236, opened 2017-08-22
DDF-3220 Disable local catalog searching from Intrigue.

troymohl:
#### What does this PR do?
Disables searching the local catalog from Intrigue. 

#### Who is reviewing it? 
@adimka  
@albany551  

@andrewkfiedler
@bdeining

Install DDF
Configure Catalog UI Search and check `Disable Local Catalog`
Configure a local source (e.g. CSW)
Ingest some data via CSW
Launch Intrigue and create a new search. Select providers. 
Verify that `ddf.distribution` or similar is not present but the CSW source is.

This allows integrators to use Intrigue while providing separate middleware for query.

[DDF-3220](https://codice.atlassian.net/browse/DDF-3220)

#### Checklist:
- [X] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/docs/src/main/resources/_contents/_tables/org.codice.ddf.catalog.ui.config-table-contents.adoc
+++ b/distribution/docs/src/main/resources/_contents/_tables/org.codice.ddf.catalog.ui.config-table-contents.adoc
@@ -193,6 +193,14 @@ TMS example (3d map support only): `[{"type" : "TMS", "url" : "https://cesiumjs.
 |25
 |true
 
+|Disable Local Catalog
+|disableLocalCatalog
+|Boolean
+|Disables querying the local catalog from UI only.
+|false
+|true
+
+
 |Enable Query Feedback
 |queryFeedbackEnabled
 |Boolean
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -124,6 +124,8 @@ public class ConfigurationApplication implements SparkApplication {
         return readOnly;
     }
 
+    private boolean disableLocalCatalog = false;
+
     private boolean queryFeedbackEnabled = false;
 
     private String queryFeedbackEmailSubjectTemplate;
@@ -253,6 +255,7 @@ public class ConfigurationApplication implements SparkApplication {
         config.put("scheduleFrequencyList", scheduleFrequencyList);
         config.put("isEditingAllowed", isEditingAllowed);
         config.put("isCacheDisabled", isCacheDisabled);
+        config.put("disableLocalCatalog", disableLocalCatalog);
         config.put("queryFeedbackEnabled", queryFeedbackEnabled);
         config.put("queryFeedbackEmailSubjectTemplate", queryFeedbackEmailSubjectTemplate);
         config.put("queryFeedbackEmailBodyTemplate", queryFeedbackEmailBodyTemplate);
@@ -600,6 +603,14 @@ public class ConfigurationApplication implements SparkApplication {
 
     public void setExternalAuthentication(Boolean isExternalAuthentication) {
         this.isExternalAuthentication = isExternalAuthentication;
+    }
+
+    public boolean isDisableLocalCatalog() {
+        return disableLocalCatalog;
+    }
+
+    public void setDisableLocalCatalog(boolean disableLocalCatalog) {
+        this.disableLocalCatalog = disableLocalCatalog;
     }
 
     public void setQueryFeedbackEnabled(boolean queryFeedbackEnabled) {
--- a/catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.config.xml
+++ b/catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.config.xml
@@ -173,6 +173,12 @@
             default="25"
             required="true"/>
 
+        <AD id="disableLocalCatalog"
+            name="Disable Local Catalog"
+            description="Disables queries to the local catalog"
+            type="Boolean"
+            default="false" />
+
         <AD id="queryFeedbackEnabled"
             name="Enable Query Feedback"
             description="Enable the query comments option."
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/model/Sources.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/model/Sources.js
@@ -10,6 +10,9 @@
  *
  **/
 /*global define*/
+
+var user = require('component/singletons/user-instance');
+
 define([
     'underscore',
     'backbone',
@@ -18,6 +21,16 @@ define([
     'jquery'
 ], function (_, Backbone, poller, properties, $) {
     "use strict";
+
+    function removeLocalCatalogIfNeeded(response, localCatalog) {
+        if (properties.isDisableLocalCatalog()) {
+            response = _.filter(response, function(source) {
+                return source.id !== localCatalog;
+            });
+        }
+
+        return response;
+    }
 
     var Types = Backbone.Collection.extend({
     });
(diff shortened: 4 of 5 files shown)


distribution/docs/src/main/resources/_contents/_tables/org.codice.ddf.catalog.ui.config-table-contents.adoc
```diff
@@ -193,6 +193,14 @@ TMS example (3d map support only): `[{"type" : "TMS", "url" : "https://cesiumjs.
...
 |true
 
+|Disable Local Catalog
+|disableLocalCatalog
+|Boolean
+|Disables querying the local catalog from UI only.
+|false
+|true
+
+
```
bdeining: maybe a nit: is this extra whitespace?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -126,7 +136,30 @@ private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
...
+                sourceIds = new HashSet<>(catalogFramework.getSourceIds());
+            }
+            sourceIds.remove(catalogFramework.getId());
+            request = new QueryRequestImpl(request.getQuery(),
+                    false,
+                    sourceIds,
+                    request.getProperties());
+
+            if (CollectionUtils.isNotEmpty(sourceIds)) {
+                response = catalogFramework.query(request);
```
bdeining: It's a nit, but since we call `response = catalogFramework.query(request);` regardless, this added logic could just set the `QueryResponse response` conditionally.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -126,7 +136,30 @@ private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
...
+        if (configuration.isDisableLocalCatalog()) {
+            Set<String> sourceIds;
+            if (CollectionUtils.isNotEmpty(request.getSourceIds())) {
+                sourceIds = new HashSet<>(request.getSourceIds());
+            } else {
+                sourceIds = new HashSet<>(catalogFramework.getSourceIds());
+            }
+            sourceIds.remove(catalogFramework.getId());
+            request = new QueryRequestImpl(request.getQuery(),
+                    false,
```
rzwiefel: ❓ It looks like for a cache query we want to set `isEnterprise` to true, will this conflict with that?
https://github.com/codice/ddf/blob/9939ef34e50441223d9114ef4c320a97f840106c/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlRequest.java#L132-L135

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -126,7 +136,30 @@ private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
                 filterBuilder);
 
         Stopwatch stopwatch = Stopwatch.createStarted();
-        QueryResponse response = catalogFramework.query(request);
+        QueryResponse response = null;
```
rzwiefel: ✏️  We might want to refactor all this new block into a method that returns the overall `response`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -126,7 +136,30 @@ private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
...
+            } else {
+                sourceIds = new HashSet<>(catalogFramework.getSourceIds());
+            }
+            sourceIds.remove(catalogFramework.getId());
+            request = new QueryRequestImpl(request.getQuery(),
+                    false,
+                    sourceIds,
+                    request.getProperties());
+
+            if (CollectionUtils.isNotEmpty(sourceIds)) {
```
rzwiefel: :question: Wouldn't this break queries that explicitly set no sources in order to do an enterprise query?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -126,7 +136,30 @@ private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
                 filterBuilder);
 
         Stopwatch stopwatch = Stopwatch.createStarted();
-        QueryResponse response = catalogFramework.query(request);
+        QueryResponse response = null;
+
+        if (configuration.isDisableLocalCatalog()) {
```
rzwiefel: ❓ Since we only query for a single source at a time (https://github.com/codice/ddf/blob/9939ef34e50441223d9114ef4c320a97f840106c/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlRequest.java#L131-L140), could this block be refactored into something like (if we refactor this chunk into a function):

```java
private Response query(<args>) {
    if (configuration.isDisableLocalCatalog() && request.sourceIds().contains(catalogFramework.getId()) {
        return new QueryResponseImpl(request, new ArrayList<>(), 0);
    }
    //otherwise do normal logic
    ...
}  
```

rzwiefel:
also @troymohl Can you elaborate on some of the motivation behind this change? Also what factors decided into making this change up here instead of in the catalog? This sounds quite similar to fanout.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
troymohl: I didn't want to make any assumption since the request being sent can support multiple source ids, I wanted to make sure that the code would perform correctly.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
troymohl: Per the API documentation for Federatable, https://github.com/troymohl/ddf/blob/bb7177ea3dd800f5d267f8bf41070e704fc9b2bb/catalog/core/catalog-core-api/src/main/java/ddf/catalog/federation/Federatable.java#L45
Setting enterprise to true is supposed to query all sources, so the intent of this code was to modify the query and remove the local source id from the list and then do a query to the other sources. In the case that a source is provided that is the local source (which shouldn't be possible since the UI doesn't show it), then this code would return an empty response. Otherwise, it performs the query as requested.

rzwiefel: I just mean the existing behavior of the catalog says if there are no source ids and its not enterprise we do have a default,
https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/QueryOperations.java#L772-L777

and the ui currently does allow you to send a query with no sources

troymohl:
@rzwiefel This is to support being able to store content & metadata in the local Solr and use CSW/OpenSearch/etc to query that, while at the same time allowing the UI to be pointed at a source that could be a broker to other searches, etc. E.g. the UI will query an externally managed source, and that source would then query the local Solr index.

troymohl:
I have pushed an update that removes the changes to the QueryApplication.java as the real goal was met by just removing the source from the front-end/javascript code.

mojogitoverhere:
Hero build was successful
Steps:
1. Built and started DDF
2. Checked "Disable Local Catalog" in the Catalog UI settings
3. Created and enabled a new CSW Specification Profile Federated Source with the default settings
4. Ingested some data
5. Created a new search with source set to my new CSW source
6. Verified all the results were from the CSW source with no ddf.distribution sources

NOTES:
Ran into an unrelated code coverage error while building without the `-DskipStatic=true` flag

> [INFO] --- jacoco-maven-plugin:0.7.4.201502262128:check (default-check) @ security-handler-pki ---
[INFO] Analyzed bundle 'DDF :: Security :: Handler :: PKI' with 5 classes
[WARNING] Rule violated for bundle DDF :: Security :: Handler :: PKI: branches covered ratio is 0.70, but expected minimum is 0.75
[WARNING] Rule violated for bundle DDF :: Security :: Handler :: PKI: complexity covered ratio is 0.71, but expected minimum is 0.73

bdeining:
build now

PR #2239, opened 2017-08-22
DDF-3228 Add logout notification prompts

brendan-hofmann:
- Fixes LoginFilter threshold to be a percentage
 instead of a hardcoded threshold that could result
 in thrashing for sufficiently low session lengths
 - Adds a backend service to check, extend,
 and invalidate a user's credentials
 - Adds an option to auto-execute logout actions
 in the logout view, so clicking the logout buttons
 is not required
 - Adds an admin ui modal that uses this backend
 - Adds a new intrigue modal component
 - Adds an intrigue modal using the new backend
 service and intrigue modal

#### What does this PR do?
Smooths out the user experience for a number of session timeout edge cases.
#### Who is reviewing it? 
@SmithJosh @djblue @emanns95 @AzGoalie @ahoffer 
https://github.com/orgs/codice/teams/security
@andrewkfiedler
@stustison
Hero process:

1. Build and start DDF
2. Update the session length in platform ui config to something like 70000.
3. Verify:
   1. Admin UI popup appears
   2. Admin UI popup logs out successfully if timeout expires
   3. Admin UI popup extends session successfully if selected
   4. Admin UI popup won't appear if user keeps interacting
   5. Catalog UI popup appears
   6. Catalog UI popup logs out successfully if timeout expires
   7. Catalog UI popup extends session successfully if selected
   8. Catalog UI popup won't appear if user keeps interacting
4. Try logging in with two guest sessions. Verify that when one guest session is expired, the other guest session continues to work.

Note: The inactivity timer will reset if the user clicks, moves the mouse, or presses a key. So make sure you don't interact with the window or you'll never see the timeout modals.
#### DDF Ticket
[DDF-3228](https://codice.atlassian.net/browse/DDF-3228)
image
image
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/admin/ui/src/main/webapp/js/application.js
+++ b/platform/admin/ui/src/main/webapp/js/application.js
@@ -29,7 +29,7 @@ define([
     'js/controllers/SystemUsage.controller',
     'text!templates/moduleTab.handlebars',
     'properties'
-    ],function (_, Backbone, Marionette, ich, $, poller, wreqr, Module, tabs, appHeader, header, footer, ModalController, SystemUsageController, moduleTab, Properties) {
+], function (_, Backbone, Marionette, ich, $, poller, wreqr, Module, tabs, appHeader, header, footer, ModalController, SystemUsageController, moduleTab, Properties) {
     'use strict';
 
     var Application = {};
@@ -37,7 +37,9 @@ define([
     // This was moved from the main.js file into here.
     // Since this modules has ui components, and it gets loaded before main.js, we need to init the renderer here for now until we sort this out.
     Marionette.Renderer.render = function (template, data) {
-        if(!template){return '';}
+        if (!template) {
+            return '';
+        }
         return ich[template](data);
     };
 
@@ -62,6 +64,7 @@ define([
         mainRegion: 'main',
         appHeader: '#appHeader',
         modalRegion: '#modalRegion',
+        sessionTimeoutModalRegion: '#sessionTmeoutModalRegion',
         alertsRegion: '.alerts'
     });
 
@@ -70,12 +73,12 @@ define([
         delay: 30000
     };
 
-    var addModuleRegions = function() {
+    var addModuleRegions = function () {
         //add tab regions
-        Application.ModuleModel.get('value').each(function(module) {
+        Application.ModuleModel.get('value').each(function (module) {
             var obj = {};
             obj[module.get('id')] = '#' + module.get('id');
-            if(!Application.App.getRegion(module.get('id'))) {
+            if (!Application.App.getRegion(module.get('id'))) {
                 Application.App.addRegions(obj);
             }
         });
@@ -101,7 +104,7 @@ define([
 
     modulePoller.start();
 
-    wreqr.vent.on('modulePoller:stop', function(){
+    wreqr.vent.on('modulePoller:stop', function () {
         modulePoller.stop();
     });
 
--- a/platform/admin/ui/src/main/webapp/js/controllers/Modal.controller.js
+++ b/platform/admin/ui/src/main/webapp/js/controllers/Modal.controller.js
@@ -13,8 +13,10 @@
 define(['jquery',
         'underscore',
         'marionette',
-        'js/wreqr'
-    ], function ($, _, Marionette, wreqr) {
+        'js/wreqr',
+        'js/views/SessionTimeoutModal.view',
+        'js/models/SessionTimeout'
+    ], function ($, _, Marionette, wreqr, SessionTimeoutView, SessionTimeout) {
         'use strict';
         var ModalController;
 
@@ -22,10 +24,18 @@ define(['jquery',
             initialize: function (options) {
                 this.application = options.application;
                 this.listenTo(wreqr.vent, "showModal", this.showModal);
+                this.listenTo(SessionTimeout, 'change:showPrompt', this.showSessionTimeoutModal);
             },
-            showModal: function(modalView) {
+            showModal: function (modalView) {
                 this.application.modalRegion.show(modalView);
                 modalView.show();
+            },
+            showSessionTimeoutModal: function () {
+                if (SessionTimeout.get('showPrompt')) {
+                    var modalView = new SessionTimeoutView();
+                    this.application.sessionTimeoutModalRegion.show(modalView);
+                    modalView.show();
+                }
             }
 
         });
--- a/platform/security/servlet/security-servlet-logout/src/main/webapp/js/main.js
+++ b/platform/security/servlet/security-servlet-logout/src/main/webapp/js/main.js
@@ -20,31 +20,39 @@
 
         var actions = JSON.parse(data);
 
+        var doLogout = function (action) {
+            $('iframe').attr('src', action.url);
+            $('#modal').removeClass('is-hidden');
+
+        };
+
         if (actions.length !== 0) {
             $('#noActions').toggleClass('is-hidden', actions.length !== 0);
             $('#actions').toggleClass('is-hidden', actions.length === 0);
             actions.forEach(function (action) {
-                var $row = $('<tr></tr>');
-                var $realm = $('<td></td>');
-                $realm.html(action.realm);
-                var $realmDescription = $('<a href="#" class="description" tabindex="-1"></a>');
-                $realmDescription.data('title', action.description);
-                $realmDescription.data('placement', 'right');
-                $realmDescription.append($('<i class="glyphicon glyphicon-question-sign"></i>'));
-                $realm.append($realmDescription);
-                var $user = $('<td></td>');
-                $user.html(action.auth);
-                var $logout = $('<td></td>');
-                var $button = $('<button>').text('Logout')
-                    .addClass('btn btn-primary float-right')
-                    .click(function () {
-                        $('iframe').attr('src', action.url);
-                        $('#modal').removeClass('is-hidden');
-
-                    });
-                $logout.append($button);
-                $row.append($realm).append($user).append($logout);
-                $('#actions tbody').append($row);
+                if ($.url().param('noPrompt')) {
+                    doLogout(action);
+                } else {
+                    var $row = $('<tr></tr>');
+                    var $realm = $('<td></td>');
+                    $realm.html(action.realm);
+                    var $realmDescription = $('<a href="#" class="description" tabindex="-1"></a>');
+                    $realmDescription.data('title', action.description);
+                    $realmDescription.data('placement', 'right');
+                    $realmDescription.append($('<i class="glyphicon glyphicon-question-sign"></i>'));
+                    $realm.append($realmDescription);
+                    var $user = $('<td></td>');
+                    $user.html(action.auth);
+                    var $logout = $('<td></td>');
+                    var $button = $('<button>').text('Logout')
+                        .addClass('btn btn-primary float-right')
+                        .click(function () {
+                            doLogout(action);
+                        });
+                    $logout.append($button);
+                    $row.append($realm).append($user).append($logout);
+                    $('#actions').find('tbody').append($row);
+                }
             });
 
             $('.description').tooltip();
--- a/platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
+++ b/platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
@@ -1,0 +1,161 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.security.servlet.expiry;
+
+import ddf.security.SecurityConstants;
+import ddf.security.Subject;
+import ddf.security.assertion.SecurityAssertion;
+import ddf.security.assertion.impl.SecurityAssertionImpl;
+import ddf.security.common.SecurityTokenHolder;
+import ddf.security.service.SecurityManager;
+import ddf.security.service.SecurityServiceException;
+import java.io.ByteArrayInputStream;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.time.Clock;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpSession;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import org.apache.cxf.ws.security.tokenstore.SecurityToken;
+import org.codice.ddf.security.handler.api.SAMLAuthenticationToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Path("/")
+public class SessionManagementService {
+  private static final Logger LOGGER = LoggerFactory.getLogger(SessionManagementService.class);
+
+  private SecurityManager securityManager;
+
+  private Clock clock = Clock.systemUTC();
+
+  private String rootContext;
+
+  @GET
+  @Path("/expiry")
+  public Response getExpiry(@Context HttpServletRequest request) {
+    HttpSession session = request.getSession(false);
+    long timeLeft = 0;
+    if (session != null) {
+      Object securityToken = session.getAttribute(SecurityConstants.SAML_ASSERTION);
+      if (securityToken instanceof SecurityTokenHolder) {
+        timeLeft = getTimeLeft((SecurityTokenHolder) securityToken);
+      }
+    }
+    return Response.ok(
+            new ByteArrayInputStream(Long.toString(timeLeft).getBytes(StandardCharsets.UTF_8)))
+        .build();
+  }
+
+  private long getTimeLeft(SecurityTokenHolder securityToken) {
+    List<SecurityAssertionImpl> values =
+        securityToken
+            .getRealmTokenMap()
+            .values()
+            .stream()
+            .map(SecurityAssertionImpl::new)
+            .collect(Collectors.toList());
+    values.sort(
+        (o1, o2) -> {
+          long l = o1.getNotOnOrAfter().getTime() - o2.getNotOnOrAfter().getTime();
+          if (l > 0) {
+            return 1;
+          } else if (l < 0) {
+            return -1;
+          } else {
+            return 0;
+          }
+        });
+    return values.isEmpty()
+        ? 0
+        : Math.max(values.get(0).getNotOnOrAfter().getTime() - clock.millis(), 0);
+  }
+
+  @GET
+  @Path("/renew")
+  public Response getRenewal(@Context HttpServletRequest request) {
+    HttpSession session = request.getSession(false);
+    Response[] response = new Response[1];
+    if (session != null) {
+      Object securityToken = session.getAttribute(SecurityConstants.SAML_ASSERTION);
+      if (securityToken instanceof SecurityTokenHolder) {
+        SecurityTokenHolder tokenHolder = (SecurityTokenHolder) securityToken;
+        Map<String, SecurityToken> realmTokenMap = tokenHolder.getRealmTokenMap();
+        realmTokenMap
+            .keySet()
+            .forEach(
+                s -> {
+                  try {
+                    doRenew(s, realmTokenMap.get(s), tokenHolder);
+                  } catch (SecurityServiceException e) {
+                    response[0] = Response.serverError().build();
+                    LOGGER.error("Failed to renew", e);
+                  }
+                });
+        if (response[0] == null) {
+          response[0] =
+              Response.ok(
+                      new ByteArrayInputStream(
+                          Long.toString(getTimeLeft(tokenHolder)).getBytes(StandardCharsets.UTF_8)))
+                  .build();
+        }
+      }
+    }
+    return response[0];
+  }
+
+  @GET
+  @Path("/invalidate")
+  public Response getInvalidate(@Context HttpServletRequest request) {
+    StringBuffer requestURL = request.getRequestURL();
+    String requestQueryString = request.getQueryString();
+    return Response.temporaryRedirect(
+            URI.create(
+                requestURL
+                    .substring(0, requestURL.indexOf(rootContext))
+                    .concat("/logout?noPrompt=true")
+                    .concat(requestQueryString != null ? "&" + requestQueryString : "")))
+        .build();
+  }
+
+  private void doRenew(String realm, SecurityToken securityToken, SecurityTokenHolder tokenHolder)
+      throws SecurityServiceException {
+    SAMLAuthenticationToken samlToken =
+        new SAMLAuthenticationToken(securityToken.getPrincipal(), securityToken, realm);
+    Subject subject = securityManager.getSubject(samlToken);
+    for (Object principal : subject.getPrincipals().asList()) {
+      if (principal instanceof SecurityAssertion) {
+        tokenHolder.addSecurityToken(realm, ((SecurityAssertion) principal).getSecurityToken());
+      }
+    }
+  }
+
+  public void setSecurityManager(SecurityManager securityManager) {
+    this.securityManager = securityManager;
+  }
+
+  public void setClock(Clock clock) {
+    this.clock = clock;
+  }
+
+  public void setRootContext(String rootContext) {
+    this.rootContext = rootContext;
+  }
+}
--- a/platform/security/servlet/security-servlet-session-expiry/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/platform/security/servlet/security-servlet-session-expiry/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- /**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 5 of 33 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.hbs
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+ --}}
+<div id="message">
+    <div>Your session will expire in <label id="timer">{{time}}</label> seconds.
+    Press "Continue working" if you need more time.</div>
```
emanns95: I feel like it's more common to say something along the lines of 
"Press 'Continue working' to remain logged in."

brendan-hofmann: I'll make them both consistent.

brendan-hofmann:
build now

brendan-hofmann:
build now

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,94 @@
...
+
+                        var blockingLightbox = BlockingLightbox.generateNewLightbox();
+                        blockingLightbox.model.updateTitle("Session Expiring");
+                        blockingLightbox.model.open();
+                        blockingLightbox.lightboxContent.show(new SessionTimeoutModal({
+                            time: idleNoticeDuration
+                        }));
+                    }
+                    else
+                        timer = setTimeout(onInterval, 1000 - ((now - start) % 1000));
```
SmithJosh: There's a problem here - the timer stops whenever the timeout modal is displayed. Initially, this timer only kept track of when the modal should be shown, and this was ok. Now it also has the responsibility of keeping track of when the session will expire on the backend, so the controller can keep it alive.

Now, the controller refreshes a session when there are min {sessionLength * 0.3, 60} seconds left. For a SAML assertion lifespan of < 3.5 minutes or so, there are < 60 seconds left when the controller decides to refresh, and this window could coincide with the 60 seconds where the modal is displayed - i.e. when the timer is stopped. So the session might expire before the user clicks "Continue working", and the session does not get extended.

We need to keep this timer running always in the controller so the session can always be refreshed on the backend. Same in the admin UI

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
```diff
@@ -0,0 +1,89 @@
...
+                    }
+
+                    if (now >= idleTimeoutDate) {
+                        $(document).off('keydown mousedown scroll');
+                        wreqr.vent.trigger('showModal', new SessionTimeoutModal({
+                            time: idleNoticeDuration
+                        }));
+                    }
+                    else
+                        timer = setTimeout(onInterval, 1000 - ((now - start) % 1000));
```
SmithJosh: Timer must be running always. It cannot stop when the modal is displayed.

platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
```diff
@@ -0,0 +1,156 @@
...
+                .concat("logout?noPrompt=true")
+                .concat(requestQueryString != null ? "&" + requestQueryString : "")))
+                .build();
+    }
+
+    private void doRenew(String realm, SecurityToken securityToken, SecurityTokenHolder tokenHolder)
+            throws SecurityServiceException {
+        SAMLAuthenticationToken samlToken =
+                new SAMLAuthenticationToken(securityToken.getPrincipal(), securityToken, realm);
+        Subject subject = securityManager.getSubject(samlToken);
```
stustison: We probably need a ticket to enhance the code behind this to handle the IdP case correctly. Right now this will work fine if DDF is serving as the IdP. However, if we're using some other IdP, we will most likely not have direct access to their STS and this renewal will fail. If the assertion we're working with is clearly from an IdP, then we would need to use one of the SSO profile bindings (like the SOAP one) that can be done without a user in the loop, to renew the assertion or request a new one.

brendan-hofmann: Agreed. Created https://codice.atlassian.net/browse/DDF-3246 for this issue.

SmithJosh:
Hero process:
1. Build and start DDF
2. Find the bundle ID of the UI component that you want to test
* `list | grep -i admin`
3. Watch that bundle
* `bundle:watch --start \<ID\>`
4. Change the timeout value to something shorter (say 2 minutes = 120000ms)
* [Search UI](https://github.com/brendan-hofmann/ddf/blob/8f564b13a45d5af651ed044ec3e2ab2b530c7e7d/catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js#L29)
* [Admin UI](https://github.com/brendan-hofmann/ddf/blob/8f564b13a45d5af651ed044ec3e2ab2b530c7e7d/platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js#L28)
5. Rebuild the bundle, making sure that it gets pulled in by Karaf
6. Test!

Note: The inactivity timer will reset if the user clicks, scrolls, or presses a key. So make sure you don't interact with the window or you'll never see the timeout modals

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/lightbox/lightbox.view.js
```diff
@@ -0,0 +1,70 @@
...
+            this.$el.toggleClass('is-open', this.model.isOpen());
+        },
+        handleOutsideClick: $.noop,
+        close: function () {
+            this.model.close();
+            this.lightboxContent.empty();
+        }
+    },{
+        generateNewLightbox: function(){
+            console.log("Creating model");
```
djblue: Was this left here for debugging?

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+        'underscore',
+        'marionette',
+        'js/wreqr',
+        './lightbox/lightbox.view',
+        './session-timeout.view'
+    ], function ($, _, Marionette, wreqr, BlockingLightbox, SessionTimeoutModal) {
+        'use strict';
+        var SessionTimeoutController;
+
+        var _this = null;
```
djblue: What is this variable being used for?

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+                    });
+                }
+
+                function onInterval() {
+                    var now = Date.now();
+
+                    if (now >= sessionRenewDate) {
+                        $.get("/services/session/renew", function () {
+                            console.log("session renewed");
+                            setSessionRenewDate();
```
djblue: Do we need to make two ajax requests, or can `/services/session/renew` return the expiry timeout?

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+                        modalShown = true;
+                    }
+                }
+
+                function onUserActivity() {
+                    idleTimeoutDate = idleTimeoutThreshold - idleNoticeDuration + Date.now();
+                }
+
+                setSessionRenewDate();
+                timer = setTimeout(onInterval, 1000);
```
djblue: The callback passed to `setTimeout` is called `onInterval`. Did we mean to have `setInterval`?

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.view.js
```diff
@@ -0,0 +1,92 @@
...
+                        var deltaTime = msRemaining % 1000;
+
+                        $(_this.$el).find('#timer').text(Math.ceil(msRemaining / 1000));
+                        timer = setTimeout(onInterval, (deltaTime === 0) ? 1000 : deltaTime);
+                    }
+                    else
+                        _this.logoutUser();
+                }
+
+                timer = setTimeout(onInterval, (deltaTime === 0) ? 1000 : deltaTime);
```
djblue: Same as the comment above.

platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
```diff
@@ -0,0 +1,156 @@
...
+                timeLeft - clock.millis(), 0))
+                .getBytes(StandardCharsets.UTF_8)))
+                .build();
+    }
+
+    @GET
+    @Path("/renew")
+    public Response getRenewal(@Context HttpServletRequest request) {
+        HttpSession session = request.getSession(false);
+        final Response[] response = new Response[1];
```
pklinef: ❓ Why are you using an array of responses when you always just use a single index?

platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
```diff
@@ -0,0 +1,156 @@
...
+        return response[0];
+    }
+
+    @GET
+    @Path("/invalidate")
+    public Response getInvalidate(@Context HttpServletRequest request) {
+        StringBuffer requestURL = request.getRequestURL();
+        String requestQueryString = request.getQueryString();
+        return Response.temporaryRedirect(URI.create(requestURL.substring(0,
+                requestURL.indexOf("services"))
```
pklinef: ❗️ Should use `org.codice.ddf.system.rootContext` system property instead of hardcoding.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/lightbox/lightbox.view.js
SmithJosh: Yes, thanks for finding that. It should be removed.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
SmithJosh: It's no longer being used. It can be removed.

platform/security/servlet/security-servlet-session-expiry/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,32 @@
...
+
+
+    <reference id="securityManager" interface="ddf.security.service.SecurityManager"/>
+
+    <bean id="sessionService"
+          class="org.codice.ddf.security.servlet.expiry.SessionManagementService">
+        <property name="securityManager" ref="securityManager"/>
+    </bean>
+
+    <jaxrs:server id="restService" address="/session">
```
pklinef: ❓ Do we want to make it clear this is really only for internal use?  If so then we should consider moving this to `/internal/session`.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
SmithJosh: No, `onInterval` is correct. This code is basically an `setInterval` implementation, but it uses `setTimeout` so that the timer inaccuracies can be corrected at each interval - hence the `onInterval` callback.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.view.js
SmithJosh: `onInterval` is correct, as above

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
```diff
@@ -0,0 +1,92 @@
...
+                var modalShown = false;
+
+                wreqr.vent.on("continueClicked", function () {
+                    modalShown = false;
+                });
+
+                function setSessionRenewDate() {
+                    $.get("/services/session/expiry", function (res) {
+                        var msUntilTimeout = parseInt(res);
+                        if (msUntilTimeout === 0)
```
pklinef: ✏️  See previous comment about else blocks.

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
```diff
@@ -0,0 +1,92 @@
...
+        SessionTimeoutController = Marionette.Object.extend({
+            initialize: function () {
+                _this = this;
+                _this.initTimer();
+
+                wreqr.vent.on("sessionRenewed", _this.initTimer);
+            },
+
+            initTimer: function () {
+                var timer = null;
```
pklinef: ❗️  It looks like you can end up with a timer that is not cleared if sessionRenewed is called and timer is set to null while active.

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
```diff
@@ -0,0 +1,92 @@
...
+        'marionette',
+        'js/wreqr',
+        'js/views/SessionTimeoutModal.view'
+    ], function ($, _, Marionette, wreqr, SessionTimeoutModal) {
+        'use strict';
+        var SessionTimeoutController;
+
+        var _this = null;
+        var idleNoticeDuration = 60000;
+        var idleTimeoutThreshold = 900000;  // Length of inactivity that will trigger user timeout
```
pklinef: ❓ Would it be possible to make this configurable?

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
```diff
@@ -0,0 +1,92 @@
...
+                                            // See STIG V-69243
+
+        SessionTimeoutController = Marionette.Object.extend({
+            initialize: function () {
+                _this = this;
```
pklinef: ✏️  `_this` is only used in the `initialize` method.  I think it can be removed.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.view.js
```diff
@@ -0,0 +1,92 @@
...
+                    var now = Date.now();
+
+                    if (dateOfTimeout > now) {
+                        var msRemaining = dateOfTimeout - now;
+                        var deltaTime = msRemaining % 1000;
+
+                        $(_this.$el).find('#timer').text(Math.ceil(msRemaining / 1000));
+                        timer = setTimeout(onInterval, (deltaTime === 0) ? 1000 : deltaTime);
+                    }
+                    else
```
pklinef: ✏️  else block

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+                var modalShown = false;
+
+                wreqr.vent.on("continueClicked", function () {
+                    modalShown = false;
+                });
+
+                function setSessionRenewDate() {
+                    $.get("/services/session/expiry", function (res) {
+                        var msUntilTimeout = parseInt(res);
+                        if (msUntilTimeout === 0)
```
pklinef: ✏️  `If` (also `else`) should always be in a block `{ }`.  This avoids future bugs if someone tries to add another line but does not notice it is not in a block.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.view.js
```diff
@@ -0,0 +1,92 @@
...
+            ui: {
+                continueWorkingBtn: "#continueWorkingBtn"
+            },
+
+            events: {
+                'click @ui.continueWorkingBtn': 'refreshSession'
+            },
+
+            initialize: function (options) {
+                _this = this;
```
pklinef: ❓ Are you sure you need `_this`?  Marionette should be properly binding this.  Most uses of `_this` in this file could be replaced with `this` which should properly bind those methods as well to the proper context.  The `_this` in the `$.get` would need to use the context parameter.

http://api.jquery.com/jquery.ajax/

platform/admin/ui/src/main/webapp/index.html
```diff
@@ -39,6 +39,7 @@
 
     </div>
     <div id="modalRegion"></div>
+    <div id="sessionTimeoutModalRegion"></div>
```
alexaabrd: I dont think this is being used anywhere - the modal appends to modalRegion on `wreqr.vent.trigger('showModal', new SessionTimeoutModal)`

platform/admin/ui/src/main/webapp/js/application.js
```diff
@@ -62,6 +64,7 @@ define([
         mainRegion: 'main',
         appHeader: '#appHeader',
         modalRegion: '#modalRegion',
+        sessionTimeoutModalRegion: '#sessionTimeoutModalRegion',
```
alexaabrd: I dont think this is being used anywhere - the modal appends to modalRegion on wreqr.vent.trigger('showModal', new SessionTimeoutModal)

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.view.js
SmithJosh: In the places where `_this` is used, the scope is different than the `initialize` function, so I thought it was necessary. But I was not aware of the jQuery context parameter, or that Marionette affects the binding of `this`.

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
SmithJosh: Sure. It's a STIG requirement, though, so it's not something we want modified after the build, right (what happens if an admin changes it to 20min of inactivity)? What's the appropriate place for the configurable value?

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
SmithJosh: Yes, it can be removed. Thanks.

platform/admin/ui/src/main/webapp/index.html
SmithJosh: You're right. The region is there so we can add the timeout modal without affecting modals already being displayed, but I think that is a non-issue and we can remove this.

This brings up another point, though - do we handle stacked modals correctly? The timeout modal should always be on top, but I'm not sure that it is. We should verify this.

platform/admin/ui/src/main/webapp/js/application.js
SmithJosh: See above. I think we can remove this.

alexaabrd:
Why implement a completely new lightbox instead of creating a specialized/custom one under components > lightbox, similar to blocking-lightbox? Since it is just css changes you should be able to override the default-lightbox css with a unique id or classname for you custom-lightbox

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
djblue: I can't seem to find where `onInterval` is being run after the initial timeout. How are we achieving the goal of checking on an interval?

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+ **/
+/* jshint browser: true */
+/* global define */
+define(['jquery',
+        'underscore',
+        'marionette',
+        'js/wreqr',
+        './lightbox/lightbox.view',
+        './session-timeout.view'
+    ], function ($, _, Marionette, wreqr, BlockingLightbox, SessionTimeoutModal) {
```
alexaabrd: I would rename "BlockingLightbox" to something more accurate for your custom lightbox like "sessionTimeoutLightbox". The BlockingLightbox variable name came from pulling in `component/lightbox/blocking/lightbox.blocking.view`

platform/admin/ui/src/main/webapp/index.html
alexaabrd: Well right now for the admin modals only one modal will display at a time because the .show() 
 on this [line](https://github.com/brendan-hofmann/ddf/blob/8f564b13a45d5af651ed044ec3e2ab2b530c7e7d/platform/admin/ui/src/main/webapp/js/controllers/Modal.controller.js#L27) will automatically destroy the previous view.
Im not sure about the modals in intrigue though

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
brendan-hofmann: It's also highly unlikely that a user would ever encounter this during the course of normal usage. In fact, the renewal aspect makes it *less* likely for them to get logged out unexpectedly. Given that, I don't think it needs to be configurable.

platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
brendan-hofmann: Lambda below requires effectively final reference.

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+
+        var idleNoticeDuration = 60000;
+        var idleTimeoutThreshold = 900000;  // Length of inactivity that will trigger user timeout
+        // See STIG V-69243
+
+        SessionTimeoutController = Marionette.Object.extend({
+            initialize: function () {
+                this.initTimer();
+
+                wreqr.vent.on("sessionRenewed", _this.initTimer);
```
SmithJosh: ❗️ Undefined reference to _this

catalog/ui/catalog-ui-search/src/main/webapp/component/session-timeout-modal/session-timeout.controller.js
```diff
@@ -0,0 +1,96 @@
...
+                    var msUntilTimeout = parseInt(res);
+                    if (msUntilTimeout === 0) {
+                        clearTimeout(timer);
+                    }
+
+                    var msUntilAutoRenew = Math.max(msUntilTimeout * 0.7, msUntilTimeout - 60000);
+                    sessionRenewDate = Date.now() + msUntilAutoRenew;
+                }
+
+                function onInterval() {
```
SmithJosh: ❗️ The `onInterval` logic is broken. It needs a `setTimeout` call at the end so the timer stays alive. Right now it is only called once and the modal will never display.

brendan-hofmann: This is fixed.

platform/security/servlet/security-servlet-logout/src/main/webapp/js/main.js
```diff
@@ -20,31 +20,39 @@
...
-                    .click(function () {
-                        $('iframe').attr('src', action.url);
-                        $('#modal').removeClass('is-hidden');
-
-                    });
-                $logout.append($button);
-                $row.append($realm).append($user).append($logout);
-                $('#actions tbody').append($row);
+                if ($.url().param('noPrompt')) {
+                    doLogout(action);
```
brendan-hofmann: ❓ is this a blocking call? In other words, if there is more than one logout to perform, will the next logout wait for the first logout's redirects to complete inside the iframe?

emanns95:
I manually tested and verified the following paths:

* Admin UI popup appears
* Admin UI popup logs out successfully if button is pressed
* Admin UI popup logs out successfully if timeout expires
* Admin UI popup extends session successfully if selected
* Admin UI popup won't appear if user keeps interacting
* Catalog UI popup appears
* Catalog UI popup logs out successfully if timeout expires
* Catalog UI popup extends session successfully if selected
* Catalog UI popup won't appear if user keeps interacting

platform/admin/ui/src/main/webapp/js/controllers/SessionTimeout.controller.js
pklinef: Different organizations and companies might have different requirements.  It is also possible that they do not want it at all and should be able to disable it.

platform/admin/ui/src/main/webapp/js/controllers/Modal.controller.js
```diff
@@ -13,19 +13,26 @@
...
+        'js/views/SessionTimeoutModal.view'
+    ], function ($, _, Marionette, wreqr, SessionTimeoutView) {
         'use strict';
         var ModalController;
 
         ModalController = Marionette.Controller.extend({
             initialize: function (options) {
                 this.application = options.application;
                 this.listenTo(wreqr.vent, "showModal", this.showModal);
+                this.listenTo(wreqr.vent, "showSessionTimeoutModal", this.showSessionTimeoutModal);
```
andrewkfiedler: Similar to how we eliminated the `wreqr` in Intrigue, we should do the same here. 

Require in the sessionTimeout model since it's a singleton and listen to the change on `showPrompt`.  If it's true, show the modal, if it's not hide it.

brendan-hofmann:
build now

emanns95:
Was able to successfully hero for all situations except ii, since there is no log out button on the pop up in the admin ui 👍

brendan-hofmann:
Thanks @emanns95 , I forgot to delete that step when I removed the button.

platform/platform-configuration/src/main/java/org/codice/ddf/configuration/PlatformUiConfiguration.java
```diff
@@ -70,6 +72,8 @@
 
     private String background;
 
+    private Integer timeout;
```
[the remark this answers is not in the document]
brendan-hofmann: No.

bdeining:
build now

brendan-hofmann:
build now

PR #2247, opened 2017-08-23
DDF-3225 Fix edge case handling in XStreamPathConverter

adimka:
#### What does this PR do?
Fixes an edge case causing indexed fields to not be considered equal in cases where they should be, and adds a unit test covering this edge case.
#### Who is reviewing it? 
@glenhein @mackncheesiest 
@bdeining
@millerw8
Full build should suffice.
Fixes and adds test coverage for a previously untested edge case.
[DDF-2335](https://codice.atlassian.net/browse/DDF-2335)
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/XstreamPathConverter.java
+++ b/catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/XstreamPathConverter.java
@@ -148,13 +148,15 @@ public class XstreamPathConverter implements Converter {
                 break;
             }
         }
+        i = countPastTag(path1, i);
+        j = countPastTag(path2, j);
         return (i == path1.length && j == path2.length) || endsMatch(path1, i, path2, j);
     }
 
     /**
      * Checks if 2 paths who have matched up to the provided indices, with 1 index being at or
-     * beyond the end of the path, match by accounting for ending count indexes and attribute
-     * value presence. This finishes the comparison of 2 paths once one path has been traversed.
+     * beyond the end of the path, match by accounting for attribute value presence. This finishes
+     * the comparison of 2 paths once one path has been traversed.
      *
      * @param path1 The first path
      * @param index1 The current index in path1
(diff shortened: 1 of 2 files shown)


adimka:
build now

bdeining:
build now

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/XstreamPathConverter.java
```diff
@@ -164,13 +166,11 @@ protected boolean doBasicPathsMatch(final Path pathObj1, final Path pathObj2) {
      */
     private boolean endsMatch(char[] path1, int index1, char[] path2, int index2) {
```
glenhein: I'm wondering if we should take a different approach on this. Instead of this complex logic to walk both strings for comparison, we could have simpler code that efficiently strips out the xpath index characters and the xpath attribute indicator. Then just compare those strings.

glenhein: Also, I believe a lot the strings being compared are repeated, so maybe caching would be a good option as well.

adimka: That was my original approach. It worked pretty well (ran at about 1/3rd the original runtime), but a single pass approach reduced that improved runtime by another 1/3rd down to < 1/5th the original runtime. I think that kind of increase is worth it.
I agree this is a considerably more cognitively complex, but this change isolates the tag skipping logic from the attribute handling (where I had foolishly combined them before -resulting in the need for this fix).
As far as caching goes, I think that would be worth looking into.

adimka:
build now

catalog/spatial/csw/spatial-csw-transformer/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/converter/XstreamPathConverter.java
mackncheesiest: iirc that "simplify and then compare those strings" was the first approach implemented, but the difference in runtime still actually ended up being pretty significant in that you go from essentially two traversals of each to a single traversal of each string in the current approach.

adimka:
build now

adimka:
build now

adimka:
build now

PR #2254, opened 2017-08-24
DDF-3236 Added window file path support

emanns95:
#### What does this PR do?
Walks the file system in order to parse escaped Windows file paths so they can be used in console commands.

#### Who is reviewing it? 
@ahoffer 
@vinamartin 
@paouelle 

@bdeining
@clockard

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
@@ -19,6 +19,8 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
+import org.apache.commons.lang.SystemUtils;
+
 import ddf.security.common.audit.SecurityLogger;
 
 /*
@@ -34,17 +36,18 @@ public class CrossPlatformFilePathEvaluator {
             return file;
         }
 
-        //look for drive letter
-        if (path.contains(":")) {
-            String[] splitPath = path.split(":");
-            Optional<File> result = handleWindowsPath(splitPath[1], new File(splitPath[0] + ":\\"));
-            if (result.isPresent()) {
-                file = result.get();
-            }
-        } else { //relative path
-            Optional<File> result = handleWindowsPath(path, new File(System.getProperty("ddf.home")));
-            if (result.isPresent()) {
-                file = result.get();
+        //If on Windows, check Windows file path
+        if (SystemUtils.IS_OS_WINDOWS) {
+            //look for drive letter
+            if (path.contains(":")) {
+                String[] splitPath = path.split(":");
+                Optional<File> result = handleWindowsPath(splitPath[1],
+                        new File(splitPath[0] + ":" + File.separatorChar));
+                return result.orElse(file);
+            } else { //relative path
+                Optional<File> result = handleWindowsPath(path,
+                        new File(System.getProperty("ddf.home")));
+                return result.orElse(file);
             }
         }
         return file;
@@ -60,16 +63,13 @@ public class CrossPlatformFilePathEvaluator {
         for (File file : fileList) {
             if (path.equals(file.getName())) {
                 //found the exact match
-                SecurityLogger.audit("Found file with path : {}",
-                        file.getAbsolutePath());
+                SecurityLogger.audit("Found file with path : {}", file.getAbsolutePath());
                 return Optional.of(file);
             } else if (path.startsWith(file.getName())) {
                 Optional<File> result = handleWindowsPath(path.replace(file.getName(), ""), file);
                 if (result.isPresent()) {
                     return result;
                 }
-            } else {
-                continue;
             }
         }
         return Optional.empty();
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+
+package org.codice.ddf.commands.util;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+import ddf.security.common.audit.SecurityLogger;
+
+/*
+ * Handles converting given Windows file paths (whose backslashes get escaped) into something the
+ * commands can handle.
+ */
+public class CrossPlatformFilePathEvaluator {
+
+    public static File handlePath(String path) {
+        //return java can find the file using the path
+        File file = new File(path);
+        if (file.exists()) {
+            return file;
+        }
+
+        //look for drive letter
+        if (path.contains(":")) {
+            String[] splitPath = path.split(":");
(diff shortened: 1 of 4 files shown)


emanns95:
I have unit tests written out for this, but I'm unsure how I can unit test it on temporary files because the directory names appear to differ slightly from how it's represented in the File object and in the actual file system. Going to push the tests I currently have as a second commit for review

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+                SecurityLogger.audit("Found file with path : {}",
+                        file.getAbsolutePath());
+                return Optional.of(file);
+            } else if (path.startsWith(file.getName())) {
+                Optional<File> result = handleWindowsPath(path.replace(file.getName(), ""), file);
+                if (result.isPresent()) {
+                    return result;
+                }
+            } else {
+                continue;
```
emanns95: I don't need this case. Idk why I put it there

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+
+import ddf.security.common.audit.SecurityLogger;
+
+/*
+ * Handles converting given Windows file paths (whose backslashes get escaped) into something the
+ * commands can handle.
+ */
+public class CrossPlatformFilePathEvaluator {
+
+    public static File handlePath(String path) {
```
lessarderic: ❓ Any reason we couldn't have used `Paths.get()`, `Path.isAbsolute()`, etc., instead of writing our own class? My assumption was that most of the OS specific differences in paths were taken care of by those new classes (see https://docs.oracle.com/javase/tutorial/essential/io/pathOps.html).

emanns95: The command gets the path string with the backslashes removed, so there's no way to tell the actual path structure without walking through the file system and trying to figure it out. For example, `C:\Users\greatUser\Downloads\example.txt` is received as `C:UsersgreatUserDownloadsexample.txt` and nothing can really figure out that's a path

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+        //return java can find the file using the path
+        File file = new File(path);
+        if (file.exists()) {
+            return file;
+        }
+
+        //look for drive letter
```
paouelle: ❗️ Let's protect this code so it only runs if Windows (FilenameUtils.isSystemWindows())

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+        if (path.contains(":")) {
+            String[] splitPath = path.split(":");
+            Optional<File> result = handleWindowsPath(splitPath[1], new File(splitPath[0] + ":\\"));
+            if (result.isPresent()) {
```
paouelle: ✏️ Can be changed to return result.orElse(file);

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+                file = result.get();
+            }
+        } else { //relative path
+            Optional<File> result = handleWindowsPath(path, new File(System.getProperty("ddf.home")));
+            if (result.isPresent()) {
```
paouelle: ✏️ Can be changed to return result.orElse(file);

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+        //return java can find the file using the path
+        File file = new File(path);
+        if (file.exists()) {
+            return file;
+        }
+
+        //look for drive letter
+        if (path.contains(":")) {
+            String[] splitPath = path.split(":");
+            Optional<File> result = handleWindowsPath(splitPath[1], new File(splitPath[0] + ":\\"));
```
paouelle: ✏️ For the \\, use File.separatorChar

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    private static Optional<File> handleWindowsPath(String path, File pwd) {
+        if (!pwd.exists()) {
+            return Optional.empty();
+        }
+
+        List<File> fileList = Arrays.asList(pwd.listFiles());
+
+        for (File file : fileList) {
```
paouelle: 👍 Nice recursion to handle cases where you might match a directory but nothing found underneath and yet you continue with the next sibling which might match a longer directory name and match the rest underneath it.

catalog/core/catalog-core-commands/src/test/java/org/codice/ddf/commands/catalog/CrossPlatformFilePathEvaluatorTest.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import org.codice.ddf.commands.util.CrossPlatformFilePathEvaluator;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class CrossPlatformFilePathEvaluatorTest {
```
paouelle: might want to add test cases when the file is not found anywhere. If it is found in 2 places (e.g. down/load/a and download/a), tests with Linux files and Windows files.

We might also want to make sure that these work on both Linux/Mac and Windows which might require creating tests cases with assumptions for Windows and for Linux/Mac only.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/util/CrossPlatformFilePathEvaluator.java
lessarderic: Are you saying that the `filePath` member variable in the command class (and therefore the `path` parameter here) doesn't have the `\`s in it when Karaf executes the command?

emanns95: Yes, they are removed before we get to any code in here.

clockard: We should probably write a karaf bug for this. Can you get a valid path by escaping the `\` characters in the console?

emanns95: You can get a valid path by replacing `\` with `\\\\`. I'm not sure why it requires four backslashes to escape and get one...

paouelle: I have seen this before somewhere else, it is because it gets through a double interpretation. The first one replaces \\\\ with \\ and then the second one goes from \\ to \

emanns95: I made a Karaf issue https://issues.apache.org/jira/browse/KARAF-5320

lessarderic: Thanks for the clarification @emanns95.

emanns95:
The root cause of this issue should be resolved in https://issues.apache.org/jira/browse/KARAF-5320.

Once the DDF is updated to use Karaf `4.1.3+`, Windows path support can be revisited. I think it will work out of the box after that Karaf update occurs

PR #2256, opened 2017-08-25
DDF-3242 Updates Twitter source to work out of the box

vinamartin:
#### What does this PR do?
Updates Twitter source to implement ConfiguredService interface so it will work when deployed. Currently even though the availability check works, the source will not show up as available without this change.

#### Who is reviewing it? 
@ahoffer 

@clockard
@coyotesqrl

Build and deploy the twitter source into any running DDF instance. 
Configure the source via the sources tab with an API key and secret, these can be obtained from Twitter's website. 

#### Review Comment Legend:

### DIFF
--- a/distribution/sdk/sample-twitter-source/src/main/java/org/codice/ddf/catalog/twitter/source/TwitterSource.java
+++ b/distribution/sdk/sample-twitter-source/src/main/java/org/codice/ddf/catalog/twitter/source/TwitterSource.java
@@ -40,6 +40,7 @@ import ddf.catalog.operation.impl.SourceResponseImpl;
 import ddf.catalog.resource.ResourceNotFoundException;
 import ddf.catalog.resource.ResourceNotSupportedException;
 import ddf.catalog.resource.ResourceReader;
+import ddf.catalog.service.ConfiguredService;
 import ddf.catalog.source.FederatedSource;
 import ddf.catalog.source.SourceMonitor;
 import ddf.catalog.source.UnsupportedQueryException;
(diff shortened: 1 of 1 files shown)


clockard:
build now

bdeining:
Is there a ticket associated with this PR?

distribution/sdk/sample-twitter-source/src/main/java/org/codice/ddf/catalog/twitter/source/TwitterSource.java
```diff
@@ -183,10 +186,9 @@ private Metacard getMetacard(Status status) {
         metacard.setSourceId(id);
         metacard.setId(String.valueOf(status.getId()));
         metacard.setTitle(status.getText());
-        metacard.setMetadata("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>" +
-                "<Resource>" +
-                "<name>" + status.getText() + "</name>" +
-                "</Resource>");
+        metacard.setMetadata(
+                "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>" + "<Resource>"
```
coyotesqrl: ✏️ I would consider making a static final const format string, or even moving it to a resource. The string would look like
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?><Resource><name>%s</name></Resource>
```
and you'd set it here with `metacard.setMetadata(String.format(formatString, status.getText())`

I suggest a resource only so you can avoid all of the escapes for quotation marks.

coyotesqrl:
This should have an associated Jira ticket and the commit message should be updated to reflect that.

vinamartin:
build now

PR #2258, opened 2017-08-28
DDF-3247 Adds granular permissions checking to Configurator operations.

coyotesqrl:
#### What does this PR do?
Adds granular checks before allowing a user to make feature/bundle changes.

#### Who is reviewing it? 
@tbatie @Lambeaux 

@codice/security

@clockard
@stustison

1. Full build and test. 
2. Install of the downstream Admin Console Beta kar and run through its operations. 
3. Attempt to use the Configurator to perform feature/bundle operations with an unauthorized user.

N/A

[DDF-3247](https://codice.atlassian.net/browse/DDF-3247)

N/A

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
+++ b/platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
@@ -20,6 +20,8 @@ import java.util.Arrays;
 
 import org.apache.karaf.bundle.core.BundleState;
 import org.apache.karaf.bundle.core.BundleStateService;
+import org.apache.shiro.SecurityUtils;
+import org.apache.shiro.subject.Subject;
 import org.codice.ddf.admin.configurator.ConfiguratorException;
 import org.codice.ddf.admin.configurator.Operation;
 import org.codice.ddf.admin.configurator.Result;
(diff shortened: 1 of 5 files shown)


coyotesqrl:
Pushed some additional changes to fix and expand unit tests.

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
```diff
@@ -129,4 +155,36 @@ private BundleStateService getBundleStateService() {
...
+        return true;
+    }
+
+    private boolean isPermittedToViewService(String servicePid) {
+        KeyValueCollectionPermission serviceToCheck = new KeyValueCollectionPermission(
+                "view-service.pid",
+                new KeyValuePermission("service.pid", Sets.newHashSet(servicePid)));
+
+        try {
+            return security.runWithSubjectOrElevate(() -> SecurityUtils.getSubject()
```
stustison: You should only need to call `runWithSubjectOrElevate` if this code is called from the karaf command line. Is that the case here? Otherwise you should just check for a Subject and if != null, check permissions, otherwise return false.

coyotesqrl: I'll update and push shortly...hold off reviewing more.

coyotesqrl:
build now

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
```diff
@@ -129,4 +152,30 @@ private BundleStateService getBundleStateService() {
...
+
+    private boolean isPermittedToViewBundle(Bundle bundle) {
+        ServiceReference<?>[] services = bundle.getRegisteredServices();
+        if (services == null || services.length == 0) {
+            return true;
+        }
+
+        for (ServiceReference<?> service : services) {
+            Object pidProperty = service.getProperty("service.pid");
+            if (pidProperty instanceof String) {
```
clockard: Just curious if we need this instanceof check. Will the service.pid ever be anything other than a string?

platform/admin/configurator/admin-configurator-impl/src/main/java/org/codice/ddf/admin/configurator/impl/BundleOperation.java
```diff
@@ -129,4 +152,30 @@ private BundleStateService getBundleStateService() {
...
+                if (!isPermittedToViewService((String) pidProperty)) {
```
clockard: Could we just do `return isPermittedToViewService((String) pidProperty)` and remove the if statement?

coyotesqrl: Could get a class cast in that case. Shouldn't. Ever. But could.

clockard: If you keep the outer instanceof if statement there shouldn't be a class cast issue with removing the inner if

coyotesqrl: Duh. Yeah. When I read your comment earlier, it was without the context of what's in the if block and I had forgotten it was exactly as dumb as it was. Of course I can change that.

coyotesqrl: Nope. Scratch that. Looking at context I can't do that. I need to run that in a loop and I'm breaking out if and only if there is a service to which the Subject does not have permissions. If I change it to a return, I'll always return the response for the first service and never check the rest.

djblue:
Manual verification was successful :+1:

PR #2270, opened 2017-08-30
DDF-3251 Allow descriptive tooltips when selecting an attribute in advanced search

albany551:
#### What does this PR do?
Lets descriptions for metacard attributes be edited and stored in system configuration, to be shown as tooltips in the attribute selector dropdown (in advanced search).

#### Who is reviewing it? 
@bdeining 
@andrewkfiedler 
@millerw8 

@bdeining 
@millerw8

In admin console, navigate to Standard Search UI -> Configuration -> Catalog UI Search and add an entry to the new `Attribute Descriptions` section.
For example, `title=Descriptive title for metacard`
In Intrigue, navigate to the advanced search editor. Choose the attribute you created a description for and verify that a tooltip appears over it with the description text.

https://codice.atlassian.net/browse/DDF-3251

screenshot from 2017-08-29 15-00-07
screenshot from 2017-08-29 15-09-34

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -147,6 +147,13 @@ public class ConfigurationApplication implements SparkApplication {
         return hiddenAttributes;
     }
 
+    public List<String> getAttributeDescriptions() {
+        return attributeDescriptions.entrySet()
+                .stream()
+                .map(pair -> String.format("%s=%s", pair.getKey(), pair.getValue()))
+                .collect(Collectors.toList());
+    }
+
     public void setScheduleFrequencyList(List<Long> scheduleFrequencyList) {
         this.scheduleFrequencyList = scheduleFrequencyList;
     }
@@ -173,6 +180,12 @@ public class ConfigurationApplication implements SparkApplication {
         this.hiddenAttributes = hiddenAttributes;
     }
 
+    public void setAttributeDescriptions(List<String> attributeDescriptions) {
+        this.attributeDescriptions = attributeDescriptions.stream()
+                .map(str -> str.split("="))
+                .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
+    }
+
     private List<String> readOnly = ImmutableList.of("checksum",
             "checksum-algorithm",
             "id",
@@ -189,6 +202,8 @@ public class ConfigurationApplication implements SparkApplication {
     private Map<String, String> attributeAliases = Collections.emptyMap();
 
     private List<String> hiddenAttributes = Collections.emptyList();
+
+    private Map<String, String> attributeDescriptions = Collections.emptyMap();
 
     private int sourcePollInterval = 60000;
 
@@ -252,6 +267,7 @@ public class ConfigurationApplication implements SparkApplication {
         config.put("summaryShow", summaryShow);
         config.put("resultShow", resultShow);
         config.put("hiddenAttributes", hiddenAttributes);
+        config.put("attributeDescriptions", attributeDescriptions);
         config.put("attributeAliases", attributeAliases);
         config.put("sourcePollInterval", sourcePollInterval);
         config.put("scheduleFrequencyList", scheduleFrequencyList);
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -171,18 +171,7 @@ public class ConfigurationApplication implements SparkApplication {
     }
 
     public void setAttributeAliases(List<String> attributeAliases) {
-        this.attributeAliases = attributeAliases.stream()
-                .map(str -> str.split("=", 2))
-                .filter((list) -> {
-                    if (list.length <= 1) {
-                        if (LOGGER.isDebugEnabled()) {
-                            LOGGER.debug("Filtered out invalid attribute/alias: {}", list[0]);
-                        }
-                        return false;
-                    }
-                    return true;
-                })
-                .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
+        this.attributeAliases = parseAttributeAndValuePairs(attributeAliases);
     }
 
     public void setHiddenAttributes(List<String> hiddenAttributes) {
@@ -190,18 +179,7 @@ public class ConfigurationApplication implements SparkApplication {
     }
 
     public void setAttributeDescriptions(List<String> attributeDescriptions) {
-        this.attributeDescriptions = attributeDescriptions.stream()
-                .map(str -> str.split("=", 2))
-                .filter((list) -> {
-                    if (list.length <= 1) {
-                        if (LOGGER.isDebugEnabled()) {
-                            LOGGER.debug("Filtered out invalid attribute/description: {}", list[0]);
-                        }
-                        return false;
-                    }
-                    return true;
-                })
-                .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
+        this.attributeDescriptions = parseAttributeAndValuePairs(attributeDescriptions);
     }
 
     private List<String> readOnly = ImmutableList.of("checksum",
@@ -519,6 +497,19 @@ public class ConfigurationApplication implements SparkApplication {
         }
 
         return config;
+    }
+
+    private Map<String, String> parseAttributeAndValuePairs(List<String> pairs) {
+        return pairs.stream()
+            .map(str -> str.split("=", 2))
+            .filter((list) -> {
+                if (list.length <= 1) {
+                    LOGGER.debug("Filtered out invalid attribute/value pair: {}", list[0]);
+                    return false;
+                }
+                return true;
+            })
+            .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
     }
 
     public HttpProxyService getHttpProxy() {
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -173,7 +173,16 @@ public class ConfigurationApplication implements SparkApplication {
     public void setAttributeAliases(List<String> attributeAliases) {
         this.attributeAliases = attributeAliases.stream()
                 .map(str -> str.split("=", 2))
-                .collect(Collectors.toMap(list -> list[0].trim(), list -> list.length > 1 ? list[1].trim() : ""));
+                .filter((list) -> {
+                    if (list.length <= 1) {
+                        if (LOGGER.isDebugEnabled()) {
+                            LOGGER.debug("Filtered out invalid attribute/alias: {}", list[0]);
+                        }
+                        return false;
+                    }
+                    return true;
+                })
+                .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
     }
 
     public void setHiddenAttributes(List<String> hiddenAttributes) {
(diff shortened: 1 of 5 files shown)


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -173,6 +180,12 @@ public void setHiddenAttributes(List<String> hiddenAttributes) {
         this.hiddenAttributes = hiddenAttributes;
     }
 
+    public void setAttributeDescriptions(List<String> attributeDescriptions) {
+        this.attributeDescriptions = attributeDescriptions.stream()
+                .map(str -> str.split("="))
+                .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
```
bdeining: What happens if there is no `=` in the attributeDescription?  Might be worth a check to avoid the index out of bounds exception.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -173,6 +180,12 @@ public void setHiddenAttributes(List<String> hiddenAttributes) {
         this.hiddenAttributes = hiddenAttributes;
     }
 
+    public void setAttributeDescriptions(List<String> attributeDescriptions) {
+        this.attributeDescriptions = attributeDescriptions.stream()
+                .map(str -> str.split("="))
```
bdeining: What happens if there is an `=` in the attributeDescription?

catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.config.xml
```diff
@@ -151,6 +151,13 @@
             default="^sortOrder$,^sortField$,^cql$,^polling$,^cached$"
             required="false"/>
 
+        <AD id="attributeDescriptions"
+            name="Attribute Descriptions"
+            description="List of friendly attribute descriptions. Example 'checksum-algorithm=Method for generating a small-sized datum from a block of digital data for the purpose of detecting errors'"
```
bdeining: We may want to emphasize that there must be an `=` sign for the attribute=attribute description

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
albany551: this was as all copied over from the existing attribute aliases code, so I guess I should add a fix there too?

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -165,14 +172,20 @@ public void setResultShow(List<String> resultShow) {
 
     public void setAttributeAliases(List<String> attributeAliases) {
         this.attributeAliases = attributeAliases.stream()
-                .map(str -> str.split("="))
-                .collect(Collectors.toMap(list -> list[0].trim(), list -> list[1].trim()));
+                .map(str -> str.split("=", 2))
+                .collect(Collectors.toMap(list -> list[0].trim(), list -> list.length > 1 ? list[1].trim() : ""));
```
rzwiefel: :question: In the case that someone accidentally puts in an invalid value and it results in the list being only `1` in size do we still want to return an empty string (meaning do we want to silently ignore bad fields)? Or should we put in an explicit check to either filter out invalid values or throw an exception?

Not sure what our "preferred" course of action is (error reporting to admin UI not a strong point as of right now) but I have multiple times seen us prefer to throw an exception (and fail to save) rather than silently succeed and trim data. Interested if any others have thoughts on this.

bdeining: IMO, It may be best to filter out the attribute descriptions that are bad instead of throwing an exception for the whole config.

rzwiefel: So if we go with the option of just validating/filtering out bad data we should probably just add a log message that we filtered out an attribute (Maybe at debug? and probably a good idea to wrap it in a `LOGGER.isDebugEnabled()` since this is inside something being streamed)

bdeining:
manually verified tooltips 🏁 , the styling looks different than the provided screenshots 
screen shot 2017-09-05 at 4 04 31 pm

albany551:
They're the browser/OS native tooltips so will look different across platforms. We could add a tooltip library or custom code going forward if that needs to change.

bdeining:
We can consider that in the future; seems out of scope for this PR

rzwiefel:
Another thing we might consider in the future is adding them to the contextual help we already have implemented (the help that comes up when you click the `?`)

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -165,14 +172,38 @@ public void setResultShow(List<String> resultShow) {
...
+        this.attributeDescriptions = attributeDescriptions.stream()
+                .map(str -> str.split("=", 2))
+                .filter((list) -> {
+                    if (list.length <= 1) {
+                        if (LOGGER.isDebugEnabled()) {
+                            LOGGER.debug("Filtered out invalid attribute/description: {}", list[0]);
+                        }
+                        return false;
+                    }
+                    return true;
```
rzwiefel: ✏️ If you wish, you could refactor this filter out into a predicate since its almost exactly the same for 192-202 and 177-183

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -165,14 +172,38 @@ public void setResultShow(List<String> resultShow) {
 
     public void setAttributeAliases(List<String> attributeAliases) {
         this.attributeAliases = attributeAliases.stream()
-                .map(str -> str.split("="))
+                .map(str -> str.split("=", 2))
```
Variadicism: This would allow "`=`" signs inside the right side of the alias without error. Could that potentially corrupt aliases?

catalog/ui/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
```diff
@@ -78,6 +78,7 @@ define([
                 }).map(function(metacardType){
                     return {
                         label: metacardType.alias || metacardType.id,
+                        description: (properties.attributeDescriptions || {})[metacardType.id],
```
mackncheesiest: So the description tooltip ends up as either the user entered description or the `metacardType.id`?

catalog/ui/catalog-ui-search/src/main/webapp/component/select/select.view.js
```diff
@@ -41,11 +41,21 @@ define([
                 'data-help': this.model.get('help')
             };
         },
+        onRender: function(){
+            if (this.model.get('description')) {
+                this.$el.attr('data-help', this.model.get('description'));
```
mackncheesiest: Or is there just no tooltip if there's no entry in the AttributeDescriptions map?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
mackncheesiest: What do you mean by "corrupt aliases"?

catalog/ui/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
Variadicism: This is used in a `div.title`. Correct me if I'm wrong, but I think `{}[metacardType.id]` will become "`undefined`" in the `div.title`. Is this the desired behavior? Maybe use the following instead?
`description: properties.attributeDescriptions[metacardType.id] || "",`

albany551: `metacardType.id` is the lookup key. If there are no attribute descriptions, or there isn't one for a particular metacard id, this will return `undefined`.

catalog/ui/catalog-ui-search/src/main/webapp/component/select/select.view.js
albany551: Aye. The `if` will be false for an `undefined` value and no `data-help` atribute will be set. This particular line sets the context-sensitive help, incidentally (when you press the ? button). Tooltips are set in this view's template (the HTML `title` attribute).

catalog/ui/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
mackncheesiest: Ah, makes sense.

catalog/ui/catalog-ui-search/src/main/webapp/component/select/select.view.js
```diff
@@ -41,11 +41,21 @@ define([
...
+        },
         serializeData: function(){
             var modelJSON = this.model.toJSON();
             if (modelJSON.label.constructor === Array){
                 modelJSON.label = modelJSON.label.join(' | ');
             }
+            if (modelJSON.description) {
+                // add line breaks to separate the description from the label
+                // within the tooltip
+                modelJSON.description = '\n\n' + modelJSON.description;
```
Variadicism: :question: If something is made to use this description in the future, they will also receive the two extra line breaks. Is this tightly coupled to ensure that this is only used for the tooltip in which this formatting trick is needed?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
albany551: Yep the intention was to just go ahead and allow an equals sign in the alias' value, which shouldn't mess anything up all the way through to the ui.

catalog/ui/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
albany551: This is intentional/desired. As written it handles the case where `properties.attributeDescriptions` is `undefined`, which would throw an error in your snippet. Also handlebars resolves the undefined to an empty string when rendering.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -165,14 +172,38 @@ public void setResultShow(List<String> resultShow) {
...
+                .filter((list) -> {
+                    if (list.length <= 1) {
+                        if (LOGGER.isDebugEnabled()) {
```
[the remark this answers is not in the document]
albany551: I added the check on @rzwiefel's advice - " probably a good idea to wrap it in a LOGGER.isDebugEnabled() since this is inside something being streamed"

catalog/ui/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
Variadicism: Fair point. I should have done the following instead. Anyway, if undefined is rendered as empty string, it should be fine.
`description: (properties.attributeDescriptions || {})[metacardType.id] || "",`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
Variadicism: @mackncheesiest, I wasn't sure how the aliases were used and if they were expected in any later code to not contain an "`=`". It just seemed like something that might be a potential problem, though I was guessing that without being sure about the context. According to the previous comment, it's not a problem.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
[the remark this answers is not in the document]
Variadicism: Oh, I see that comment. @rzwiefel, can you explain your advice?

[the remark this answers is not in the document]
rzwiefel: Since this is something that is being looped over (it is in a stream) it could (theoretically) have a large number of items in it, and in that case we would be logging to debug **N** number of times ... And relatively an if statement is much cheaper than the log statement. So we only actually pay the cost of logging it if the logs are set at that level.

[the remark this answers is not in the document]
Variadicism: Note: as per online discussion, this `LOGGER.isDebugEnabled()` check can be removed.

bdeining:
build now

bdeining:
build now

bdeining:
manually verified again ✅

PR #2255, opened 2017-08-24
DDF-3241 Added thread factory builder utility

mcalcote:
#### What does this PR do?
This adds the ability to add a thread factory to all blueprint executors, capture individual executor thread pools and configure them to use the general utility

#### Who is reviewing it? 
(please choose AT LEAST two reviewers that need to approve the PR before it can get merged)
@clockard @emmberk @rzwiefel @ahoffer @vinamartin @lessarderic

PR #2277, opened 2017-08-31
DDF-3139 Resolves issue where CrlChecker opened a new pool on every construction

coyotesqrl:
#### What does this PR do?
`CrlChecker` now uses a single static pool for its scheduled checks for CRL lists. A better solution would be to convert the checker into a service; however, this should be sufficient for now.

This PR also resolves a small, related, issue with a thread pool that was constructed and never used for the `ReliableResourceDownloader`.

When running this evening, I found that while the issue at hand was resolved, I had three itest failures on my local development box. I do not believe these are related to this change, and in fact might be a misleading result of the development branch upon which I had done my development (I based off of #2255 because of the improved ability to track down the culprit branches/pools). However, it is possible there is a bug introduced that caused these tests to fail.

```
Failed tests: 
  TestSolrCommands.testSolrBackupNumToKeep:102->waitForBackupDirsToBeCreated:158->waitFor:139 Timed out after 10000 seconds waiting for correct number of backups in /projects/combo/ddf/distribution/test/itests/test-itests-ddf/target/exam/c649aee4-9a3b-4fcf-8bc3-c42fb08b3c4f/data/solr/catalog/data
  TestSolrCommands.testSolrBackupCommand:81 
Expected: a string containing "Backup of [catalog] complete"
     but: was ""
  TestSolrCommands.testSolrBackupBadCoreName:90 
Expected: a string containing "Error backing up Solr core: [blah]"
     but: was ""
```

#### Who is reviewing it? 
@rzwiefel @garrettfreibott @kcwire @mcalcote 

@codice/security

@pklinef
@stustison

Full build and itest. Running on an OSX machine with limited resources (an MBP, for example) should be sufficient. The issue was the number of orphaned threads growing so high that the ability of the JVM to get native threads was compromised.

N/A

[DDF-3139](https://codice.atlassian.net/browse/DDF-3139)

N/A

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
+++ b/platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
@@ -50,16 +50,13 @@ public class CrlChecker {
     private static String encryptionPropertiesLocation = new AbsolutePathResolver(
             "etc/ws-security/server/encryption.properties").getPath();
 
-    private AtomicReference<CRL> crlCache = new AtomicReference<>();
+    private static AtomicReference<CRL> crlCache = new AtomicReference<>();
 
-    private CrlRefresh refresh = new CrlRefresh();
+    private static final CrlRefresh REFRESH = new CrlRefresh();
 
-    /**
-     * Constructor method. Reads encryption.properties and sets CRL location
-     */
-    public CrlChecker() {
+    static {
         Executors.newScheduledThreadPool(1)
-                .scheduleWithFixedDelay(refresh, 0, 1, TimeUnit.HOURS);
+                .scheduleWithFixedDelay(REFRESH, 0, 1, TimeUnit.HOURS);
     }
 
     /**
@@ -110,10 +107,10 @@ public class CrlChecker {
      *                 check certificate revocation.
      */
     public void setCrlLocation(String location) {
-        refresh.setCrlLocation(location);
+        REFRESH.setCrlLocation(location);
     }
 
-    URL urlFromPath(String location) {
+    static URL urlFromPath(String location) {
         try {
             return new URL(location);
         } catch (MalformedURLException e) {
@@ -127,7 +124,7 @@ public class CrlChecker {
      * @param location location of properties file
      * @return Properties from the file
      */
-    Properties loadProperties(String location) {
+    static Properties loadProperties(String location) {
         return PropertiesLoader.loadProperties(location);
     }
 
@@ -135,7 +132,7 @@ public class CrlChecker {
      * Runnable to refresh the CRL from the URL when either the nextUpdate time has elapsed
      * or it has rolled over to the next day.
      */
-    private class CrlRefresh implements Runnable {
+    private static class CrlRefresh implements Runnable {
         private Lock lock = new ReentrantLock();
 
         private Calendar start = Calendar.getInstance();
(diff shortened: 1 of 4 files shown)


coyotesqrl:
build now

rzwiefel:
I would definitely say merge now since it's a developer blocker, but we should probably do a test of the CRL updating post-merge.

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -50,16 +50,13 @@
     private static String encryptionPropertiesLocation = new AbsolutePathResolver(
             "etc/ws-security/server/encryption.properties").getPath();
 
-    private AtomicReference<CRL> crlCache = new AtomicReference<>();
+    private static AtomicReference<CRL> crlCache = new AtomicReference<>();
```
brendan-hofmann: ✏️ Nit, looks like this could be final as well? At least I don't see it being reassigned anywhere else.

coyotesqrl: That's a good point. It could have been final before when it was a member field as well. Missed it. But I think I'm going to let this slide for now so we don't have to run CI again, with the thinking that this should be re-written as a service.

lessarderic: ❓ Could this be `final`?

brendan-hofmann: I'm ok with that reasoning. 👍

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -50,16 +50,13 @@
...
 
-    private CrlRefresh refresh = new CrlRefresh();
+    private static final CrlRefresh REFRESH = new CrlRefresh();
 
-    /**
-     * Constructor method. Reads encryption.properties and sets CRL location
-     */
-    public CrlChecker() {
+    static {
```
lessarderic: ❓ Should we move this in the `CrlRefresh` constructor instead of having a `static` block here?

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -110,10 +107,10 @@ private boolean passesCrl(X509Certificate[] certs) {
      *                 check certificate revocation.
      */
     public void setCrlLocation(String location) {
-        refresh.setCrlLocation(location);
+        REFRESH.setCrlLocation(location);
     }
 
-    URL urlFromPath(String location) {
+    static URL urlFromPath(String location) {
```
lessarderic: ❓ Move inside `CrlRefresh` instead of making `static`?

platform/security/handler/security-handler-pki/src/main/java/org/codice/ddf/security/handler/pki/CrlChecker.java
```diff
@@ -127,15 +124,15 @@ URL urlFromPath(String location) {
      * @param location location of properties file
      * @return Properties from the file
      */
-    Properties loadProperties(String location) {
+    static Properties loadProperties(String location) {
```
lessarderic: ❓ Same question.
(I noticed that those are used by the unit tests, but I don't think it's a good enough reason to expose them like we do, especially since they are so short. And depending on the implementation of the class under test to write the test feels like a smell).

lessarderic:
Oh well, looks like I missed the deadline :-(

PR #2280, opened 2017-09-01
DDF-3262 Increase Solr string indexing size and exclude attributes

spearskw:
#### What does this PR do?
Allow Solr tokenized txt field to index more data than txt string field.
Support excluding attributes from Solr response.

@jlcsmith
@pklinef

Configure Tika and PDF transformers to a higher character limit (e.g. 1000000). Ingest large text files and verify that all text up to the character limit is indexed but only the first 32k of text is returned. Verify that metadata and lux fields are not returned.

[DDF-3262](https://codice.atlassian.net/browse/DDF-3262)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
@@ -175,6 +175,9 @@ public class DynamicSchemaResolver {
         anyTextFieldsCache.addAll(basicTextAttributes);
         fieldsCache.add(Validation.VALIDATION_ERRORS + SchemaFields.TEXT_SUFFIX);
         fieldsCache.add(Validation.VALIDATION_WARNINGS + SchemaFields.TEXT_SUFFIX);
+
+        fieldsCache.add(SchemaFields.METACARD_TYPE_FIELD_NAME);
+        fieldsCache.add(SchemaFields.METACARD_TYPE_OBJECT_FIELD_NAME);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


spearskw:
build now

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
```diff
@@ -336,6 +347,35 @@ public void addFields(Metacard metacard, SolrInputDocument solrInputDocument)
         solrInputDocument.addField(SchemaFields.METACARD_TYPE_OBJECT_FIELD_NAME, metacardTypeBytes);
     }
 
+    public static String truncateAsUTF8(String value, int maximumBytes) {
```
rzwiefel: ❓ @spearskw Can you go into a little more detail on what this method is doing and what it means? Some comments in the code might not hurt. 

I presume we're doing manual byte counting because the actual size of bytes per character can vary in UTF-8 Encoding,  Just not sure on the specifics of reasons why the `value.length()` doesn't work and such.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -377,9 +382,57 @@ protected SolrQuery postAdapt(QueryRequest request, SolrFilterDelegate filterDel
...
 
+        filterAttributes(request, query);
+
         return query;
     }
 
+    private void filterAttributes(QueryRequest request, SolrQuery query) {
+        if (!request.hasProperties() || !request.containsPropertyName(EXCLUDE_ATTRIBUTES)
+                || !(request.getPropertyValue(EXCLUDE_ATTRIBUTES) instanceof Set) ||
+                "true".equals(System.getProperty("solr.client.filterAttributes.disable"))) {
```
rzwiefel: ✏️ This condition (the system property check) seems a little awkward here with the property existence check and instanceof check. Maybe we could pull it out into its' own if statement above this one, that way we leave the logic relating to the exclude attributes grouped together and the system property check on its own.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
spearskw: I'll be honest, I don't know what it's doing. @pklinef ?

pklinef:
Replaced by https://github.com/codice/ddf/pull/2355

PR #2281, opened 2017-09-01
DDF-3261 Updating pipeline to support multi-branch jobs

oconnormi:
#### What does this PR do?
Updates Jenkinsfile to support using multi-branch pipeline in jenkins.
Various steps were updated to detect when a pull request is being built vs a regular branch
Several settings were tweaked to bring build closer to the legacy jobs in configuration.

  * Updated maven memory options
  * Added property to override random device on linux
  * Made deploy step conditional
  * Made coverity step conditional on master branch
  * Made Incremental vs Full build conditional
  * Set git longpath support on windows
  * Added stage to run pomfix before actual build steps

#### Who is reviewing it? 
@mcalcote @LinkMJB @clockard 

@codice/build 

@lessarderic 
@shaundmorris 

[DDF-3261](https://codice.atlassian.net/browse/DDF-3261)
PR Build
image

Branch Build
image

Tasks:
- [x] Add master branch requirement to coverity step
- [x] Determine whether the cron trigger is necessary with multi-branch pipelines
- [x] Add quick build as first step of incremental stage
- [x] Add pom-fix as first stage. (Stage Name?)
- [x] Reformat indentations
- [x] Share common maven options

#### Review Comment Legend:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,6 +38,7 @@ pipeline {
                                 }
                                 timeout(time: 3, unit: 'HOURS') {
                                     withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
+                                        sh 'mvn install -pl !$DOCS -DskipStatic=true -DskipTests=true -T 1C'
                                         sh 'mvn clean install -B -T 1C -pl !$ITESTS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
                                         sh 'mvn install -B -Dmaven.test.redirectTestOutputToFile=true -pl $ITESTS -nsu'
                                     }
@@ -52,6 +53,7 @@ pipeline {
                                 }
                                 timeout(time: 3, unit: 'HOURS') {
                                     withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS}', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
+                                        bat 'mvn install -pl !%DOCS% -DskipStatic=true -DskipTests=true -T 1C'
                                         bat 'mvn clean install -B -T 1C -pl !%ITESTS% -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/%CHANGE_TARGET%'
                                         bat 'mvn install -B -Dmaven.test.redirectTestOutputToFile=true -pl %ITESTS% -nsu'
                                     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,79 +20,79 @@ pipeline {
         }
         // The incremental build will be triggered only for PRs. It will build the differences between the PR and the target branch
         stage('Incremental Build') {
-          when {
-            allOf {
-              expression { env.CHANGE_ID != null }
-              expression { env.CHANGE_TARGET != null }
+            when {
+                allOf {
+                    expression { env.CHANGE_ID != null }
+                    expression { env.CHANGE_TARGET != null }
+                }
             }
-          }
-          // TODO DDF-2971 refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
-          steps {
-            parallel(
-              linux: {
-                node('linux-large') {
-                  retry(3) {
-                    checkout scm
-                  }
-                  timeout(time: 3, unit: 'HOURS') {
-                    withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -Djava.security.egd=file:/dev/./urandom', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
-                      sh 'mvn clean install -B -T 1C -pl !$ITESTS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
-                      sh 'mvn install -B -Dmaven.test.redirectTestOutputToFile=true -pl $ITESTS -nsu'
-                    }
-                  }
-                }
-              },
-              windows: {
-                node('proxmox-windows') {
-                  bat 'git config --system core.longpaths true'
-                  retry(3) {
-                    checkout scm
-                  }
-                  timeout(time: 3, unit: 'HOURS') {
-                    withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
-                      bat 'mvn clean install -B -T 1C -pl !%ITESTS% -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/%CHANGE_TARGET%'
-                      bat 'mvn install -B -Dmaven.test.redirectTestOutputToFile=true -pl %ITESTS% -nsu'
-                    }
-                  }
-                }
-              }
-            )
-          }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -69,30 +112,37 @@ pipeline {
...
                                 }
                             }
                         },
+                        // Coverity will be skipped on all PR builds
                         coverity: {
                             node('linux-medium') {
-                                retry(3) {
+                              script {
+                                if (env.CHANGE_ID != null) {
+                                  echo "Skippng Coverity on PR build"
```
clockard: anyway to 'return' here so we don't need to nest the rest of this in an else statement?

Jenkinsfile
```diff
@@ -18,7 +18,50 @@ pipeline {
...
+          }
+          // TODO DDF-2971 refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
+          steps {
+            parallel(
+              linux: {
+                node('linux-large') {
+                  retry(3) {
+                    checkout scm
+                  }
+                  timeout(time: 3, unit: 'HOURS') {
```
shaundmorris: It would probably be a build smell if we ever started exceeding 3 hours for this kind of build :-)

Jenkinsfile
```diff
@@ -18,7 +18,50 @@ pipeline {
...
+                    withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -Djava.security.egd=file:/dev/./urandom', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
+                      sh 'mvn clean install -B -T 1C -pl !$ITESTS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
+                      sh 'mvn install -B -Dmaven.test.redirectTestOutputToFile=true -pl $ITESTS -nsu'
+                    }
+                  }
+                }
+              },
+              windows: {
```
shaundmorris: Nice to see windows on the PR builds

Jenkinsfile
shaundmorris: A green checkmark still shows up on the PR builds, that might not be a problem, but may confuse people as if the PR had no major coverity issues.

Jenkinsfile
LinkMJB: @oconnormi IIRC, @lessarderic identified that a quick-build is required prior to performing the GIB enabled build otherwise it fails. I would do a quick compare against https://jenkins.phx.connexta.com/job/DDF-Linux-PR/configure

Jenkinsfile
```diff
@@ -18,7 +18,50 @@ pipeline {
...
+                node('proxmox-windows') {
+                  bat 'git config --system core.longpaths true'
+                  retry(3) {
+                    checkout scm
+                  }
+                  timeout(time: 3, unit: 'HOURS') {
+                    withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
+                      bat 'mvn clean install -B -T 1C -pl !%ITESTS% -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/%CHANGE_TARGET%'
```
LinkMJB: @oconnormi IIRC, @lessarderic identified that a quick-build is required prior to performing the GIB enabled build otherwise it fails. I would do a quick compare against https://jenkins.phx.connexta.com/job/DDF-Linux-PR/configure

Jenkinsfile
LinkMJB: @shaundmorris we get pretty close sometimes: https://jenkins.codice.org/job/DDF-Linux-PR/1914/

Jenkinsfile
oconnormi: @LinkMJB any chance we know how it fails? Is it something with snapshot dependency resolution by any chance?

Jenkinsfile
oconnormi: @LinkMJB see above question 😄

Jenkinsfile
oconnormi: Yeah, I think theres not much we can do currently to avoid a green check mark. Though I'm curious if we will be able to have conditional parallel steps when the declarative pipelines fully support parallel stages (Should  be in the next release of the pipeline-model-definition)

Jenkinsfile
lessarderic: We need to run a quick build to make sure we're not pulling down stale dependencies from Nexus when doing the incremental build, or pull down dependencies while they are being uploaded to Nexus from the master build pipeline.

We also used to run the pomfix tool prior to anything else to fail fast if the maven dependencies were not in-sync with the feature files.

oconnormi: Alright, I'll add a quickbuild step prior to the incremental build. For the pomfix part i'm thinking of actually having that be an initial stage before any other build is run. this would help visualize it as a discrete gate in the process that could fail really early. Thoughts? 

Related I've actually also thought about doing the same thing with checkstyle, but that is probably out of scope for this pr

Jenkinsfile
```diff
@@ -69,30 +112,37 @@ pipeline {
...
                             node('linux-medium') {
-                                retry(3) {
+                              script {
+                                if (env.CHANGE_ID != null) {
+                                  echo "Skippng Coverity on PR build"
+                                } else {
+                                  retry(3) {
                                     checkout scm
+                                  }
+                                  withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '-Djava.security.egd=file:/dev/./urandom') {
```
lessarderic: ✏️  These files are deeply indented and fairly difficult to read. We may want to make sure our indentations are consistent throughout. That might help a little.

Jenkinsfile
```diff
@@ -116,10 +166,21 @@ pipeline {
                 )
             }
         }
+        /*
+          Deploy stage will only be executed for deployable branches. These include master and any patch branch matching M.m.x format (i.e. 2.10.x, 2.9.x, etc...).
+          It will also only deploy in the presence of an environment variable JENKINS_ENV = 'prod'. This can be passed in globally from the jenkins master node settings.
```
lessarderic: ❓ Curious to know, when do we have a deployable branch that matches the pattern but shouldn't be deployed? Couldn't we always use the `JENKINS_ENV` environment variable and not use the pattern match?

oconnormi: We could, only the plan was to set that env variable globally per jenkins master instance. This would allow us to test pipelines while auto-skipping the deploy phase when it is run on the test jenkins environment, vs the prod jenkins environment. 

As for the deployable branches. The idea here is that we want to avoid building and deploying on branches unless they are going to have unique versions in their poms (which the 2.10.x, 2.9.x, etc...) branches do. All work is currently done using forks, however that doesn't necessarily stop anyone with commit privileges from pushing an arbitrary branch to this repo. Lets say for example sake that it was branched from master (currently at 2.11.0-SNAPSHOT) and the version number wasn't changed, without this check jenkins would automatically discover, build, and deploy it if a jenkinsfile was present. This check allows for jenkins to discover and build it, but to skip the deploy step. 

The benefit of doing it this way is that, 
* local branches with potentially conflicting versions don't cause issues
* patch fix branches can be built, verified, and deployed based on scm changes
* all deploy steps can be disabled on a per jenkins master basis

Jenkinsfile
shaundmorris: @oconnormi, @LinkMJB  do we think this timeout is appropriate or should we adjust or try to improve it on a later?

Jenkinsfile
```diff
@@ -18,7 +18,50 @@ pipeline {
...
+                }
+              },
+              windows: {
+                node('proxmox-windows') {
+                  bat 'git config --system core.longpaths true'
+                  retry(3) {
+                    checkout scm
+                  }
+                  timeout(time: 3, unit: 'HOURS') {
+                    withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
```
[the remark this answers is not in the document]
oconnormi: I could probably just but them in variable for now. I think eventually we will want to encapsulate some of our build process in some shared libraries

Jenkinsfile
oconnormi: I probably need to run them through a formatter

oconnormi: should be fixed now

Jenkinsfile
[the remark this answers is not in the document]
oconnormi: @lessarderic shared them using env vars

Jenkinsfile
oconnormi: Yeah i think we should leave it this way since we get close sometimes. Hopefully things will improve in the future

Jenkinsfile
oconnormi: Quickbuild added as first step for both windows and linux incremental builds

Jenkinsfile
oconnormi: fixed

Jenkinsfile
```diff
@@ -10,90 +10,105 @@ pipeline {
     }
     environment {
         DOCS = 'distribution/docs'
-        ITESTS= 'distribution/test/itests/test-itests-ddf'
+        ITESTS = 'distribution/test/itests/test-itests-ddf'
+        POMFIX = 'libs/libs-pomfix,libs/libs-pomfix-run'
+        LARGE_MVN_OPTS = '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC '
```
shaundmorris: ✏️  might be worth a comment as to why we have some of these values for people who may contribute to the file later.

PR #2278, opened 2017-08-31
DDF-3253 Updated CswRecordMapperFilterVisitor to use a configurable map

AzGoalie:
#### What does this PR do?
Exposes and allows `CswRecordMapperFilterVisitor` to use a configurable map.
#### Who is reviewing it? 
@emanns95 @ahoffer 
#### Select relevant component teams: 
https://github.com/orgs/codice/teams/ogc
https://github.com/orgs/codice/teams/core-apis
#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)
@coyotesqrl
@bdeining 
#### What are the relevant tickets?
[DDF-3253](https://codice.atlassian.net/browse/DDF-3253)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests
#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2289, opened 2017-09-05
DDF-3259 Moved SortBy normalization logic after the query filter transforms

mojogitoverhere:
#### What does this PR do?
Moves the SortBy normalization logic after the QueryFilterTransformers have run. This makes it possible to update invalid SortBy parameters in the QueryFilterTransformer.

#### Who is reviewing it? 
@adimka @brianfelix @mackncheesiest @brendan-hofmann 

@bdeining
@millerw8

Full build with unit and itests.

Previously, the SortBy parameters were normalized before the QueryRequest object was created and passed to the QueryFilterTransformers. If a SortBy parameter was an invalid CSW attribute, then it was stripped out completely. This change allows the QueryFilterTransformers to update the SortBy before it is normalized like usual.

[DDF-3259](https://codice.atlassian.net/browse/DDF-3259)
#### Checklist:
- [ ] Documentation Updated
- [* ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
@@ -39,8 +39,8 @@ import org.apache.commons.io.IOUtils;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.CswConstants;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.CswException;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.PropertyIsFuzzyFunction;
-import org.codice.ddf.spatial.ogc.csw.catalog.common.converter.DefaultCswRecordMap;
 import org.codice.ddf.spatial.ogc.csw.catalog.endpoint.mappings.SourceIdFilterVisitor;
+import org.codice.ddf.spatial.ogc.csw.catalog.endpoint.transformer.CswRecordMap;
 import org.geotools.feature.NameImpl;
 import org.geotools.filter.AttributeExpressionImpl;
 import org.geotools.filter.FilterFactoryImpl;
@@ -95,13 +95,17 @@ public class CswQueryFactory {
 
     private final FilterAdapter adapter;
 
+    private final CswRecordMap cswRecordMap;
+
     private Map<String, Set<String>> schemaToTagsMapping = new HashMap<>();
 
     private AttributeRegistry attributeRegistry;
 
     private QueryFilterTransformerProvider queryFilterTransformerProvider;
 
-    public CswQueryFactory(FilterBuilder filterBuilder, FilterAdapter adapter) {
+    public CswQueryFactory(CswRecordMap cswRecordMap, FilterBuilder filterBuilder,
+            FilterAdapter adapter) {
+        this.cswRecordMap = cswRecordMap;
         this.builder = filterBuilder;
         this.adapter = adapter;
     }
@@ -298,25 +302,22 @@ public class CswQueryFactory {
             return null;
         }
 
-        String propertyName = cswSortBy.getPropertyName().getPropertyName();
+        String propertyName = cswSortBy.getPropertyName()
+                .getPropertyName();
         if (propertyName == null) {
             LOGGER.debug("Property in SortBy Field is null");
             return null;
         }
 
-        NamespaceSupport namespaceContext = cswSortBy.getPropertyName().getNamespaceContext();
+        NamespaceSupport namespaceContext = cswSortBy.getPropertyName()
+                .getNamespaceContext();
         if (!attributeRegistry.lookup(propertyName)
-                .isPresent() && !DefaultCswRecordMap.hasDefaultMetacardFieldForPrefixedString(
-                propertyName,
-                namespaceContext)) {
-            LOGGER.debug("Property {} is not a valid SortBy Field",
-                    propertyName);
+                .isPresent() && !cswRecordMap.hasProperty(propertyName, namespaceContext)) {
+            LOGGER.debug("Property {} is not a valid SortBy Field", propertyName);
             return null;
         }
 
-        String name =
-                DefaultCswRecordMap.getDefaultMetacardFieldForPrefixedString(propertyName,
-                        namespaceContext);
+        String name = cswRecordMap.getProperty(propertyName, namespaceContext);
 
         PropertyName propName = new AttributeExpressionImpl(new NameImpl(name));
         return new SortByImpl(propName, cswSortBy.getSortOrder());
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -232,6 +233,90 @@ public class CswQueryFactory {
         return filter;
     }
 
+    private QueryRequest normalizeSort(QueryRequest request) {
+        if (request == null) {
+            return null;
+        }
+
+        Query query = request.getQuery();
+        if (query == null) {
+            return request;
+        }
+
+        List<SortBy> sortBys = new ArrayList<>();
+
+        // Primary sort parameter
+        SortBy sortBy = query.getSortBy();
+        SortBy newSortBy = normalizeSortBy(sortBy);
+        if (newSortBy != null) {
+            sortBys.add(newSortBy);
+        }
+
+        // Additional sort parameters
+        Serializable extraSortBys = request.getPropertyValue(EXT_SORT_BY);
+        if (extraSortBys != null && extraSortBys instanceof SortBy[]) {
+            List<SortBy> extraSortBysList = Arrays.asList((SortBy[]) extraSortBys);
+            extraSortBysList.stream()
+                    .map(this::normalizeSortBy)
+                    .filter(Objects::nonNull)
+                    .forEach(sortBys::add);
+        }
+
+        Map<String, Serializable> newProperties = request.getProperties();
+        SortBy normalizedSortBy = null;
+        if (sortBys.size() > 0) {
+            normalizedSortBy = sortBys.get(0);
+        }
+        if (sortBys.size() > 1) {
+            SortBy[] extraSortBysArray = sortBys.subList(1, sortBys.size())
+                    .toArray(new SortBy[0]);
+            newProperties.put(EXT_SORT_BY, extraSortBysArray);
+        } else {
+            newProperties.remove(EXT_SORT_BY);
+        }
+
+        Query newQuery = new QueryImpl(query,
+                query.getStartIndex(),
+                query.getPageSize(),
+                normalizedSortBy,
+                query.requestsTotalResultsCount(),
+                query.getTimeoutMillis());
+
+        return new QueryRequestImpl(newQuery,
+                request.isEnterprise(),
+                request.getSourceIds(),
+                newProperties);
+    }
+
+    private SortBy normalizeSortBy(SortBy cswSortBy) {
+        if (cswSortBy == null) {
+            return null;
+        }
+
+        if (cswSortBy.getPropertyName() != null
+                && !attributeRegistry.lookup(cswSortBy.getPropertyName()
+                .getPropertyName())
+                .isPresent() && !DefaultCswRecordMap.hasDefaultMetacardFieldForPrefixedString(
+                cswSortBy.getPropertyName()
+                        .getPropertyName(),
+                cswSortBy.getPropertyName()
+                        .getNamespaceContext())) {
+            LOGGER.debug("Property {} is not a valid SortBy Field",
+                    cswSortBy.getPropertyName()
+                            .getPropertyName());
+            return null;
+        }
+
+        String name =
+                DefaultCswRecordMap.getDefaultMetacardFieldForPrefixedString(cswSortBy.getPropertyName()
+                                .getPropertyName(),
+                        cswSortBy.getPropertyName()
+                                .getNamespaceContext());
+
+        PropertyName propName = new AttributeExpressionImpl(new NameImpl(name));
+        return new SortByImpl(propName, cswSortBy.getSortOrder());
+    }
+
     private SortBy[] buildSort(SortByType sort) throws CswException {
         if (sort == null || sort.getSortProperty() == null) {
             return null;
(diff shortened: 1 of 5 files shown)


catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -1034,9 +1040,9 @@ private DistanceBufferType createDistanceBufferType() {
 
         QueryImpl frameworkQuery = (QueryImpl) queryFactory.getQuery(grr)
                 .getQuery();
-        assertThat(frameworkQuery.getFilter(), instanceOf(clz));
+        assertThat(((QueryImpl) frameworkQuery.getFilter()).getFilter(), instanceOf(clz));
```
Variadicism: [pencil] This whole QueryImpl.getFilter() call with case is used again two lines later. Maybe it would be better off as a local variable.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -1150,16 +1156,17 @@ private Filter generateTemporalFilter(JAXBElement<BinaryComparisonOpType> tempor
         QueryImpl frameworkQuery = (QueryImpl) queryFactory.getQuery(grr)
                 .getQuery();
         N temporal = null;
+        Filter queryFilter = ((QueryImpl) frameworkQuery.getFilter()).getFilter();
         if (classes.length > 1) {
-            assertThat(frameworkQuery.getFilter(), instanceOf(Or.class));
+            assertThat(queryFilter, instanceOf(Or.class));
             int i = 0;
-            for (Filter filter : ((Or) frameworkQuery.getFilter()).getChildren()) {
+            for (Filter filter : ((Or) queryFilter).getChildren()) {
```
Variadicism: [pencil] This would be better off as a classic indexed for loop instead of using an outer index variable and a for-each loop.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -1150,16 +1156,17 @@ private Filter generateTemporalFilter(JAXBElement<BinaryComparisonOpType> tempor
...
                 assertThat(filter, instanceOf(classes[i++]));
```
Variadicism: There should be an assertion that the length of `classes` is equal to the number of `queryFilter`'s children.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -1150,16 +1156,17 @@ private Filter generateTemporalFilter(JAXBElement<BinaryComparisonOpType> tempor
...
                 temporal = (N) filter;
             }
         } else {
-            assertThat(frameworkQuery.getFilter(), instanceOf(classes[0]));
-            temporal = (N) frameworkQuery.getFilter();
+            assertThat(queryFilter, instanceOf(classes[0]));
```
Variadicism: There should be an assertion that the length of `classes` is not empty.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -232,6 +233,90 @@ private Filter transformCustomFunctionToFilter(Filter filter) {
...
+        Query query = request.getQuery();
+        if (query == null) {
+            return request;
+        }
+
+        List<SortBy> sortBys = new ArrayList<>();
+
+        // Primary sort parameter
+        SortBy sortBy = query.getSortBy();
+        SortBy newSortBy = normalizeSortBy(sortBy);
```
Variadicism: [pencil] `newSortBy` doesn't seem like a useful name for the way in which you're using it. Maybe `normSortBy`? Maybe `sortBy` and inline the current `sortBy` variable, i.e. the following?
```
SortBy sortBy = normalizeSortBy(query.getSortBy());
if (sortBy != null) {
    sortBys.add(newSortBy);
}
```

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -232,6 +233,90 @@ private Filter transformCustomFunctionToFilter(Filter filter) {
...
+        if (newSortBy != null) {
+            sortBys.add(newSortBy);
+        }
+
+        // Additional sort parameters
+        Serializable extraSortBys = request.getPropertyValue(EXT_SORT_BY);
+        if (extraSortBys != null && extraSortBys instanceof SortBy[]) {
```
Variadicism: Maybe it would it be better to make this instance check an assertion. Otherwise, if this property is ever changed to something besides a `SortBy[]`, it will _silently_ be unable to read the extra sort parameters.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -232,6 +233,90 @@ private Filter transformCustomFunctionToFilter(Filter filter) {
...
+                    .forEach(sortBys::add);
+        }
+
+        Map<String, Serializable> newProperties = request.getProperties();
+        SortBy normalizedSortBy = null;
+        if (sortBys.size() > 0) {
+            normalizedSortBy = sortBys.get(0);
+        }
+        if (sortBys.size() > 1) {
+            SortBy[] extraSortBysArray = sortBys.subList(1, sortBys.size())
```
Variadicism: [pencil] If you accept my proposal on line 257 to make the `SortBy[]` instance check an assertion, then you can cast `extraSortBys` to a `SortBy[]`, then you can replace this tricky `sortBys.subList(...).toArray(...)` with `extraSortBys`.
@ line 256:
```
Serializable extraSortBysRaw = request.getPropertyValue(EXT_SORT_BY);
assert(extraSortBysRaw, instanceOf(SortBy[].class));
SortBy[] extraSortBys = (SortBy[]) extraSortBysRaw;
if (extraSortBys != null) {...
```
@ line 270:
```
if (sortBys.size() > 1) {
    newProperties.put(EXT_SORT_BY, extraSortBys);
}
```

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
bdeining: I believe `assert` shouldn't be used in production.  `null` is not an instance of `SortBy`, so I think you could remove the null check here.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -232,6 +233,90 @@ private Filter transformCustomFunctionToFilter(Filter filter) {
...
+                    .toArray(new SortBy[0]);
+            newProperties.put(EXT_SORT_BY, extraSortBysArray);
```
bdeining: what happens if `newProperties` is null at this point?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -232,6 +233,90 @@ private Filter transformCustomFunctionToFilter(Filter filter) {
...
+                request.getSourceIds(),
+                newProperties);
+    }
+
+    private SortBy normalizeSortBy(SortBy cswSortBy) {
+        if (cswSortBy == null) {
+            return null;
+        }
+
+        if (cswSortBy.getPropertyName() != null
```
bdeining: There seem to be some potential NPEs in this conditional.  In particular : `cswSortBy.getPropertyName().getPropertyName()`

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
```diff
@@ -232,6 +233,90 @@ private Filter transformCustomFunctionToFilter(Filter filter) {
...
+                && !attributeRegistry.lookup(cswSortBy.getPropertyName()
+                .getPropertyName())
+                .isPresent() && !DefaultCswRecordMap.hasDefaultMetacardFieldForPrefixedString(
```
bdeining: This PR makes some changes to the way the DefaultCswRecordMap is used.  There may be some conflicts here https://github.com/codice/ddf/pull/2278

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
mojogitoverhere: I like the inline suggestion

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
mojogitoverhere: This logic isn't just used to cast `extraSortBys` to an array though. It's used to strip off the leading element.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
mojogitoverhere: The original logic only accounted for a null `cswSortBy.getPropertyName()`, but I added the check just in case.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
Variadicism: Fair enough about the assertion. Still, shouldn't there be some sort of error if `extraSortBys` is not `null` and not a `SortBy[]`?

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
Variadicism: I forgot that the primary `sortBy` might not be added, so that sublist is not necessarily equivalent to `extraSortBys`. My bad.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
mojogitoverhere: If `newProperties` is null then `sortBys.size()` will never be greater than 1 so this `put()` will never happen. The `remove()` 2 lines down though will blow up. I refactored the logic to account for a null `newProperties`.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
mojogitoverhere: I think we tend to try to do the next best thing if any errors occur. In this case, the extra sort bys are silently ignored. I'll add a debug log message though for some kind of record.

bdeining: that sounds reasonable to me

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
mojogitoverhere: I see where you are coming from but I'd argue your suggestion isn't necessary because if the length of classes is not equal to the number of `queryFilter`s then the test will fail anyways with some index out of range error.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
mojogitoverhere: See my reply to your comment on line 1157

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
millerw8: It feels like tit shouldn't be hardcoded to use the defaultCswRecordMap at this level.  But instead, be able to substitute it for another map.  It appears the other PR @bdeining mentioned does that for the FilterVisitor and we would want to do something similar here.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactory.java
Variadicism: If it's at least logged, that should be fine.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
Variadicism: There will be no `IndexOutOfBoundsException` if there are more classes than `queryFilter` children. Besides that, an assertion failing is more informative and compliant with testing infrastructure than an `Exception` or `Error` thrown during a test. I'll make it a :pencil2:, though.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
Variadicism: [] -> :pencil2:

mojogitoverhere:
build now

mojogitoverhere:
build now

bdeining:
build now

PR #2290, opened 2017-09-06
DDF-3162 Updated Latest/Search command with ResultIterable

rzwiefel:
#### What does this PR do?
updates latest & search command to use resultIterable so it is not capped at 1000 results
#### Who is reviewing it? 
anyone

@clockard
@coyotesqrl

running commands

[](https://codice.atlassian.net/browse/DDF-3162)

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/LatestCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/LatestCommand.java
@@ -49,7 +49,7 @@ public class LatestCommand extends CatalogCommands {
 
     private static final String NUMBER = "#";
 
-    @Argument(name = "NUMBER_OF_ITEMS", description = "Number of maximum records to display. 0 returns 1000 records.", index = 0, multiValued = false, required = false)
+    @Argument(name = "NUMBER_OF_ITEMS", description = "Maximum number of records to display. 0 returns 1000 records.", index = 0, multiValued = false, required = false)
     int numberOfItems = DEFAULT_NUMBER_OF_ITEMS;
 
     @Override
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/LatestCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/LatestCommand.java
@@ -14,6 +14,7 @@
 package org.codice.ddf.commands.catalog;
 
 import static org.apache.commons.lang.StringUtils.isNotBlank;
+import static ddf.catalog.util.impl.ResultIterable.resultIterable;
 
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/LatestCommand.java
```diff
@@ -48,7 +49,7 @@
 
     private static final String NUMBER = "#";
 
-    @Argument(name = "NUMBER_OF_ITEMS", description = "Number of maximum records to display. 0 returns all records.", index = 0, multiValued = false, required = false)
+    @Argument(name = "NUMBER_OF_ITEMS", description = "Number of maximum records to display. 0 returns 1000 records.", index = 0, multiValued = false, required = false)
```
brendan-hofmann: Why not just default to 1000 and eliminate the magical 0 logic? It would let you get rid of the ternary below as well. Same goes for both commands. If it's for the sake of backwards-compatibility, the behavior is already changing, right?

rzwiefel:
@brendan-hofmann  I already completely refactored one of the other commands... but scrapped it until after release since we should try and keep our changes low impact as the release looms near. We can add that refactor to the list 📓

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/LatestCommand.java
rzwiefel: Well... I think @kcwire has mentioned commands are technically a public api.. so maybe we should do a best effort backwards compatibility and treat 0 as int.max_value? Honestly i'm fine either way, because right now they are limited to 1000 results without this fix anyways. Do you have a preference @brendan-hofmann ?

jrnorth: This should say "Maximum number of records to display".

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/SearchCommand.java
```diff
@@ -77,28 +81,23 @@ private Object executeSearchStore(Filter filter) throws Exception {
...
         if (numberOfItems > -1) {
             query.setPageSize(numberOfItems);
         }
 
         long start = System.currentTimeMillis();
 
+        List<Result> results = Lists.newArrayList(resultIterable(catalogProvider::query,
+                new QueryRequestImpl(query),
+                numberOfItems > 0 ? numberOfItems : 1000));
         SourceResponse response = catalogProvider.query(new QueryRequestImpl(query));
```
jrnorth: Can this be deleted now?

peterhuffer:
Hero'd. Received all results from the `latest` command, but `search` was still returning only 1000. Tried `search * 2000` with roughly ~1500 documents, but only got 1000.

peterhuffer:
Successfully hero'd. Uploaded 1074 Metacards and verified both `search * 1500` and `latest 1500` returned all 1074 Metacards. Also verified if I specified 500 it only returned 500.

rzwiefel:
build now

rzwiefel:
build now

PR #2294, opened 2017-09-07
DDF-3272 fixed dump command by adding pagination

rzwiefel:
#### What does this PR do?
Adds pagination to dump command queries
#### Who is reviewing it? 
anyone

@clockard
@coyotesqrl

Running dump command with  > 1000 results
https://codice.atlassian.net/browse/DDF-3272

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/DumpCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/DumpCommand.java
@@ -30,6 +30,7 @@ import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
@@ -56,9 +57,11 @@ import ddf.catalog.data.impl.MetacardImpl;
 import ddf.catalog.operation.SourceResponse;
 import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
+import ddf.catalog.operation.impl.SourceResponseImpl;
 import ddf.catalog.transform.CatalogTransformerException;
 import ddf.catalog.transform.MetacardTransformer;
 import ddf.catalog.transform.QueryResponseTransformer;
+import ddf.catalog.util.impl.ResultIterable;
 import ddf.security.common.audit.SecurityLogger;
 
 //TODO DDF-3116 The catalog:dump shuts down DDF when it fails to transform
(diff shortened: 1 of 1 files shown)


catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/DumpCommand.java
```diff
@@ -110,7 +113,8 @@
             "Multithreaded"}, multiValued = false, description =
             "Number of threads to use when dumping. Setting "
                     + "this value too high for your system can cause performance degradation.")
-    int multithreaded = 20;
+    int multithreaded = Runtime.getRuntime()
+            .availableProcessors();
```
rzwiefel: @pklinef (@coyotesqrl) I ran some tests with (with the paging updates) to see the effect that multithreading/batch size has on the dump command. At a minimum I'd say we should adjust batch size to 512. The thread count i'm not so sure about. 

I ingested 38,809 records and dumped them using variations of multithreading and batch size.

------
[Chart Data Here](https://gist.github.com/rzwiefel/f97c8b7fe76117c2a481851aaaf1a904)
(X: batch size, Y: Dump time in seconds)
screenshot 2017-09-08 13 56 13

rzwiefel:
Manually verified paging works :+1:

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/DumpCommand.java
coyotesqrl: Both dials - particularly thread count - are very sensitive to environment. However, I think 512 seems like a good default value and your analysis supports that. I'd probably stick with a conservative default thread count proportionate to available cores and keep it in the smaller side. One or two per core. This will leave resources available for other connections.

pklinef: What datasets did you use?  How much memory did the JVM have?  I have seen very different impacts adjusting threads and batch size depending on if the ingested files create large or small metacards.

rzwiefel: @pklinef I used the wikipedia dataset, so very large metacard's  (which I would presume would have benefited the most from a large amount of threads). Left the JVM at 4GB aka the **default** setting, which means our commands should probably reflect that default.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/DumpCommand.java
```diff
@@ -239,9 +248,8 @@ protected Object executeWithSubject() throws Exception {
                 break;
             }
 
-            if (pageSize > 0) {
-                query.setStartIndex(query.getStartIndex() + pageSize);
-            }
+            query.setStartIndex(query.getStartIndex() + pageSize);
```
coyotesqrl: ❓ So...you're doing your own paging on top of the paging being run by the `ResultIterable`. Might it make more sense to just partition the results from the iterable and let it handle the pagination internally?

rzwiefel:
build now

PR #2307, opened 2017-09-12
DDF-3121 Fixed NPE in the Platform Scheduler

emmberk:
#### What does this PR do?
This PR adds a work-around that fixes a NPE when scheduling a command from the `Platform Command Scheduler`.
#### Who is reviewing it? 
@vinamartin @mcalcote @glenhein 
@clockard
@figliold 
Add a `Platform Command Scheduler` configuration, and confirm that there is no `java.lang.NullPointerException: "in" is null!` entry in the log. Confirm that the command is executed and that the scheduler behaves as expected.
[KARAF-5362](https://issues.apache.org/jira/browse/KARAF-5362) NPE Creating Session with a null "in" parameter from a SessionFactory
[DDF-3121](https://codice.atlassian.net/browse/DDF-3121)
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
+++ b/platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
@@ -15,6 +15,7 @@ package ddf.platform.scheduler;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.PrintStream;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.StandardCharsets;
@@ -49,13 +50,12 @@ public class CommandJob implements Job {
 
     private static final Security SECURITY = Security.getInstance();
 
-    public Subject getSystemSubject() {
+    protected Subject getSystemSubject() {
         return SECURITY.getSystemSubject();
     }
 
     @Override
     public void execute(final JobExecutionContext context) throws JobExecutionException {
-
         SECURITY.runAsAdmin(() -> {
             Subject subject = getSystemSubject();
 
@@ -84,8 +84,7 @@ public class CommandJob implements Job {
         return bundleContext.getService(bundleContext.getServiceReference(SessionFactory.class));
     }
 
-    public void doExecute(JobExecutionContext context) throws JobExecutionException {
-
+    private void doExecute(JobExecutionContext context) {
         String commandInput;
         try {
             commandInput = checkInput(context);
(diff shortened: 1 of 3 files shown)


platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
```diff
@@ -104,9 +103,18 @@ public void doExecute(JobExecutionContext context) throws JobExecutionException
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
         Session session = null;
 
-        try (PrintStream output = getPrintStream(byteArrayOutputStream)) {
+        // TODO DDF-3280 remove work-around for NPE when creating session with a null "in" parameter from a SessionFactory
+        try (PrintStream output = getPrintStream(byteArrayOutputStream);
+                InputStream emptyInputStream = new InputStream() {
+                    @Override
+                    public int read() throws IOException {
+                        LOGGER.warn(
```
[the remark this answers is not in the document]
emmberk: It's not necessary, but if the `org.apache.felix.gogo` library is updated again and this `InputStream` is actually `read`, maybe this log statement would indicate that this work-around needs to be updated.

platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
```diff
@@ -104,9 +103,18 @@ public void doExecute(JobExecutionContext context) throws JobExecutionException
         ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
         Session session = null;
 
-        try (PrintStream output = getPrintStream(byteArrayOutputStream)) {
+        // TODO DDF-3280 remove work-around for NPE when creating session with a null "in" parameter from a SessionFactory
+        try (PrintStream output = getPrintStream(byteArrayOutputStream);
```
mcalcote: ❓  Is there any reason not to move the ByteArrayOutputStream into this try-with-resources? It looks like you might be able to move the session inside as well and completely remove the `finally`.

emmberk: :+1: The finally now only contains:
```java
        } finally {
            if (session != null) {
                session.close();
            }
        }
```

emmberk:
build now

platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
```diff
@@ -49,13 +50,12 @@
 
     private static final Security SECURITY = Security.getInstance();
 
-    public Subject getSystemSubject() {
+    protected Subject getSystemSubject() {
```
figliold: :question: Can this be private?

vinamartin:
Manually verified ✅

emmberk:
build now

platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
```diff
@@ -130,16 +138,14 @@ public void doExecute(JobExecutionContext context) throws JobExecutionException
             }
         } catch (UnsupportedEncodingException e) {
             LOGGER.info("Unable to produce output", e);
+        } catch (IOException e) {
+            LOGGER.info(
```
figliold: :question: If this IOException happened, wouldn't it happen on the implicit call to close() in your try-with-resources? It doesn't sound like that from your log message.

platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
emmberk: This method is redefined in the unit tests.

platform/platform-scheduler/src/main/java/ddf/platform/scheduler/CommandJob.java
emmberk: I updated the message to `"Error with the emptyInputStream used as a work-around for a Karaf bug. This should never happen. There is an issue with the Platform Command Scheduler implementation."`

emmberk:
build now

PR #2317, opened 2017-09-14
DDF-3207 Updated CSW delete to collect all IDs before batching

emanns95:
#### What does this PR do?
Collects all IDs matching the filter and holds them in memory before batching out delete requests. The reason for this is because if we send a batch for deletion and then iterate on the query request, the resultIterable will update the index to the next set even though the deletion will result in the results left-shifting. This will result in records that should be deleted being skipped when batching is used.

#### Who is reviewing it? 
@jhunzik 
@rzwiefel 
@ahoffer 

@coyotesqrl
@lessarderic

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
@@ -70,6 +70,7 @@ import org.codice.ddf.spatial.ogc.csw.catalog.common.transaction.DeleteAction;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.transaction.InsertAction;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.transaction.UpdateAction;
 import org.codice.ddf.spatial.ogc.csw.catalog.common.transformer.TransformerManager;
+import org.geotools.filter.text.cql2.CQLException;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
 import org.slf4j.Logger;
@@ -589,7 +590,7 @@ public class CswEndpoint implements Csw {
         for (DeleteAction deleteAction : request.getDeleteActions()) {
             try {
                 numDeleted += deleteRecords(deleteAction);
-            } catch (CswException | FederationException | IngestException | SourceUnavailableException | UnsupportedQueryException | CatalogQueryException e) {
+            } catch (Exception e) {
                 LOGGER.debug("Unable to delete record(s)", e);
                 throw new CswException("Unable to delete record(s).",
                         CswConstants.TRANSACTION_FAILED,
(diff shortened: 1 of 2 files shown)


emanns95:
build now

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswEndpoint.java
```diff
@@ -670,25 +670,26 @@ private int deleteRecords(DeleteAction deleteAction)
         queryRequest = queryFactory.updateQueryRequestTags(queryRequest,
                 schemaTransformerManager.getTransformerSchemaForId(deleteAction.getTypeName()));
 
-        Iterable<List<Result>> resultLists = Iterables.partition(ResultIterable.resultIterable(
-                framework,
-                queryRequest), DEFAULT_BATCH);
+        ResultIterable resultIterable = ResultIterable.resultIterable(framework, queryRequest);
```
lessarderic: ❓ To really make sure we're not causing OOM, should we do something like this instead?
```
Iterable<Result> results = Iterables.limit(ResultIterable(
         framework,
         queryRequest), DEFAULT_BATCH);

while (results.hasNext()) {
    Set<String> idsToDelete = results.next().stream()....;

    if (!idsToDelete.isEmpty()) {
        // Delete batch
    }

    results = Iterables.limit(ResultIterable(
         framework,
         queryRequest), DEFAULT_BATCH);
}
```

rzwiefel: If you wish to limit the amount of results there is a method now to pass a maximum result limit to fetch 
https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java#L97-L98

rzwiefel: @lessarderic In regards to the actual comment, I suggested your way as an alternative but @coyotesqrl thought that it would be best to do it wholesale like it is being done here. 

We are only mapping to IDs, which means that they would have to be requesting... a very very large amount of ids in order to overwhelm the system purely with id strings.

lessarderic: Yeah, I did think about that. There are two things to consider here imho:

1. Multiple operations could be going on at the same time, so if someone does a large delete and bloats the memory with IDs, that could cause other requested being processed to blow up the memory.
1. We've never tried to see how many IDs can be passed down to the Catalog Framework/Solr before things blow up like they did on the request side.

rzwiefel: The Max Boolean clauses limit of solr usually blows up (1024 max clauses). (For a single request)

emanns95:
build now

clockard:
build now

emanns95:
build now

jhunzik:
Successfully deleted 968 records as intended. 👍

PR #2321, opened 2017-09-14
DDF-3281 Add default geoname data and default to offline gazetteer.

clockard:
#### What does this PR do?
Switches the default gazetteer from the online version to the offline. Adds a default dataset for the offline gazetteer.
#### Who is reviewing it? 
@emmberk @mcalcote 

@bdeining
@rzwiefel

Install DDF and turn off the option Standard Search UI -> Catalog UI Search-> Use Online Gazetteer.
Go to the Intrigue ui and
Search for a location like "Arizona" in the search map search bar in the upper right and verify that the location is found.
[DDF-3281](https://codice.atlassian.net/browse/DDF-3281)
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
+++ b/catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
@@ -1,0 +1,103 @@
+/**
+ * Copyright (c) Codice Foundation
+ * <p>
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or any later version.
+ * <p>
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ **/
+package org.codice.ddf.spatial.geocoding.index;
+
+import java.io.File;
+import java.nio.file.Paths;
+import java.util.concurrent.ExecutorService;
+
+import org.codice.ddf.spatial.geocoding.GeoEntryExtractionException;
+import org.codice.ddf.spatial.geocoding.GeoEntryExtractor;
+import org.codice.ddf.spatial.geocoding.GeoEntryIndexer;
+import org.codice.ddf.spatial.geocoding.GeoEntryIndexingException;
+import org.codice.ddf.spatial.geocoding.GeoNamesRemoteDownloadException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IndexInitializer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(IndexInitializer.class);
+
+    private String defaultGeonamesDataPath;
+
+    private String indexLocationPath;
+
+    private GeoEntryExtractor extractor;
+
+    private GeoEntryIndexer indexer;
+
+    private ExecutorService executor;
+
+    public void init() {
+        File defaultGeonamesDataFile = Paths.get(defaultGeonamesDataPath)
+                .toFile();
+        File indexLocationDir = Paths.get(indexLocationPath)
+                .toFile();
+        if (!defaultGeonamesDataFile.exists()) {
+            LOGGER.warn(
+                    "Could not locate default geonames data file at {}. Check that your distribution is no corrupted");
+            return;
+        }
+
+        if (indexLocationDir.exists() && indexLocationDir.list().length > 0) {
+            LOGGER.info("Geoname index already exists. Not indexing default data set.");
+            return;
+        }
+        executor.submit(() -> {
+            try {
+                LOGGER.info("Indexing default geoname data at {}.", defaultGeonamesDataPath);
+
+                indexer.updateIndex(defaultGeonamesDataFile.getAbsolutePath(),
+                        extractor,
+                        true,
+                        progress -> {
+                            if (progress >= 100) {
+                                LOGGER.info("Default geoname data indexed successfully");
+                            }
+                        });
+            } catch (GeoEntryIndexingException | GeoEntryExtractionException | GeoNamesRemoteDownloadException e) {
+                LOGGER.error(
+                        "Failed to index default geonames data. Try using the geonames:update command to generate the index manually.",
+                        e);
+            }
+        });
+
+        executor.shutdown();
+    }
+
+    public void destroy() {
+        executor.shutdownNow();
+    }
+
+    public void setDefaultGeonamesDataPath(String defaultGeonamesDataPath) {
+        this.defaultGeonamesDataPath = defaultGeonamesDataPath;
+    }
+
+    public void setExtractor(GeoEntryExtractor extractor) {
+        this.extractor = extractor;
+    }
+
+    public void setIndexLocationPath(String indexLocationPath) {
+        this.indexLocationPath = indexLocationPath;
+    }
+
+    public void setIndexer(GeoEntryIndexer indexer) {
+        this.indexer = indexer;
+    }
+
+    public void setExecutor(ExecutorService executor) {
+        this.executor = executor;
+    }
+
+}
--- a/catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -28,6 +28,21 @@
         <property name="indexLocation" value="data/geonames-index"/>
     </bean>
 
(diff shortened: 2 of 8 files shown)


distribution/docs/src/main/jdocs/content/_application-references/managing-spatial-contents.adoc
```diff
@@ -8,10 +8,10 @@ The ${ddf-spatial} Application provides KML transformer and a KML network link e
 
 ==== Offline Gazetteer Service
 
-In the ${ddf-spatial} Application, you have the option to install a feature called `offline-gazetteer`.
+In the ${ddf-spatial} Application, the `offline-gazetteer` is installed by default.
 This feature enables you to use an offline index of GeoNames data (as an alternative to the GeoNames Web service enabled by the `webservice-gazetteer` feature) to perform searches via the gazetteer search box in the Search UI.
 
-To use the offline gazetteer, you will need to create an index using the `geonames:update` command.
+By default a small index set is included with the offline gazetteer. This index can be expanded by using the `geonames:update` command.
```
emmberk: `can be expanded or updated by`?

clockard: Updated

clockard:
build now

clockard:
build now

catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.config.xml
```diff
@@ -87,7 +87,7 @@
 
         <AD id="onlineGazetteer"
             name="Use Online Gazetteer"
-            description="Should the online gazetteer be used? If unchecked a local one will be used."
+            description="Should the online gazetteer be used? If unchecked a local gazetteer service will be used. This only applies to the search gazetteer in Intrigue."
             type="Boolean"
             default="true"
```
emanns95: Should this default be left to true?

clockard: This is only for the UI gazetteer so this still should be true. For the UI online gazetteer means using a built in online gazetteer. Not online or local gazetteer means us the ddf gazetteer service which could be and offline or online gazetteer.

catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.config.xml
```diff
@@ -87,7 +87,7 @@
 
         <AD id="onlineGazetteer"
             name="Use Online Gazetteer"
-            description="Should the online gazetteer be used? If unchecked a local one will be used."
+            description="Should the online gazetteer be used? If unchecked a local gazetteer service will be used. This only applies to the search gazetteer in Intrigue."
```
emmberk: Thanks for adding this change:+1:

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -28,6 +28,21 @@
...
+    <reference id="geoExtractor" interface="org.codice.ddf.spatial.geocoding.GeoEntryExtractor"/>
+
+    <bean id="executorService" class="java.util.concurrent.Executors"
+          factory-method="newSingleThreadExecutor">
+    </bean>
+
+    <bean id="indexInitializer" class="org.codice.ddf.spatial.geocoding.index.IndexInitializer"
+          init-method="init" destroy-method="destroy">
+        <property name="indexLocationPath" value="data/geonames-index"/>
+        <property name="defaultGeonamesDataPath" value="data/default_geonames_data.zip"/>
```
rzwiefel: ❓  will there be any issue with these paths on other systems like Windows?

rzwiefel: Instead of hard coding the character here we could use the extras parameters on `Path.get(String, String...)` so that it builds the path for us

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private GeoEntryIndexer indexer;
+
+    private ExecutorService executor;
+
+    public void init() {
+        File defaultGeonamesDataFile = Paths.get(defaultGeonamesDataPath)
+                .toFile();
+        File indexLocationDir = Paths.get(indexLocationPath)
+                .toFile();
```
rzwiefel: ❓ I'm not quite familiar with how blueprint handles/communicates exceptions during an init method, but are we fine with the default behavior if `Paths.get`throws an `InvalidPathException`?

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/resources/OSGI-INF/blueprint/blueprint.xml
clockard: I will double check that this will work on windows. It is following the same pattern already used in the file right now but that doesn't make it right

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
```diff
@@ -0,0 +1,96 @@
...
+        if (!defaultGeonamesDataFile.exists() || (indexLocationDir.exists()
+                && indexLocationDir.list().length > 0)) {
+            LOGGER.info("Not indexing default geonames.");
```
rzwiefel: ✏️  Could we give a little more information in this log message as to why we are not indexing them? (perhaps based on why the logic above was true).  Also maybe a little more on what effects this will have (E.g, using the local gazetteer in Intrigue will not function)

Also since file is supposed to be there and isn't (and there's really nothing an admin can do about this -- unless we tell them to verify distribution install integrity) should we have this statement at WARN ?

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
```diff
@@ -0,0 +1,96 @@
...
+            return;
+        }
+        executor.submit(() -> {
+            try {
+                LOGGER.info("Indexing default geoname data at {}.", defaultGeonamesDataPath);
```
rzwiefel: ❓ Do we want these log statements at info (line 54 & 61)? Seems like this is more of a debug or trace log.

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
clockard: I think I might break this up into two different statements and log different messages. If the default file is not there then i think you are right that we should log a warning.

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -28,6 +28,21 @@
         <property name="indexLocation" value="data/geonames-index"/>
     </bean>
 
+    <reference id="geoExtractor" interface="org.codice.ddf.spatial.geocoding.GeoEntryExtractor"/>
+
+    <bean id="executorService" class="java.util.concurrent.Executors"
+          factory-method="newSingleThreadExecutor">
```
rzwiefel: Just a reminder comment to make sure we clean up this executor/single thread after we are done with initialization.

clockard: Good call. Added executor.shutdown() at the end of the init() method

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
clockard: If the distribution is good then this shouldn't really ever happen but I did run into this when developing and everything came up fine except for the this class.

catalog/spatial/geocoding/spatial-geocoding-lucene/src/main/java/org/codice/ddf/spatial/geocoding/index/IndexInitializer.java
clockard: Since these only will be logged once I was thinking it would be ok at the info level.

clockard:
build now

clockard:
build now

clockard:
build now

clockard:
build now

clockard:
Manually verified.

PR #2334, opened 2017-09-18
DDF-3094 Updates histogram to bin dates by ranges instead of categories, and to show the entire result set

andrewkfiedler:
- Added new method to calculate categories for date type histograms.
 - Updated single click methods to utilize the getCategories methods instead of trying to calculate the single bin.  This is more reliable.
 - Do note that plotly uses a special convention sometimes for date bin sizes.  Instead of a number, they give a string like so: 'M3'.  Where M means months, and the number following M means the number of months.
 - Updates histogram to use full result set rather than page
 - Adds guard for attributes that have no matching data.
 - Updates new selection interface attribute to be cleared when going back to searches from results.
 - Updates metacard.js to account for new selection interface attribute
 - Updates the histogram tooltips to be more consistent with the rest of the UI, and readable.

@jlcsmith 
@kcwire 
@pklinef

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/alert/alert.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/alert/alert.js
@@ -44,6 +44,11 @@ define([
                 type: Backbone.Many,
                 key: 'activeSearchResults',
                 relatedModel: Metacard.MetacardResult
+            },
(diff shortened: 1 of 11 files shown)


jlcsmith:
Verified histogram binning of dates as expected
capture

PR #2340, opened 2017-09-19
DDF-3300 Makes searching easier for users

andrewkfiedler:
#### What does this PR do?
- Adds "adhoc" search capability to the workspaces route.
 - Refactors the workspace route to be more friendly towards adhoc searching.
 - Effectively removes the slideout panel for adding / editing searches.  This is done through
dropdowns now.
 - Users have to initiate drawing now.  In drawing mode, dropdowns are hidden to allow the full use of the map.  Once drawing finishes the dropdowns return.  Users can also cancel drawings.
 - Updates logic for adding queries when max number has been reached.  Now the user gets the op
tion to replace a current query or to start a new workspace with the desired query.
 - Basic and Advanced views collapsed to make it so the user doesn't have to remember if their
search was advanced or not.
 - Query item view refactored to allow easy access to editing the query, seeing status, adding
notifications, etc.
 - Adding a search is done from the top navigation bar for the workspace route now.
 - Users can now tell at a glance whether a query has notifications turned on or not.
 - Updated workspace item view to include owner information
 - Moves logic for not closing on drawing to the base dropdown companion, as it makes more sense there is eliminates the need for extensions of the view.  It also helps keep the close logic in sync, as the extensions had drifted a bit.
 - Fixes a spelling issue.
 - Updates cesium styling to ensure drawing tools are hidden during failover
 - Update label for federation to be sources
 - Adds context help to query button
 - Updates adhoc search construction, default titles for searches
 - Adhoc searches now only surround text with wildcards if no wildcard is present
 - Default titles when adding searches is now more consistent / "smarter"
 - Lowers point size on maps to be closer to pre-icon size
 - In the future we should try to tie these settings into the user's chosen theme.  For now the default is 12px, in a circle 24px in diameter.
 - Updates search to be more noticable in top bar
 - Adds text to search button in top bar.  The text will naturally disappear when the screen width is too small or the workspace name too big.  In the future we can make it hide the moment it starts to disappear, but this will work for the time being.
 - Fixes the location component css to deal with width correctly.
 - Adds form to adhoc component to enable autocomplete.
 - Adds type="button" to some buttons to prevent them grabbing the enter key when pressing enter in the text box.
 - Default blank search only is a single wildcard now.
 - Blank search titles are set to wildcard now, similarly with workspace titles from adhoc search.
 - Removes old visualization selector
 - Updates other routes to have their own golden layouts
 - And a number of other things 😉  (read the commit logs)

#### Who is reviewing it? 
@jlcsmith
@kcwire
@pklinef


kcwire:
@millerw8 FYSA.

pklinef:
@andrewkfiedler Can you add some screenshots and/or videos of the more major changes the PR is bring in?

jlcsmith:
Workspaces view and quick search
workspaces

New Search
new-search

Search Edit
search-edit

Edit Location Search
location-search

Drawing Mode
drawing-mode

Search Status
search-status

Search Notification Settings
search-notifications

Search Settings
search-settings

Search List
search-list

New Search by Result Location
search-by-metacard-location

clockard:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/workspace/QueryMetacardTypeImpl.java
```diff
@@ -126,6 +128,15 @@
             false /* tokenized */,
             false /* multivalued */,
             BasicTypes.STRING_TYPE));
+
+    DESCRIPTORS.add(
+        new AttributeDescriptorImpl(
+            QUERY_IS_ADVANCED,
```
pklinef: ❓ Have we done anything to check if this impacts backwards compatibility with old workspaces?

andrewkfiedler: It won't have any impact.  Old workspace queries will default to false for isAdvanced, so they will act as they would have previously (basic view, with the user having to know to switch to advanced if necessary).

clockard:
build now

PR #2344, opened 2017-09-20
DDF-3034 Prevent duplicate results during result pagination.

coyotesqrl:
#### What does this PR do?
Because we simulate a cursor by executing sequential queries with an incremented start index position, it is possible that ingests can occur, interleaved within subsequent queries. This could result in duplicate results appearing in multiple pages of the results. This will remove those duplicates.

#### Who is reviewing it? 
@rzwiefel @emanns95 

@codice/core-apis

@clockard
@lessarderic

Full build and test run should be sufficient. Rerunning some of the manual heroing steps from recent pagination-related command PRs would be good as well.

N/A

N/A

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
@@ -18,6 +18,7 @@ import static org.apache.commons.lang.Validate.isTrue;
 import static org.apache.commons.lang.Validate.notNull;
 
 import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.operation.Query;
@@ -27,13 +28,19 @@ import ddf.catalog.operation.impl.QueryImpl;
 import ddf.catalog.operation.impl.QueryRequestImpl;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import javax.annotation.Nullable;
 
 /**
  * Class used to iterate over the {@link Result} objects contained in a {@link
@@ -52,6 +59,22 @@ public class ResultIterable implements Iterable<Result> {
   private final QueryRequest queryRequest;
 
   private final int maxResultCount;
+
+  private ResultIterable(
+      CatalogFramework catalogFramework, QueryRequest queryRequest, int maxResultCount) {
+    this(catalogFramework::query, queryRequest, maxResultCount);
+  }
+
+  private ResultIterable(
+      QueryFunction queryFunction, QueryRequest queryRequest, int maxResultCount) {
+    notNull(queryFunction, "Query function cannot be null");
+    notNull(queryRequest, "Query request cannot be null");
+    isTrue(maxResultCount >= 0, "Max Results cannot be negative", maxResultCount);
+
+    this.queryFunction = queryFunction;
+    this.queryRequest = queryRequest;
+    this.maxResultCount = maxResultCount;
+  }
 
   /**
    * Creates an iterable that will call the {@link CatalogFramework} to retrieve the results that
@@ -112,20 +135,9 @@ public class ResultIterable implements Iterable<Result> {
     return new ResultIterable(queryFunction, queryRequest, maxResultCount);
   }
 
-  private ResultIterable(
-      CatalogFramework catalogFramework, QueryRequest queryRequest, int maxResultCount) {
-    this(catalogFramework::query, queryRequest, maxResultCount);
-  }
-
-  private ResultIterable(
-      QueryFunction queryFunction, QueryRequest queryRequest, int maxResultCount) {
-    notNull(queryFunction, "Query function cannot be null");
-    notNull(queryRequest, "Query request cannot be null");
-    isTrue(maxResultCount >= 0, "Max Results cannot be negative", maxResultCount);
-
-    this.queryFunction = queryFunction;
-    this.queryRequest = queryRequest;
-    this.maxResultCount = maxResultCount;
+  private static Stream<Result> stream(Iterator<Result> iterator) {
+    return StreamSupport.stream(
+        Spliterators.spliteratorUnknownSize(iterator, Spliterator.ORDERED), false);
   }
 
   @Override
@@ -140,23 +152,14 @@ public class ResultIterable implements Iterable<Result> {
     return stream(iterator());
   }
 
-  private static Stream<Result> stream(Iterator<Result> iterator) {
-    return StreamSupport.stream(
-        Spliterators.spliteratorUnknownSize(iterator, Spliterator.ORDERED), false);
-  }
-
   private static class ResultIterator implements Iterator<Result> {
 
     private final QueryFunction queryFunction;
-
+    private final Set<String> foundIds = new HashSet<>(2048);
     private int currentIndex;
-
     private QueryImpl queryCopy;
-
     private QueryRequestImpl queryRequestCopy;
-
     private Iterator<Result> results = Collections.emptyIterator();
-
     private boolean finished = false;
 
     ResultIterator(QueryFunction queryFunction, QueryRequest queryRequest) {
@@ -179,7 +182,8 @@ public class ResultIterable implements Iterable<Result> {
 
       fetchNextResults();
 
-      return results.hasNext();
+      // Recurse to ensure we continue querying even if we get 1 or more completely filtered pages.
+      return hasNext();
     }
 
     @Override
@@ -201,19 +205,51 @@ public class ResultIterable implements Iterable<Result> {
       return results.next();
     }
 
+    @SuppressWarnings("squid:CommentedOutCodeLine")
     private void fetchNextResults() {
       queryCopy.setStartIndex(currentIndex);
 
       try {
         SourceResponse response = queryFunction.query(queryRequestCopy);
 
-        if (response.getResults().size() == 0) {
+        final List<Result> resultList = response.getResults();
+
+        // Because some of the results may be filtered out by the catalog framework's
+        // plugins, we need a way to know the actual page size and increment currentIndex based
+        // on that number instead of using the result list size.
+        // If the property is not present, we will have no option but to fallback to the size
+        // of the (potentially filtered) resultList.
+        //
+        // This means that if the filtered results size is zero, but the raw number of results
+        // had been greater than zero, we will not find results beyond the filtered gap. In practice
+        // this should not happen, as queries will run through the QueryOperations.query() method;
+        // however, should a user ever construct a QueryFunction that does NOT rely on that method,
+        // there is no guarantee that this property will be properly set.
+        int actualResultSize =
+            Optional.ofNullable(response.getProperties())
+                .map(m -> m.get("actualResultSize"))
+                .filter(Integer.class::isInstance)
+                .map(Integer.class::cast)
+                .orElse(resultList.size());
+
+        if (actualResultSize == 0) {
           finished = true;
           return;
         }
+        currentIndex += actualResultSize;
 
-        results = response.getResults().iterator();
-        currentIndex += response.getResults().size();
+        List<Result> dedupedResults = new ArrayList<>(resultList.size());
+        for (Result result : resultList) {
+          if (isDistinctResult(result)) {
+            dedupedResults.add(result);
+          }
+          Optional.ofNullable(result)
+              .map(Result::getMetacard)
+              .map(Metacard::getId)
+              .ifPresent(foundIds::add);
+        }
+
+        this.results = dedupedResults.iterator();
 
         if (response.getHits() >= 0 && currentIndex > response.getHits()) {
           finished = true;
@@ -221,6 +257,13 @@ public class ResultIterable implements Iterable<Result> {
       } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
         throw new CatalogQueryException(e);
       }
+    }
+
+    private boolean isDistinctResult(@Nullable Result result) {
+      return result != null
+          && (result.getMetacard() == null
+              || result.getMetacard().getId() == null
+              || !foundIds.contains(result.getMetacard().getId()));
     }
 
     private void copyQueryRequestAndQuery(QueryRequest queryRequest) {
@@ -234,7 +277,8 @@ public class ResultIterable implements Iterable<Result> {
               query.getStartIndex(),
               pageSize,
               query.getSortBy(),
-              true, // always get the hit count
+              true,
+              // always get the hit count
               query.getTimeoutMillis());
 
       this.queryRequestCopy =
--- a/catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
+++ b/catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
@@ -14,12 +14,15 @@
 package ddf.catalog.util.impl
 
 import ddf.catalog.CatalogFramework
+import ddf.catalog.data.Metacard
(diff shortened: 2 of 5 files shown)


catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -212,16 +220,32 @@ private void fetchNextResults() {
...
-                results = response.getResults()
-                        .iterator();
-                currentIndex += response.getResults()
-                        .size();
+                foundIds.addAll(dedupedResults.stream()
+                        .map(Result::getMetacard)
+                        .filter(Objects::nonNull)
+                        .map(Metacard::getId)
+                        .filter(Objects::nonNull)
+                        .collect(Collectors.toSet()));
```
rzwiefel: ❓ Do we want to incur the overhead of the extra set creation instead of just iterating/adding to the `foundResults` set?
Not sure if its worth it, but could be since this could be used very heavily

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -164,6 +170,8 @@ private ResultIterable(QueryFunction queryFunction, QueryRequest queryRequest,
 
         private boolean finished = false;
 
+        private final Set<String> foundIds = new HashSet<>(2048);
```
rzwiefel: ❓ Rational?
Many queries might just be 20 items. or 250.  2048 default set size seems a little extreme since it will be very difficult to gauge how much it will actually need

coyotesqrl: If the resultset _is_ large, the last thing we want to do is thrash and repeatedly perform mem copies. If the resultset is small, this is going to be very short-lived. Also, it's not realIy all that much memory being allocated up front. I weighed the options and leaned toward the more memory-intensive solution.

I could be persuaded to change the default size.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
coyotesqrl: Good suggestion.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -212,16 +220,32 @@ private void fetchNextResults() {
...
-                results = response.getResults()
-                        .iterator();
-                currentIndex += response.getResults()
-                        .size();
+                dedupedResults.stream()
+                        .map(Result::getMetacard)
+                        .filter(Objects::nonNull)
+                        .map(Metacard::getId)
+                        .filter(Objects::nonNull)
+                        .forEach(foundIds::add);
```
coyotesqrl: Per @rzwiefel's suggestion, skipped the interim set creation.

coyotesqrl:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -207,13 +215,32 @@ private void fetchNextResults() {
...
+        // TODO RAP 19 Sep 17: This test is insufficient, as it does not account
+        // for large blocks of filtered results
+        if (resultCount == 0) {
           finished = true;
           return;
         }
 
-        results = response.getResults().iterator();
-        currentIndex += response.getResults().size();
+        dedupedResults
```
lessarderic: ❓ Since we have to assume that results will always be returned in the same order for paging to work properly, I was thinking that we might be able to only keep the ID of the _last_ metacard returned and drop everything that comes prior and including that metacard. Thoughts?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -159,6 +165,8 @@ private ResultIterable(
 
     private boolean finished = false;
 
+    private final Set<String> foundIds = new HashSet<>(2048);
```
coyotesqrl: Two reviewers have rightly pointed out that this large set represents what we expect to be the less common case. My feeling is that the short-lived large collection is a good tradeoff for using too small a collection by default and doing too many memcopies for it to expand. @lessarderic, do you have any thoughts on this? Maybe a smaller, but still copious initial size around 250 would be a good compromise?

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
ahoffer: It is possible to get duplicate results because the catalog changes between successive queries. Is it also possible for the provider's index to be out-of-sync with the expected results?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -223,6 +250,12 @@ private void fetchNextResults() {
       }
     }
 
+    private boolean isDistinctResult(Result result) {
+      return result.getMetacard() == null
+          || result.getMetacard().getId() == null
+          || !foundIds.contains(result.getMetacard().getId());
+    }
```
ahoffer: If null metacards/ids were considered to be not distinct, could the stream filters on lines 234 and 236 be removed? Line 222 would have already eliminated them from the result set.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
coyotesqrl: The index could absolutely be out-of-sync. Have discussed with @lessarderic and we're going to stick with the more expensive but surer deduping logic that's already present.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
coyotesqrl: @lessarderic What conditions, if any, can result in a result with a null metacard or a metacard with a null id?

coyotesqrl:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
ahoffer: I think Ethan put those checks in because he couldn't be sure they wouldn't be null.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -275,7 +278,44 @@ class QueryResultPaginatorSpec extends Specification {
...
+        }
+        def id = actualResults*.metacard*.id
+        def dedupedCount = id.unique().size()
+
+        catalogFramework.query(_ as QueryRequest) >>
+                { qr -> buildQueryResponse(actualResults, 0..4)} >>
+                { qr -> buildQueryResponse(actualResults, 5..9)} >>
+                { qr -> buildQueryResponse(actualResults, 10..14)} >>
+                { qr -> buildQueryResponse(actualResults, 15..19)} >>
+                { qr -> buildQueryResponse(actualResults, 20..24)}
```
rzwiefel: This should technically end with an empty response after all results have been returned. Currently it will end because the hit count is present (so it knows to stop). However in the absence of a hit count this test will spin indefinitely (or maybe stack overflow 🤔 ) because it will continue try try and query for results but always get those final 5 results. 

You should add a
`{ QueryRequest queryRequest -> buildEmptyQueryResponse(actualResults) }` 
at the end similiar to the other results.

rzwiefel:
Have we verified that no other places will need to be updated with this new property to correctly page through filtered results?

coyotesqrl:
@rzwiefel I haven't scoured the codebase for anyone doing any pagination manually, but I believe those have all been caught in the pagination updates.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -207,13 +216,43 @@ private void fetchNextResults() {
...
+
+        // Because some of the results may be filtered out by the catalog framework's
+        // plugins, we need a way to know the actual page size and increment currentIndex based
+        // on that number instead of using the result list size.
+        // If the property is not present, we will have no option but to fallback to the size
+        // of the (potentially filtered) resultList.
+        //
+        // This means that if the filtered results size is zero, but the raw number of results
+        // had been greater than zero, we will not find results beyond the filtered gap. In practice
+        // this should not happen, as queries will run through the QueryOperations.query() method;
```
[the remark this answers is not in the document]
rzwiefel: Wrong again, Kiddo.

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: fixed

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: Reduced initial capacity to 256.

rzwiefel:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
rzwiefel: Changed back to 2048

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -201,19 +210,50 @@ public Result next() {
...
+            .filter(Objects::nonNull)
+            .map(Metacard::getId)
+            .filter(Objects::nonNull)
+            .forEach(foundIds::add);
 
-        results = response.getResults().iterator();
-        currentIndex += response.getResults().size();
+        this.results = dedupedResults.iterator();
 
         if (response.getHits() >= 0 && currentIndex > response.getHits()) {
```
lessarderic: ❓ Do we still need this check here?

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
```diff
@@ -28,10 +29,12 @@ import spock.lang.Unroll
 import static ddf.catalog.util.impl.ResultIterable.resultIterable
 import static java.util.stream.Collectors.toList
 
-class QueryResultPaginatorSpec extends Specification {
+class ResultIterableSpec extends Specification {
```
lessarderic: ❓ Do we have all the tests required to cover the code related to the new `actualResultSize` property, and all the other possible scenarios, e.g., empty result list returned by more results are still available?

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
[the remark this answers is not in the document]
coyotesqrl: I'm REALLY NOT HAPPY that SonarQube 
a) has any opinion whatsoever about inline comments where are a valid and useful tool to explicate complex or otherwise confusing code
b) is potentially choking on my perfectly correct use of a semi-colon to mark out an independent clause.

Would SonarQube rather I used a grammatically incorrect _comma splice_?!?

Joking aside, if there's a way to tell SonarQube to NEVER EVER EVER comment about the existence of an inline comment, we need to do that.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -201,19 +210,50 @@ public Result next() {
...
+        if (actualResultSize == 0) {
           finished = true;
           return;
         }
+        currentIndex += actualResultSize;
+
+        final List<Result> dedupedResults =
+            resultList.stream().filter(this::isDistinctResult).collect(Collectors.toList());
+
+        dedupedResults
```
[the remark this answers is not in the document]
coyotesqrl: So something like...
resultList.iter...
if !isDistinct then remove
else add metacardId
?

That might be much tighter if the list we get support iterator.remove(). And if not, could do something similar with an add to a new list.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
coyotesqrl: This catches the case where we got the very last result in this batch. If we remove it, we'd just execute one more query that returned nothing. It's a minor performance enhancement.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
```diff
@@ -201,19 +210,50 @@ public Result next() {
...
+                    .map(m -> m.get("actualResultSize"))
+                    .orElse(resultList.size());
+
+        if (actualResultSize == 0) {
           finished = true;
           return;
         }
+        currentIndex += actualResultSize;
+
+        final List<Result> dedupedResults =
```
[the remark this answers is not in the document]
rzwiefel: I try to not mutate collections returned from api's since it can produce some nasty issues, so I will pre create the list the same size as the results

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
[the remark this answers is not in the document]
rzwiefel: Condensed to:

```java
        final List<Result> dedupedResults =
            resultList
                .stream()
                .filter(this::isDistinctResult)
                .peek(
                    result ->
                        Optional.ofNullable(result)
                            .map(Result::getMetacard)
                            .map(Metacard::getId)
                            .ifPresent(foundIds::add))
                .collect(Collectors.toCollection(() -> new ArrayList<>(resultList.size())));
```

catalog/core/catalog-core-api-impl/src/test/groovy/ddf/catalog/util/impl/ResultIterableSpec.groovy
rzwiefel: added

rzwiefel:
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/util/impl/ResultIterable.java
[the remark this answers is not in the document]
lessarderic: Even though this works, `peek` is mostly meant for debugging purposes (see API Notes here: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#peek-java.util.function.Consumer-) and not to enable stream "side-effects". Any reason we shouldn't just use a simple for-each loop here?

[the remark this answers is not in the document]
coyotesqrl: Agreed. I would much rather the stream be replaced with the iterative pseudocode. It would be clearer for sure.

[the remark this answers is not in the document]
rzwiefel: Oh yes we can replace with an iterative solution, I (mistakenly) assumed the wrong intention to change to a condensed functional solution. 😄

rzwiefel:
build now

clockard:
build now

clockard:
build now

PR #2347, opened 2017-09-20
DDF-3297 added security manager for DDF container

stustison:
#### Who is reviewing it? 
@pklinef @garrettfreibott @ahoffer @SmithJosh 
@coyotesqrl
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/bin/setenv.bat
+++ b/distribution/ddf-common/src/main/resources/bin/setenv.bat
@@ -56,10 +56,11 @@ rem SET KARAF_DATA
 rem Karaf base folder
 rem SET KARAF_BASE
 rem Additional available Karaf options
-rem SET KARAF_OPTS=-Dderby.system.home="..\data\derby"  -Dderby.storage.fileSyncTransactionLog=true -Dfile.encoding=UTF8 -Dddf.home=%DDF_HOME% -XX:+DisableAttachMechanism
+rem SET KARAF_OPTS=
 
 rem comment out the line below to enable cxf logging interceptors
 rem set EXTRA_JAVA_OPTS="-Dcom.sun.xml.ws.transport.http.HttpAdapter.dump=true"
 
-set JAVA_OPTS=-Xms2g -Xmx4g -Dderby.system.home="%DDF_HOME%\data\derby"  -Dderby.storage.fileSyncTransactionLog=true -Dfile.encoding=UTF8 -Dddf.home=%DDF_HOME% -XX:+DisableAttachMechanism
-:: set JAVA_OPTS=-Xms2g -Xmx4g -Dfile.encoding=UTF8 -Djavax.net.ssl.keyStore=../etc/keystores/serverKeystore.jks -Djavax.net.ssl.keyStorePassword=changeit -Djavax.net.ssl.trustStore=../etc/keystores/serverTruststore.jks -Djavax.net.ssl.trustStorePassword=changeit -Dddf.home=%DDF_HOME%
+set DDF_HOME_POLICY=%DDF_HOME:/bin/..=/%
+
+set JAVA_OPTS=-Xms2g -Xmx4g -Dderby.system.home="%DDF_HOME%\data\derby"  -Dderby.storage.fileSyncTransactionLog=true -Dfile.encoding=UTF8 -Dddf.home=%DDF_HOME% -Dddf.home.policy=%DDF_HOME_POLICY% -XX:+DisableAttachMechanism -Djava.security.manager=net.sourceforge.prograde.sm.ProGradeJSM -Djava.security.policy==%DDF_HOME%\etc\default.policy -DproGrade.getPermissions.override=sun.rmi.server.LoaderHandler:loadClass,org.apache.jasper.compiler.JspRuntimeContext:initSecurity
--- a/distribution/ddf-common/src/main/resources/etc/default.policy
+++ b/distribution/ddf-common/src/main/resources/etc/default.policy
@@ -1,0 +1,79 @@
+priority "grant";
+
+deny {
+  // Read, write to, or execute any file.
+  permission java.io.FilePermission "<<ALL FILES>>", "read, write, execute";
+
+  permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
+
+  // Change current security manager
+  permission java.lang.RuntimePermission "setSecurityManager";
+
+  // Modify application permissions at will.
+  permission java.security.SecurityPermission "setPolicy";
+
+  // Load classes into any protection domain
+  permission java.lang.RuntimePermission "createClassLoader";
+};
+
+grant codeBase "file:/${ddf.home.policy}system/ddf/-" {
+  permission java.io.FilePermission "${ddf.home}${/}-", "read, write";
+  permission java.io.FilePermission "${ddf.home}", "read, write";
+  permission java.util.PropertyPermission "*", "read, write";
+};
+
+grant codeBase "file:/${ddf.home.policy}system/org/codice/-" {
+  permission java.io.FilePermission "${ddf.home}${/}-", "read, write";
+  permission java.io.FilePermission "${ddf.home}", "read, write";
+  permission java.util.PropertyPermission "*", "read, write";
+};
+
+grant codeBase "file:/${ddf.home.policy}system/org/eclipse/platform/-" {
+  permission java.lang.RuntimePermission "createClassLoader";
+  permission java.util.PropertyPermission "*", "read, write";
+};
+
+grant codeBase "file:/${ddf.home.policy}system/org/apache/karaf/-" {
+  permission java.lang.RuntimePermission "createClassLoader";
+  permission java.util.PropertyPermission "*", "read, write";
+};
+
+grant codeBase "file:/${ddf.home.policy}system/org/apache/aries/blueprint/-" {
+  permission java.lang.RuntimePermission "createClassLoader";
+};
+
+grant codeBase "file:/${ddf.home.policy}lib/-" {
+  permission java.lang.RuntimePermission "createClassLoader";
+  permission java.util.PropertyPermission "*", "read, write";
+};
+
+grant principal org.apache.karaf.jaas.boot.principal.RolePrincipal "manager", principal org.apache.karaf.jaas.boot.principal.RolePrincipal "admin" {
+  permission java.io.FilePermission "${ddf.home}${/}-", "read, write";
+  permission java.io.FilePermission "${ddf.home}", "read, write";
+
+  permission javax.security.auth.AuthPermission "getSubject";
+};
+
+grant {
+  permission java.util.PropertyPermission "java.*", "read";
+  permission java.util.PropertyPermission "os.name", "read";
+  permission java.util.PropertyPermission "os.version", "read";
+  permission java.util.PropertyPermission "os.arch", "read";
+  permission java.util.PropertyPermission "file.separator", "read";
+  permission java.util.PropertyPermission "path.separator", "read";
+  permission java.util.PropertyPermission "line.separator", "read";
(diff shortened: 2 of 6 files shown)


stustison:
build now

stustison:
build now

stustison:
build now

distribution/ddf-common/src/main/resources/bin/setenv.bat
```diff
@@ -56,10 +56,11 @@ rem SET KARAF_DATA
...
 rem Additional available Karaf options
-rem SET KARAF_OPTS=-Dderby.system.home="..\data\derby"  -Dderby.storage.fileSyncTransactionLog=true -Dfile.encoding=UTF8 -Dddf.home=%DDF_HOME% -XX:+DisableAttachMechanism
+rem SET KARAF_OPTS=
 
 rem comment out the line below to enable cxf logging interceptors
 rem set EXTRA_JAVA_OPTS="-Dcom.sun.xml.ws.transport.http.HttpAdapter.dump=true"
 
-set JAVA_OPTS=-Xms2g -Xmx4g -Dderby.system.home="%DDF_HOME%\data\derby"  -Dderby.storage.fileSyncTransactionLog=true -Dfile.encoding=UTF8 -Dddf.home=%DDF_HOME% -XX:+DisableAttachMechanism
-:: set JAVA_OPTS=-Xms2g -Xmx4g -Dfile.encoding=UTF8 -Djavax.net.ssl.keyStore=../etc/keystores/serverKeystore.jks -Djavax.net.ssl.keyStorePassword=changeit -Djavax.net.ssl.trustStore=../etc/keystores/serverTruststore.jks -Djavax.net.ssl.trustStorePassword=changeit -Dddf.home=%DDF_HOME%
+set DDF_HOME_POLICY=%DDF_HOME:/bin/..=/%
```
coyotesqrl: ❓ Shouldn't this be `%DDF_HOME:\bin\..=\%`?

stustison:
@paouelle

distribution/ddf-common/src/main/resources/bin/setenv.bat
stustison: No the security policy codeBase grants require all forward slashes whether you're on windows or *nix.

stustison: And there is code up above this that changes all of the back slashes to forward slashes.

paouelle:
❓ Does the security manager get installed if the customer used the documented command to generate a service from the console?

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,79 @@
+priority "grant";
+
+deny {
```
SmithJosh: ❓ Is this deny block necessary? Since you're using the double-equals notation when setting the -Djava.security.policy, aren't these permissions are denied unless explicitly granted?

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,79 @@
...
+  // Change current security manager
+  permission java.lang.RuntimePermission "setSecurityManager";
+
+  // Modify application permissions at will.
+  permission java.security.SecurityPermission "setPolicy";
+
+  // Load classes into any protection domain
+  permission java.lang.RuntimePermission "createClassLoader";
+};
+
```
paouelle: ❗️ The migration:import will need delete permissions and readlink under ddf.home. It will also need read permission outside ddf.home.

distribution/kernel/pom.xml
```diff
@@ -38,6 +38,7 @@
         <win.setup.folder>target/temp/win</win.setup.folder>
         <!-- The name of the distribution zip file to generate in the /target directory -->
         <distribution.file.name>ddf-${project.artifactId}-${project.version}</distribution.file.name>
+        <pro-grade.version>1.1.2</pro-grade.version>
```
paouelle: ✏️  We should move version properties in the to parent pom

platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
```diff
@@ -37,29 +38,37 @@
 
   @Override
   public void configurationEvent(ConfigurationEvent event) {
-    try {
-      String type = getType(event);
-      Configuration[] configuration =
-          configurationAdmin.listConfigurations("(service.pid=" + event.getPid() + ')');
+    AccessController.doPrivileged(
```
paouelle: ❓ Why do we need to elevate the security privileges here? And should we be doing something similar during migration export:import as we are also reading and writing configurations to config admin.

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,79 @@
...
+  permission java.io.FilePermission "${ddf.home}${/}-", "read";
+  permission java.io.FilePermission "${ddf.home}${/}data${/}-", "write";
+  permission java.io.FilePermission "${ddf.home}${/}system${/}-", "write";
+  permission java.io.FilePermission "${ddf.home}${/}instances${/}-", "write";
+  permission java.io.FilePermission "${ddf.home}${/}workspace${/}-", "write";
+  permission java.io.FilePermission "${ddf.home}${/}karaf.pid", "write";
+  permission java.io.FilePermission "${ddf.home}${/}solr${/}-", "write";
+  permission java.io.FilePermission "${user.home}${/}-", "read, write";
+  permission java.io.FilePermission "${java.home}${/}release", "read";
+  permission java.io.FilePermission "${java.home}${/}lib${/}-", "read";
```
ryeats: Do we need some documentation on this? when would an admin have to update this would it need to be updated  if a directory monitor to ingest metacards is used.

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,79 @@
...
+  permission java.lang.RuntimePermission "createClassLoader";
+};
+
+grant codeBase "file:/${ddf.home.policy}lib/-" {
+  permission java.lang.RuntimePermission "createClassLoader";
+  permission java.util.PropertyPermission "*", "read, write";
+};
+
+grant principal org.apache.karaf.jaas.boot.principal.RolePrincipal "manager", principal org.apache.karaf.jaas.boot.principal.RolePrincipal "admin" {
+  permission java.io.FilePermission "${ddf.home}${/}-", "read, write";
```
ryeats: Since this gives read write to etc does this mean if someone manages to get admin or manager access they can just edit this policy file in etc and then gain access to anywhere on the system not just ddf.home?

distribution/ddf-common/src/main/resources/etc/default.policy
stustison: Yes, you would not be able to set up a directory monitor outside of ddf.home or user.home with these changes without modifying the policy.

distribution/ddf-common/src/main/resources/etc/default.policy
stustison: We are using these roles as our "sshRole" right now, which is added back in with the next version of Karaf. The policy can be adjusted after that. It would allow you to adjust the policy file and then you could read in the new policy if the system was restarted. We are taking the additional measure of restricting ssh to the local system anyways, however. So you'd already need another account for the machine to get in.

distribution/ddf-common/src/main/resources/etc/default.policy
stustison: This is how the pro-grade security manager works and because I set the priority to grant to avoid additional denied permissions.

distribution/ddf-common/src/main/resources/etc/default.policy
stustison: We can't do read outside ddf.home. We should be able to add the others for that command, however.

paouelle: We need the import to be able to read outside of ddf.home to compute checksum for cases where the customer has referenced files (e.g. certs or otherwise outside of ddf.home)
@lessarderic How do we handle that?

platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
stustison: You need permission to access the local subject, if one is available and this runs the block with all of the permissions available to the caller's protection domain.

paouelle: ❓ How about the second question? Would that be something that needs to be done to the import and the export command to allow them to properly access all configurations when they are exporting and re-importing?

distribution/ddf-common/src/main/resources/etc/default.policy
lessarderic: @paouelle, how does the code handle read failures when doing a checksum on a file during export and import? Do we just warn the user and move or fail the export/import?

platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
stustison: It should not be necessary when running a command.

distribution/ddf-common/src/main/resources/etc/default.policy
paouelle: We do generate warnings if we cannot read it but that would still break the migration import/export functionality here.

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,79 @@
+priority "grant";
```
paouelle: This whole file now needs to be exported and imported for the migration to work properly. This is to account for changes made by the administrator to allow read access (among other things) to certs that would be placed outside ddf.home.

We also need special permissions added for migration export/import otherwise the functionality will be broken and invalidated in so many place.

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,79 @@
...
+};
+
+grant principal org.apache.karaf.jaas.boot.principal.RolePrincipal "manager", principal org.apache.karaf.jaas.boot.principal.RolePrincipal "admin" {
+  permission java.io.FilePermission "${ddf.home}${/}-", "read, write";
+  permission java.io.FilePermission "${ddf.home}", "read, write";
+
+  permission javax.security.auth.AuthPermission "getSubject";
+};
+
+grant {
```
paouelle: migration export/import needs read access to:
- javax.net.ssl.keyStore
- javax.net.ssl.trustStore

platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
paouelle: The reason I am asking is because export/import will require additional privileges that shouldn't be granted to anything else and since that code will be called by the Karaf shell and the commands we register there, we would need to extend our privileges.

distribution/ddf-common/src/main/resources/etc/default.policy
paouelle: Karaf restart functionality requires read and write access to karaf.restart.jvm from both karaf's code and also from our own code where we implement the restart hook for the UI and restart in migration:import.

coyotesqrl:
Superseded by #2373

PR #2302, opened 2017-09-11
DDF-3268 Search text of indexed Tika parsed documents

gordocanchola:
Adding ext.extracted.text attribute to store the extracted text
Adding XmlMetadataContentHandler used by tika transformers to extract the body of the document

Updating TikaMetadataExtractor to use the XmlMetadataContentHandler
Updating PdfInpuTransformer and TikaInputTransformer
  extract the body of the document and store it in the new ext.extracted.text attribute
  store the Tika parsed metadata in metacard.METADATA
  adding configurable value for number of bytes to store the ext.extracted.text

Updating VideoInputTransformer to handle changes to TikaMetadataExtractor

#### What does this PR do?
This PR affects Tika input transformers.
It stores the document's text extracted by the tika parser in a new metacard attribute, ext.extracted.text.
The length of the extracted text is configurable for each of the input transformers. 
The tika parsed metadata will be stored in the metacard's metadata attribute.

#### Who is reviewing it? 
@dcruver @spearskw 
#### Select relevant component teams: 
https://github.com/orgs/codice/teams
#### Choose 2 committers to review/merge the PR. 
@jlcsmith 
@pklinef
@clockard 
#### How should this be tested? (List steps with links to updated documentation)
Ingest a document (pdf, doc/x, ppt/x, xls/x, mhtml).
Perform a search
Verify the ext.extracted.text attribute is populated in the Inspector->Details
Perform a text search using a word that is in the extracted bytes of the document.
Verify expected results.
Perform a text search using a word that is not in the extracted text.
Verify no result is returned.
Using the Admin Console, change the length of extracted text.
Ingest a document.
Verify the ext.extracted.text attribute contains the expected number of bytes.

#### What are the relevant tickets?
[DDF-3268](https://codice.atlassian.net/browse/DDF-3268)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests
#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2355, opened 2017-09-22
DDF-3262 Increase Solr string indexing size and exclude attributes

pklinef:
#### What does this PR do?
Allow Solr tokenized txt field to index more data than txt string field.
Support excluding attributes from Solr response.

#### Who is reviewing it? 
@brendan-hofmann 
@bdeining 

@codice/solr 

@jlcsmith
@rzwiefel

Create or modify a record with a string field (e.g. title, description) that contains over 32k of text.  At the end of the text (past 32k) include a unique word.  Verify that the record ingests, the unique word past the 32k is searchable, and the first 32k of the string is returned.

This provides the ability to search more than 32k of text that is extracted by the Tika transformer after the DDF-3268 changes to the Tika transformer.
https://github.com/codice/ddf/pull/2302

[DDF-3262](https://codice.atlassian.net/browse/DDF-3262)

#### Checklist:
- [x] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
@@ -43,6 +43,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -90,6 +91,8 @@ public class DynamicSchemaResolver {
   public static final String LUX_XML_FIELD_NAME = "lux_xml";
 
   public static final String SCORE_FIELD_NAME = "score";
+
+  public static final int TOKEN_MAXIMUM_BYTES = 32766;
 
   protected static final char FIRST_CHAR_OF_SUFFIX = '_';
 
@@ -148,7 +151,7 @@ public class DynamicSchemaResolver {
 
   private Processor processor = new Processor(new Config());
 
-  public DynamicSchemaResolver() {
+  public DynamicSchemaResolver(List<String> additionalFields) {
     this.schemaFields = new SchemaFields();
 
     fieldsCache.add(Metacard.ID + SchemaFields.TEXT_SUFFIX);
@@ -168,6 +171,19 @@ public class DynamicSchemaResolver {
     anyTextFieldsCache.addAll(basicTextAttributes);
     fieldsCache.add(Validation.VALIDATION_ERRORS + SchemaFields.TEXT_SUFFIX);
     fieldsCache.add(Validation.VALIDATION_WARNINGS + SchemaFields.TEXT_SUFFIX);
+
+    fieldsCache.add(SchemaFields.METACARD_TYPE_FIELD_NAME);
+    fieldsCache.add(SchemaFields.METACARD_TYPE_OBJECT_FIELD_NAME);
+
+    for (String field : additionalFields) {
+      if (StringUtils.isNotBlank(field)) {
+        fieldsCache.add(field);
+      }
+    }
+  }
+
+  public DynamicSchemaResolver() {
+    this(Collections.emptyList());
   }
 
   /**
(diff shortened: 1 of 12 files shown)


jlcsmith:
build now

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
```diff
@@ -226,33 +231,41 @@ public void addFields(Metacard metacard, SolrInputDocument solrInputDocument)
...
+                    .stream()
+                    .map(
+                        value ->
+                            value != null
+                                ? truncateAsUTF8(value.toString(), TOKEN_MAXIMUM_BYTES)
+                                : value)
+                    .collect(Collectors.toList());
+            // *_txt
+            solrInputDocument.addField(
+                ad.getName() + getFieldSuffix(AttributeFormat.STRING), truncatedValues);
```
rzwiefel: ❓ @pklinef Are we still also indexing this field in addition to the tokenized version below? Not if there's a reason to keep both indexed or if we could disable with our current field definitions anyways -- but hopefully this doesn't have a negative effect on the index size.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
```diff
@@ -325,6 +342,39 @@ public void addFields(Metacard metacard, SolrInputDocument solrInputDocument)
...
+        skip = 1;
+      } else {
+        more = 3;
+      }
+
+      if (b + more > maximumBytes) {
+        return value.substring(0, i);
+      }
+      b += more;
+      i += skip;
```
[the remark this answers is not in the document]
rzwiefel: ❗️ We'll need to deal with this

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -367,9 +371,67 @@ protected SolrQuery postAdapt(
...
+            .collect(Collectors.toSet());
+
+    Set<String> fields = Sets.union(includedFields, wildcardFields);
+
+    if (query.getFields() != null && query.getFields().length() > 2) {
+      fields = Sets.union(fields, Sets.newHashSet(query.getFields().substring(2).split(",")));
+    }
+
+    if (!fields.isEmpty()) {
+      query.setFields(fields.toArray(new String[fields.size()]));
```
rzwiefel: ✏️ Shouldn't we do 
```java
query.setFields(fields.toArray(new String[0]));
```
here instead of the size? 😄

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -367,9 +371,67 @@ protected SolrQuery postAdapt(
...
+                        .stream()
+                        .filter(Objects::nonNull)
+                        .filter(field -> field.endsWith(suffix))
+                        .filter(field -> excludedAttributes.stream().noneMatch(field::startsWith)))
+            .collect(Collectors.toSet());
+
+    Set<String> fields = Sets.union(includedFields, wildcardFields);
+
+    if (query.getFields() != null && query.getFields().length() > 2) {
+      fields = Sets.union(fields, Sets.newHashSet(query.getFields().substring(2).split(",")));
```
rzwiefel: ❓ Should we worry about stripping spaces after splitting on `,`?

From their own code what I can tell is they [sometimes have spaces](https://github.com/apache/lucene-solr/blob/master/solr/solrj/src/java/org/apache/solr/client/solrj/SolrQuery.java#L1074) but then in other comments and code [don't have spaces](https://github.com/apache/lucene-solr/blob/master/solr/core/src/java/org/apache/solr/search/grouping/distributed/requestfactory/StoredFieldsShardRequestFactory.java#L70) and in the end looks like they split on [commas or spaces](https://github.com/apache/lucene-solr/blob/master/solr/core/src/java/org/apache/solr/handler/admin/LukeRequestHandler.java#L339) but don't filter out any empty values... 

So it _seems_ like nothing should happen even in odd cases (like somehow getting a double space) but it might be safest to just trim the fields anyways.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -367,9 +371,67 @@ protected SolrQuery postAdapt(
...
+    if (!fields.isEmpty()) {
+      query.setFields(fields.toArray(new String[fields.size()]));
+    }
+  }
+
+  private boolean skipFilteredAttributes(QueryRequest request) {
+    return !request.hasProperties()
+        || !request.containsPropertyName(EXCLUDE_ATTRIBUTES)
+        || !(request.getPropertyValue(EXCLUDE_ATTRIBUTES) instanceof Set)
+        || "true".equals(System.getProperty("solr.client.filterAttributes.disable"));
```
rzwiefel: ❓ Might we also add one additional check about the set and ensure that it is also not empty? This would ensure none of the `filterAttributes` logic runs unecessarily.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -367,9 +371,67 @@ protected SolrQuery postAdapt(
...
+    Set<String> excludedAttributes = (Set<String>) request.getPropertyValue(EXCLUDE_ATTRIBUTES);
+
+    Set<String> excludedFields =
+        resolver
+            .fieldsCache
+            .stream()
+            .filter(Objects::nonNull)
+            .filter(field -> excludedAttributes.stream().anyMatch(field::startsWith))
+            .collect(Collectors.toSet());
+
```
rzwiefel: At this point we just pulled out all the fields we know about (via the `fieldsCache`) that they want excluded. 

❓ If for some reason they did happen to match any fields (whether valid case or mistake), do we want to add a short circuiting return here? It would essentially just prevent us from sending all our suffixes wildcarded in the field list.

rzwiefel:
@pklinef Since this will essentially filter out certain fields from response -- It will appear to whomever looks at those metacard responses that some attributes are simply missing. 


❓ 1: Have we ensured that this cannot be used maliciously to hide sensitive data or needed attributes? 

One example I checked (and ensured will not be a problem) is the `populateMetacards` method in delete/update operations classes. When we create this query to fetch the metacards (so that we can apply security policies to them) we do **not** forward the properties from the parent request. If we accidentally (or purposely) had forwarded the request properties then it *would have* been able to shadow out attributes in metacards given to the security policies. Along with this line of thought, have we double checked this wont be able to be misused?

❓ 2: Have we considered (or maybe do we even care) about the ramifications of clients removing critical data ( data we generally 'assume' will be there)? One example is metacard ID. I'm sure there are many other attributes as well that, if removed, would cause Exceptions throughout the framework. 

One possible solution/addition we may want to include anyways ✏️ : Is there Any attributes we want to Blacklist from exclusions? (eg Metacard ID) (aka attributes we always want to include if we are setting the field list)

rzwiefel:
Another example of critical fields we may want to blacklist from the exclusion list is the "private" solr fields that we always need to be there to reconstruct metacard type, etc.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
pklinef: We are indexing the tokenized and the un-tokenized version.  The un-tokenized version is indexed as a single token for equality searches.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
pklinef: My understanding is Java will use that new array if it is the correct size or larger.

https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#toArray(T[])

rzwiefel: It is slower because it must zero out the sized array and cannot optimize out that optimization

rzwiefel:
build now

clockard:
build now

pklinef:
Regarding the remaining unanswered questions from @rzwiefel, The excluded attributes query property is not exposed by any endpoints.  Since it is internal, I am not as concerned with adding the additional guard code at this time.

pklinef:
Looks like the build passed but Sonar failed.

> 00:28:58 [INFO] DDF :: Distribution :: DDF ......................... SUCCESS [01:46 min]
00:28:58 [INFO] DDF :: Integration :: Test ......................... SUCCESS [34:59 min]
00:28:58 [INFO] ------------------------------------------------------------------------
00:28:58 [INFO] BUILD SUCCESS
00:28:58 [INFO] ------------------------------------------------------------------------

>[ERROR] sonar-scanner-engine-shaded-6.6.0.30458.jar|61390bfaa81aaf8bdd213a18122bde19: Fail to download sonar-scanner-engine-shaded-6.6.0.30458.jar to /root/.sonar/cache/_tmp/fileCache3415727485403028624.tmp: timeout: Socket closed

clockard:
build now

gordocanchola:
Manual testing successfully hero'd

PR #2362, opened 2017-09-25
DDF-3315 - Add protection code for invalid config admin objects during migration export/import

paouelle:
#### What does this PR do?
Currently the export/import process for config admin objects relies on the pid and the factory pid properties to be properly defined. However, they can easily get out of sync in config admin if they do not get updated properly by client code. We actually seen in some weird cases these not set properly. As such, we should make sure to protect the code by ensuring to store the right pid and factory pid in the properties set we export and at import time, we should also protect ourselves against missing pid.
#### Who is reviewing it? 
@emanns95 
@Lambeaux 
@brjeter 
@coyotesqrl
@figliold
The only way to test the export would be to modify a config object directly at run-time. For the import, it would require tweaking the exported file by updating a config file where the pid could be removed. I would suggest just verifying that export and import still works.
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
@@ -14,10 +14,13 @@
 package org.codice.ddf.configuration.admin;
 
 import java.nio.file.Path;
+import java.util.Dictionary;
 import org.apache.commons.lang.Validate;
 import org.codice.ddf.configuration.persistence.PersistenceStrategy;
 import org.codice.ddf.migration.ExportMigrationEntry;
+import org.osgi.framework.Constants;
 import org.osgi.service.cm.Configuration;
+import org.osgi.service.cm.ConfigurationAdmin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -26,6 +29,7 @@ import org.slf4j.LoggerFactory;
  * object in memory.
  */
 public class ExportMigrationConfigurationAdminEntry {
+
   private static final Logger LOGGER =
       LoggerFactory.getLogger(ExportMigrationConfigurationAdminEntry.class);
 
@@ -55,7 +59,28 @@ public class ExportMigrationConfigurationAdminEntry {
     if (!stored) {
       LOGGER.debug(
           "Exporting configuration [{}] to [{}]...", configuration.getPid(), entry.getPath());
-      this.stored = entry.store((r, out) -> persister.write(out, configuration.getProperties()));
+      final Dictionary<String, Object> props = configuration.getProperties();
+
+      // just make sure the pid, bundle location, and factory pid are correct as sometimes they
+      // might
+      // get out of sync by an invalid update to the properties and we do rely on those at import
+      // time
+      final String factoryPid = configuration.getFactoryPid();
+
+      if (factoryPid != null) {
+        props.put(ConfigurationAdmin.SERVICE_FACTORYPID, factoryPid);
+      } else {
+        props.remove(ConfigurationAdmin.SERVICE_FACTORYPID);
+      }
+      props.put(Constants.SERVICE_PID, configuration.getPid());
+      final String location = configuration.getBundleLocation();
+
+      if (location != null) {
+        props.put(ConfigurationAdmin.SERVICE_BUNDLELOCATION, location);
+      } else {
+        props.remove(ConfigurationAdmin.SERVICE_BUNDLELOCATION);
+      }
+      this.stored = entry.store((r, out) -> persister.write(out, props));
     }
     return stored;
   }
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ImportMigrationConfigurationAdminContext.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ImportMigrationConfigurationAdminContext.java
@@ -53,6 +53,7 @@ import org.slf4j.LoggerFactory;
  * memory.
  */
 public class ImportMigrationConfigurationAdminContext {
+
   private static final Logger LOGGER =
       LoggerFactory.getLogger(ImportMigrationConfigurationAdminContext.class);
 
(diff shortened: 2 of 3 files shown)


paouelle:
build now

paouelle:
build now

paouelle:
build now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
```diff
@@ -55,7 +58,19 @@ public boolean store() {
...
       LOGGER.debug(
           "Exporting configuration [{}] to [{}]...", configuration.getPid(), entry.getPath());
-      this.stored = entry.store((r, out) -> persister.write(out, configuration.getProperties()));
+      final Dictionary<String, Object> props = configuration.getProperties();
+
+      // just make sure the pid and factory pid are correct as sometimes they might get out of
+      // sync by an invalid update to the properties and we do rely on those at import time
+      final String factoryPid = configuration.getFactoryPid();
+
+      if (factoryPid != null) {
```
lessarderic: ❓ Should we log something if we have to update the factoryPid?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
```diff
@@ -55,7 +58,19 @@ public boolean store() {
...
+        props.put(ConfigurationAdmin.SERVICE_FACTORYPID, factoryPid);
+      } else {
+        props.remove(ConfigurationAdmin.SERVICE_FACTORYPID);
+      }
+      props.put(Constants.SERVICE_PID, configuration.getPid());
```
lessarderic: ❓ Same question for the service PID.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
```diff
@@ -55,7 +58,19 @@ public boolean store() {
     if (!stored) {
       LOGGER.debug(
           "Exporting configuration [{}] to [{}]...", configuration.getPid(), entry.getPath());
-      this.stored = entry.store((r, out) -> persister.write(out, configuration.getProperties()));
+      final Dictionary<String, Object> props = configuration.getProperties();
```
lessarderic: ❓Instead of adding this logic, should we consider storing the content of the `Dictionary` as-is and store the factory PID, service PID and bundle location as separate attributes of the entry? This would keep all the values in-sync in case there was a good reason for the difference between what's in the `Dictionary` and what the methods return.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
```diff
@@ -55,7 +58,19 @@ public boolean store() {
...
+      // sync by an invalid update to the properties and we do rely on those at import time
+      final String factoryPid = configuration.getFactoryPid();
+
+      if (factoryPid != null) {
+        props.put(ConfigurationAdmin.SERVICE_FACTORYPID, factoryPid);
+      } else {
+        props.remove(ConfigurationAdmin.SERVICE_FACTORYPID);
+      }
+      props.put(Constants.SERVICE_PID, configuration.getPid());
+      this.stored = entry.store((r, out) -> persister.write(out, props));
```
lessarderic: ❗️We also need to handle the `ConfigurationAdmin.SERVICE_BUNDLELOCATION` / `getBundleLocation()` the same way here and in the import.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ImportMigrationConfigurationAdminContext.java
```diff
@@ -272,23 +272,33 @@ private ImportMigrationConfigurationAdminEntry proxy(ImportMigrationEntry entry)
...
+      // those later
+      final String pid = Objects.toString(properties.remove(Constants.SERVICE_PID), null);
+
+      if (pid == null) {
+        // this should never happen unless someone created the zip file manually and messed up the
+        // config file
+        context
+            .getReport()
+            .record(
+                new MigrationException("Import error: missing pid from configuration [%s].", path));
```
lessarderic: ❓ Do we have a unit test to make sure a `MigrationException` is recorded if the service PID is `null`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
paouelle: I don't think we need to add that complexity. The fact is that these 2 properties and the felix install file re never re-imported on purpose. This follows the paradigm already used by Felix in both the config installer and cellar. As such, it simplifies to simply re-use these properties like this.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
paouelle: not really. We end up not caring about these in the dictionary to start with. This is in sync with how felix's config installer works already where it actually removes these properties from the dictionary before saving back to etc or before re-importing from etc.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
paouelle: same answer

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ExportMigrationConfigurationAdminEntry.java
paouelle: nice! this was actually one of my question on the original PR. I didn't know wether or not the location should be restored. I will make these changes.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/admin/ImportMigrationConfigurationAdminContext.java
paouelle: The config admin migratable only unit test at a higher level by running import followed by export. Since this case would only ever happen if someone was to manually tweak the zip file, I am less concern. In any case, we are recording an error which will abort the import by design. Not too worried here.

paouelle:
build now

vinamartin:
Manually verified ✅

PR #2363, opened 2017-09-26
DDF-3317 Add better handling of Tika extraction errors

gordocanchola:
Adding a catch throwable around the Tika metadata extraction that logs the error
Changing the flow so that the transform will continue given a Tika extraction failure

#### What does this PR do?
This PR provides better Tika exception handling. 
A catch throwable is used to wrap the Tika extraction because it's hard to capture all of the possible Tika exceptions.
The flow was changed so that an exception in the Tika extraction will not prevent the transform.

#### Who is reviewing it? 
@dcruver @beyelerb @kcwire 
@jlcsmith
@pklinef

[DDF-3317](https://codice.atlassian.net/browse/DDF-3317)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/transformer/catalog-transformer-pptx/src/main/java/ddf/catalog/transformer/input/pptx/PptxInputTransformer.java
+++ b/catalog/transformer/catalog-transformer-pptx/src/main/java/ddf/catalog/transformer/input/pptx/PptxInputTransformer.java
@@ -29,6 +29,7 @@ import java.io.InputStream;
 import java.util.Objects;
 import org.apache.commons.io.IOUtils;
 import org.apache.pdfbox.tools.imageio.ImageIOUtil;
+import org.apache.poi.EncryptedDocumentException;
 import org.apache.poi.sl.usermodel.SlideShow;
 import org.apache.poi.sl.usermodel.SlideShowFactory;
 import org.apache.poi.xslf.usermodel.XMLSlideShow;
@@ -111,7 +112,11 @@ public class PptxInputTransformer implements InputTransformer {
       Metacard metacard =
           extractInitialMetadata(fileBackedOutputStream.asByteSource().openStream());
 
-      extractThumbnail(metacard, fileBackedOutputStream.asByteSource().openStream());
+      try {
+        extractThumbnail(metacard, fileBackedOutputStream.asByteSource().openStream());
+      } catch (EncryptedDocumentException e) {
+        LOGGER.debug("Unable to generate thumbnail", e);
+      }
       return metacard;
     }
   }
@@ -140,6 +145,7 @@ public class PptxInputTransformer implements InputTransformer {
    * @param metacard
    * @param input
    * @throws IOException
+   * @throws EncryptedDocumentException
    */
   private void extractThumbnail(Metacard metacard, InputStream input) throws IOException {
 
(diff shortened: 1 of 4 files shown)


catalog/transformer/catalog-transformer-tika-input/src/main/java/ddf/catalog/transformer/input/tika/TikaInputTransformer.java
```diff
@@ -351,46 +352,48 @@ public Metacard transform(InputStream input, String id)
...
-      String contentType;
+      Metacard metacard = new MetacardImpl(commonTikaMetacardType);
+      String contentType = DataType.DATASET.name();
       TikaMetadataExtractor extractor = null;
       try (InputStream inputStreamCopy = fileBackedOutputStream.asByteSource().openStream()) {
         extractor = new TikaMetadataExtractor(inputStreamCopy, previewMaxLength, metadataMaxLength);
 
-      } catch (TikaException e) {
-        throw new CatalogTransformerException(e);
+      } catch (Throwable t) {
```
paouelle: ❗️ We should not blindly catch all throwables like that. We should first do a catch(VirtualMachineError) and throw that back without even attempting to log.

paouelle:
❓ How about adding new unit tests to cover the new paths in the code?

catalog/transformer/catalog-transformer-tika-input/src/main/java/ddf/catalog/transformer/input/tika/TikaInputTransformer.java
lessarderic: Actually, why not just catch `RuntimeException` and any checked exceptions thrown by Tika?

paouelle: Definitely better! We should only resort to catching more when really required.

gordocanchola: Changed to catch checked exceptions and RuntimeException

gordocanchola:
Added test

gordocanchola:
build now

gordocanchola:
build now

gordocanchola:
build now

jlcsmith:
Successfully hero'd

catalog/transformer/catalog-transformer-pptx/src/main/java/ddf/catalog/transformer/input/pptx/PptxInputTransformer.java
```diff
@@ -111,7 +111,11 @@ private Metacard transformLogic(InputStream input)
       Metacard metacard =
           extractInitialMetadata(fileBackedOutputStream.asByteSource().openStream());
 
-      extractThumbnail(metacard, fileBackedOutputStream.asByteSource().openStream());
+      try {
+        extractThumbnail(metacard, fileBackedOutputStream.asByteSource().openStream());
+      } catch (IllegalStateException e) {
```
paouelle: Since the above code actually throws EncryptedDocumentException, I think we should reduce that catch down to that exception instead. We should also document in the javadoc for the extractThumbnail() method that it can throw EncryptedDocumentException.

gordocanchola:
build now

PR #2364, opened 2017-09-26
DDF-3318 Update URLResourceReader usages to include basic auth info

kcrowe01:
#### What does this PR do?
Update CSW and OpenSearch sources to include basic auth info in the retrieveResource properties map.
#### Who is reviewing it? 
@vinamartin @mcalcote 
@clockard
@rzwiefel
Install DDF and setup federation to query over OpenSearch and CSW endpoints. The DDF being queried should be setup to only allow basic authentication for the OpenSearch and CSW endpoints. Verify that a resource can be downloaded over both of these endpoints only when a valid user/pass is provided in the source setup. 
[DDF-3318](https://codice.atlassian.net/browse/DDF-3318)
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
+++ b/catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
@@ -738,7 +738,7 @@ public class OpenSearchSource implements FederatedSource, ConfiguredService {
     if (serializableId != null) {
       String metacardId = serializableId.toString();
       WebClient restClient = newRestClient(null, metacardId, true, null);
-      if (StringUtils.isNotBlank(username) && StringUtils.isNotBlank(password)) {
+      if (StringUtils.isNotBlank(username)) {
         requestProperties.put(USERNAME_PROPERTY, username);
         requestProperties.put(PASSWORD_PROPERTY, password);
       }
--- a/catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
+++ b/catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
@@ -117,6 +117,10 @@ public class OpenSearchSource implements FederatedSource, ConfiguredService {
 
   private static final String LOCAL_SEARCH_PARAMETER = "local";
 
+  private static final String USERNAME_PROPERTY = "username";
+
+  private static final String PASSWORD_PROPERTY = "password";
+
   private static final Logger LOGGER = LoggerFactory.getLogger(OpenSearchSource.class);
 
   private final EncryptionService encryptionService;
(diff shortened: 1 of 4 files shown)


kcrowe01:
build now

kcrowe01:
build now

kcrowe01:
build now

catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
```diff
@@ -117,6 +117,10 @@
 
   private static final String LOCAL_SEARCH_PARAMETER = "local";
 
+  private static final String USERNAME_PROPERTY = "username";
+
+  private static final String PASSWORD_PROPERTY = "password";
```
rzwiefel: (out of scope for this pr, but for the record) It would be a nice future improvement  to extract out the username/password string as an exported constant so that these changes and the previous change (https://codice.atlassian.net/browse/DDF-2547) wouldn't both have copies of the same two variables.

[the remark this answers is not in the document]
rzwiefel: ❗️ Like @mcalcote mentioned we should add a suppression for this.

catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
```diff
@@ -733,6 +737,10 @@ public ResourceResponse retrieveResource(URI uri, Map<String, Serializable> requ
     if (serializableId != null) {
       String metacardId = serializableId.toString();
       WebClient restClient = newRestClient(null, metacardId, true, null);
+      if (StringUtils.isNotBlank(username) && StringUtils.isNotBlank(password)) {
```
rzwiefel: ❓ Do we need to account for the case if the password is blank? (E.g, a nonblank username with blank password)

kcrowe01: Is a blank password valid? If so, I can account for this.

rzwiefel: Hmm 🤔 . I don't think its illegal to be blank depending on what the client is connecting to. maybe @stustison might know?

rzwiefel:
build now

catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
kcrowe01: @rzwiefel The URLResourceReader checks for the username and password properties being null before it uses them https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java#L552. I think it should be safe to just remove this if statement which will allow blank username/passwords. I believe this should also fix the SonarQube Cognitive Complexity error. What do you think?

kcrowe01:
build now

josephthweatt:
Built and verified manually with @garrettfreibott and @figliold. Everything works fine!

rzwiefel:
Also @kcrowe01 was there any unit tests we could add for this change?

kcrowe01:
build now

PR #2373, opened 2017-09-27
DDF-3297 Adds security manager to container

coyotesqrl:
#### What does this PR do?
Adds ProGrade security manager and policy file to container.

#### Who is reviewing it? 
@paouelle @Lambeaux @SmithJosh @garrettfreibott @ryeats 

@codice/security 

@clockard
@lessarderic
@stustison

Full regression test of DDF. This will potentially impact downstream projects; they will need to add/update policy files and `setenv` scripts accordingly.

N/A

[DDF-3297](https://codice.atlassian.net/browse/DDF-3297)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/bin/setenv
+++ b/distribution/ddf-common/src/main/resources/bin/setenv
@@ -33,7 +33,7 @@
 DIRNAME=`dirname "$0"`
 
 # Set environment variable for DDF home directory (used by DDF Java code)
-DDF_HOME=`cd $DIRNAME/..; pwd`
+DDF_HOME=`cd $DIRNAME/..; pwd -P`
 
 # NOTE: export on separate line to support caveat of Solaris OS
 export DDF_HOME
@@ -64,5 +64,22 @@ export EXTRA_JAVA_OPTS
 # export KARAF_DATA # Karaf data folder
 # export KARAF_BASE # Karaf base folder
 
-export KARAF_OPTS="-Dfile.encoding=UTF8 -Dddf.home=$DDF_HOME"
+#
+# Admins can uncomment the following line and comment out the other definition of
+# the KARAF_SYSTEM_OPTS variable in order to determine any missing security permissions
+# should they install third-party bundles requiring additional access.
+#
+# Please note that turning on the PolicyFileGeneratorJSM has the side-effect of turning off
+# security on your system. It should be used with caution, only to ascertain missing
+# permissions to be added to the default.policy file.
+#
+# N.B. The use of the double equals on the 'java.security.policy' property is intentional.
+# See http://docs.oracle.com/javase/7/docs/technotes/guides/security/PolicyFiles.html#DefaultLocs
+# for more information.
+#
+# export KARAF_SYSTEM_OPTS="-Dprograde.generated.policy=$DDF_HOME/generated.policy -Dprograde.use.own.policy=true -Dpolicy.provider=net.sourceforge.prograde.policy.ProGradePolicy -Djava.security.manager=net.sourceforge.prograde.sm.PolicyFileGeneratorJSM -Djava.security.policy==$DDF_HOME/security/default.policy -DproGrade.getPermissions.override=sun.rmi.server.LoaderHandler:loadClass,org.apache.jasper.compiler.JspRuntimeContext:initSecurity"
+
+export KARAF_SYSTEM_OPTS="-Dpolicy.provider=net.sourceforge.prograde.policy.ProGradePolicy -Djava.security.manager=net.sourceforge.prograde.sm.ProGradeJSM -Djava.security.policy==$DDF_HOME/security/default.policy -DproGrade.getPermissions.override=sun.rmi.server.LoaderHandler:loadClass,org.apache.jasper.compiler.JspRuntimeContext:initSecurity"
+
+export KARAF_OPTS="-Dfile.encoding=UTF8 -Dddf.home=$DDF_HOME -Dddf.home.policy=$DDF_HOME/"
 export JAVA_OPTS="-Xms2g -Xmx4g -XX:+UnlockDiagnosticVMOptions -XX:+DisableAttachMechanism"
--- a/distribution/ddf-common/src/main/resources/security/default.policy
+++ b/distribution/ddf-common/src/main/resources/security/default.policy
@@ -1,0 +1,169 @@
+priority "grant";
+
+deny {
+    // Read, write, or execute any file
+    permission java.io.FilePermission "<<ALL FILES>>", "read, write, execute";
+
+    // Deny deletion of the security policy file through any Java code
+    permission java.io.FilePermission "${ddf.home}${/}security${/}default.policy", "read, write, execute, delete";
+
+    permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
+    permission java.util.PropertyPermission "java.io.tmpdir", "write";
+    permission java.util.PropertyPermission "user.home", "write";
+
+    // Change current security manager
+    permission java.lang.RuntimePermission "setSecurityManager";
+
+    // Modify application permissions at will.
+    permission java.security.SecurityPermission "setPolicy, insertProvider, removeProvider.*, clearProviderProperties.*, putProviderProperty.*, removeProviderProperty.*";
+
+    // Load classes into any protection domain
+    permission java.lang.RuntimePermission "createClassLoader";
+};
+
+//
+// DDF Libraries have broad file permissions within the install directory
+//
+grant
+  codeBase "file:${ddf.home.policy}system/ddf/-" {
+    permission java.io.FilePermission "${ddf.home}", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}-", "read, write";
+
+    permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
+};
+grant
+  codeBase "file:${ddf.home.policy}system/org/codice/-" {
+    permission java.io.FilePermission "${ddf.home}", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}-", "read, write";
+
+    permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
+};
+
+//
+// DDF Migration Framework and Config Admin Migratable
+//
+grant
+  codeBase "file:${ddf.home.policy}system/ddf/platform/platform-migratable/-" {
+    // some permissions might be repeated here to ensure that they are granted to the migration
+    // framework if we ever tighten the permissions above
+
+    permission java.io.FilePermission "<<ALL FILES>>", "read, readlink";
+    permission java.io.FilePermission "${ddf.home}", "read, write";
+
+    permission java.util.PropertyPermission "*", "read";
+    permission java.util.PropertyPermission "ddf.home", "read";
+    permission java.util.PropertyPermission "karaf.restart.jvm", "write";
+};
+
+//
+// DDF Platform Migratable
+//
+grant
+  codeBase "file:${ddf.home.policy}system/ddf/platform/platform-migratable/-" {
+    // some permissions might be repeated here to ensure that they are granted to the migration
+    // framework if we ever tighten the permissions above
+
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}ws-security", "read";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}ws-security${/}-", "read, delete";
+
+    permission java.util.PropertyPermission "javax.net.ssl.keyStore", "read";
+    permission java.util.PropertyPermission "javax.net.ssl.trustStore", "read";
+};
+
+//
+// DDF Security Migratable
+//
+grant
+  codeBase "file:${ddf.home.policy}system/ddf/security/migration/security-migratable/-" {
+    // some permissions might be repeated here to ensure that they are granted to the migration
+    // framework if we ever tighten the permissions above
+
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}pdp", "read";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}pdp${/}-", "read, delete";
+};
+
+//
+// The lib directory contains root, trusted jars for DDF, OSGi, and security
+//
+grant
+  codeBase "file:${ddf.home.policy}lib/-" {
+    permission java.io.FilePermission "${ddf.home}", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}*", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}-", "read";
+
+    permission java.lang.RuntimePermission "createClassLoader";
+    permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
+};
+grant
+  codeBase "file:${ddf.home.policy}lib/boot/-" {
+    permission java.io.FilePermission "${ddf.home}${/}*", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}config.properties", "write";
+    permission java.util.PropertyPermission "karaf.restart.jvm", "write";
+};
+grant
+  codeBase "file:${ddf.home.policy}lib/boot/org.osgi.core-6.0.0.jar" {
+    permission java.io.FilePermission "${ddf.home}${/}*", "read, execute";
+    permission java.io.FilePermission "${ddf.home}${/}deploy${/}-", "read, execute";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}-", "read, execute";
+};
+
+//
+// Core OSGi libraries
+//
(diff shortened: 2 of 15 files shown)


distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,120 @@
...
+// todo symlinks
+
+deny {
+    // Read, write, or execute any file
+    permission java.io.FilePermission "<<ALL FILES>>", "read, write, execute";
+
+    // Deny deletion of the security policy file through any Java code
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}default.policy", "delete";
+
+    permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
```
paouelle: migration export/import needs read access to:

javax.net.ssl.keyStore
javax.net.ssl.trustStore

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,120 @@
+priority "grant";
+// todo symlinks
+
+deny {
+    // Read, write, or execute any file
+    permission java.io.FilePermission "<<ALL FILES>>", "read, write, execute";
+
+    // Deny deletion of the security policy file through any Java code
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}default.policy", "delete";
```
paouelle: Should we block write too?
import will need write access though to restore that file.

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,120 @@
+priority "grant";
+// todo symlinks
+
+deny {
+    // Read, write, or execute any file
+    permission java.io.FilePermission "<<ALL FILES>>", "read, write, execute";
```
paouelle: Should also be blocking delete?

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,120 @@
...
+    permission java.io.FilePermission "${java.home}${/}lib${/}-", "read";
+
+    // System libraries
+    permission java.io.FilePermission "/dev/urandom", "read";
+    permission java.io.FilePermission "/lib", "read";
+    permission java.io.FilePermission "/proc/self/exe", "read";
+    permission java.io.FilePermission "/usr/lib", "read";
+
+
+    // Distribution File Permissions
```
figliold: @paouelle Wouldn't the export/import need access to the directory the exported zip was saved in?

paouelle: and a lot more. I am compiling the required changes in a separate spike.

distribution/ddf-common/src/main/resources/bin/setenv
```diff
@@ -64,5 +64,14 @@ export EXTRA_JAVA_OPTS
...
 
-export KARAF_OPTS="-Dfile.encoding=UTF8 -Dddf.home=$DDF_HOME"
+#
+# Admins can uncomment the following line and comment out the other definition of
+# the KARAF_SYSTEM_OPTS variable in order to determine any missing security permissions
+# should they install third-party bundles requiring additional access.
+#
+# export KARAF_SYSTEM_OPTS="-Dprograde.generated.policy=$DDF_HOME/generated.policy -Dprograde.use.own.policy=true -Djava.security.manager=net.sourceforge.prograde.sm.PolicyFileGeneratorJSM -Djava.security.policy==$DDF_HOME/etc/default.policy -DproGrade.getPermissions.override=sun.rmi.server.LoaderHandler:loadClass,org.apache.jasper.compiler.JspRuntimeContext:initSecurity"
+
+export KARAF_SYSTEM_OPTS="-Djava.security.manager=net.sourceforge.prograde.sm.ProGradeJSM -Djava.security.policy==$DDF_HOME/etc/default.policy -DproGrade.getPermissions.override=sun.rmi.server.LoaderHandler:loadClass,org.apache.jasper.compiler.JspRuntimeContext:initSecurity"
```
coyotesqrl: Using `KARAF_SYSTEM_OPTS` here and in the Windows script in order to provide differentiation between `client` and `ddf` scripts. There are other ways we could handle the different needs - the client script does _not_ need a security manager - but they would require managing more files separate from the Karaf distro.

distribution/ddf-common/src/main/resources/etc/default.policy
coyotesqrl: It should, and future iterations (likely beyond 2.11.0) will. Could also consider blocking `readLink`, but that's of lower concern.

distribution/ddf-common/src/main/resources/etc/default.policy
coyotesqrl: The permission line above this is already globally blocking read, write, and execute.

distribution/ddf-common/src/main/resources/etc/default.policy
coyotesqrl: Let me know which bundle(s) need those permissions and I'll grant them.

paouelle: It's a place holder for now. I will compile all the requirements for export/import separatly

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,111 @@
...
+    // Java libraries
+    permission java.io.FilePermission "${java.home}${/}release", "read";
+    permission java.io.FilePermission "${java.home}${/}lib${/}-", "read";
+
+    // System libraries
+    permission java.io.FilePermission "/dev/urandom", "read";
+    permission java.io.FilePermission "/lib", "read";
+    permission java.io.FilePermission "/proc/self/exe", "read";
+    permission java.io.FilePermission "/usr/lib", "read";
+
```
lessarderic: ❓ How about temporary folders, e.g., `/tmp`, `/var/tmp`, `C:\TEMP`, etc.? The `FileBackedOutputStream` class (which is used in quite a few places in our code) needs to be able to create temporary files: https://github.com/google/guava/blob/master/guava/src/com/google/common/io/FileBackedOutputStream.java#L195.

coyotesqrl: We will need to test this across platforms, but it is almost definitely broken right now.

distribution/ddf-common/src/main/resources/etc/default.policy
coyotesqrl: We do grant many `codeBase`s access to `etc`, but it's _mostly_ restricted. We will lock it down more, but this should be sufficiently safe for now.

paouelle:
We need to update the documentation with respect to service creation since the changes to setenv won't propagate there. We should also re-test the security manager when running as a service

distribution/ddf-common/src/main/resources/etc/default.policy
paouelle: Can you add the following permissions for export/import?

//
// DDF Migration Framework and Config Admin Migratable
//
grant
  codeBase "file:${ddf.home.policy}system/ddf/platform/platform-migratable/-" {

    permission java.io.FilePermission "<<ALL FILES>>", "read, readlink";
    permission java.io.FilePermission "${ddf.home}", "read, write";

    permission java.util.PropertyPermission "*", "read";
    permission java.util.PropertyPermission "ddf.home", "read";
    permission java.util.PropertyPermission "karaf.restart.jvm", "write";
};

//
// DDF Platform Migratable
//
grant
  codeBase "file:${ddf.home.policy}system/ddf/platform/platform-migratable/-" {

    permission java.io.FilePermission "${ddf.home}${/}etc${/}ws-security", "read";
    permission java.io.FilePermission "${ddf.home}${/}etc${/}ws-security${/}-", "read, delete";

    permission java.util.PropertyPermission "javax.net.ssl.keyStore", "read";
    permission java.util.PropertyPermission "javax.net.ssl.trustStore", "read";
};

//
// DDF Security Migratable
//
grant
  codeBase "file:${ddf.home.policy}system/ddf/security/migration/security-migratable/-" {

    permission java.io.FilePermission "${ddf.home}${/}etc${/}pdp", "read";
    permission java.io.FilePermission "${ddf.home}${/}etc${/}pdp{/}-", "read, delete";
};

coyotesqrl: Will do, @paouelle

coyotesqrl:
build now

rzwiefel:
@coyotesqrl This PR is marked with `2.11.0` but is pointed into codice:master.. Is this PR supposed to be opened against `codice:2.11.x` instead?

coyotesqrl:
build now

coyotesqrl:
@rzwiefel It should have the milestone removed. It was intended to be merged prior to the branch; a backport will be put up after this is approved and merged.

distribution/ddf-common/src/main/resources/etc/default.policy
lessarderic: Even though things worked because `FileBackedOutStream` defaults to a directory under `ddf.home`, we may still want to grant read access to the `java.io.tmpdir` system property and allow read/write access to the directory it points to. This would prevent any unexpected issues if a third-party library decides to create some temporary files there.

coyotesqrl:
@lessarderic I should have just looked to begin with. The `karaf` and `karaf.bat` scripts set `java.io.tmpdir` to `${KARAF_DATA}/tmp`. I'll explicitly add that path to the global grants, but it should already be covered.

coyotesqrl:
build now

coyotesqrl:
Build now

coyotesqrl:
Build now

distribution/ddf-common/src/main/resources/etc/default.policy
paouelle: don't forget delete too

coyotesqrl: Thanks, @paouelle. Fixed.

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,157 @@
...
+    permission java.lang.RuntimePermission "createClassLoader";
+};
+
+//
+// DDF Libraries have broad file permissions within the install directory
+//
+grant
+  codeBase "file:${ddf.home.policy}system/ddf/-" {
+    permission java.io.FilePermission "${ddf.home}", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}etc${/}-", "read, write";
```
paouelle: Felix config installer (and Steve's replacement) will also need delete for .cfg and .config files located under etc (not recursive)

distribution/ddf-common/src/main/resources/etc/default.policy
```diff
@@ -0,0 +1,157 @@
...
+
+    // Java libraries
+    permission java.io.FilePermission "${java.home}${/}release", "read";
+    permission java.io.FilePermission "${java.home}${/}lib${/}-", "read";
+
+    // System libraries
+    permission java.io.FilePermission "/dev/urandom", "read";
+    permission java.io.FilePermission "/lib", "read";
+    permission java.io.FilePermission "/proc/self/exe", "read";
+    permission java.io.FilePermission "/usr/lib", "read";
```
stustison: Are there any Windows equivalents for these that we need to worry about?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
```diff
@@ -585,7 +585,12 @@ private static Integer findPortNumber(Integer portToTry) {
         vmOption("-Xmx2048M"),
         // avoid integration tests stealing focus on OS X
         vmOption("-Djava.awt.headless=true"),
-        vmOption("-Dfile.encoding=UTF8"));
+        vmOption("-Dfile.encoding=UTF8"),
+        vmOption("-Djava.security.manager=net.sourceforge.prograde.sm.ProGradeJSM"),
+        vmOption("-Dpolicy.provider=net.sourceforge.prograde.policy.ProGradePolicy"),
+        vmOption("-Djava.security.policy==etc/all.policy"),
```
stustison: I'd feel better if this was default.policy + extra permissions for itests

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -6,7 +6,7 @@ deny {
     permission java.io.FilePermission "<<ALL FILES>>", "read, write, execute";
 
     // Deny deletion of the security policy file through any Java code
```
paouelle: ✏️  might want to update the comment to reflect what we are denying

coyotesqrl:
build now

paouelle:
re-tested latest code (i.e. default.policy) with PR 2409 (i.e. export/import changes for security manager) and it is 👍

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/catalog/TestFederation.java
```diff
@@ -703,54 +703,6 @@ public void testFederatedRetrieveProductInvalidResourceUrl() throws Exception {
         .body(containsString("Unable to transform Metacard."));
   }
 
-  /**
```
coyotesqrl: The two tests removed from this class no longer make sense. They fail early because of permissions that protect the system and there's no reason for the tests to continue.

coyotesqrl:
build now

distribution/ddf-common/src/main/resources/bin/setenv
```diff
@@ -64,5 +64,18 @@ export EXTRA_JAVA_OPTS
 # export KARAF_DATA # Karaf data folder
 # export KARAF_BASE # Karaf base folder
 
-export KARAF_OPTS="-Dfile.encoding=UTF8 -Dddf.home=$DDF_HOME"
+#
+# Admins can uncomment the following line and comment out the other definition of
+# the KARAF_SYSTEM_OPTS variable in order to determine any missing security permissions
+# should they install third-party bundles requiring additional access.
```
paouelle: ❓ If uncommenting this line grants everything by default, we might want to add a comment to that regard

coyotesqrl: Good point. It is a risk if left on. I'll expand the comment here and in the windows script.

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
```diff
@@ -794,4 +821,33 @@ protected boolean isCatalogEmpty() {
       return false;
     }
   }
+
+  /**
+   * Helper Option class to allow interpolation of {@code karaf.home} directory based on the
+   * provided {@link #UNPACK_DIRECTORY}.
+   */
+  static class HomeAwareVmOption extends VMOption {
```
tbatie: Can we leave a comment mentioning why this was made and the edge case it's addressing? I don't anyone else should really ever be using this.

paouelle:
❓ Should we undo the PR that disabled SSH as part of this one?

paouelle:
❗️ The security manager stuff will be problematic when running as a service. Checkout [My application is throwing security exceptions when I run with the Wrapper.](https://wrapper.tanukisoftware.com/doc/english/faq.html#5)

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
coyotesqrl: Considering it's itest code and does have a short javadoc, I don't know we need too much more. I doubt anyone would duplicate this approach.

coyotesqrl:
@paouelle The SSH change is going to be separate work that follows this. It will also include shutting off permissions to some Felix console commands.

coyotesqrl:
@paouelle We'll need to run as a service on Windows and Linux and determine how the security manager interplays. I would probably be fine granting the wrapper all permissions such that our codeBases would get the intersection of their restricted permissions and all permissions.

paouelle:
@coyotesqrl the code base for the wrapper is lib/wrapper.

coyotesqrl:
build now

coyotesqrl:
build now

oconnormi:
Build Now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

coyotesqrl:
build now

oconnormi:
:tada:

PR #2379, opened 2017-09-28
DDF-3323 - Adds ability for admin to configure the map home in Intrigue

andrewkfiedler:
- Adds configuration for map home to Intrigue, allowing admins to customize it.
 - Adds map home button to openlayers, removes the default map home button from cesium and replaces it with our own for consistency.
 - Default map home is a rough bounding box of the USA.
 - The help for map home specifies it needs to be in a particular order, but really it works so long as a series of coordinates in lon, lat order are supplied.

@jlcsmith 
@djblue 
@kcwire

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/visualization/maps/map.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/visualization/maps/map.view.js
@@ -27,12 +27,11 @@ var LocationModel = require('component/location-old/location-old');
 var user = require('component/singletons/user-instance');
 var LayersDropdown = require('component/dropdown/layers/dropdown.layers.view');
 var DropdownModel = require('component/dropdown/dropdown');
-
-var DropdownModel = require('component/dropdown/dropdown');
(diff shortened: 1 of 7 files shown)


jlcsmith:
build now

catalog/ui/catalog-ui-search/src/main/webapp/component/visualization/maps/map.view.js
```diff
@@ -41,6 +40,89 @@ function wrapNum(x, range) {
     return ((x - min) % d + d) % d + min;
 }
 
+function findExtreme({objArray, property, comparator}) {
+    if (objArray.length === 0) {
+        return undefined;
+    }
+    return objArray.reduce((extreme, coordinateObj) => {
+        extreme = comparator(extreme, coordinateObj[property]);
```
djblue: Do we need to assign the result before returning it? Also, if we remove the assignment it would be a single line, and we could simply do:

`(extreme, coordinateObj) => comparator(extreme, coordinateObj[property])`

PR #2396, opened 2017-10-02
DDF-3324 Catch VirtualMachineError before Throwable

jhunzik:
#### What does this PR do?
Fix dangerous condition where Throwable is caught by catching VirtualMachineErrors beforehand and rethrowing them so that important JVM errors that require action are not hidden
#### Who is reviewing it? @paouelle @figliold @tbatie @djblue @emanns95 
@coyotesqrl
@figliold 
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
+++ b/platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
@@ -62,8 +62,10 @@ public class ConfigurationSecurityLogger implements SynchronousConfigurationList
                   SecurityLogger.auditWarn(
                       "Configuration {} via filesystem: {}", type, updatedConfiguration);
                 }
+              } catch (VirtualMachineError vme) {
+                throw vme;
               } catch (Throwable e) {
-                LOGGER.error("Error auditing config update for {}", event.getPid(), e);
+                LOGGER.error("Error auditing config update for " + event.getPid(), e);
                 SecurityLogger.audit("Error auditing config update for {}", event.getPid());
               }
 
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheBulkProcessor.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheBulkProcessor.java
@@ -86,6 +86,8 @@ public class CacheBulkProcessor {
 
                 lastBulkAdd = new Date();
               }
+            } catch (VirtualMachineError vme) {
+              throw vme;
             } catch (Throwable throwable) {
               LOGGER.warn("Scheduled bulk ingest to cache failed", throwable);
             }
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CachingFederationStrategy.java
@@ -38,6 +38,7 @@ import ddf.catalog.plugin.PreFederatedQueryPlugin;
 import ddf.catalog.plugin.StopProcessingException;
 import ddf.catalog.source.CatalogProvider;
 import ddf.catalog.source.Source;
+import ddf.catalog.source.UnsupportedQueryException;
 import ddf.catalog.util.impl.RelevanceResultComparator;
 import ddf.catalog.util.impl.Requests;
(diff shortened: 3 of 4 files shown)


platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
```diff
@@ -56,6 +56,8 @@ public void configurationEvent(ConfigurationEvent event) {
         // there was no subject change was caused by an update to the config file on the filesystem
         SecurityLogger.auditWarn("Configuration {} via filesystem: {}", type, updatedConfiguration);
       }
+    } catch (VirtualMachineError vme) {
+      throw vme;
     } catch (Throwable e) {
       LOGGER.error("Error auditing config update for {}", event.getPid(), e);
```
paouelle: ✏️  I know it is outside the scope of this PR, but the error log will not be logging the exception here. Replace with: ```LOGGER.error("Error auditing config update for " + event.getPid(), e);```

jhunzik: Referencing Lombardi's comment on the ticket, should we try to be more specific and catch OutOfMemory, StackOverflow, and Thread Death as well?

paouelle: After discussing this with @lessarderic, @pklinef, and @figliold we concluded in only catching VirtualMachineError and throwing these back without doing anything or swallowing them.

jhunzik: Ah ok, I didn't realize this was the **only** error that needed to be caught.

platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
```diff
@@ -56,9 +56,11 @@ public void configurationEvent(ConfigurationEvent event) {
         // there was no subject change was caused by an update to the config file on the filesystem
         SecurityLogger.auditWarn("Configuration {} via filesystem: {}", type, updatedConfiguration);
       }
+    } catch (VirtualMachineError vme) {
+      throw vme;
     } catch (Throwable e) {
-      LOGGER.error("Error auditing config update for {}", event.getPid(), e);
-      SecurityLogger.audit("Error auditing config update for {}", event.getPid());
+      LOGGER.error("Error auditing config update for " + event.getPid(), e);
+      SecurityLogger.audit("Error auditing config update for ", event.getPid());
```
coyotesqrl: ❗️ Line 62 is correct with your change, as you can't intermix interpolation params and exceptions. However, your change to line 63 is incorrect as you've removed the `{}` and not changed to string concatenation. I would probably just leave the `SecurityLogger` call as it was and only fix the `LOGGER` call. 

If the inconsistency worries you (that someone may misread in the future and break it), you can always leave a short in-line comment explaining the difference.

jhunzik: Oops, fixed!

jhunzik:
build now

jhunzik:
build now

platform/security/core/security-core-impl/src/main/java/ddf/security/config/impl/ConfigurationSecurityLogger.java
```diff
@@ -56,8 +56,10 @@ public void configurationEvent(ConfigurationEvent event) {
         // there was no subject change was caused by an update to the config file on the filesystem
         SecurityLogger.auditWarn("Configuration {} via filesystem: {}", type, updatedConfiguration);
       }
+    } catch (VirtualMachineError vme) {
+      throw vme;
     } catch (Throwable e) {
-      LOGGER.error("Error auditing config update for {}", event.getPid(), e);
+      LOGGER.error("Error auditing config update for " + event.getPid(), e);
```
shaundmorris: ❓ why was this change made, the log format seemed correct for slf4j?

paouelle: because it wasn't logging the error. The prototype for the error() method that accepts a throwable doesn't accept a format message and parameters

jhunzik:
build now

jhunzik:
build now

PR #2400, opened 2017-10-02
[2.11.x] DDF-3279 Updates installer to force restart even if no changes are made

vinamartin:
#### What does this PR do?
Updates installer UI to suggest the admin to restart even though no changes have been made.

#### Who is reviewing it? 
@garrettfreibott @clockard @djblue @tbatie 

@figliold
@lessarderic

Run through the installer without making any changes.
Verify the shutdown button shuts down the system.
Verify the restart button restarts the system. 

This is related to a potential issue with lock files in Solr.

[DDF-3279](https://codice.atlassian.net/browse/DDF-3279)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/platform/admin/ui/src/main/webapp/templates/installer/finish.handlebars
+++ b/platform/admin/ui/src/main/webapp/templates/installer/finish.handlebars
@@ -26,14 +26,12 @@
             <br/>
             <br/>
             A system restart is required each time these files are modified in order for the changes to take effect.
-            {{#if modified}}
                 <br/>
                 <br/>
                 Click the 'Restart Now' button below to conclude the setup and restart the system.
                 <br/>
                 <br/>
-                Click the 'Restart Later' button below to conclude the setup. A system restart will be needed for changes to take effect.
-            {{/if}}
+                Click the 'Shutdown Now" button below to shutdown now. A manual restart will be required to use the system.
         </p>
     </div>
 </div>
--- a/platform/admin/ui/src/main/webapp/templates/installer/navigationButtons.handlebars
+++ b/platform/admin/ui/src/main/webapp/templates/installer/navigationButtons.handlebars
@@ -42,20 +42,12 @@
     {{/ifNotAnd}}
 {{else}}
     <li>
-        {{#if modified}}
-            {{#if busy}}
-                <a id="shutdownLaterStep" class="disabled-next" href="#">Restart Later</a>
-                <a id="shutdownNowStep" class="disabled-next" href="#">Restart Now</a>
-            {{else}}
-                <a id="shutdownLaterStep" class="next shutdown-later" href="#">Restart Later</a>
-                <a id="shutdownNowStep" class="next shutdown-now" href="#">Restart Now</a>
-            {{/if}}
+        {{#if busy}}
+            <a id="shutdownNowStep" class="disabled-next" href="#">Shutdown Now</a>
+            <a id="restartNowStep" class="disabled-next" href="#">Restart Now</a>
         {{else}}
-            {{#if busy}}
-                <a id="finishStep" class="disabled-next" href="#">Finish</a>
-            {{else}}
-                <a id="finishStep" class="next finish" href="#">Finish</a>
-            {{/if}}
+            <a id="shutdownNowStep" class="next shutdown-now" href="#">Shutdown Now</a>
+            <a id="restartNowStep" class="next restart-now" href="#">Restart Now</a>
         {{/if}}
     </li>
 {{/if}}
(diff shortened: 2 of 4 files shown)


clockard:
It sounds like the intent is to force a restart but really this pr will just provide the option to in the installer. The user could still click restart later.

vinamartin:
@clockard My wording wasn't super clear here. 
I can see value in users still being allowed to restart later if they need to configure some more stuff that may require restart. 
Otherwise I can remove the other button.

clockard:
That's fine. Just wanted to make sure that was the intent.

platform/admin/ui/src/main/webapp/templates/installer/finish.handlebars
```diff
@@ -26,14 +26,12 @@
             <br/>
             <br/>
```
figliold: ❓ Is a restart required all of the time now? This makes is sound like a restart is only required if something is modified.

platform/admin/ui/src/main/webapp/templates/installer/finish.handlebars
```diff
@@ -26,14 +26,12 @@
...
             A system restart is required each time these files are modified in order for the changes to take effect.
-            {{#if modified}}
                 <br/>
                 <br/>
```
figliold: ❓ I wonder if we force the restart "now" since solr is in a corrupt state until the system is restarted.

lessarderic: Good point. We may just want to keep a single `Restart` button.

platform/admin/ui/src/main/webapp/templates/installer/navigationButtons.handlebars
```diff
@@ -42,20 +42,12 @@
...
-        {{#if modified}}
-            {{#if busy}}
-                <a id="shutdownLaterStep" class="disabled-next" href="#">Restart Later</a>
-                <a id="shutdownNowStep" class="disabled-next" href="#">Restart Now</a>
-            {{else}}
-                <a id="shutdownLaterStep" class="next shutdown-later" href="#">Restart Later</a>
-                <a id="shutdownNowStep" class="next shutdown-now" href="#">Restart Now</a>
-            {{/if}}
+        {{#if busy}}
+            <a id="shutdownLaterStep" class="disabled-next" href="#">Restart Later</a>
```
lessarderic: ❗️Same comment as above.

vinamartin:
Because a user may still want to configure the system (in a way that would require restart), I'm going to change the wording of the message to warn them they **must** restart for the system to work properly.

vinamartin:
The new functionality is to either Shutdown or Restart.

clockard:
Wonder if we can just remove the shutdown option and only have a restart now button. What is the use case for shutting down now without restarting now as well?

vinamartin:
@clockard The thought is that if a user wants to configure something in a config file after installation, one may need an additional restart. 
There may be other corner cases such in which someone may want to shutdown instead of going live right away.

platform/admin/ui/src/main/webapp/templates/installer/navigationButtons.handlebars
```diff
@@ -42,20 +42,12 @@
...
+        {{#if busy}}
+            <a id="shutdownLaterStep" class="disabled-next" href="#">Shutdown Now</a>
+            <a id="shutdownNowStep" class="disabled-next" href="#">Restart Now</a>
         {{else}}
-            {{#if busy}}
-                <a id="finishStep" class="disabled-next" href="#">Finish</a>
-            {{else}}
-                <a id="finishStep" class="next finish" href="#">Finish</a>
-            {{/if}}
+            <a id="shutdownLaterStep" class="next shutdown-later" href="#">Shutdown Now</a>
```
lessarderic: ❗️ The `id` and `class` for these two elements don't match with the actual text, which may cause confusion and mislead people making changes to the css, JavaScript or tests in the future. We should consider using `id="shutdownNowStep" class="next shutdown-now"` and `id="restartNowStep" class="next restart-now"` respectively.

vinamartin: I'm in agreement here, thanks for catching this!

vinamartin:
@figliold , @lessarderic addressed your concerns.

vinamartin:
build now

rzwiefel:
Heroed. Both buttons work as expected.

PR #2406, opened 2017-10-03
[2.11.x] DDF-3323 - Adds ability for admin to configure the map home in Intrigue

andrewkfiedler:
- Adds configuration for map home to Intrigue, allowing admins to customize it.
 - Adds map home button to openlayers, removes the default map home button from cesium and replaces it with our own for consistency.
 - Default map home is a rough bounding box of the USA.
 - The help for map home specifies it needs to be in a particular order, but really it works so long as a series of coordinates in lon, lat order are supplied.

@djblue 
@jlcsmith
@clockard


jlcsmith:
build now

PR #2408, opened 2017-10-03
DDF-3261 Updated nsp Static Analysis stage to only scan projects using browserify or webpack

oconnormi:
#### What does this PR do?

Changes the node security scan to only scan projects that are actually packaging dependencies

#### Who is reviewing it? 

@mcalcote  @djblue 

@codice/build 

@clockard @shaundmorris 

#### Review Comment Legend:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -152,12 +152,8 @@ pipeline {
                                         def packageFile = readJSON file: 'package.json'
                                         if (packageFile.scripts =~ /.*webpack.*/ || packageFile.containsKey("browserify")) {
                                             nodejs(configId: 'npmrc-default', nodeJSInstallationName: 'nodejs') {
-                                                try {
-                                                    echo "Scanning ${packageFiles[i].path}"
-                                                    sh 'nsp check'
-                                                } catch (err) {
-                                                    echo "Failed for ${packageFiles[i].path}"
-                                                }
+                                                echo "Scanning ${packageFiles[i].path}"
+                                                sh 'nsp check'
                                             }
                                         }
                                     }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -149,9 +149,16 @@ pipeline {
...
-                                        nodejs(configId: 'npmrc-default', nodeJSInstallationName: 'nodejs') {
-                                            sh 'nsp check'
+                                        def packageFile = readJSON file: 'package.json'
+                                        if (packageFile.scripts =~ /.*webpack.*/ || packageFile.containsKey("browserify")) {
+                                            nodejs(configId: 'npmrc-default', nodeJSInstallationName: 'nodejs') {
+                                                try {
+                                                    echo "Scanning ${packageFiles[i].path}"
+                                                    sh 'nsp check'
+                                                } catch (err) {
+                                                    echo "Failed for ${packageFiles[i].path}"
```
clockard: If we catch the error will this still fail the build?

oconnormi: It will not fail the build this way. The reason I'm proposing we add this is because nsp does not currently seem to have a way to suppress vulnerabilities. The only vulnerabilities that are being reported at the moment are from a transitive that should actually have been listed as a non-production dependency. Without a way to suppress this there is no way to gracefully handle incorrectly declared production dependencies inherited transitively

oconnormi: Strike that, it actually describes how to do it in the nsp tools readme, there were just no docs on the website... I'll remove the catch statement and instead add an exception for the library in question

clockard: We can add suppressions via an .nsprc file for the module in question. We currently do this for the catalog-ui-search and admin ui. https://github.com/nodesecurity/nsp

oconnormi: yep, just found that when I double checked 👍

clockard:
This is preventing the build from completing successfully. Merging now.

PR #2411, opened 2017-10-03
DDF-3338 upgrade maven-dependency-plugin to 3.0.2

oconnormi:
#### What does this PR do?

Upgrades maven-dependency-plugin to 3.0.2

#### Who is reviewing it? 

@mcalcote @LinkMJB 

@codice/build 

@lessarderic 
@shaundmorris 

[DDF-3338](https://codice.atlassian.net/browse/DDF-3338)

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-federationstrategy/pom.xml
+++ b/catalog/core/catalog-core-federationstrategy/pom.xml
@@ -48,7 +48,10 @@
         <dependency>
             <groupId>ddf.catalog.core</groupId>
             <artifactId>catalog-core-api-impl</artifactId>
-            <type>test-jar</type>
+            <type>jar</type>
+            <!-- Bug in dependency-plugin 3.0.2 requires using classifier for test-jars instead of type.
+                See: https://issues.apache.org/jira/browse/MDEP-587 -->
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/catalog/core/catalog-core-solr/pom.xml
+++ b/catalog/core/catalog-core-solr/pom.xml
@@ -131,7 +131,10 @@
             <groupId>ddf.platform.solr</groupId>
             <artifactId>platform-solr-server-standalone</artifactId>
             <version>${project.version}</version>
-            <type>test-jar</type>
+            <type>jar</type>
+            <!-- Bug in dependency-plugin 3.0.2 requires using classifier for test-jars instead of type.
+                See: https://issues.apache.org/jira/browse/MDEP-587 -->
+            <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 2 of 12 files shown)


oconnormi:
build now

oconnormi:
build now

oconnormi:
Seems like this version might hurt more than help, already got two build failures with issues resolving/downloading dependencies

LinkMJB:
@oconnormi it looks like those dependency resolution failures were legitimate, it was the same dependency that failed in both builds. The flag that led me looking deeper into it, is that BFA didn't identify it as a Nexus issue, meaning that the artifact was actually missing rather than it timed-out resolving. AFAICT, it looks like the dependency is available now (Nexus negative cache invalidation timing?)

oconnormi:
@LinkMJB well in that case we can try again 😄 
Build Now

oconnormi:
@LinkMJB looks like its still having problems finding certain dependencies. I'll have to look into it more

LinkMJB:
@oconnormi I think that's the same dependency it was failing to find before 😢

oconnormi:
build now

oconnormi:
I'm starting to think there is a bug in this version of the maven-dependency-plugin. It continuously fails to download the same artifact each time the build is run. The error message it provides is `Unable to find/resolve artifact. Could not find artifact ddf.platform.solr:platform-solr-server-standalone:jar:2.12.0-SNAPSHOT` when building the `catalog-core-solr` module. The pom for that module actually depends on a `test-jar` not a `jar` artifact from the `platform-solr-server-standalone` module. The `platform-solr-server-standalone` module declares a main artifact of type `war` and its build does not produce a regular jar or attach one anywhere to the deploy. The only artifacts created by this modules build are the war, some xml files, and the test-jar. I think this version of the dependency plugin is either incorrectly looking for a jar instead of a test-jar, or it is just expecting that a jar (which is the default packaging type) exists for each dependency even if it is not the requested artifact.

oconnormi:
I can confirm there is a bug in dependency plugin 3.0.2. It does is not correctly translating the type `test-jar` to the real values of `type:jar`, `classifier:tests` the way it used to. instead it is just getting the jar artifact. This would result in always getting the incorrect artifact when requesting test-jars, since it would always grab the production jar instead.

oconnormi:
Created a bug for this: https://issues.apache.org/jira/browse/MDEP-587, but I'm thinking we won't be upgrading to this version 😄

oconnormi:
Added a workaround for the bug I linked above

Build Now

catalog/core/catalog-core-federationstrategy/pom.xml
```diff
@@ -48,7 +48,10 @@
         <dependency>
             <groupId>ddf.catalog.core</groupId>
             <artifactId>catalog-core-api-impl</artifactId>
-            <type>test-jar</type>
+            <type>jar</type>
+            <!-- Bug in dependency-plugin 3.0.2 requires using classifier for test-jars instead of type.
```
shaundmorris: thanks for adding the comments in here

oconnormi: its the type of thing that I myself would have changed back if I didn't see a comment like that 😄

oconnormi:
@shaundmorris I found them all by running `ag '<type>test-jar'` from the root of the project

catalog/core/catalog-core-federationstrategy/pom.xml
shaundmorris: yes, I would have forgotten by then

oconnormi:
not sure why those tests failed, trying again.

Build Now

PR #2413, opened 2017-10-04
[2.11.x] DDF-3317 Add better handling of Tika extraction errors

gordocanchola:
2.11.x backport of PR https://github.com/codice/ddf/pull/2363

Adding a catch throwable around the Tika metadata extraction that logs the error
Changing the flow so that the transform will continue given a Tika extraction failure
Changing the PPTX input transformer to not fail ingest on an exception when generating the thumbnail

@dcruver 
@pklinef 
@jlcsmith


gordocanchola:
build nocw

gordocanchola:
build now

PR #2435, opened 2017-10-06
[2.11.x] DDF-3315 - Add protection code for invalid config admin objects during migration export/import

paouelle:
Master [PR 2362](https://github.com/codice/ddf/pull/2362)

#### What does this PR do?
Currently the export/import process for config admin objects relies on the pid and the factory pid properties to be properly defined. However, they can easily get out of sync in config admin if they do not get updated properly by client code. We actually seen in some weird cases these not set properly. As such, we should make sure to protect the code by ensuring to store the right pid and factory pid in the properties set we export and at import time, we should also protect ourselves against missing pid.
@figliold
The only way to test the export would be to modify a config object directly at run-time. For the import, it would require tweaking the exported file by updating a config file where the pid could be removed. I would suggest just verifying that export and import still works.
[DDF-3215](https://codice.atlassian.net/browse/DDF-3215)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:


paouelle:
build now

PR #2437, opened 2017-10-06
DDF-2931: Close inputstream if product retrieval fails in CSW

tbatie:
#### What does this PR do?
Closes the products inputstream if product retrieval fails

anyone

@coyotesqrl
@stustison
Successful full build

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
+++ b/catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
@@ -914,6 +914,9 @@ public abstract class AbstractCswSource extends MaskableImpl
   }
 
   @Override
+  @SuppressWarnings("squid:S3776"
+  /* The codes complexity is not unreasonable in this case. Breaking up this method would serve little value as of this writing */
+  )
   public ResourceResponse retrieveResource(
       URI resourceUri, Map<String, Serializable> requestProperties)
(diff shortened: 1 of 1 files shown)


tbatie:
build now

coyotesqrl:
Regarding the SQ finding, I would prefer it _not_ be addressed in this PR as that would be a case of scope creep. The finding will be addressed in short order, I'm sure.

tbatie:
build now

tbatie:
build now

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
```diff
@@ -955,26 +958,31 @@ public ResourceResponse retrieveResource(
...
-            responseBytesSkipped =
-                (Long) recordCollection.getResourceProperties().get(BYTES_SKIPPED);
-          }
-          alignStream(resource.getInputStream(), requestedBytesToSkip, responseBytesSkipped);
+        long responseBytesSkipped = 0L;
+        if (recordCollection.getResourceProperties().get(BYTES_SKIPPED) != null) {
+          responseBytesSkipped = (Long) recordCollection.getResourceProperties().get(BYTES_SKIPPED);
+        }
+
+        try (InputStream resourceIs = resource.getInputStream()) {
```
clockard: After looking at this closer, I don't think we can put this stream in the try-with-resource. When the try finishes without error it will still close the stream but the caller of this method is expecting to use that stream to read the resource. We could still close the stream if there is an exception but shouldn't close it otherwise.

PR #2441, opened 2017-10-07
[2.11.x] DDF-3354 CDM verify file size is stable before ingesting

clockard:
#### What does this PR do?
Makes sure a file has finished copying before trying to ingest it with the CDM.
Also fixed a potential NPE in FileSystemStorageProvider
Also fix multiple monitored directory issue.

#### Who is reviewing it? 
@brendan-hofmann @emmberk 
@figliold
@rzwiefel

Install DDF and create a CDM for an empty directory.
Copy this script to a file and make it executable and changing 'cdm-dir' to the monitored directory configured for the CDM. The script simulates a slowly copying file. 
```
#!/bin/bash
for i in {1..60}
do
   echo "Line $i" >> cdm-dir/cdm-changing-file.txt
   sleep 1
done
```
Start the script and verify that the cdm-changing-file.txt doesn't show up in the catalog until after the script has finished. 
Run the script again and verify that the cdm-changing-file.txt metacard doesn't get updated until after the script has finished.
Add a second CDM for a different directory and copy a text file into it. Verify that the text file is ingested.
[DDF-3354](https://codice.atlassian.net/browse/DDF-3354)
#### Checklist:
- [X] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
+++ b/catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
@@ -1,0 +1,126 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.content.monitor;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import javafx.util.Pair;
+import javax.validation.constraints.NotNull;
+import org.apache.commons.io.monitor.FileAlterationListenerAdaptor;
+import org.codice.ddf.platform.util.StandardThreadFactoryBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DurableFileAlterationListener extends FileAlterationListenerAdaptor {
+
+  public static final long DEFAULT_PERIOD = 5;
+
+  public static final String CDM_FILE_CHECK_PERIOD_PROPERTY = "org.codice.ddf.cdm.fileCheckPeriod";
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DurableFileAlterationListener.class);
+
+  private Map<File, Pair<Long, WatchEvent.Kind<Path>>> fileMap = new ConcurrentHashMap<>();
+
+  private ScheduledExecutorService executorService;
+
+  private AbstractDurableFileConsumer consumer;
+
+  public DurableFileAlterationListener(@NotNull AbstractDurableFileConsumer consumer) {
+    this.consumer = consumer;
+    executorService =
+        Executors.newSingleThreadScheduledExecutor(
+            StandardThreadFactoryBuilder.newThreadFactory("directoryMonitorFileChecker"));
+    this.startExecutor();
+  }
+
+  public DurableFileAlterationListener(
+      @NotNull AbstractDurableFileConsumer consumer,
+      @NotNull ScheduledExecutorService executorService) {
+    this.consumer = consumer;
+    this.executorService = executorService;
+    this.startExecutor();
+  }
+
+  private void startExecutor() {
+    long period = DEFAULT_PERIOD;
+    try {
+      period =
+          Long.parseLong(
+              System.getProperty(CDM_FILE_CHECK_PERIOD_PROPERTY, Long.toString(DEFAULT_PERIOD)));
+      if (period < 1) {
+        period = DEFAULT_PERIOD;
+      }
+    } catch (NumberFormatException e) {
+      LOGGER.debug(
+          "Invalid value for system property org.codice.ddf.cdm.fileCheckPeriod. Expected an integer but was {}. Defaulting to {}",
+          System.getProperty(CDM_FILE_CHECK_PERIOD_PROPERTY),
+          period);
+    }
+    executorService.scheduleAtFixedRate(this::checkFiles, 10, period, TimeUnit.SECONDS);
+  }
+
+  public void destroy() {
+    // copied from the Executor javadocs
+    executorService.shutdown();
+    try {
+      if (!executorService.awaitTermination(60, TimeUnit.SECONDS)) {
+        executorService.shutdownNow();
+        if (!executorService.awaitTermination(60, TimeUnit.SECONDS))
+          LOGGER.debug("Error terminating scheduled executor service");
+      }
+    } catch (InterruptedException ie) {
+      executorService.shutdownNow();
+      Thread.currentThread().interrupt();
+    }
+  }
+
+  @Override
+  public void onFileChange(File file) {
+    if (!fileMap.containsKey(file)) {
+      fileMap.put(file, new Pair<>(-1L, StandardWatchEventKinds.ENTRY_MODIFY));
+    }
+  }
+
+  @Override
+  public void onFileCreate(File file) {
+    fileMap.put(file, new Pair<>(-1L, StandardWatchEventKinds.ENTRY_CREATE));
+  }
+
+  @Override
+  public void onFileDelete(File file) {
+    consumer.createExchangeHelper(file, StandardWatchEventKinds.ENTRY_DELETE);
+  }
+
+  void checkFiles() {
+    List<File> completedFiles = new ArrayList<>();
+    for (Map.Entry<File, Pair<Long, WatchEvent.Kind<Path>>> entry : fileMap.entrySet()) {
+      if (entry.getKey().length() == entry.getValue().getKey()) {
+        completedFiles.add(entry.getKey());
+        consumer.createExchangeHelper(entry.getKey(), entry.getValue().getValue());
+      } else {
+        entry.setValue(new Pair<>(entry.getKey().length(), entry.getValue().getValue()));
+      }
+    }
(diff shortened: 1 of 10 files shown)


clockard:
build now

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
```diff
@@ -0,0 +1,115 @@
...
+  public void onFileCreate(File file) {
+    fileMap.put(file, new Pair<>(-1L, StandardWatchEventKinds.ENTRY_CREATE));
+  }
+
+  @Override
+  public void onFileDelete(File file) {
+    consumer.createExchangeHelper(file, StandardWatchEventKinds.ENTRY_DELETE);
+  }
+
+  void checkFiles() {
```
figliold: :question: Seems like this could still be a problem when files are copied slowly.  It's too bad that we can't have the code that is doing to copying of the file into the CDM directory name the file with a different extension until it is fully copied.  This way knowing when the file was completely copied would be under the control of the thing actually writing the file.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import javafx.util.Pair;
+import javax.validation.constraints.NotNull;
+import org.apache.commons.io.monitor.FileAlterationListenerAdaptor;
+import org.codice.ddf.platform.util.StandardThreadFactoryBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DurableFileAlterationListener extends FileAlterationListenerAdaptor {
```
figliold: :question: Does this work when the file system being monitored is a network mounted drive?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
brendan-hofmann: Unfortunately we have no control over that code, as it could be anything writing to that directory. There are some conventions we could additionally check for, but those are implementation-dependent, so we would still need a fall-back logic.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
brendan-hofmann: As long as the period is adjusted and the current file size is retrievable, I think it would work. The logic is just waiting for the size of the file to stabilize.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
```diff
@@ -0,0 +1,115 @@
...
+
+  public static final long DEFAULT_PERIOD = 5;
+
+  public static final String CDM_FILE_CHECK_PERIOD_PROPERTY = "org.codice.ddf.cdm.fileCheckPeriod";
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DurableFileAlterationListener.class);
+
+  private Map<File, Pair<Long, WatchEvent.Kind<Path>>> fileMap = new ConcurrentHashMap<>();
```
brendan-hofmann: ❓ does this map need to be filesystem-backed? What happens if the system is restarted while we are waiting for a file to stabilize? It looks like we initialize the map entries to `-1L`, so I think the watcher might cover this.

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
```diff
@@ -397,7 +397,10 @@ private ContentItem readContent(URI uri) throws StorageException {
       extension = FilenameUtils.getExtension(FilenameUtils.removeExtension(filename));
       try {
         reference = new URI(new String(Files.readAllBytes(file), Charset.forName("UTF-8")));
-        if (reference.getScheme().equalsIgnoreCase("file")) {
+
+        if (reference.getScheme() == null) {
```
brendan-hofmann: ❓ Is this branch necessary? What was producing a null reference scheme?

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
clockard: Like you said we will just start over with a -1 size and start monitoring from there so we shouldn't need to persist any information.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
```diff
@@ -0,0 +1,115 @@
...
+    } catch (NumberFormatException e) {
+      LOGGER.warn(
+          "Invalid value for system property org.codice.ddf.cdm.fileCheckPeriod. Expected an integer but was {}. Defaulting to {}",
+          System.getProperty(CDM_FILE_CHECK_PERIOD_PROPERTY),
+          period);
+    }
+    executorService.scheduleAtFixedRate(this::checkFiles, 10, period, TimeUnit.SECONDS);
+  }
+
+  public void destroy() {
```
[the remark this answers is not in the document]
clockard: Probably, it is just an ugly block of code that I don't really like :). I will update this to use it though.

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
clockard: When a monitored file was updated the URI had only the path and no scheme causing a NPE here. Add this check to prevent this. File updates seem to work fine with this change.

brendan-hofmann: These files are only written and read by the storage provider IIRC. It sounds like it is either receiving a schemeless URI or not persisting that URI correctly. I would prefer to address that issue, it sounds like a bug to me.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
clockard: I did test with a remote file system mount and it worked as expected.

clockard:
build now

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
clockard: Seems like we would still want at least a check here so we don't throw a NPE. It it a requirement that the Path we get back from getContentFilePath has a scheme? Would you be ok with keeping this change for now and have a new ticket look into this?

brendan-hofmann: It used to be required, hence my concern. You might try testing a filesystem monitor and a webdav monitor, I suspect something will break in that case.

clockard:
build now

catalog/core/catalog-core-localstorageprovider/src/main/java/org/codice/ddf/catalog/content/impl/FileSystemStorageProvider.java
clockard: After the latest update, I tested with both a directory and a webdav endpoint both worked for ingest and update with this change.

clockard:
Manually tested both webdav, local dir, mounted remote dir. Verified that updates in all cases are also processed.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileComponent.java
```diff
@@ -22,9 +22,6 @@
 import org.apache.commons.lang.StringUtils;
 
 public class DurableFileComponent extends GenericFileComponent<EventfulFileWrapper> {
-  Boolean isDav;
```
clockard: Removed and pushed down into the DurableFileEndpoint since this is a singleton from camels perspective.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileComponent.java
```diff
@@ -41,9 +38,8 @@
       throw new IllegalArgumentException("Location to monitor must be specified");
     }
 
-    this.remaining = remaining;
     String davParam = String.valueOf(parameters.get("isDav"));
-    isDav = Boolean.valueOf(davParam);
+    boolean isDav = Boolean.valueOf(davParam);
```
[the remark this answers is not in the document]
brendan-hofmann: This subtlety was lost on me at first; it's complaining about using one of the Boolean methods that returns a boxed primitive, when it's being assigned to a primitive variable, which means it is immediately unboxed.

clockard:
buidl now

clockard:
build now

clockard:
build now

clockard:
build now

AzGoalie:
Verified on Windows 👍

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/DurableFileAlterationListener.java
```diff
@@ -0,0 +1,126 @@
...
+    List<File> completedFiles = new ArrayList<>();
+    for (Map.Entry<File, Pair<Long, WatchEvent.Kind<Path>>> entry : fileMap.entrySet()) {
+      if (entry.getKey().length() == entry.getValue().getKey()) {
+        completedFiles.add(entry.getKey());
+        consumer.createExchangeHelper(entry.getKey(), entry.getValue().getValue());
+      } else {
+        entry.setValue(new Pair<>(entry.getKey().length(), entry.getValue().getValue()));
+      }
+    }
+    completedFiles.stream().forEach(file -> fileMap.remove(file));
```
rzwiefel: Not necessary to change but for future reference this could be 
```java
completedFiles.stream().forEach(fileMap::remove);
```

PR #2445, opened 2017-10-09
DDF-3359 Fix installer port change issue

clockard:
#### What does this PR do?
Update the SystemPropertiesAdmin to update the pax web ports when needed

Also addressed some minor sonarqube findings in some of the files.

#### Who is reviewing it? 
@peterhuffer @emmberk 

@bdeining
@lessarderic

Install DDF and change the ports in the installer to https=9993 http=9181. 
Verify that the system restarts properly and the UIs are accessible.
[DDF-3359](https://codice.atlassian.net/browse/DDF-3359)
#### Checklist:
- [x] Update / Add Unit Tests

### DIFF
--- a/platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/SystemPropertiesAdmin.java
+++ b/platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/SystemPropertiesAdmin.java
@@ -23,6 +23,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import javax.management.InstanceAlreadyExistsException;
+import javax.management.InstanceNotFoundException;
+import javax.management.MBeanRegistrationException;
 import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
 import javax.management.NotCompliantMBeanException;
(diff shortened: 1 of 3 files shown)


platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/SystemPropertiesAdmin.java
```diff
@@ -192,6 +183,45 @@ public void writeSystemProperties(Map<String, String> updatedSystemProperties) {
 
     writeOutUsersDotPropertiesFile(userPropertiesFile);
     writeOutUsersDotAttributesFile(userAttributesFile);
+
+    // We need to update the pax web config when the ports are changed. However if we change it
+    // when the system is running some parts might not correctly handle the update. We need to
+    // make the change as the system is coming down so that it will be in place when the system
+    // comes back up.
+    LOGGER.info("Adding shutdown hoot to update {} port numbers.", PAX_WEB_FILE);
```
shaundmorris: ✏️  `hook` ?

clockard: Ooops :) fixed

lessarderic:
So it turns out that the Felix file installer code supports variables substitutions (e.g., `${var}`) out-of-the-box, which means that if you use those in any `.cfg` or `.config` file, the variables will be replaced when the file is read in and the actual values will be written back out.

The only way I've found to fix this is is to escape the substitutions, like this: `$\{var\}`. This prevents the substitution when the file is read in, and it gets written back out as `${var}`. This all works as long as the file isn't changed and needs to be re-read. In that case, the variables would be substituted and the actual value written back out.

clockard:
Closing this pr since a better solution is being worked.

PR #2447, opened 2017-10-09
[2.11.x] DDF-3346 Added OutOfMemory and VM error support on Linux

paouelle:
#### What does this PR do?
This PR adds out of memory and VM error support to automatically restart DDF. It also fixes the service wrapper documentation by simplifying the installation process. It also fixes the restart support from the UI or from migration:import when running using the service wrapper.
Changed escape string from backslash to `_DO_NOT_EXPAND_` when processing files in ddf-common to allow the backslash from actually being used to escape env vars in bash scripts. Upgraded the maven resource plugin to fix a corruption that was happening in ddf-common randomly on some of the filtered files.
#### Who is reviewing it? 
@figliold
@emanns95 
@tbatie
@clockard
@coyotesqrl
@ricklarsen - Documentation
1. Unzip and install the distribution on Linux
2. Go through the installation and make sure that the restart step at the end still works
3. Force an out of memory (see below). The process should restart
4. run a migration:export followed by a migration:import , the process should restart automatically
5. Go through the documentation to install the service wrapper
6. re-do step #2 and #4 via the client tool
7. Make sure that the setenv-XXX-wrapper.conf files are exported too.

The only way to verify the out of memory is to manually modify the code somewhere to create a leak. Something like this:
```final List list = new ArrayList();

      try {
        while (true) {
          list.add(new int[Integer.MAX_VALUE]);
        }
      } catch (OutOfMemoryError e) { // swallow
      }
```
in the code of a command that you can trigger from the shell would do the trick.

This ticket addresses both the ddf script and the wrapper service on Linux. The design is to use a tag or command file in between the normal starting execution path and the error path. The error path will either touch that file when invoked as a script or add a RESTART command when invoked via the wrapper as soon as an error is detected. 

When running as a script, we introduce a new ddf_exec script in between karaf's script and the JVM which monitors that restart.jvm file and if it is present when the JVM exist, it changes the exit code to 10 (10 is used by Karaf as a special way to tell the karaf script to restart). When started via the wrapper, the wrapper is configured to monitor that restart.jvm file every 5 seconds and process all commands that are written to it. In our case, the error script simply writes out RESTART in the file to get the wrapper to restart the JVM.

Additional code in the error script is added to monitor for up to 1 minute the JVM process and if it doesn't stop normally, it then forcibly kills it (i.e. kill -9).

The restart.jvm file should not be writeable by anything inside the JVM.
[DDF-3346](https://codice.atlassian.net/browse/DDF-3346)
#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:


paouelle:
build now

paouelle:
build now

paouelle:
build now

paouelle:
build now

paouelle:
build now

paouelle:
build now

distribution/ddf-common/src/main/resources/bin/setenv.bat
```diff
@@ -58,7 +58,7 @@ rem SET KARAF_BASE
 rem Additional available Karaf options
 rem SET KARAF_OPTS=
 
-rem comment out the line below to enable cxf logging interceptors
+rem Uncomment out the line below to enable cxf logging interceptors
```
figliold: ❓ Windows?

distribution/ddf-common/src/main/resources/bin/setenv-windows-wrapper.conf
```diff
@@ -0,0 +1,58 @@
+# Enables restart of JVM when it exits with 10. 10 is normally used by the karaf script to
```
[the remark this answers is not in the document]
paouelle: ok, you got me 😄 
I had to provide this file since I was changing the documentation on how to install the wrapper. That being said, the file doesn't define the error handling support yet; it is commented out and left for the next ticket.

distribution/ddf-common/src/main/resources/bin/setenv.bat
paouelle: Simple update to a comment that was wrong to start with.

paouelle:
build now

paouelle:
build now

paouelle:
build now

paouelle:
build now

distribution/ddf-common/src/main/resources/bin/ddf_on_error.sh
```diff
@@ -0,0 +1,50 @@
...
+  # When the process is started via karaf's script, our ddf.sh script which invokes karaf's script
+  # will check upon exit Karaf's script if the restart.jvm file is present and if it will restart
+  # karaf's script which will end up restarting the JVM
+  touch bin/restart.jvm
+  # Try to terminate the process normally to give it a chance
+  kill $PID
+fi
+
+# Wait for the process to die
+waitForPid $PID
```
coyotesqrl: ❓ Wouldn't it be possible for the process to already be dead by the time we got to here, meaning we'd go into our 60-second loop for no reason and then try to send the SIGKILL to a pid that was already long gone?

distribution/ddf-common/src/main/resources/bin/setenv-wrapper.conf
```diff
@@ -0,0 +1,67 @@
...
+wrapper.on_exit.10 = RESTART
+
+# Defines a command file to be used by the ddf_on_error.sh script when executed following
+# a non-recoverable system error and the JVM should be restarted
+wrapper.commandfile = %KARAF_HOME%/bin/restart.jvm
+
+# Additional JVM parameters
+# note that n is the parameter number starting from 11 (1-10 is defined in ddf-wrapper.conf).
+# make sure there is no gap
+wrapper.java.additional.11 = -Dddf.home=%KARAF_HOME%
```
coyotesqrl: ❓ Windows-style variable interpolation? This shouldn't be `$KARAF_HOME` or `${KARAF_HOME}`?

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -53,12 +56,16 @@ grant
     // some permissions might be repeated here to ensure that they are granted to the migration
     // framework if we ever tighten the permissions above
 
+    permission javax.management.MBeanServerPermission "createMBeanServer";
+
     permission java.io.FilePermission "<<ALL FILES>>", "read, readlink";
     permission java.io.FilePermission "${ddf.home}", "read, write";
+    permission java.io.FilePermission "${ddf.home}${/}-", "read, write, delete";
```
coyotesqrl: ✏️ This change is likely to have huge repercussions and will require a full regression pass. I'm particularly worried about blocking the `delete` permission on files beneath `$ddf.home` globally without granting that permission to several specific `codeBase`s.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -56,6 +62,8 @@
 
   private static final String INVALID_NULL_EXPORT_DIR = "invalid null export directory";
 
+  @VisibleForTesting static final Path RESTART_JVM = Paths.get("bin", "restart.jvm");
```
coyotesqrl: ❗️ We should specify the root of this path at `$ddf.home` instead of using a relative path.

distribution/ddf-common/src/main/resources/bin/ddf_on_error.sh
paouelle: good question. If that was the case, the first `kill -0` would return right away indicating the process is not running anymore. As such, we would not be waiting 60 seconds.

distribution/ddf-common/src/main/resources/bin/setenv-wrapper.conf
paouelle: This is a service wrapper configuration and not a bash or command. These are expanded by the wrapper.

distribution/ddf-common/src/main/resources/security/default.policy
paouelle: This permission is granted only to the migration framework.

coyotesqrl: I read that diff wrong and thought that was in the global deny block. My mistake.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
paouelle: That is actually a left over from a previous implementation attempt. It was being resolved from ddf_home at the time when used. I have removed the line.

distribution/ddf-common/src/main/resources/security/default.policy
paouelle: I would rather we discuss it now then later when it is too late ;-)

paouelle:
build now

emanns95:
I pulled down and manually confirmed the changes on CentOS

paouelle:
build now

paouelle:
build now

paouelle:
build now

paouelle:
build now

paouelle:
build now

PR #2486, opened 2017-10-12
DDF-3376 Add encryption support to the migration framework

oconnormi:
#### What does this PR do?

Encrypts/Decrypts the export created by the migration framework

#### Who is reviewing it? 

@paouelle @peterhuffer @emmberk @AzGoalie  

@clockard @coyotesqrl 

1. Install DDF
2. run `migration:export`
3. confirm that `<ddf_home>/exported/<ddf-export-file-name>.dar` and `<ddf_home>/exported/<ddf-export-file-name>.dar.key` and `<ddf_home>/exported/<ddf-export-file-name>.dar.sha256` exist
4. manually unzip the file and verify that files have encrypted contents
5. copy the zip and the key to a new system
6. run `migration:import` and confirm success

Each entry in the migration zip is individually encrypted. The zip can be unzipped with any zip utility, however the contents of each file will be unreadable unless decrypted using the correct key

[DDF-3376](https://codice.atlassian.net/browse/DDF-3376)

#### Review Comment Legend:

### DIFF
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
@@ -1,0 +1,201 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.configuration.migration;
+
+import static org.apache.commons.codec.binary.Hex.decodeHex;
+import static org.apache.commons.codec.binary.Hex.encodeHex;
+import static org.apache.commons.io.FileUtils.writeStringToFile;
+import static org.codice.ddf.configuration.migration.MigrationZipConstants.getDefaultChecksumPathFor;
+import static org.codice.ddf.configuration.migration.MigrationZipConstants.getDefaultKeyPathFor;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Charsets;
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import javax.crypto.Cipher;
+import javax.crypto.CipherOutputStream;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.DecoderException;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.Validate;
+import org.codice.ddf.migration.MigrationException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CipherUtils {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CipherUtils.class);
+
+  private final Path zipPath;
+  private final Path keyPath;
+  private final Path checksumPath;
+  private final Cipher cipher;
+  private SecretKey secretKey;
+
+  public CipherUtils(Path zipPath) {
+    this(zipPath, getDefaultKeyPathFor(zipPath), getDefaultChecksumPathFor(zipPath));
+  }
+
+  public CipherUtils(Path zipPath, Path keyPath, Path checksumPath) {
+    Validate.notNull(zipPath, "Null zip path");
+    Validate.notNull(keyPath, "Null key path");
+    Validate.notNull(checksumPath, "Null checksum path");
+    this.zipPath = zipPath;
+    this.keyPath = keyPath;
+    this.checksumPath = checksumPath;
+    cipher = createEncryptionCipher();
+  }
+
+  /**
+   * Wraps a given {@link OutputStream} in a {@link CipherOutputStream}
+   *
+   * @param outputStream
+   * @return a {@link CipherOutputStream} for writing encrypted {@link java.util.zip.ZipEntry}
+   *     contents
+   */
+  public CipherOutputStream getCipherOutputStream(OutputStream outputStream) {
+    return new CipherOutputStream(outputStream, cipher);
+  }
+
+  /**
+   * Creates a {@link Cipher} for use during file encryption and stores the secret key in a file
+   *
+   * @throws MigrationException when an invalid key is provided
+   * @throws MigrationException when an invalid key padding is used
+   * @throws MigrationException when and invalid cipher algorithm is used
+   * @return a {@link Cipher} in encrypt mode
+   */
+  private Cipher createEncryptionCipher() {
+    try {
+      Cipher iCipher = Cipher.getInstance(MigrationZipConstants.CIPHER_ALGORITHM);
+      IvParameterSpec ivParameterSpec = new IvParameterSpec(MigrationZipConstants.CIPHER_IV);
+      iCipher.init(Cipher.ENCRYPT_MODE, initKey(keyPath), ivParameterSpec);
+      return iCipher;
+    } catch (InvalidKeyException
+        | NoSuchPaddingException
+        | NoSuchAlgorithmException
+        | InvalidAlgorithmParameterException e) {
+      throw new MigrationException(MigrationZipConstants.KEY_INVALID_ERROR, keyPath, e);
+    }
+  }
+
+  /**
+   * Creates a file containing the checksum for the zip file
+   *
+   * @throws IOException
+   * @throws NoSuchAlgorithmException
+   */
+  public void createZipChecksumFile() throws IOException, NoSuchAlgorithmException {
+    String hash = getChecksum(zipPath);
+    writeStringToFile(checksumPath.toFile(), hash, Charsets.UTF_8);
+  }
+
+  public Path getZipPath() {
+    return zipPath;
+  }
+
+  public Path getKeyPath() {
+    return keyPath;
+  }
+
+  public Path getChecksumPath() {
+    return checksumPath;
+  }
+
+  @VisibleForTesting
+  SecretKey getSecretKey() {
+    return secretKey;
+  }
+
+  @VisibleForTesting
+  Cipher getCipher() {
+    return cipher;
+  }
+
+  private SecretKey initKey(Path keyPath) throws NoSuchAlgorithmException {
+    if (keyPath.toFile().exists()) {
+      SecretKey key;
+      try {
+        byte[] keyBytes = decodeHex(loadKeyDataFrom(keyPath).toCharArray());
+        key = new SecretKeySpec(keyBytes, MigrationZipConstants.KEY_ALGORITHM);
+      } catch (DecoderException e) {
+        String message = String.format(MigrationZipConstants.KEY_DECODE_ERROR, keyPath, e);
+        LOGGER.warn(message);
+        throw new MigrationException(message);
+      }
+      this.secretKey = key;
+      return key;
+    } else {
+      return createSecretKey(keyPath);
+    }
+  }
+
+  private String loadKeyDataFrom(Path keyPath) {
+    try (FileInputStream fis = new FileInputStream(keyPath.toFile())) {
+      return IOUtils.toString(fis, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      String message = String.format(MigrationZipConstants.FILE_IO_ERROR, keyPath, e);
+      LOGGER.warn(message);
+      throw new MigrationException(message);
+    }
+  }
+
+  /**
+   * Creates a secret key and stores it in a file
+   *
+   * @param keyPath the path for storing the secret key
+   * @throws MigrationException when an invalid key algorithm is used
+   * @throws MigrationException when the key can not be written to a file
+   * @throws NoSuchAlgorithmException when an invalid algorithm is used for the {@link KeyGenerator}
+   * @return a {@link SecretKey}
+   */
+  private SecretKey createSecretKey(Path keyPath) throws NoSuchAlgorithmException {
+    KeyGenerator keyGenerator = null;
+    try {
+      keyGenerator = KeyGenerator.getInstance(MigrationZipConstants.KEY_ALGORITHM);
+      secretKey = keyGenerator.generateKey();
+      char[] hexKey = encodeHex(secretKey.getEncoded());
+      writeStringToFile(keyPath.toFile(), String.valueOf(hexKey), Charsets.UTF_8);
+      return secretKey;
+    } catch (IOException e) {
+      throw new MigrationException(String.format(MigrationZipConstants.FILE_IO_ERROR, keyPath, e));
+    }
+  }
+
+  private String getChecksum(Path path) throws IOException, NoSuchAlgorithmException {
+    try (BufferedInputStream bif = new BufferedInputStream(new FileInputStream(path.toFile()))) {
+      MessageDigest digest =
+          MessageDigest.getInstance(MigrationZipConstants.CHECKSUM_DIGEST_ALGORITHM);
+      byte[] fileBytes = IOUtils.toByteArray(bif);
+      digest.update(fileBytes);
+      return new String(encodeHex(digest.digest()));
+    } catch (FileNotFoundException e) {
+      LOGGER.error("File {} does not exist", path.toString(), e);
+      throw e;
+    }
+  }
+}
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
@@ -15,17 +15,20 @@ package org.codice.ddf.configuration.migration;
 
 import com.google.common.annotations.VisibleForTesting;
 import ddf.security.common.audit.SecurityLogger;
+import java.io.FileNotFoundException;
 import java.io.IOError;
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.security.NoSuchAlgorithmException;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.BinaryOperator;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
+import javax.crypto.NoSuchPaddingException;
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanException;
 import javax.management.MalformedObjectNameException;
@@ -169,23 +172,24 @@ public class ConfigurationMigrationManager implements ConfigurationMigrationServ
   }
 
   @VisibleForTesting
-  void delegateToImportMigrationManager(MigrationReportImpl report, Path exportFile) {
+  void delegateToImportMigrationManager(MigrationReportImpl report, MigrationZipFile zip)
+      throws NoSuchAlgorithmException, NoSuchPaddingException {
     final ImportMigrationManagerImpl mgr =
-        new ImportMigrationManagerImpl(report, exportFile, migratables.stream());
-
+        new ImportMigrationManagerImpl(report, zip, migratables.stream());
     try {
-      report.record(Messages.IMPORTING_DATA, productBranding, exportFile);
+      report.record(Messages.IMPORTING_DATA, productBranding, zip.getZipPath());
       mgr.doImport(productBranding, productVersion);
     } finally {
-      IOUtils.closeQuietly(mgr); // do not care if we fail to close the mgr/zip file!!!
+      IOUtils.closeQuietly(mgr);
     }
   }
 
   @VisibleForTesting
-  void delegateToExportMigrationManager(MigrationReportImpl report, Path exportFile)
-      throws IOException {
+  void delegateToExportMigrationManager(
+      MigrationReportImpl report, Path exportFile, CipherUtils cipherUtils)
+      throws IOException, NoSuchPaddingException, NoSuchAlgorithmException {
     try (final ExportMigrationManagerImpl mgr =
-        new ExportMigrationManagerImpl(report, exportFile, migratables.stream())) {
+        new ExportMigrationManagerImpl(report, exportFile, cipherUtils, migratables.stream())) {
       report.record(Messages.EXPORTING_DATA, productBranding, exportFile);
       mgr.doExport(productBranding, productVersion);
     }
@@ -212,8 +216,13 @@ public class ConfigurationMigrationManager implements ConfigurationMigrationServ
                 + productVersion
                 + ConfigurationMigrationManager.EXPORT_EXTENSION);
 
+    final CipherUtils cipherUtils = new CipherUtils(exportFile);
+
     try {
-      delegateToExportMigrationManager(report, exportFile);
+      delegateToExportMigrationManager(report, exportFile, cipherUtils);
+      if (report.wasSuccessful()) {
+        cipherUtils.createZipChecksumFile();
+      }
     } catch (MigrationException e) {
       report.record(e);
     } catch (IOException e) {
@@ -222,12 +231,18 @@ public class ConfigurationMigrationManager implements ConfigurationMigrationServ
       report.record(new MigrationException(Messages.EXPORT_SECURITY_ERROR, exportFile, e));
     } catch (RuntimeException e) {
       report.record(new MigrationException(Messages.EXPORT_INTERNAL_ERROR, exportFile, e));
+    } catch (NoSuchPaddingException | NoSuchAlgorithmException e) {
+      report.record(
+          new MigrationException(Messages.EXPORT_INTERNAL_ERROR, cipherUtils.getKeyPath(), e));
     }
     report.end();
     if (report.hasErrors()) {
       SecurityLogger.audit("Errors exporting configuration settings to file {}", exportFile);
       // don't leave the zip file there if the export failed
       PathUtils.deleteQuietly(exportFile, ConfigurationMigrationManager.EXPORT_DIR);
+      PathUtils.deleteQuietly(cipherUtils.getKeyPath(), ConfigurationMigrationManager.EXPORT_DIR);
+      PathUtils.deleteQuietly(
+          cipherUtils.getChecksumPath(), ConfigurationMigrationManager.EXPORT_DIR);
       report.record(new MigrationException(Messages.EXPORT_FAILURE, exportFile));
     } else if (report.hasWarnings()) {
       SecurityLogger.audit("Warnings exporting configuration settings to file {}", exportFile);
@@ -250,29 +265,36 @@ public class ConfigurationMigrationManager implements ConfigurationMigrationServ
                 + productVersion
                 + ConfigurationMigrationManager.EXPORT_EXTENSION);
 
+    MigrationZipFile zip = null;
     try {
-      delegateToImportMigrationManager(report, exportFile);
+      zip = newZipFileFor(exportFile);
+      if (!zip.isValidChecksum()) {
+        throw new MigrationException(Messages.IMPORT_ZIP_CHECKSUM_INVALID, exportFile);
+      }
+      delegateToImportMigrationManager(report, zip);
     } catch (MigrationException e) {
       report.record(e);
     } catch (SecurityException e) {
       report.record(new MigrationException(Messages.IMPORT_SECURITY_ERROR, exportFile, e));
     } catch (RuntimeException e) {
       report.record(new MigrationException(Messages.IMPORT_INTERNAL_ERROR, exportFile, e));
+    } catch (NoSuchPaddingException | NoSuchAlgorithmException e) {
+      report.record(new MigrationException(Messages.IMPORT_INTERNAL_ERROR, zip.getKeyPath(), e));
     }
     report.end();
-    if (report.hasErrors()) {
+    if ((zip == null) || (report.hasErrors())) {
       SecurityLogger.audit("Errors importing configuration settings from file {}", exportFile);
       report.record(new MigrationException(Messages.IMPORT_FAILURE, exportFile));
     } else if (report.hasWarnings()) {
       SecurityLogger.audit("Warnings importing configuration settings from file {}", exportFile);
       // don't leave the zip file there if the import succeeded
-      PathUtils.deleteQuietly(exportFile, ConfigurationMigrationManager.EXPORT_DIR);
+      zip.deleteQuitetly();
       report.record(new MigrationWarning(Messages.IMPORT_SUCCESS_WITH_WARNINGS, exportFile));
       report.record(new MigrationWarning(Messages.RESTART_SYSTEM_WHEN_WARNINGS));
     } else {
       SecurityLogger.audit("Exported configuration settings from file {}", exportFile);
       // don't leave the zip file there if the import succeeded
-      PathUtils.deleteQuietly(exportFile, ConfigurationMigrationManager.EXPORT_DIR);
+      zip.deleteQuitetly();
       report.record(new MigrationSuccessfulInformation(Messages.IMPORT_SUCCESS, exportFile));
       // force a JVM restart
       restart(report);
@@ -310,5 +332,16 @@ public class ConfigurationMigrationManager implements ConfigurationMigrationServ
       return true;
     }
     return false;
+  }
+
+  private static MigrationZipFile newZipFileFor(Path exportFile) {
+    Validate.notNull(exportFile, "invalid null export file");
+    try {
+      return new MigrationZipFile(exportFile);
+    } catch (FileNotFoundException e) {
+      throw new MigrationException(Messages.IMPORT_FILE_MISSING_ERROR, exportFile, e);
+    } catch (SecurityException | IOException e) {
+      throw new MigrationException(Messages.IMPORT_FILE_OPEN_ERROR, exportFile, e);
+    }
   }
 }
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
@@ -29,6 +29,7 @@ import java.util.function.BiPredicate;
 import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
+import javax.crypto.CipherOutputStream;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.filefilter.TrueFileFilter;
 import org.apache.commons.io.output.ClosedOutputStream;
@@ -59,6 +60,8 @@ public class ExportMigrationContextImpl extends MigrationContextImpl<ExportMigra
 
   private final ZipOutputStream zipOutputStream;
 
+  private final CipherUtils cipherUtils;
+
   private volatile OutputStream currentOutputStream;
 
   /**
@@ -71,14 +74,17 @@ public class ExportMigrationContextImpl extends MigrationContextImpl<ExportMigra
    * </code> is <code>null</code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  ExportMigrationContextImpl(MigrationReport report, Migratable migratable, ZipOutputStream zos) {
+  ExportMigrationContextImpl(
+      MigrationReport report, Migratable migratable, ZipOutputStream zos, CipherUtils cipherUtils) {
     super(
         new ExportMigrationReportImpl(report, migratable),
         migratable,
         ExportMigrationContextImpl.validateNotNull(migratable, "invalid null migratable")
             .getVersion());
     Validate.notNull(zos, "invalid null zip output stream");
+    Validate.notNull(cipherUtils, "invalid null cipher utils");
     this.zipOutputStream = zos;
+    this.cipherUtils = cipherUtils;
   }
 
   private static <T> T validateNotNull(T t, String msg) {
@@ -220,8 +226,9 @@ public class ExportMigrationContextImpl extends MigrationContextImpl<ExportMigra
             }
           };
 
-      this.currentOutputStream = oos;
-      return oos;
+      CipherOutputStream cos = cipherUtils.getCipherOutputStream(oos);
+      this.currentOutputStream = cos;
+      return cos;
     } catch (IOException e) {
       throw new UncheckedIOException(e);
     }
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
@@ -29,6 +29,8 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
+import javax.crypto.CipherOutputStream;
+import org.apache.commons.io.output.CloseShieldOutputStream;
 import org.apache.commons.lang.Validate;
 import org.codice.ddf.migration.Migratable;
 import org.codice.ddf.migration.MigrationException;
@@ -60,6 +62,8 @@ public class ExportMigrationManagerImpl implements Closeable {
 
   private final Path exportFile;
 
+  private CipherUtils cipherUtils;
+
   private boolean closed = false;
 
   /**
@@ -75,10 +79,14 @@ public class ExportMigrationManagerImpl implements Closeable {
    *     is <code>null</code>
    */
   public ExportMigrationManagerImpl(
-      MigrationReport report, Path exportFile, Stream<? extends Migratable> migratables) {
+      MigrationReport report,
+      Path exportFile,
+      CipherUtils cipherUtils,
+      Stream<? extends Migratable> migratables) {
     this(
         report,
         exportFile,
+        cipherUtils,
         migratables,
         ExportMigrationManagerImpl.newZipOutputStreamFor(exportFile));
   }
@@ -86,21 +94,26 @@ public class ExportMigrationManagerImpl implements Closeable {
   ExportMigrationManagerImpl(
       MigrationReport report,
       Path exportFile,
+      CipherUtils cipherUtils,
       Stream<? extends Migratable> migratables,
       ZipOutputStream zos) {
     Validate.notNull(report, "invalid null report");
     Validate.isTrue(
         report.getOperation() == MigrationOperation.EXPORT, "invalid migration operation");
     Validate.notNull(migratables, "invalid null migratables");
+    Validate.notNull(cipherUtils, "invalid null cipher utils");
     this.report = report;
     this.exportFile = exportFile;
+    this.cipherUtils = cipherUtils;
     this.zipOutputStream = zos;
     // pre-create contexts for all registered migratables
     this.contexts =
         migratables.collect(
             Collectors.toMap(
                 Migratable::getId,
-                m -> new ExportMigrationContextImpl(report, m, zipOutputStream),
+                m ->
+                    new ExportMigrationContextImpl(
+                        report, m, zipOutputStream, new CipherUtils(exportFile)),
                 ConfigurationMigrationManager.throwingMerger(),
                 LinkedHashMap::new)); // to preserved ranking order and remove duplicates
   }
@@ -156,20 +169,31 @@ public class ExportMigrationManagerImpl implements Closeable {
     LOGGER.debug("Exported metadata: {}", metadata);
   }
 
+  @SuppressWarnings({
+    "squid:S2093" /* Not using try-with-resources with CipherOutputStream. Stream is closed inside finally block */
+  })
   @Override
   public void close() throws IOException {
     if (!closed) {
       this.closed = true;
+      CipherOutputStream cos = null;
       try {
         zipOutputStream.closeEntry();
         try {
           zipOutputStream.putNextEntry(
               new ZipEntry(MigrationContextImpl.METADATA_FILENAME.toString()));
-          JsonUtils.MAPPER.writeValue(zipOutputStream, metadata);
+          CloseShieldOutputStream shieldOutputStream = new CloseShieldOutputStream(zipOutputStream);
+          cos = cipherUtils.getCipherOutputStream(shieldOutputStream);
+          JsonUtils.MAPPER.writeValue(cos, metadata);
         } catch (IOException e) {
           throw new MigrationException(Messages.EXPORT_METADATA_CREATE_ERROR, e);
+        } finally {
+          if (cos != null) {
+            cos.close();
+          }
         }
       } finally {
+
         zipOutputStream.close();
       }
     }
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
@@ -28,7 +28,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.stream.Stream;
-import java.util.zip.ZipFile;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.Validate;
 import org.codice.ddf.migration.ImportMigrationContext;
@@ -61,7 +60,7 @@ public class ImportMigrationContextImpl extends MigrationContextImpl<MigrationRe
 
   private final Set<String> files = new HashSet<>();
 
-  private final ZipFile zip;
+  private final MigrationZipFile zip;
 
   private final List<InputStream> inputStreams = new ArrayList<>();
 
@@ -74,7 +73,7 @@ public class ImportMigrationContextImpl extends MigrationContextImpl<MigrationRe
    * </code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  public ImportMigrationContextImpl(MigrationReport report, ZipFile zip) {
+  public ImportMigrationContextImpl(MigrationReport report, MigrationZipFile zip) {
     super(report);
     Validate.notNull(zip, ImportMigrationContextImpl.INVALID_NULL_ZIP);
     this.zip = zip;
@@ -90,7 +89,7 @@ public class ImportMigrationContextImpl extends MigrationContextImpl<MigrationRe
    *     is <code>null</code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  public ImportMigrationContextImpl(MigrationReport report, ZipFile zip, String id) {
+  public ImportMigrationContextImpl(MigrationReport report, MigrationZipFile zip, String id) {
     super(report, id);
     Validate.notNull(zip, ImportMigrationContextImpl.INVALID_NULL_ZIP);
     this.zip = zip;
@@ -106,7 +105,8 @@ public class ImportMigrationContextImpl extends MigrationContextImpl<MigrationRe
    * </code> is <code>null</code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  public ImportMigrationContextImpl(MigrationReport report, ZipFile zip, Migratable migratable) {
+  public ImportMigrationContextImpl(
+      MigrationReport report, MigrationZipFile zip, Migratable migratable) {
     super(report, migratable);
     Validate.notNull(zip, ImportMigrationContextImpl.INVALID_NULL_ZIP);
     this.zip = zip;
@@ -222,7 +222,7 @@ public class ImportMigrationContextImpl extends MigrationContextImpl<MigrationRe
   }
 
   @VisibleForTesting
-  ZipFile getZip() {
+  MigrationZipFile getZip() {
     return zip;
   }
 
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
@@ -15,7 +15,6 @@ package org.codice.ddf.configuration.migration;
 
 import com.google.common.annotations.VisibleForTesting;
 import java.io.Closeable;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
@@ -25,7 +24,6 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import java.util.zip.ZipFile;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.Validate;
 import org.codice.ddf.migration.ImportMigrationEntry;
@@ -51,7 +49,7 @@ public class ImportMigrationManagerImpl implements Closeable {
    */
   private final Map<String, ImportMigrationContextImpl> contexts;
 
-  private final ZipFile zip;
+  private final MigrationZipFile zip;
 
   private final String version;
 
@@ -65,7 +63,7 @@ public class ImportMigrationManagerImpl implements Closeable {
    * Creates a new migration manager for an import operation.
    *
    * @param report the migration report where warnings and errors can be recorded
-   * @param exportFile the exported zip file
+   * @param zip the exported zip file
    * @param migratables a stream of all migratables in the system
    * @throws MigrationException if a failure occurs while processing the zip file (the error will
    *     not be recorded with the report)
@@ -74,21 +72,19 @@ public class ImportMigrationManagerImpl implements Closeable {
    *     is <code>null</code>
    */
   public ImportMigrationManagerImpl(
-      MigrationReport report, Path exportFile, Stream<? extends Migratable> migratables) {
-    this(report, exportFile, migratables, ImportMigrationManagerImpl.newZipFileFor(exportFile));
+      MigrationReport report, MigrationZipFile zip, Stream<? extends Migratable> migratables) {
+    this(report, migratables, zip);
   }
 
   ImportMigrationManagerImpl(
-      MigrationReport report,
-      Path exportFile,
-      Stream<? extends Migratable> migratables,
-      ZipFile zip) {
+      MigrationReport report, Stream<? extends Migratable> migratables, MigrationZipFile zip) {
     Validate.notNull(report, "invalid null report");
+    Validate.notNull(zip, "null zip file");
     Validate.isTrue(
         report.getOperation() == MigrationOperation.IMPORT, "invalid migration operation");
     Validate.notNull(migratables, "invalid null migratables");
     this.report = report;
-    this.exportFile = exportFile;
+    this.exportFile = zip.getZipPath();
     this.zip = zip;
     Map<String, Object> metadata;
 
@@ -129,17 +125,6 @@ public class ImportMigrationManagerImpl implements Closeable {
     } catch (RuntimeException e) {
       IOUtils.closeQuietly(zip);
       throw e;
-    }
-  }
-
-  private static ZipFile newZipFileFor(Path exportFile) {
-    Validate.notNull(exportFile, "invalid null export file");
-    try {
-      return new ZipFile(exportFile.toFile());
-    } catch (FileNotFoundException e) {
-      throw new MigrationException(Messages.IMPORT_FILE_MISSING_ERROR, exportFile, e);
-    } catch (SecurityException | IOException e) {
-      throw new MigrationException(Messages.IMPORT_FILE_OPEN_ERROR, exportFile, e);
     }
   }
 
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
@@ -75,6 +75,25 @@ public final class Messages {
   public static final String IMPORT_MISMATCH_DDF_HOME_ERROR =
       "Import error: mismatched ddf.home [%s]; expecting local system to be installed under [%s].";
 
+  public static final String IMPORT_ZIP_CHECKSUM_INVALID =
+      "Import error: incorrect checksum for export file [%s].";
+
+  public static final String IMPORT_KEY_OPEN_ERROR = "Import error: failed to open key file [%s].";
+
+  public static final String IMPORT_KEY_DECODE_ERROR =
+      "Import error: could not decode key file [%s].";
+
+  public static final String IMPORT_KEY_INVALID_ERROR =
+      "Import error: invalid key file used for import [%s].";
+
+  public static final String EXPORT_KEY_INVALID_ERROR = "Export error: invalid key used for export";
+
+  public static final String EXPORT_KEY_IO_ERROR =
+      "Export error: failed to write decryption key file [%s].";
+
+  public static final String IMPORT_ENTRY_READ_ERROR =
+      "Import error: unable to read entry from zip";
+
   public static final String EXPORT_METADATA_CREATE_ERROR =
       "Export error: failed to create metadata; %s.";
 
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipConstants.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipConstants.java
@@ -1,0 +1,69 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.configuration.migration;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public final class MigrationZipConstants {
+
+  @VisibleForTesting static final String CHECKSUM_EXTENSION = ".sha256";
+  @VisibleForTesting static final String KEY_EXTENSION = ".key";
+
+  @VisibleForTesting
+  static final String KEY_DECODE_ERROR = "could not decode key from file [%s] [%s].";
+
+  @VisibleForTesting
+  static final String KEY_INVALID_ERROR = "invalid key file used for import [%s] [%s].";
+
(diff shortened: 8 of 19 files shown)


paouelle:
❓ Should we still named the file with the .zip extension now that it is encrypted?

paouelle:
the exported dir is under ddf.home and not under etc

oconnormi:
@paouelle I fixed the reference in the description here, took out the mention of `etc`. Also I had thought adding a `.enc` on the end of the file to denote that it is encrypted

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
```diff
@@ -96,19 +98,20 @@ public ImportMigrationManagerImpl(
...
       // add a system contexts
-      contexts.put(null, new ImportMigrationContextImpl(report, zip));
-      zip.stream()
-          .filter(ze -> !ze.isDirectory())
-          .map(ze -> new ImportMigrationEntryImpl(this::getContextFor, ze))
+      contexts.put(null, new ImportMigrationContextImpl(report, zipContents));
+      zipContents
+          .entrySet()
+          .stream()
+          .filter(k -> !k.getKey().isDirectory())
```
paouelle: ❗️ might no longer be needed to filter here as we should filter when we actually go through the content of the zip such that we don't try to read the content (empty) of directory entries. I would suggest you retest the directory filtering by taking the zip, decrypting it, unzipping it, re-zipping from the command-line which will introduce directory entries and then re-encrypting it.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * Streams the contents of a {@link ZipInputStream} as a {@link Map} where the content of each
+   * {@link ZipEntry} is stored inside a key corresponding to each {@link ZipEntry}'s name
+   *
+   * @param in the {@link ZipInputStream} to retrieve the entries from
+   * @return the stream of entries
+   */
+  public Map<ZipEntry, byte[]> entriesFrom(ZipInputStream in) {
+    ZipEntry entry;
+    HashMap<ZipEntry, byte[]> entries = new HashMap<>();
```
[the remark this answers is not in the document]
paouelle: :question: are zip entries hashable so they can be used as keys in a map

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
```diff
@@ -96,19 +98,20 @@ public ImportMigrationManagerImpl(
...
+          .map(k -> new ImportMigrationEntryImpl(this::getContextFor, k.getKey()))
```
paouelle: ❓ Would it be better to pass in the entry ctor the byte array for the content (or the error if we couldn't read it) as opposed to relying on the entry to get back here to get it later? It was done like that before because we were using the ZipFile and we only wanted to get the input stream on-demand.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImplTest.java
```diff
@@ -65,30 +62,36 @@ public ImportMigrationManagerImplTest() {
...
-            })
-        .when(zip)
-        .stream();
-
-    mgr = new ImportMigrationManagerImpl(report, exportFile, Stream.of(migratables), zip);
+            zipContents,
+            Optional.of(MigrationContextImpl.CURRENT_VERSION),
+            Optional.of(PRODUCT_VERSION));
+
+    //    // use answer to ensure we create a new stream each time if called multiple times
```
paouelle: ✏️  might want to remove the code if it is no longer needed

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImplTest.java
```diff
@@ -139,23 +142,27 @@ private ZipEntry getMetadataZipEntry(
     Mockito.when(ze.getName()).thenReturn(MigrationContextImpl.METADATA_FILENAME.toString());
     Mockito.when(ze.isDirectory()).thenReturn(false);
     // use answer to ensure we create a new stream each time if called multiple times
-    Mockito.doAnswer(
-            AdditionalAnswers.answer(
-                zea -> new ByteArrayInputStream(sb.toString().getBytes(Charsets.UTF_8))))
-        .when(zip)
-        .getInputStream(ze);
+    //    Mockito.doAnswer(
```
paouelle: ✏️  remove if no longer needed

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImplTest.java
```diff
@@ -139,23 +142,27 @@ private ZipEntry getMetadataZipEntry(
...
+      throws IOException {
     final ZipEntry ze = Mockito.mock(ZipEntry.class);
 
     Mockito.when(ze.getName()).thenReturn(Paths.get(migratableId).resolve(path).toString() + '/');
     Mockito.when(ze.isDirectory()).thenReturn(true);
     // use answer to ensure we create a new stream each time if called multiple times
-    Mockito.doAnswer(AdditionalAnswers.answer(zea -> new NullInputStream(0L)))
-        .when(zip)
-        .getInputStream(ze);
+    //    Mockito.doAnswer(AdditionalAnswers.answer(zea -> new NullInputStream(0L)))
```
paouelle: ✏️  remove if no longer needed

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImplTest.java
```diff
@@ -291,25 +302,13 @@ public void testConstructorWhenZipFileNotFound() throws Exception {
   }
 
   @Test
```
paouelle: ❗️ Might want to re-add this test case but since we no longer have the input stream, we still need to test the failure scenario since the code can still be invoked for other errors

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -27,7 +27,6 @@
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.Validate;
 import org.apache.karaf.system.SystemService;
 import org.codice.ddf.migration.Migratable;
```
paouelle: ❗️ Make sure to delete the key file when export failure occurs and when import succeeds just as we delete the zip.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,214 @@
...
+import javax.crypto.CipherOutputStream;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+import org.apache.commons.codec.DecoderException;
+import org.apache.commons.io.IOUtils;
+import org.codice.ddf.migration.MigrationException;
+
+public class ZipUtils {
```
paouelle: ❗️ This class doesn't maintain any state. So why not make all methods static, add a private ctor and remove the need from instantiating it when it is required?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,214 @@
...
+
+  @SuppressWarnings("PMD.DefaultPackage" /* designed as an internal service within this package */)
```
paouelle: ❓ If you are putting them package private for testing only then you shouldn't need to suppress warnings just keep the @VisibleForTesting annotation. Would protected be better if that is the case.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -68,55 +69,57 @@
...
    * @throws IllegalArgumentException if <code>report</code> or <code>zip</code> is <code>null
    * </code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  public ImportMigrationContextImpl(MigrationReport report, ZipFile zip) {
+  public ImportMigrationContextImpl(MigrationReport report, Map<ZipEntry, byte[]> zipContents) {
     super(report);
-    Validate.notNull(zip, ImportMigrationContextImpl.INVALID_NULL_ZIP);
-    this.zip = zip;
+    Validate.notNull(zipContents, ImportMigrationContextImpl.INVALID_NULL_ZIP);
```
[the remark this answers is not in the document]
oconnormi: yeah probably not necessary now, since the zipContents will always "exist" it just may not contain anything now instead of being `null`. Removing.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -68,55 +69,57 @@
...
    *     is <code>null</code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  public ImportMigrationContextImpl(MigrationReport report, ZipFile zip, String id) {
+  public ImportMigrationContextImpl(
+      MigrationReport report, Map<ZipEntry, byte[]> zipContents, String id) {
     super(report, id);
-    Validate.notNull(zip, ImportMigrationContextImpl.INVALID_NULL_ZIP);
-    this.zip = zip;
+    Validate.notNull(zipContents, ImportMigrationContextImpl.INVALID_NULL_ZIP);
```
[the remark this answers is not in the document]
oconnormi: Removing

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -68,55 +69,57 @@
...
    * </code> is <code>null</code>
    * @throws java.io.IOError if unable to determine ${ddf.home}
    */
-  public ImportMigrationContextImpl(MigrationReport report, ZipFile zip, Migratable migratable) {
+  public ImportMigrationContextImpl(
+      MigrationReport report, Map<ZipEntry, byte[]> zipContents, Migratable migratable) {
     super(report, migratable);
-    Validate.notNull(zip, ImportMigrationContextImpl.INVALID_NULL_ZIP);
-    this.zip = zip;
+    Validate.notNull(zipContents, ImportMigrationContextImpl.INVALID_NULL_ZIP);
```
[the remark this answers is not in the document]
oconnormi: Removing

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -263,7 +266,7 @@ InputStream getInputStreamFor(ImportMigrationEntryImpl entry, boolean checkAcces
         sm.checkRead(entry.getFile().getPath());
       }
     }
-    final InputStream is = zip.getInputStream(entry.getZipEntry());
+    final InputStream is = new ByteArrayInputStream(zipContents.get(entry.getZipEntry()));
```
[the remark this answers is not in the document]
oconnormi: Removing exception

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
oconnormi: Yeah that would make more sense, I'll move the filtering out to the `ZipUtils` class instead

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
oconnormi: I meant to ask you about that today actually 😄 . I had a feeling that was why it was done this way originally, but I wanted to check before I just went and changed it.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
```diff
@@ -72,6 +72,33 @@
   public static final String IMPORT_MISMATCH_DDF_HOME_ERROR =
       "Import error: mismatched ddf.home [%s]; expecting local system to be installed under [%s].";
 
+  public static final String IMPORT_KEY_OPEN_ERROR = "Import error: failed to open key file [%s].";
+
+  public static final String IMPORT_KEY_DECODE_ERROR = "Import error: could nod decode key [%s].";
+
+  public static final String IMPORT_KEY_PADDING_ERROR =
```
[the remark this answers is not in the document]
oconnormi: Yeah I already wasn't a fan of the key padding and algorithm errors, since there should be no way of causing them to occur. The only one that should be possible to cause without making code changes is the `INVALID_KEY` error which could occur if the key was corrupted somehow, or if the zip was actually encrypted with a different key. I'm thinking i'll just change the padding and algorithm errors to log statements and keep them stored in the `ZipUtils` class instead

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
oconnormi: yep, I like that idea 😄

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
oconnormi: Wasn't sure at the time, removing.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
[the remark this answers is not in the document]
oconnormi: It should be, it has a `hashCode()` method at any rate

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,207 @@
...
+   * {@link ZipEntry} is stored inside a key corresponding to each {@link ZipEntry}'s name
+   *
+   * @param in the {@link ZipInputStream} to retrieve the entries from
+   * @return the stream of entries
+   */
+  public Map<ZipEntry, byte[]> entriesFrom(ZipInputStream in) {
+    ZipEntry entry;
+    HashMap<ZipEntry, byte[]> entries = new HashMap<>();
+    try {
+      while ((entry = in.getNextEntry()) != null) {
```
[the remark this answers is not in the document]
oconnormi: Will add this 👍

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,207 @@
...
+   * @throws MigrationException when an invalid key padding is used
+   * @throws MigrationException when and invalid cipher algorithm is used
+   * @return a {@link Cipher} in encrypt mode
+   */
+  public Cipher createEncryptionCipher(Path keyPath) {
+    try {
+      Cipher cipher = Cipher.getInstance(CIPHER_ALGORITHM);
+      cipher.init(Cipher.ENCRYPT_MODE, createSecretKey(keyPath));
+      return cipher;
+    } catch (InvalidKeyException e) {
```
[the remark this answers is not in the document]
oconnormi: I think the `InvalidKey` error is actually one of the few in the decryption process that could be caused by external factors (tampering, file corruption, etc) while the other ones in here couldn't be caused by an admin's actions. Thoughts?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,207 @@
...
+   */
+  public SecretKey createSecretKey(Path keyPath) {
+    KeyGenerator keyGenerator = null;
+    try {
+      keyGenerator = KeyGenerator.getInstance(KEY_ALGORITHM);
+      SecretKey secretKey = keyGenerator.generateKey();
+      char[] hexKey = encodeHex(secretKey.getEncoded());
+      writeStringToFile(keyPath.toFile(), String.valueOf(hexKey), Charsets.UTF_8);
+      return secretKey;
+    } catch (NoSuchAlgorithmException e) {
```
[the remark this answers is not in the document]
oconnormi: Yeah, I think that is sorta what I was trying to say above. The algorithm errors can probably be logged since they shouldn't actually occur.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
oconnormi: yep, will add this now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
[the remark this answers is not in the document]
paouelle: ❓ Why would NoSuchAlgorythm be triggered by something the admin do? And even if it was, do we need the console output to be that specific? I think logs would be sufficient but for the user, indicating more something to what they have do would be.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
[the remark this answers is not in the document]
paouelle: Just have those 2 other exception bubble out as system errors so they end up being caught in the cfg mgr and log as internal errors

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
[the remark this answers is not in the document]
paouelle: Yup, see above reply

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
```diff
@@ -120,22 +129,21 @@ public ImportMigrationManagerImpl(
...
+      throws NoSuchAlgorithmException, NoSuchPaddingException {
     Validate.notNull(exportFile, "invalid null export file");
+    ZipInputStream zis = null;
     try {
-      return new ZipFile(exportFile.toFile());
-    } catch (FileNotFoundException e) {
-      throw new MigrationException(Messages.IMPORT_FILE_MISSING_ERROR, exportFile, e);
-    } catch (SecurityException | IOException e) {
-      throw new MigrationException(Messages.IMPORT_FILE_OPEN_ERROR, exportFile, e);
+      zis = getCipherZipInputStreamFor(exportFile);
```
clockard: Could we just do a try with resource here instead of the finally block

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -200,6 +206,10 @@ private MigrationReportImpl doExport(
       report.record(new MigrationException(Messages.EXPORT_SECURITY_ERROR, exportFile, e));
     } catch (RuntimeException e) {
       report.record(new MigrationException(Messages.EXPORT_INTERNAL_ERROR, exportFile, e));
+    } catch (NoSuchPaddingException e) {
+      report.record(new MigrationException(Messages.EXPORT_INTERNAL_ERROR, keyFile, e));
+    } catch (NoSuchAlgorithmException e) {
```
coyotesqrl: ✏️ This branch is identical to the one above it and could be collapsed.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -231,14 +246,23 @@ private MigrationReport doImport(
...
       delegateToImportMigrationManager(report, exportFile);
     } catch (MigrationException e) {
       report.record(e);
     } catch (SecurityException e) {
       report.record(new MigrationException(Messages.IMPORT_SECURITY_ERROR, exportFile, e));
     } catch (RuntimeException e) {
       report.record(new MigrationException(Messages.IMPORT_INTERNAL_ERROR, exportFile, e));
+    } catch (NoSuchPaddingException e) {
+      report.record(new MigrationException(Messages.IMPORT_INTERNAL_ERROR, keyFile, e));
+    } catch (NoSuchAlgorithmException e) {
```
coyotesqrl: ✏️ This branch identical to the one above it.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -231,14 +246,23 @@ private MigrationReport doImport(
...
+      report.record(new MigrationException(Messages.IMPORT_INTERNAL_ERROR, keyFile, e));
+    } catch (IOException e) {
```
coyotesqrl: ✏️ This branch identical to the `RuntimeException` branch.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -210,6 +220,11 @@ private MigrationReportImpl doExport(
       } else {
         SecurityLogger.audit("Failed to delete exported file {}", exportFile);
       }
+      if (FileUtils.deleteQuietly(keyFile.toFile())) {
```
coyotesqrl: ❓ Any reason you're not using your private `deleteQuietly` method for these deletions-with-audits?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,284 @@
...
+   * @param in the {@link ZipInputStream} to retrieve the entries from
+   * @return the stream of entries
+   */
+  public static Map<ZipEntry, byte[]> entriesFrom(ZipInputStream in) {
+    ZipEntry entry;
+    Map<ZipEntry, byte[]> entries = new HashMap<>();
+    try {
+      while ((entry = in.getNextEntry()) != null) {
+        if (!entry.isDirectory()) {
+          byte[] content = IOUtils.toByteArray(in);
```
coyotesqrl: ❓ Do we have a ballpark figure on how large the memory footprint of this map will be during its lifetime?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
paouelle: We should not use try-with-resource with input streams. It can throw IOExeptions when closing the streams and we do not care about failing to close once we are done reading.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -148,21 +154,18 @@ public MigrationReport doImport(Path exportDirectory, Consumer<MigrationMessage>
   @SuppressWarnings(
       "squid:S2093" /* try-with-resource will throw IOException with InputStream and we do not care to get that exception */)
```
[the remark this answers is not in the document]
paouelle: It is no longer required

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
paouelle: @coyotesqrl I think it is just simpler since it handles all errors and return false.
❓ That being said, do we need to audit the deletion
❗️ We should make sure to delete all 3 files in case of failures and not just one.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -231,14 +248,23 @@ private MigrationReport doImport(
                 + productVersion
                 + ConfigurationMigrationManager.EXPORT_EXTENSION);
 
+    final Path keyFile = exportDirectory.resolve(ConfigurationMigrationManager.EXPORT_KEY_NAME);
+
     try {
+      Validate.isTrue(checksumIsValid(exportFile));
```
paouelle: ❓ Should we really let an invalid checksum be reported as an internal error below? I think a specific error messages is warranted here.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -248,12 +274,14 @@ private MigrationReport doImport(
       SecurityLogger.audit("Warnings importing configuration settings from file {}", exportFile);
       // don't leave the zip file there if the import succeeded
       deleteQuietly(exportFile.toFile());
```
paouelle: ❓ Should we also delete the checksum file?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -248,12 +274,14 @@ private MigrationReport doImport(
...
+      deleteQuietly(keyFile.toFile());
       report.record(new MigrationWarning(Messages.IMPORT_SUCCESS_WITH_WARNINGS, exportFile));
       report.record(new MigrationWarning(Messages.RESTART_SYSTEM_WHEN_WARNINGS));
     } else {
       SecurityLogger.audit("Exported configuration settings from file {}", exportFile);
       // don't leave the zip file there if the import succeeded
       deleteQuietly(exportFile.toFile());
+      deleteQuietly(keyFile.toFile());
```
paouelle: ❓ How about the checksum file?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -107,14 +109,10 @@ public ExportMigrationManagerImpl(
 
   @SuppressWarnings(
       "squid:S2095" /* up to the caller of this method to close it; exception is thrown if the stream cannot be created in the first place */)
```
paouelle: ❓ Not sure the suppression is still warranted here

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -68,55 +67,54 @@
...
   }
 
   /**
    * Creates a new migration context for an import operation.
    *
    * @param report the migration report where warnings and errors can be recorded
-   * @param zip the zip file associated with the import
+   * @param zipContents the contents of the zip file associated with the import
    * @param migratable the migratable this context is for
    * @throws IllegalArgumentException if <code>report</code>, <code>zip</code> or <code>migratable
```
paouelle: comment is out of date ... no more `zip`

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
```diff
@@ -72,6 +72,21 @@
   public static final String IMPORT_MISMATCH_DDF_HOME_ERROR =
       "Import error: mismatched ddf.home [%s]; expecting local system to be installed under [%s].";
 
+  public static final String IMPORT_KEY_OPEN_ERROR = "Import error: failed to open key file [%s].";
+
+  public static final String IMPORT_KEY_DECODE_ERROR = "Import error: could nod decode key [%s].";
```
paouelle: `key` to `key file`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
```diff
@@ -72,6 +72,21 @@
...
+
+  public static final String IMPORT_KEY_INVALID_ERROR =
+      "Import error: invalid key used for import [%s].";
```
paouelle: `key` to `key file`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
```diff
@@ -72,6 +72,21 @@
...
+
+  public static final String EXPORT_KEY_INVALID_ERROR = "Export error: invalid key used for export";
+
+  public static final String EXPORT_KEY_IO_ERROR =
+      "Export error: failed to write decryption key to [%s].";
```
paouelle: `key` to `key file`?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
paouelle: No bigger than it was when using the ZipFile. But as big as the number of files we export. The biggest zip file I saw so far was about 200KB

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,284 @@
...
+      throw e;
+    } catch (NoSuchPaddingException e) {
+      LOGGER.error(IMPORT_KEY_PADDING_ERROR, keyPath, e);
+      throw e;
+    } catch (InvalidKeyException e) {
+      throw new MigrationException(Messages.IMPORT_KEY_INVALID_ERROR, keyPath, e);
+    }
+  }
+
+  public static String getChecksum(Path path) throws IOException, NoSuchAlgorithmException {
```
paouelle: ❗️ Can we not use PathUtils.getChecksumFor()?

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
```diff
@@ -80,9 +80,9 @@
  * <p>// This is the system imported into: ./ddf ./ddf/etc ./ddf/Version.txt
  *
  * <p>// The backup from the imported system will look similar to this:
- * ./exported-1.0-20170816T142400.zip
+ * ./exported-1.0-20170816T142400.enc
```
paouelle: ❓ `.enc` stands for?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
oconnormi: fixed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
oconnormi: fixed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
oconnormi: fixed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/Messages.java
```diff
@@ -72,6 +72,33 @@
   public static final String IMPORT_MISMATCH_DDF_HOME_ERROR =
       "Import error: mismatched ddf.home [%s]; expecting local system to be installed under [%s].";
 
+  public static final String IMPORT_KEY_OPEN_ERROR = "Import error: failed to open key file [%s].";
+
+  public static final String IMPORT_KEY_DECODE_ERROR = "Import error: could nod decode key [%s].";
```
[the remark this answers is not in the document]
oconnormi: fixed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
oconnormi: it gets removed now on both failure and success

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
oconnormi: Collapsed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
oconnormi: Collapsed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
oconnormi: I updated it so that each of the files is now deleted using the private method

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
oconnormi: Collapsed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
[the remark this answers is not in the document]
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
oconnormi: fixed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,207 @@
...
+   * @param zipPath the exported zip path
+   * @throws MigrationException when the export zip file can not be created
+   * @return the {@link ZipOutputStream} for the exported zip file
+   */
+  public ZipOutputStream getCipherOutputStreamFor(Path zipPath) {
+    try {
+      return new ZipOutputStream(
+          new BufferedOutputStream(
+              new CipherOutputStream(
+                  new FileOutputStream(zipPath.toFile()),
```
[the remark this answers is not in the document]
oconnormi: Done

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
oconnormi: encrypted

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
oconnormi: I didn't even realize you already had that in there. I can change it

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
paouelle: 🤔  Maybe we should have an extension for DDF archive, or DDF export. Indicating the file is encrypted, is just asking for it.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
paouelle: Less code to maintain and test ;-)

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
oconnormi: Any ideas what we should call it then @paouelle? `.dar` (ddf archive)?

paouelle: That would work with me. @lessarderic What do you think?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
oconnormi: I'm actually going to keep this I think @paouelle , since the PathUtils stuff is all non-static and also uses a different hash algorithm for the checksum method

oconnormi:
build now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
```diff
@@ -0,0 +1,284 @@
...
+    ZipEntry entry;
+    Map<ZipEntry, byte[]> entries = new HashMap<>();
+    try {
+      while ((entry = in.getNextEntry()) != null) {
+        if (!entry.isDirectory()) {
+          byte[] content = IOUtils.toByteArray(in);
+          entries.put(entry, content);
+        }
+      }
+    } catch (IOException e) {
```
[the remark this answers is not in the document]
oconnormi: @paouelle at the moment I'm kinda torn on this one. Part of me feels that it might be better to handle the IOExceptions from reading the next ZipEntry up front. I can also see the benefit of deferring it though. Im curious would you be alright if the deferral of the error handling could be implemented in a follow-on PR so that we can get the first iteration of the encrypted migration into the baseline first? 
If you feel this behavior is essential for this pr however I will work on finishing my implementation of a ZipFile-like class that will wrap the CipherInputStream for each entry instead of creating a ticket to iterate further on this approach

oconnormi:
Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ZipUtils.java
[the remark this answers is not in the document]
paouelle: My issue is that it changes the current design.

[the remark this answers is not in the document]
lessarderic: I agree with @paouelle here. Even though we technically comply with the API's methods, we're changing the intended behavior, which is to abstract how the entries are stored and fail only when they are accessed, not at the very beginning.

If we were to replace the zip with a directory structure for instance, I would not expect the migration to fail at the very beginning if one of the files is missing or corrupted. Only when a migratable tries to access such a file. And even then, it should be up to the migratable to decide whether it should fail or not.

Switching to this implementation changes that behavior, which would have to change again if we were to postpone the change Patrick is asking for.

As a side-note, we may want to document this behavior somewhere in the API to make this clear and ensure that implementations always work consistently.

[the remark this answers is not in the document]
oconnormi: @lessarderic that is a good point about how the behavior would look if the migration was changed to use a directory structure. I really hadn't thought of it that way, I was just thinking of the zip as a single entity. I will continue working on an implementation that will allow for this to work as intended.

oconnormi:
Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -94,13 +105,16 @@ public ExportMigrationManagerImpl(
     Validate.notNull(migratables, "invalid null migratables");
```
paouelle: ❓ Should we validate that the cipher utils is not null too?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
```diff
@@ -71,14 +74,16 @@
...
+  ExportMigrationContextImpl(
+      MigrationReport report, Migratable migratable, ZipOutputStream zos, CipherUtils cipherUtils) {
     super(
         new ExportMigrationReportImpl(report, migratable),
         migratable,
         ExportMigrationContextImpl.validateNotNull(migratable, "invalid null migratable")
             .getVersion());
     Validate.notNull(zos, "invalid null zip output stream");
     this.zipOutputStream = zos;
+    this.cipherUtils = cipherUtils;
```
paouelle: ❓ Should we validate that the cipher utils is not null too?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -160,16 +174,18 @@ public void doExport(String productBranding, String productVersion) {
   public void close() throws IOException {
     if (!closed) {
       this.closed = true;
+      CipherOutputStream cos = cipherUtils.getCipherOutputStream(zipOutputStream);
```
paouelle: ❓ Shouldn't this be done after the putEntry()?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -160,16 +174,18 @@ public void doExport(String productBranding, String productVersion) {
...
         try {
           zipOutputStream.putNextEntry(
               new ZipEntry(MigrationContextImpl.METADATA_FILENAME.toString()));
-          JsonUtils.MAPPER.writeValue(zipOutputStream, metadata);
+          JsonUtils.MAPPER.writeValue(cos, metadata);
         } catch (IOException e) {
           throw new MigrationException(Messages.EXPORT_METADATA_CREATE_ERROR, e);
         }
       } finally {
+        cos.close();
```
paouelle: ❗️ 🤔  Not sure we should be closing this one. If you need to because of the cipher then we should wrap the output stream passed to the cipher output stream using the CloseShieldOutputStream class. If you check above, we created a ProxyOutputStream to protect the zip stream closing. The next line will properly close the entry and the zip stream.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationManagerImpl.java
```diff
@@ -123,6 +120,8 @@ public ImportMigrationManagerImpl(
       // process migratables' metadata
       JsonUtils.getMapFrom(metadata, MigrationContextImpl.METADATA_MIGRATABLES)
           .forEach((id, o) -> getContextFor(id).processMetadata(JsonUtils.convertToMap(o)));
+    } catch (FileNotFoundException e) {
```
paouelle: ❓ Which part of the code now throws this FileNotFoundException? It used to be thrown below when the zip file was created which you are now doing when you are creating the MigrationZipFile which is done before we get here.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipConstants.java
```diff
@@ -0,0 +1,69 @@
...
+
+  private MigrationZipConstants() {}
+
+  /**
+   * Computes the default key location based on the location of the zip file
+   *
+   * @param zipPath path to the zip file
+   * @return the default location of the key
+   */
+  public static Path getDefaultKeyPathFor(Path zipPath) {
```
paouelle: 👍  So it will be XXXX.dar.key?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipConstants.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Computes the default checkusm file location based on the location of the zip file
+   *
+   * @param zipPath path to the zip file
+   * @return the default location of the checksum file
+   */
+  public static Path getDefaultChecksumPathFor(Path zipPath) {
+    return Paths.get(zipPath.toString() + MigrationZipConstants.CHECKSUM_EXTENSION);
```
paouelle: ❓  So it will be XXXX.dar.sha256? Will this be ok for Windows to have an extension of 6 characters?

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationSupport.java
```diff
@@ -83,19 +97,33 @@
...
+  public static Map<String, MigrationZipEntry> getEntriesFrom(Path path) throws IOException {
     return AbstractMigrationSupport.getEntriesFrom(
         new BufferedInputStream(new FileInputStream(path.toFile())));
   }
 
+  public static String decrypt(byte[] data, Path keyPath)
+      throws IOException, DecoderException, NoSuchPaddingException, NoSuchAlgorithmException,
+          InvalidKeyException, InvalidAlgorithmParameterException {
+
+    String keyData = new String(readFileToByteArray(keyPath.toFile()));
```
paouelle: ❗️ That would use the default charset. We should use UTF-8?

paouelle:
Nice version!

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipConstants.java
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param zipPath path to the zip file
+   * @return the default location of the key
+   */
+  public static Path getDefaultKeyPathFor(Path zipPath) {
+    return Paths.get(zipPath.toString() + MigrationZipConstants.KEY_EXTENSION);
+  }
+
+  /**
+   * Computes the default checkusm file location based on the location of the zip file
```
peterhuffer: ✏️ Typo: `checkusm`

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
peterhuffer: ❓ Should these comments be updated to `.dar`?

oconnormi:
Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -84,6 +86,14 @@ public MigrationZipFile(Path zipPath, Path keyPath, Path checksumPath) throws IO
     Validate.notNull(zipPath, "Null zip path");
     Validate.notNull(keyPath, "Null key path");
     Validate.notNull(checksumPath, "Null checksum path");
+    if ((!zipPath.toFile().exists())
+        || (!keyPath.toFile().exists())
+        || (!checksumPath.toFile().exists())) {
+      throw new FileNotFoundException(
+          String.format(
+              "Could not find all required files [%s], [%s], [%s]",
+              zipPath.toString(), keyPath.toString(), checksumPath.toString()));
```
paouelle: ✏️  No need for the `.toString()`

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -84,6 +86,14 @@ public MigrationZipFile(Path zipPath, Path keyPath, Path checksumPath) throws IO
     Validate.notNull(zipPath, "Null zip path");
     Validate.notNull(keyPath, "Null key path");
     Validate.notNull(checksumPath, "Null checksum path");
+    if ((!zipPath.toFile().exists())
```
paouelle: 👍

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -145,6 +155,22 @@ public boolean isClosed() {
...
+  /**
+   * Closes the open {@link ZipFile} and deletes the file, key, and checksum
+   *
+   * @return
+   */
+  public boolean deleteQuitetly() {
+    try {
+      close();
+    } catch (IOException e) {
+      throw new MigrationException("Error closing zip file [%s] [%s].", zipPath, e);
```
paouelle: ❗️ Based on previous comments, we shouldn't be closing here.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ConfigurationMigrationManager.java
```diff
@@ -276,23 +277,19 @@ private MigrationReport doImport(
       report.record(new MigrationException(Messages.IMPORT_INTERNAL_ERROR, zip.getKeyPath(), e));
     }
     report.end();
-    if (report.hasErrors()) {
+    if ((zip == null) || (report.hasErrors())) {
```
paouelle: ❗️  You should re-introduce the closeQuietly for the import manager in the delegateToImportMigrationManager() to make sure the zip file is properly closed.

oconnormi:
Rebuilding due to maven bundle plugin error.

Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -174,18 +176,22 @@ public void doExport(String productBranding, String productVersion) {
...
+          CloseShieldOutputStream shieldOutputStream = new CloseShieldOutputStream(zipOutputStream);
+          cos = cipherUtils.getCipherOutputStream(shieldOutputStream);
           JsonUtils.MAPPER.writeValue(cos, metadata);
         } catch (IOException e) {
           throw new MigrationException(Messages.EXPORT_METADATA_CREATE_ERROR, e);
         }
       } finally {
-        cos.close();
+        if (cos != null) {
+          cos.close();
```
paouelle: ❗️ Since this close might throw an exception, what will happen with the zipOutputStream one? Might want to close it in a finally close. Or if I may suggest, move this close into a finally for the previous try/catch. That way this one will always be called to close the zip output stream.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -145,6 +155,22 @@ public boolean isClosed() {
...
+    }
+    return (FileUtils.deleteQuietly(zipPath.toFile()))
```
paouelle: ❗️ Might want consider to use `&` instead of `&&` to make sure we attempt to delete all 3 before bailing out.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
```diff
@@ -81,9 +81,9 @@
  * <p>// This is the system imported into: ./ddf ./ddf/etc ./ddf/Version.txt
  *
  * <p>// The backup from the imported system will look similar to this:
- * ./exported-1.0-20170816T142400.enc
+ * ./exported-1.0-20170816T142400.dar
```
paouelle: 😄  My mistake for forgetting that one before

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/admin/ConfigurationAdminMigratableTest.java
```diff
@@ -81,9 +81,9 @@
...
  *
- * <p>// Thw exported zip from ddfExport will look similar to this: ./exported-1.0.enc
+ * <p>// Thw exported zip from ddfExport will look similar to this: ./exported-1.0.dar
```
paouelle: ✏️  `The`

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationSupport.java
```diff
@@ -106,7 +105,8 @@ public static String decrypt(byte[] data, Path keyPath)
       throws IOException, DecoderException, NoSuchPaddingException, NoSuchAlgorithmException,
           InvalidKeyException, InvalidAlgorithmParameterException {
 
-    String keyData = new String(readFileToByteArray(keyPath.toFile()));
+    BufferedInputStream bif = new BufferedInputStream(new FileInputStream(keyPath.toFile()));
```
paouelle: 😄  There was a new String() ctor that accepted a charset

oconnormi:
Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /**
+   * Closes the open {@link ZipFile} and deletes the file, key, and checksum
+   *
+   * @return
+   */
+  public boolean deleteQuitetly() {
+    return (FileUtils.deleteQuietly(zipPath.toFile()))
+        & (FileUtils.deleteQuietly(keyPath.toFile()))
```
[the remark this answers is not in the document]
paouelle: Nope. This was intentional ... time to suppress ;-)

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -0,0 +1,248 @@
...
+        & (FileUtils.deleteQuietly(checksumPath.toFile()));
```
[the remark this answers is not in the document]
paouelle: Nope. This was intentional ... time to suppress ;-)

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -0,0 +1,248 @@
...
+        | InvalidAlgorithmParameterException e) {
+      String message = String.format(MigrationZipConstants.KEY_INVALID_ERROR, keyPath, e);
+      LOGGER.error(message);
+      throw new ZipException(message);
+    }
+  }
+
+  private SecretKey initKey(Path keyPath) throws ZipException {
+    BufferedInputStream bif = null;
+    try {
```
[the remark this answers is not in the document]
paouelle: Please don't. We do not care about failing to close the stream when reading. Time to suppress.

oconnormi:
Build Now

oconnormi:
Build Now

clockard:
Manually tested and verified that functionality works as expected. :shipit:

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Creates a {@link Cipher} for use during file encryption and stores the secret key in a file
+   *
+   * @throws MigrationException when an invalid key is provided
+   * @throws MigrationException when an invalid key padding is used
+   * @throws MigrationException when and invalid cipher algorithm is used
+   * @return a {@link Cipher} in encrypt mode
+   */
+  private Cipher createEncryptionCipher() {
```
coyotesqrl: ✏️ Nit: public methods all to the top, then protected/package-private, and then finally private. You've got this here, some public methods below it, then a bunch of private methods, then finally more public.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+   * @throws NoSuchAlgorithmException
+   */
+  public void createZipChecksumFile() throws IOException, NoSuchAlgorithmException {
+    String hash = getChecksum(zipPath);
+    writeStringToFile(checksumPath.toFile(), hash, Charsets.UTF_8);
+  }
+
+  @SuppressWarnings({ //
+    "squid:S2093", /* try-with-resource would require that the input stream is initialized too early */
+    "squid:S2095" /* stream is closed in finally clause */
```
coyotesqrl: ❗️ This suppression is no longer needed/correct.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Creates a secret key and stores it in a file
+   *
+   * @param keyPath the path for storing the secret key
+   * @throws MigrationException when an invalid key algorithm is used
+   * @throws MigrationException when the key can not be written to a file
+   * @return a {@link SecretKey}
+   */
+  private SecretKey createSecretKey(Path keyPath)
+      throws MigrationException, NoSuchAlgorithmException {
```
coyotesqrl: ❗️ Remove `MigrationException` from the throws clause as it is a `RuntimeException`. It can stay in the javadoc.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+      LOGGER.warn(message);
+      throw new MigrationException(message);
+    }
+  }
+
+  /**
+   * Creates a secret key and stores it in a file
+   *
+   * @param keyPath the path for storing the secret key
+   * @throws MigrationException when an invalid key algorithm is used
```
coyotesqrl: ✏️ If we're going to javadoc this method and its possible exceptional paths, we should include an `@throws` declaration for `NoSuchAlgorithmException`.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+      }
+      this.secretKey = key;
+      return key;
+    } else {
+      return createSecretKey(keyPath);
+    }
+  }
+
+  private String loadKeyDataFrom(Path keyPath) {
+    try (BufferedInputStream bif = new BufferedInputStream(new FileInputStream(keyPath.toFile()))) {
```
coyotesqrl: ✏️ It's not of great import one way or the other, but it's a bit of overkill to decorate the stream with a `BufferedInputStream` for a file that's never going to be big.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+      char[] hexKey = encodeHex(secretKey.getEncoded());
+      writeStringToFile(keyPath.toFile(), String.valueOf(hexKey), Charsets.UTF_8);
+      return secretKey;
+    } catch (IOException e) {
+      throw new MigrationException(String.format(MigrationZipConstants.FILE_IO_ERROR, keyPath, e));
+    }
+  }
+
+  private String getChecksum(Path path) throws IOException, NoSuchAlgorithmException {
+    try (FileInputStream fis = new FileInputStream(path.toFile())) {
```
coyotesqrl: ✏️ This one, on the other hand, is going to be pretty big and buffering it might be of value.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
```diff
@@ -0,0 +1,205 @@
...
+  public Path getKeyPath() {
+    return keyPath;
+  }
+
+  public Path getChecksumPath() {
+    return checksumPath;
+  }
+
+  @VisibleForTesting
+  SecretKey getSecretKey() {
```
coyotesqrl: ✏️ I don't love this, but as the secret key is written to disk and it would take a malicious developer with access to the system to access this method, it's probably okay.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -156,20 +172,32 @@ public void doExport(String productBranding, String productVersion) {
     LOGGER.debug("Exported metadata: {}", metadata);
   }
 
+  @SuppressWarnings({ //
+    "squid:S2093", /* try-with-resource would require that the input stream is initialized too early */
+    "squid:S2095" /* stream is closed in finally clause */
```
coyotesqrl: ❗️ This suppression is no longer needed/valid.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -75,32 +81,42 @@
    *     is <code>null</code>
    */
   public ExportMigrationManagerImpl(
-      MigrationReport report, Path exportFile, Stream<? extends Migratable> migratables) {
+      MigrationReport report,
+      Path exportFile,
+      CipherUtils cipherUtils,
+      Stream<? extends Migratable> migratables)
+      throws NoSuchAlgorithmException, NoSuchPaddingException {
```
coyotesqrl: ❗️ These exceptions not thrown.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
```diff
@@ -0,0 +1,255 @@
...
+    return closed;
+  }
+
+  /**
+   * Closes the open {@link ZipFile} and deletes the file, key, and checksum
+   *
+   * @return
+   */
+  @SuppressWarnings("squid:S2178" /* using & to ensure all files are deleted before return */)
+  public boolean deleteQuitetly() {
```
coyotesqrl: ❗️ 1. Really dislike the abuse of the bitwise operator here. It _works_, but at the cost of using a construct that's fallen far out of favor.
2. And...as you're never using the value returned from this method, there's little value in calculating it.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationSupport.java
```diff
@@ -83,19 +97,33 @@
...
+      throws IOException, DecoderException, NoSuchPaddingException, NoSuchAlgorithmException,
+          InvalidKeyException, InvalidAlgorithmParameterException {
+
+    String keyData = new String(readFileToByteArray(keyPath.toFile()), StandardCharsets.UTF_8);
+    byte[] keyBytes = decodeHex(keyData.toCharArray());
+    SecretKey secretKey = new SecretKeySpec(keyBytes, MigrationZipConstants.KEY_ALGORITHM);
+    Cipher cipher = Cipher.getInstance(MigrationZipConstants.CIPHER_ALGORITHM);
+    IvParameterSpec iv = new IvParameterSpec(MigrationZipConstants.CIPHER_IV);
+    cipher.init(Cipher.DECRYPT_MODE, secretKey, iv);
+    InputStream is = new CipherInputStream(ByteSource.wrap(data).openStream(), cipher);
```
coyotesqrl: ❗️ Stream needs to be closed.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
oconnormi: Done

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
oconnormi: Added

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/CipherUtils.java
oconnormi: Added

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
oconnormi: Removed

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
oconnormi: Updated this to be a `void` method and removed the bitwise operators and the related suppression

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationSupport.java
oconnormi: Fixed

oconnormi:
Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/MigrationZipFile.java
paouelle: I am good with changing it to void. 

I wouldn't call the use of `&` versus `&&` as an abuse. It is a standard technique for doing what was required here without complicating the code.

oconnormi:
Build Now

oconnormi:
Forgot a suppression: Build Now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationManagerImpl.java
```diff
@@ -156,20 +169,31 @@ public void doExport(String productBranding, String productVersion) {
     LOGGER.debug("Exported metadata: {}", metadata);
   }
 
+  @SuppressWarnings({
+    "squid:S2093" /* Not using try-with-resources with CipherOutputStream. Stream is closed inside finally block */
```
paouelle: 👍

PR #2494, opened 2017-10-13
[2.11.x] DDF-3297: Create tests for the Security Manager policy

tbatie:
#### Who is reviewing it? 
@paouelle 
@emanns95 
@garrettfreibott 

@coyotesqrl
@lessarderic

TestSecurity test passes

#### Checklist:
- [ ] Documentation Updated
- [ x] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/security/default.policy
+++ b/distribution/ddf-common/src/main/resources/security/default.policy
@@ -19,8 +19,12 @@ deny {
     permission java.security.SecurityPermission "createAccessControlContext";
     permission java.security.SecurityPermission "setPolicy";
     permission java.security.SecurityPermission "insertProvider";
+    permission java.security.SecurityPermission "insertProvider.*";
+    permission java.security.SecurityPermission "removeProvider";
     permission java.security.SecurityPermission "removeProvider.*";
+    permission java.security.SecurityPermission "clearProviderProperties";
     permission java.security.SecurityPermission "clearProviderProperties.*";
+    permission java.security.SecurityPermission "removeProviderProperty";
     permission java.security.SecurityPermission "removeProviderProperty.*";
 
     // Load classes into any protection domain
@@ -107,9 +111,13 @@ grant
     permission java.security.SecurityPermission "createAccessControlContext";
     permission java.security.SecurityPermission "setPolicy";
     permission java.security.SecurityPermission "insertProvider";
+    permission java.security.SecurityPermission "insertProvider.*";
+    permission java.security.SecurityPermission "removeProvider";
     permission java.security.SecurityPermission "removeProvider.*";
+    permission java.security.SecurityPermission "clearProviderProperties";
     permission java.security.SecurityPermission "clearProviderProperties.*";
     permission java.security.SecurityPermission "putProviderProperty.*";
+    permission java.security.SecurityPermission "removeProviderProperty";
     permission java.security.SecurityPermission "removeProviderProperty.*";
 };
 grant
@@ -119,6 +127,7 @@ grant
     permission java.util.PropertyPermission "karaf.restart.jvm", "write";
 
     permission java.security.SecurityPermission "insertProvider";
+    permission java.security.SecurityPermission "insertProvider.*";
 };
 grant
   codeBase "file:${ddf.home.policy}lib/boot/org.osgi.core-6.0.0.jar" {
--- a/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
+++ b/distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
@@ -50,9 +50,13 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.security.AccessControlException;
+import java.security.AccessController;
 import java.security.KeyManagementException;
 import java.security.KeyStore;
 import java.security.NoSuchAlgorithmException;
+import java.security.Policy;
+import java.security.Provider;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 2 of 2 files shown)


distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -19,8 +19,12 @@ deny {
     permission java.security.SecurityPermission "createAccessControlContext";
     permission java.security.SecurityPermission "setPolicy";
     permission java.security.SecurityPermission "insertProvider";
+    permission java.security.SecurityPermission "insertProvider.*";
+    permission java.security.SecurityPermission "removeProvider";
     permission java.security.SecurityPermission "removeProvider.*";
+    permission java.security.SecurityPermission "clearProviderProperties";
```
rzwiefel: ❓  Why do we need this variation of the permission? It doesn't look like it's listed as a SecurityPermission. (neither does line 27 the removeProviderProperty one) https://docs.oracle.com/javase/8/docs/api/java/security/SecurityPermission.html

coyotesqrl: https://docs.oracle.com/javase/8/docs/technotes/guides/security/permissions.html#SecurityPermission

rzwiefel: It is not in there that I see. neither is `removeProviderProperty`

coyotesqrl: Search the page, please.

rzwiefel: It is not there, there is only this form `removeProviderProperty.{provider name}` , which from what I can tell is not the same as what is listed above

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
```diff
@@ -1679,6 +1683,102 @@ public void testAccessGroups() throws Exception {
...
+      fail("Should not have permissions to \"setSecurityManager\"");
+    } catch (AccessControlException e) {
+    }
+
+    try {
+      AccessController.getContext().getDomainCombiner();
+      fail("Should not have permissions to \"getDomainCombiner\"");
+    } catch (AccessControlException e) {
+    }
+
```
lessarderic: ❓ Can we test the `createAccessControlContext` permission?

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
```diff
@@ -1679,6 +1683,102 @@ public void testAccessGroups() throws Exception {
...
+      fail("Should not have permissions to \"clearProviderProperties.*\"");
+    } catch (AccessControlException e) {
+    }
+
+    try {
+      java.security.Security.getProviders()[0].remove("testProperty");
+      fail("Should not have permissions to \"removeProviderProperty.*\"");
+    } catch (AccessControlException e) {
+    }
+  }
```
lessarderic: ❓ Can we test the `createClassLoader` permission?

distribution/ddf-common/src/main/resources/security/default.policy
rzwiefel: They explicitly list when something has a `thing.name` vs just a `thing`, unless this is some implicit secret security policy trick that they don't list in their permissions?

I don't see anything listed here regarding any special forms that would allow this here:
http://docs.oracle.com/javase/8/docs/technotes/guides/security/PolicyFiles.html#FileSyntax

coyotesqrl: I'll look again in the morning when I'm not doing this on my phone.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
```diff
@@ -1679,6 +1683,102 @@ public void testAccessGroups() throws Exception {
     }
   }
 
+  @Test
+  public void testSecurityManagerPolicy() {
+    // TODO: tbatie - 10/12/1 - Because the security policy currently allows access to anything
```
paouelle: ❓ Could we keep them on but put an Assume that would check if the current directory is under user.home in order to skip them? That way, they would run for those (and Jenkins) that do not have their workspace under user.home.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
```diff
@@ -1679,6 +1683,102 @@ public void testAccessGroups() throws Exception {
     }
   }
 
+  @Test
+  public void testSecurityManagerPolicy() {
```
paouelle: ❓ Should these be separate test cases where we expect an exception to come out as opposed to having to catch them with a try/catch?

distribution/ddf-common/src/main/resources/security/default.policy
coyotesqrl: And now is when I apologize for trying to look at this on my phone. You are absolutely correct, @rzwiefel; these additions are incorrect permissions in some cases and do not make sense in others. E.g. there is no permission `clearProviderProperties` without a qualifying provider name; and "the `insertProvider` permission subsumes the `insertProvider.{provider name}` permission".

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
tbatie: I'd prefer to wait until the policy is more restricted to avoid unstable builds.

distribution/ddf-common/src/main/resources/security/default.policy
tbatie: Yeah I got a little carried away there, I'll clean it up.

distribution/test/itests/test-itests-ddf/src/test/java/ddf/test/itests/platform/TestSecurity.java
tbatie: I personally find this more readable than having all the tests broken out. If there is protest, I can change it.

paouelle: I think we should have one method per test. If the first one fail, the second one is never tested.

PR #2496, opened 2017-10-13
DDF-3379 Add support for adding imagery providers as trusted servers

mojogitoverhere:
#### What does this PR do?
Adds a `withCredentials` field to the imagery providers json so each provider can optionally be added as a trusted server in Cesium. Also adds a new option in the Map Layers Configuration UI and changes the Show and Transparent toggle to checkboxes for consistency.
#### Who is reviewing it? 
@djblue @bdeining
@andrewkfiedler
@millerw8

Add a new imagery provider using the Map Layers Configuration UI. Give it a name, set the url to `http://c.tile.openstreetmap.org`, set alpha to 1, set provider type to "Open Street Map", and only check "show" (leave all the other boxes unchecked). Go to `/search/catalog/` and you should see the globe with the map like normal. Now go back into the Map Layers Configuration UI and check "Use Stored Credentials" and save. Refresh the search page and you should see an empty globe without any map (because that URL isn't configured to handle CORS).

Adding a provider as a trusted server sets `withCredentials=true` on the XHRs for the tiles for CORS requests.
[DDF-3379](https://codice.atlassian.net/browse/DDF-3379)
screen shot 2017-10-12 at 6 40 43 pm

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
+++ b/catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
@@ -164,9 +164,9 @@ const ProviderEditor = ({ provider, onUpdate, buffer, onEdit, error = Map() }) =
     </div>
     <div style={{ padding: '0 16px' }}>
       <Flexbox flex='1'>
-        <Flexbox width='250px'>
+        <Flexbox width='290px'>
           <Checkbox
-            label='Use Stored Credentials'
+            label='Allow Credential Forwarding'
             checked={bool(provider.get('withCredentials'))}
             onCheck={(e, value) => {
               onUpdate(value, 'withCredentials')
@@ -174,7 +174,7 @@ const ProviderEditor = ({ provider, onUpdate, buffer, onEdit, error = Map() }) =
         </Flexbox>
         <Flexbox flex='1'>
           {bool(provider.get('withCredentials'))
-            ? <Warning>Requests will fail if the server is not configured to use CORS</Warning>
+            ? <Warning>Requests will fail if the server does not prompt for credentials</Warning>
             : null}
         </Flexbox>
       </Flexbox>
--- a/catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
+++ b/catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
@@ -6,6 +6,7 @@ import { connect } from 'react-redux'
 import CircularProgress from 'material-ui/CircularProgress'
 import ContentAdd from 'material-ui/svg-icons/content/add'
 import DeleteIcon from 'material-ui/svg-icons/action/delete'
+import WarningIcon from 'material-ui/svg-icons/alert/warning'
 import Checkbox from 'material-ui/Checkbox'
 import Dialog from 'material-ui/Dialog'
 import FlatButton from 'material-ui/FlatButton'
@@ -18,7 +19,6 @@ import RaisedButton from 'material-ui/RaisedButton'
 import SelectField from 'material-ui/SelectField'
 import Snackbar from 'material-ui/Snackbar'
 import TextField from 'material-ui/TextField'
-import Toggle from 'material-ui/Toggle'
 import muiThemeable from 'material-ui/styles/muiThemeable'
 import { List, ListItem } from 'material-ui/List'
 import SortUp from 'material-ui/svg-icons/navigation/arrow-upward'
@@ -82,6 +82,16 @@ const Title = muiThemeable()(({ children, muiTheme }) => (
 
 const DeleteIconThemed = muiThemeable()(({ muiTheme }) => (
   <DeleteIcon style={{ color: muiTheme.palette.errorColor }} />
+
+const Warning = muiThemeable()(({ children, muiTheme }) => (
+  <Flexbox style={{ color: muiTheme.palette.warningColor }} alignItems='center'>
+    <div style={{ paddingRight: 10 }}>
+      <WarningIcon color={muiTheme.palette.warningColor} />
+    </div>
+    <div>
+      {children}
+    </div>
+  </Flexbox>
 ))
 
 const implementationSupport = (implementors) => {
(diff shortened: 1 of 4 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/js/controllers/cesium.layerCollection.controller.js
```diff
@@ -60,6 +60,20 @@ define(['underscore',
                 }
 
                 var provider = new type(initObj);
+
+                /*
+                  Optionally add this provider as a TrustedServer. This sets withCredentials = true
+                  on the XmlHttpRequests for CORS.
+                 */
+                if (model.get('withCredentials')) {
```
djblue: Can we not just always add all layers to cesiums trusted servers? How is openlayers working?

andrewkfiedler: I think if it doesn't expect credentials it won't work, so we need to be explicit.

catalog/ui/catalog-ui-search/src/main/webapp/js/controllers/cesium.layerCollection.controller.js
```diff
@@ -60,6 +60,20 @@ define(['underscore',
...
+                  var parsedUrl = new URL(provider.url);
```
andrewkfiedler: ❗ `URL` is not standard and will not work in Internet Explorer.  

http://caniuse.com/#search=URL

catalog/ui/catalog-ui-search/src/main/webapp/js/controllers/cesium.layerCollection.controller.js
```diff
@@ -60,6 +60,20 @@ define(['underscore',
...
+                  var port = parsedUrl.port;
+                  if (!port) {
+                    port = (parsedUrl.protocol === "https") ? 443 : 80;
```
andrewkfiedler: Where are these numbers coming from?

mojogitoverhere: They are just the standard ports for https and http. Cesium won't let you add a trusted server without a specific port.

catalog/ui/catalog-ui-search/src/main/webapp/js/controllers/cesium.layerCollection.controller.js
mojogitoverhere: Any suggestions on what to use instead?

millerw8: Cesium uses (and should include maybe) this: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/ThirdParty/Uri.js

I wonder if we can just piggyback that, since it already needs to work for the TrustedServer stuff to work anyway.

Note in both that logic, and in ours, scheme gets parsed as "https:" or "http:", so we will either need to use a startsWith or a length-1 for our comparison in line 72

djblue: I think you can use `const url = require('url')` instead. Then do something like `url.parse(provider.url)` which returns an object with port, protocol, and hostname.

catalog/ui/catalog-ui-search/src/main/webapp/js/controllers/cesium.layerCollection.controller.js
millerw8: @djblue, @mojogitoverhere tried that originally but as @andrewkfiedler points out, it then broke on other servers when withCredentials=true was added to the XmlHttpRequest object.  That is why we strayed back to explicitly defining them

bdeining:
build now

mojogitoverhere:
build now

catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
```diff
@@ -142,13 +155,30 @@ const ProviderEditor = ({ provider, onUpdate, buffer, onEdit, error = Map() }) =
...
           onCheck={(e, value) => {
             onUpdate(value, 'proxyEnabled')
           }} />
       </Flexbox>
     </div>
+    <div style={{ padding: '0 16px' }}>
+      <Flexbox flex='1'>
+        <Flexbox width='250px'>
+          <Checkbox
+            label='Use Stored Credentials'
```
bdeining: `Allow Cross Origin Credentials` might be more clear here

andrewkfiedler:
Is there any possibility of doing the check for administrators when they are configuring the layers?  It would be great if we could avoid needing the checkbox at all.

mojogitoverhere:
@andrewkfiedler what would that look like? I'm imagining pulling a tile on save to check if the server prompts for credentials. If it does, then automatically set the flag so it can be added as a trusted server. Does that sound reasonable?

catalog/admin/module/catalog-admin-module-maplayers/src/main/webapp/map-layers/index.js
mojogitoverhere: The more I think about, the more I think that this doesn't have anything to do with CORS at all (although that's the error I was getting in my browser during testing), especially after reading @millerw8 's description https://github.com/codice/ddf/pull/2496#pullrequestreview-70053008. So `Allow Cross Origin Credentials` also doesn't sound right.

mojogitoverhere: What about `Allow Credential Forwarding` with the warning message `Requests will fail if the server does not prompt for credentials`?

bdeining: 👍

mojogitoverhere: Done

bdeining:
build now

andrewkfiedler:
@mojogitoverhere were you able to try out the auto check?

mojogitoverhere:
@andrewkfiedler no I haven't tried it. It's kind of difficult since there's no publicly available tile server with dummy authentication to test against. I would most likely have to setup my own server for it, which could be time intensive (I've never done it before). In the interest of time, can we leave this as is for now and create a new ticket to investigate this later? I know it's not the ideal solution but it's the quickest path forward.

bdeining:
Given the amount of effort to do the auto check, it might be a little out of scope for the original ticket.  Wrote a separate ticket for it here : https://codice.atlassian.net/browse/DDF-3407

PR #2498, opened 2017-10-13
DDF-3346 Added OutOfMemory and VM error support on Linux

paouelle:
#### What does this PR do?
This PR adds out of memory and VM error support to automatically restart DDF. It also fixes the service wrapper documentation by simplifying the installation process. It also fixes the restart support from the UI or from migration:import when running using the service wrapper.
Changed escape string from backslash to `_DO_NOT_EXPAND_` when processing files in ddf-common to allow the backslash from actually being used to escape env vars in bash scripts. Upgraded the maven resource plugin to fix a corruption that was happening in ddf-common randomly on some of the filtered files.
#### Who is reviewing it? 
@figliold
1. Unzip and install the distribution on Linux
2. Go through the installation and make sure that the restart step at the end still works
3. Force an out of memory (see below). The process should restart
4. run a migration:export followed by a migration:import , the process should restart automatically
5. Go through the documentation to install the service wrapper
6. re-do step #2 and #4 via the client tool
7. Make sure that the setenv-XXX-wrapper.conf files are exported too.

The only way to verify the out of memory is to manually modify the code somewhere to create a leak. Something like this:
```final List list = new ArrayList();

      try {
        while (true) {
          list.add(new int[Integer.MAX_VALUE]);
        }
      } catch (OutOfMemoryError e) { // swallow
      }
```
in the code of a command that you can trigger from the shell would do the trick.

This ticket addresses both the ddf script and the wrapper service on Linux. The design is to use a tag or command file in between the normal starting execution path and the error path. The error path will either touch that file when invoked as a script or add a RESTART command when invoked via the wrapper as soon as an error is detected. 

When running as a script, we introduce a new ddf_exec script in between karaf's script and the JVM which monitors that restart.jvm file and if it is present when the JVM exist, it changes the exit code to 10 (10 is used by Karaf as a special way to tell the karaf script to restart). When started via the wrapper, the wrapper is configured to monitor that restart.jvm file every 5 seconds and process all commands that are written to it. In our case, the error script simply writes out RESTART in the file to get the wrapper to restart the JVM.

Additional code in the error script is added to monitor for up to 1 minute the JVM process and if it doesn't stop normally, it then forcibly kills it (i.e. kill -9).

The restart.jvm file should not be writeable by anything inside the JVM.
[DDF-3346](https://codice.atlassian.net/browse/DDF-3346)
#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:


paouelle:
build now

PR #2499, opened 2017-10-13
[WIP] Fix filter plugin

rzwiefel:
@stustison 
@emanns95 
#### Review Comment Legend:


catalog/security/catalog-security-filter/src/main/java/ddf/catalog/security/filter/plugin/FilterPlugin.java
```diff
@@ -244,17 +244,24 @@ public QueryResponse processPostQuery(QueryResponse input) throws StopProcessing
       metacard = result.getMetacard();
       Attribute attr = metacard.getAttribute(Metacard.SECURITY);
       if (!checkPermissions(attr, securityPermission, subject, CollectionPermission.READ_ACTION)) {
+        boolean foundProcessor = false;
         for (FilterStrategy filterStrategy : filterStrategies.values()) {
           FilterResult filterResult = filterStrategy.process(input, metacard);
           if (filterResult.processed()) {
+            foundProcessor = true;
```
rzwiefel: This would be assuming that if a filterStrategy does not mark a result as processed it basically could not handle it.. So this would force filterStrategies to return a processed but empty FilterResult in order to redact a metacard

PR #2507, opened 2017-10-13
[2.11.x] DDF-3383 temporarily modified clean directory() not to delete directories

paouelle:
#### What does this PR do?
It temporarily modifies import's clean directory feature to only delete files and not directories in order to workaround a delete handle still opened issue on Windows which renders us incapable of re-creating a directory we just deleted at import time.

A change in the migration API will be coming up with the same ticket that replace the need for this workaround.
#### Who is reviewing it? 
@tbatie @emanns95 
@figliold
@lessarderic
Perform an export followed by an import on a Windows 2000 server machine.
[DDF-3383](https://codice.atlassian.net/browse/DDF-3383)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
@@ -17,6 +17,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 import ddf.security.common.audit.SecurityLogger;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.LinkOption;
@@ -33,7 +34,6 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.stream.Stream;
 import java.util.zip.ZipFile;
-import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.Validate;
 import org.codice.ddf.migration.ImportMigrationContext;
@@ -185,7 +185,7 @@ public class ImportMigrationContextImpl extends MigrationContextImpl<MigrationRe
       return false;
     }
     try {
-      FileUtils.cleanDirectory(fdir);
+      ImportMigrationContextImpl.cleanDirectory(fdir);
       SecurityLogger.audit("Deleted content of directory {}", fdir);
     } catch (IOException e) {
       SecurityLogger.audit("Error deleting content of directory {}", fdir);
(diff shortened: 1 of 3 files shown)


paouelle:
There is one ... it leaves un-needed directories in there 😄 
By defaults, a customer would not be creating additional directories in those we recursively export and import so there should actually be no left over directories. But even if there are, there will be no side effect for now.

paouelle:
build now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -196,6 +196,47 @@ public boolean cleanDirectory(Path path) {
     return true;
   }
 
+  /**
+   * Temporary version of cleanDirectory() that doesn't delete any directories but only files found.
```
lessarderic: 👍 Great comment! Do we have a ticket for a more permanent solution? If so, could we add it to the comment?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -196,6 +196,47 @@ public boolean cleanDirectory(Path path) {
...
+    for (final File file : files) {
+      try {
+        if (file.isDirectory()) {
+          ImportMigrationContextImpl.cleanDirectory(file);
+        } else {
+          final boolean exists = file.exists();
+
+          if (!file.delete()) {
+            if (!exists) {
+              throw new FileNotFoundException("file does not exist: " + file);
```
lessarderic: ❓Do we care if the file couldn't be deleted and it doesn't exist anymore when we check?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -196,6 +196,47 @@ public boolean cleanDirectory(Path path) {
...
+            }
+            throw new IOException("unable to delete file: " + file);
+          }
+        }
+      } catch (IOException e) {
+        ioe = e;
```
lessarderic: ❓ We're ok with only reporting the last error that occurred? What if there are multiple failures and the one that really matters to figure out what the problem is was the first one? Should we keep track of the list of files that caused problem and put that list in the `IOException` being thrown out?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
```diff
@@ -196,6 +196,47 @@ public boolean cleanDirectory(Path path) {
...
+        if (file.isDirectory()) {
+          ImportMigrationContextImpl.cleanDirectory(file);
+        } else {
+          final boolean exists = file.exists();
+
+          if (!file.delete()) {
+            if (!exists) {
+              throw new FileNotFoundException("file does not exist: " + file);
+            }
+            throw new IOException("unable to delete file: " + file);
```
lessarderic: ❓ Should we log something when these happen?

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
paouelle: Same ticket different PR. This is just a first quick solution.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
paouelle: This is an almost identical implementation of Apache's cleanDirectory(). The calling code is handling that case already.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
paouelle: Handled by the calling code. This is meant to replace Apache's cleanDirectory() method so I do not want t log in here and start complicating this method since it will go away with the next PR for this ticket.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ImportMigrationContextImpl.java
paouelle: Again, almost identical copy of Apache's cleanDirectory(). At this point, I don't think it makes much difference if you report the first or the last. If this was meant to stay here for a longer period, I would probably record in the migration report all errors and not just one.

paouelle:
build now

figliold:
Hero'd successfully on a Window Server 2012 VM.

PR #2527, opened 2017-10-18
[2.11.x] DDF-3385 Updated platform-filter-delegate to auto install

emanns95:
#### What does this PR do?
Changes the `platform-filter-delegate` feature to auto install. This helps alleviate an issue where it is not available for its dependencies during startup, which could cause ingest and actions to fail. This was observed most often in Windows Server 2012.

#### Who is reviewing it? 
@vinamartin 
@mackncheesiest 

@clockard
@lessarderic
@figliold 

Verify that installing DDF on Windows Server 2012 will have ingest and actions be available without requiring a restart.

[DDF-3385](https://codice.atlassian.net/browse/DDF-3385)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/platform/platform-app/src/main/resources/features.xml
+++ b/platform/platform-app/src/main/resources/features.xml
@@ -1046,7 +1046,7 @@
         <bundle>mvn:ddf.platform/platform-parser-xml/${project.version}</bundle>
     </feature>
 
-    <feature name="platform-filter-delegate" install="manual" version="${project.version}">
+    <feature name="platform-filter-delegate" install="auto" version="${project.version}">
         <feature prerequisite="true">commons</feature>
         <feature prerequisite="true" version="${project.version}">kernel</feature>
         <feature dependency="true" version="${spring.version}_1">spring</feature>


emanns95:
build now

platform/platform-app/src/main/resources/features.xml
```diff
@@ -1046,7 +1046,7 @@
         <bundle>mvn:ddf.platform/platform-parser-xml/${project.version}</bundle>
     </feature>
 
-    <feature name="platform-filter-delegate" install="manual" version="${project.version}">
+    <feature name="platform-filter-delegate" install="auto" version="${project.version}">
```
clockard: I was talking with @emmberk about this and we don't think this will make the feature start any sooner. The auto value is only used when a kar file is dropped in or by our own application installer but this won't affect the boot feature startup time/order.

emanns95: Then I can't figure out why I haven't been able to reproduce the issue with this change but I can as soon as I revert it 😭

emmberk: Following up for our conversation, after double checking with the Documentation (see section `4.10.12. Deployer` in https://karaf.apache.org/manual/latest/), I wouldn't think that `install="auto"` would make a difference here because we're not hot deploying the features xml file. I'd like to see other solutions explored, like testing out different configurations of prereqs and boot features. On the other hand, `install="auto"` might be working for some reason that I don't understand, and I don't think it would hurt anything to merge this in (with a code comment) if it seems to fix the issue.

lessarderic:
A different PR will be created to address this issue. Closing.

PR #2542, opened 2017-10-23
[2.11.x] DDF-2644 Silence warnings about attribute definition discrepancies

clockard:
#### What does this PR do?
Silence warnings about attribute definition discrepancies. Users can't do anything about these warnings and it doesn't affect the functionality so remove the warning from the user interface. 

#### Who is reviewing it? 
@mcalcote @djblue 

@andrewkfiedler
@lessarderic
@shaundmorris

Install and ingest a product that has known issues with attribute definitions and verify no warning message is shown.
[DDF-2644](https://codice.atlassian.net/browse/DDF-2644)

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/property/property.collection.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/property/property.collection.view.js
@@ -220,19 +220,8 @@ define([
                         if (!metacardDefinitions.metacardTypes[property.id].multivalued){
                             if (value.sort === undefined){
                                 value = [value];
-                            } else {
-                                announcement.announce({
-                                    title: 'Conflicting Attribute Definition',
-                                    message: property.id+' claims to be singlevalued by definition, but the value on the result is not.  If this problem persists, contact your Administrator.',
-                                    type: 'warn'
-                                });
                             }
                         } else if (value.sort === undefined){
-                            announcement.announce({
-                                title: 'Conflicting Attribute Definition',
-                                message: property.id+' claims to be multivalued by definition, but the value on the result is not.  If this problem persists, contact your Administrator.',
-                                type: 'warn'
-                            });
                             value = [value];
                         }
                     } else {


clockard:
@andrewkfiedler What would an admin be able to do about this? I think attribute definition issues probably will require code changes so we shouldn't bother the users or admins with it.

lessarderic:
@clockard, @andrewkfiedler, could we remove the warning if
1. the attribute is marked as multi-valued but is not, or
2. if the attribute is marked as single-valued and contains different, non-blank values
I think this would address the current problem where we show a message that is of no consequences to Intrigue users but makes it look like it's a big problem, but would keep showing up for cases when a single value attribute contains multiple values that are different. Thoughts?

rzwiefel:
@clockard We should **not** have any conflicting definitions by default. That means that if there are any conflicting its one that the system has been configured with (possibly incorrectly) and the admin should investigate.

rzwiefel:
It might actually be nice if we could leave the notifications but only enable them in "experimental features" mode (or dev mode if we had one)

andrewkfiedler:
In the past this has always been due to misconfigured definitions.  It will cause issues when users try to edit metacards.

clockard:
I have been seeing these right out of the box on downstream projects. With attribute injection I could see how you might cause this issue but it still shouldn't be shown to a user in Intrigue I think. If there is an attribute discrepancy I would think that should be handled by the attribute injector code. This could be included in a larger pr to fix the injector code but I wanted to make sure we at least removed the popup warnings by next release.

rzwiefel:
If they are happening out of the box on downstream projects - There is a problem with the downstream project we should fix. 

We definitely should implement them as admin warnings if we do remove them from here.

rzwiefel:
However I do agree this is more of an admin actionable item over a user -- though since it has **real impacts** on the user experience we either need to keep it here or move it to the admin UI.

clockard:
Agree that we still need to fix the downstream project and we should add some kind of admin warning. For the UI could we just disable editing for any attribute that has a definition conflict?

lessarderic:
@andrewkfiedler, @clockard, @rzwiefel, does this sounds like what we want to do here: (1) fix the root cause of the problem, (2) let the admins know when such a problem exists, and (3) only report this problem to Intrigue users when it actually affects them?

If we all agree on the above, do we need to have all of those resolved as part of the same PR? The current message looks pretty bad and isn't helpful to Intrigue users. So could we have this PR address that problem first? Either we make it show up only in more specific cases (see my comment above), or show up only when a user tries to edit a metacard?

We can leave the ticket open and take care of the other 2 things in separate PRs?

andrewkfiedler:
I have some free time today so I'll take a stab at tackling this.

andrewkfiedler:
Here is the ticket I'll be working off of: https://codice.atlassian.net/browse/DDF-3410

clockard:
Closing this pr since @andrewkfiedler is working on a solution.

PR #2310, opened 2017-09-12
DDF-3278 Updating Async Processing Framework with LazyProcessResource

gordocanchola:
Adding markMetacardAsModified to the ProcessResourceItem interface.
Changing InMemoryProcessingFramework's storeProcessRequest to handle the case where a resource exist but only the metacard was modified.
Changing the way we close ProcessResource input streams in InMemoryProcessingFramework to use the ProcessResource's new close method
  The original way of getting the input stream and closing it directly was causing the lazy loaded process resource to load the resource, just to close the stream.

Adding a LazyProcessResource that will only load the resource when it is accessed
  the LazyProcessResources is passed a Supplier that will be used to load the resource
Changing the ProcessingPostIngestPlugin to use the LazyProcessResource instead of ProcessResource that is loading all resources

Adding a close() method to the ProcessResource interface to be called for closing the resource's input stream

#### What does this PR do?
This PR adds a lazy loaded process resource to the async processing framework.
Also makes a small change to handle the case where a PostProcessPlugin updates the metacard but leaves the resource untouched. 

#### Who is reviewing it? 
@peterhuffer @troymohl @emmberk  @dcruver 
(please choose AT LEAST two reviewers that need to approve the PR before it can get merged)
#### Select relevant component teams: 
https://github.com/orgs/codice/teams
#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)
@jlcsmith 
@pklinef 
@clockard 
#### What are the relevant tickets?
[DDF-3278](https://codice.atlassian.net/browse/DDF-3278)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests
#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2557, opened 2017-10-25
[2.11.x] DDF-3278 Async Processing Framework

gordocanchola:
2.11.x backport of PR #2310 

Adding markMetacardAsModified to the ProcessResourceItem interface.
Changing InMemoryProcessingFramework's storeProcessRequest to handle the case where a resource exist but only the metacard was modified.
Changing the way we close ProcessResource input streams in InMemoryProcessingFramework to use the ProcessResource's new close method
  The original way of getting the input stream and closing it directly was causing the lazy loaded process resource to load the resource, just to close the stream.

LazyProcessResourceImpl:
Adding a LazyProcessResource that will only load the resource when it is accessed
  the LazyProcessResources is passed a Supplier that will be used to load the resource
Changing the ProcessingPostIngestPlugin to use the LazyProcessResource instead of ProcessResource that is loading all resources
Changing the lazy loaded processResource to only get the resource when mimeType, name, and inputStream are accesses.
Changing the qualifier to be pulled from the URI
Throwing InaccessibleResourceException on failed resource load

Adding a close() method to the ProcessResource interface to be called for closing the resource's input stream

ProcessResourceImpl
  Removing the constructors that take the modified flag in favor of manually marking the resource as modified using the new markAsModified method

ProcessingPostIngestPlugin:
Updating to populate ProcessResource with size and uri read from metacard

Adding InaccessibleResourceException to be thrown when a resource fails to load

Updating tests

@dcruver 
@jlcsmith 
@pklinef

### DIFF
--- a/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/InaccessibleResourceException.java
+++ b/catalog/async/catalog-async-data-api/src/main/java/org/codice/ddf/catalog/async/data/api/internal/InaccessibleResourceException.java
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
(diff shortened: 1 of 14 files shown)


gordocanchola:
build now

PR #2564, opened 2017-10-26
[2.11.x] DDF-3408 Removes some low-risk/low-priority SonarQube suppressions

coyotesqrl:
#### What does this PR do?
Resolves some low-risk/low-priority sonarqube findings that had been suppressed.

#### Who is reviewing it? 
@brjeter @paouelle @Lambeaux 

@clockard
@rzwiefel

Full CI run should be sufficient to validate.

[DDF-3408](https://codice.atlassian.net/browse/DDF-3408)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheStrategy.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheStrategy.java
@@ -36,10 +36,12 @@ public final class CacheStrategy {
       new CacheStrategy(
           "FEDERATED", m -> !m.getSourceId().equals(CacheStrategy.localSourceIdSupplier.get()));
 
-  // squid:S00108 -  Nested blocks of code should not be left empty
-  @SuppressWarnings(
-      "squid:S00108" /*The Consumer being supplied is intended to be a no-op consumer, which disables caching. */)
-  public static final CacheStrategy NONE = new CacheStrategy("NONE", (rs, c) -> {});
+  public static final CacheStrategy NONE =
+      new CacheStrategy(
+          "NONE",
+          (rs, c) -> {
+            /*noop*/
+          });
 
   private BiConsumer<Collection<Result>, Consumer<Metacard>> cacheStrategyFunction;
 
--- a/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
+++ b/platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
@@ -47,7 +47,6 @@ import org.slf4j.LoggerFactory;
  * The export migration context keeps track of exported migration entries for a given migratable
  * while processing an export migration operation.
  */
-@SuppressWarnings("squid:S2160" /* the base class equals() is sufficient for our needs. */)
 public class ExportMigrationContextImpl extends MigrationContextImpl<ExportMigrationReportImpl>
     implements ExportMigrationContext, Closeable {
 
@@ -243,6 +242,27 @@ public class ExportMigrationContextImpl extends MigrationContextImpl<ExportMigra
       return false;
     }
     return true;
+  }
+
+  /**
+   * The superclass implementation is sufficient for our needs.
+   *
+   * @param o the object to check
+   * @return true if equal
+   */
+  @Override
+  public boolean equals(Object o) {
+    return super.equals(o);
+  }
+
+  /**
+   * The superclass implementation is sufficient for our needs.
+   *
+   * @return the hashcode
+   */
+  @Override
+  public int hashCode() {
+    return super.hashCode();
   }
 }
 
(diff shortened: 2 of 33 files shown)


coyotesqrl:
build now

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/CacheStrategy.java
```diff
@@ -36,10 +36,12 @@
       new CacheStrategy(
           "FEDERATED", m -> !m.getSourceId().equals(CacheStrategy.localSourceIdSupplier.get()));
 
-  // squid:S00108 -  Nested blocks of code should not be left empty
-  @SuppressWarnings(
-      "squid:S00108" /*The Consumer being supplied is intended to be a no-op consumer, which disables caching. */)
-  public static final CacheStrategy NONE = new CacheStrategy("NONE", (rs, c) -> {});
+  public static final CacheStrategy NONE =
```
coyotesqrl: I apologize for this promiscuous use of whitespace now.

coyotesqrl:
build now

coyotesqrl:
build now

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
```diff
@@ -244,6 +243,27 @@ private boolean isDirectory(ExportMigrationEntryImpl entry) {
     }
     return true;
   }
+
```
paouelle: ❓ This is really better than simply suppressing that warning?

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationReportSupport.java
```diff
@@ -21,9 +21,7 @@
 import org.junit.Before;
 
 /** Base class for report-type test cases which handles setup of various messages. */
-@SuppressWarnings(
-    "squid:S2187" /* base class for report-type test cases which handles setup of various messages */)
-public class AbstractMigrationReportTest extends AbstractMigrationTest {
+public class AbstractMigrationReportSupport extends AbstractMigrationSupport {
```
paouelle: 🤔  I think that rule is bogus to start with.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
coyotesqrl: I don't love that it makes the formatting ugly, but I'd really like us to avoid developing a habit of suppressing when we don't have to.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationReportSupport.java
coyotesqrl: I think it's a slightly naive rule. For example, I doubt they have one that checks files with the suffix "Spec" for spock tests. However, it's not the worst idea in the world that your files that match a test pattern for JUnit are actually tests.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
paouelle: My issue here is that it adds code that shouldn't be there to start with. This ends up lowering the coverage and it adds maintenance. I would rather keep the suppression. As for not suppressing when we don't need to, I am in agreement. Maybe there are cases we can revisit the set of rules we have.

platform/migration/platform-migration/src/test/java/org/codice/ddf/configuration/migration/AbstractMigrationReportSupport.java
paouelle: If you read the explanation, they say that developers might get confused that there are no tests in the class! Really! The name was pretty straight forward and indicative it was an abstract class and not a test class.

platform/migration/platform-migration/src/main/java/org/codice/ddf/configuration/migration/ExportMigrationContextImpl.java
paouelle: The other part I don't like about doing this here is that the next developer will come in and thing we forgot to check these additional parameters. The suppression comment added clarification to why it was done that way.

coyotesqrl: 1. Sorry. When I replied about formatting, I mistakenly thought your initial comment was on the empty consumer.
2. This does lower coverage by adding lines of code, but that's an arbitrary decrease to be honest. It isn't _semantically_ less covered than it was before; Jacoco just isn't smart enough to know that.
3. I disagree that it adds maintenance. It adds methods that _should_ not need be touched again in order to make explicit the fact that the equality of this class is _not_ dependent on any of the fields present in it. The class-wide suppression could be missed by a developer in the future who might think this class _should_ include its own fields in the calculation of equality/hashcode (as would be the normal case) and might add them, thereby breaking logic. The suppression would then still be present, incorrectly, while the functionality was in fact wrong. Adding these method explicitly makes intent clear.

PR #2579, opened 2017-10-28
[2.11.x] DDF-3410 Updates conflicting definition warning tactic

andrewkfiedler:
Follow on to: https://github.com/codice/ddf/pull/2542

#### What does this PR do?
 - Conflicting definitions are no longer warned about in announcements, they are now warned only when the user is in edit mode.  It's similar to read only attributes, except it shows 'Conflicting Definition'.  The help prompts them to contact the admin.  We do have to disallow editing, since it will appear to succeed but not actually change the attribute correctly.
 - Fixes a few places where we weren't intersecting metacard types.

Who is reviewing it?

@mcalcote @djblue

Choose 2 committers to review/merge the PR.

@lessarderic
@shaundmorris
@clockard 

How should this be tested?
Install and ingest a product that has known issues with attribute definitions and verify no warning message is shown.  Go to the details view and verify that in edit mode the attribute with conflicting definitions is read only (it will say conflicting definitions).  Note that read only will take precedence over conflicting definitions if both are present (although the end result is the same, read only).

screen shot 2017-10-27 at 4 40 57 pm

https://codice.atlassian.net/browse/DDF-3410

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/add-attribute/add-attribute.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/add-attribute/add-attribute.view.js
@@ -29,7 +29,7 @@ function determineMissingAttributes(selectionInterface) {
     var types = _.union.apply(this, selectionInterface.getSelectedResults().map((result) => {
         return [result.get('metacardType')];
     }));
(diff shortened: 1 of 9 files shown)


rzwiefel:
Nice sample data 😉

clockard:
build now

PR #2580, opened 2017-10-28
DDF-3410 Updates conflicting definition warning tactic

andrewkfiedler:
Follow on to: https://github.com/codice/ddf/pull/2542

#### What does this PR do?
 - Conflicting definitions are no longer warned about in announcements, they are now warned only when the user is in edit mode.  It's similar to read only attributes, except it shows 'Conflicting Definition'.  The help prompts them to contact the admin.  We do have to disallow editing, since it will appear to succeed but not actually change the attribute correctly.
 - Fixes a few places where we weren't intersecting metacard types.

Who is reviewing it?

@mcalcote @djblue

Choose 2 committers to review/merge the PR.

@rzwiefel
@shaundmorris
@clockard 

How should this be tested?
Install and ingest a product that has known issues with attribute definitions and verify no warning message is shown.  Go to the details view and verify that in edit mode the attribute with conflicting definitions is read only (it will say conflicting definitions).  Note that read only will take precedence over conflicting definitions if both are present (although the end result is the same, read only).

screen shot 2017-10-27 at 4 40 57 pm

https://codice.atlassian.net/browse/DDF-3410

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/attributes-rearrange/attributes-rearrange.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/attributes-rearrange/attributes-rearrange.view.js
@@ -25,10 +25,16 @@ var Sortable = require('sortablejs');
 var metacardDefinitions = require('component/singletons/metacard-definitions');
 
 function calculateAvailableAttributesFromSelection(selectionInterface) {
+    var types = _.union.apply(this, selectionInterface.getSelectedResults().map((result) => {
+        return [result.get('metacardType')];
+    }));
+    var possibleAttributes = _.intersection.apply(this, types.map((type) => {
+        return Object.keys(metacardDefinitions.metacardDefinitions[type]);
+    }));
     return selectionInterface.getSelectedResults().reduce(function (currentAvailable, result) {
         currentAvailable = _.union(currentAvailable, Object.keys(result.get('metacard').get('properties').toJSON()));
         return currentAvailable;
-    }, []).filter(function (property) {
+    }, []).filter((attribute) => possibleAttributes.indexOf(attribute) >= 0).filter(function (property) {
         if (metacardDefinitions.metacardTypes[property]) {
             return !metacardDefinitions.metacardTypes[property].hidden;
         } else {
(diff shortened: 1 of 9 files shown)


clockard:
Lets make sure we port this the the 2.11.x branch

clockard:
build now

catalog/ui/catalog-ui-search/src/main/webapp/component/attributes-rearrange/attributes-rearrange.view.js
```diff
@@ -25,10 +25,16 @@ var Sortable = require('sortablejs');
 var metacardDefinitions = require('component/singletons/metacard-definitions');
 
 function calculateAvailableAttributesFromSelection(selectionInterface) {
+    var types = _.union.apply(this, selectionInterface.getSelectedResults().map((result) => {
```
djblue: [_.uniq](http://underscorejs.org/#uniq) might be a better function to use here.

catalog/ui/catalog-ui-search/src/main/webapp/component/property/property.collection.view.js
```diff
@@ -216,23 +216,16 @@ define([
                 metacards.forEach(function(metacard) {
                     var value = metacard[property.id];
                     var isDefined = value !== undefined;
+                    let hasConflictingDefinition = false;
                     if (isDefined) {
                         if (!metacardDefinitions.metacardTypes[property.id].multivalued){
-                            if (value.sort === undefined){
+                            if (value.constructor !== Array){
```
djblue: Are we trying to determine if the value is an array? Would `Array.isArray` be better?

catalog/ui/catalog-ui-search/src/main/webapp/component/attributes-rearrange/attributes-rearrange.view.js
andrewkfiedler: I looked into it a bit, but union still works best.

Unique is good if you end up with a single array, but I'm doing a map which produces multiple arrays.  For example: [ ['type1', 'type2'], ['type1'], ['type1']]

andrewkfiedler:
build now

PR #2593, opened 2017-10-31
DDF-3362 Update AbstractCswSource.alignStream to close resource for IOException

mcalcote:
#### What does this PR do?
In the event that alignStream throws an IOException,  the InputStream passed in will be orphaned and become a resource leak. This change closes the stream when the IOException is caught, before it is re-thrown. 

#### Who is reviewing it? 
@rzwiefel @idperez @stustison @garrettfreibott @shaundmorris 

Review changes

This was identified by a coverity scan.

[DDF-3362](https://codice.atlassian.net/browse/DDF-3362)

### DIFF
--- a/catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
+++ b/catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
@@ -109,6 +109,7 @@ import net.opengis.ows.v_1_0_0.DomainType;
 import net.opengis.ows.v_1_0_0.Operation;
 import net.opengis.ows.v_1_0_0.OperationsMetadata;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.cxf.common.util.CollectionUtils;
 import org.codice.ddf.configuration.SystemBaseUrl;
@@ -1018,6 +1019,9 @@ public abstract class AbstractCswSource extends MaskableImpl
         throw new IOException("Server skipped more bytes than requested in the range header.");
       }
     } catch (IOException e) {
+      /*In the event an IOException is thrown, the InputStream should be closed to prevent resource
+      leakage. Otherwise the stream should be kept open to pass into the ResourceImpl constructor.*/
+      IOUtils.closeQuietly(in);
       throw new IOException(
           String.format(
               "Unable to align input stream with the requested byteOffset of %d",


catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
```diff
@@ -1018,6 +1018,7 @@ private void alignStream(InputStream in, long requestedBytesToSkip, long respons
         throw new IOException("Server skipped more bytes than requested in the range header.");
       }
     } catch (IOException e) {
+      in.close();
```
coyotesqrl: ✏️ I think `IOUtils.closeQuietly` would be more appropriate here so you don't throw an exception from here and fail to generate the exception you want.

mcalcote: Good call, will update.

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
```diff
@@ -1018,6 +1019,7 @@ private void alignStream(InputStream in, long requestedBytesToSkip, long respons
         throw new IOException("Server skipped more bytes than requested in the range header.");
       }
     } catch (IOException e) {
+      IOUtils.closeQuietly(in);
```
bdeining: We only want to close the stream if an exception is thrown?

mcalcote: If the exception is not thrown the stream needs to be passed into the ResourceImpl constructor which assumes responsibility for closing it. This change only prevents the edge case of an exception being thrown and the stream staying open without being passed into that constructor.

bdeining:
build now

mcalcote:
It seems the system didn't fully come up during the itests. This would be a most impressive change to have caused this failure.

mcalcote:
build now

catalog/spatial/csw/spatial-csw-source-common/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/common/source/AbstractCswSource.java
shaundmorris: ❓  Is that worth making a comment in the code, might help with developers who look at this down the road.

mcalcote: I don't think it will hurt. Adding

mcalcote:
build now

PR #2600, opened 2017-10-31
[2.11.x] DDF-3425 Fix UpdateOperations page size

clockard:
#### What does this PR do?
Updates the page size in UpdateOperations that was missed in the ResultsIterator updates
#### Who is reviewing it? 
@emmberk @peterhuffer @AzGoalie 
@coyotesqrl
@rzwiefel

Easiest way to test is to setup to DDFs with registries. 
Publish/Subscribe ddf1 to ddf2. 
Update ddf1 identity node by adding a description and save.
Verify that ddf2 has the updated description in its remote nodes.

[DDF-3425](https://codice.atlassian.net/browse/DDF-3425)

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
@@ -638,7 +638,7 @@ public class UpdateOperations {
         new QueryImpl(
             queryOperations.getFilterWithAdditionalFilters(idFilters, updateRequest),
             1, /* start index */
-            0, /* page size */
+            updateRequest.getUpdates().size(), /* page size */
             null,
             false, /* total result count */
             0 /* timeout */);


catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/UpdateOperations.java
```diff
@@ -638,7 +638,7 @@ private QueryRequestImpl createQueryRequest(UpdateRequest updateRequest) {
         new QueryImpl(
             queryOperations.getFilterWithAdditionalFilters(idFilters, updateRequest),
             1, /* start index */
-            0, /* page size */
+            QueryOperations.MAX_PAGE_SIZE, /* page size */
```
rzwiefel: Will the case where we are updating more items than the MAX_PAGE_SIZE fail? We might consider using the ResultIterable to ensure we get all the results

clockard: Right after I put this up I started wondering the same thing. I will update to use the ResultsIterable

clockard: After looking at it a little more I think instead of using the MAX_PAGE_SIZE I will just use the size of the request for the page size. It doesn't look like we use the ResultIterable within the standard framework so I think this change is consistent with how we handle queries within the framework right now.

clockard:
build now

PR #2605, opened 2017-11-01
DDF-3427 System property for branding in log4j configs

gordocanchola:
Adding a system property for system branding

Changing the log4j config files to use the branding system property for log name
Changing the log pattern to use ISO8601 instead of ABSOLUTE

Changing the karaf logging to display the logger in place of the class and line number

#### What does this PR do?

This PR adds a system property to be used by the log4j config files to generate the log name

#### Who is reviewing it? 
@ryeats @jrnorth 
@clockard 
@pklinef 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:


distribution/ddf-common/src/main/resources/etc/system.properties
```diff
@@ -25,6 +25,11 @@
 #START DDF SETTINGS
 
 #
+# System branding
+#
+org.codice.ddf.system.branding=ddf
```
emmberk: Is there a way to use the branding defined https://github.com/codice/ddf/blob/17be0101150f311503a120f261ca709537ac4cee/pom.xml#L184, instead? Or, at least a way to make sure that this doesn't get out of sync with that property?

distribution/ddf-common/src/main/resources/etc/log4j2.config.xml
```diff
@@ -31,12 +29,12 @@
...
         <Console name="stdout" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{ABSOLUTE} | ${log-pattern}"/>
+            <PatternLayout pattern="%d{ISO8601} | ${log-pattern}"/>
         </Console>
 
         <RollingFile name="out" append="true"
-                     fileName="${sys:karaf.data}/log/ddf.log"
-                     filePattern="${sys:karaf.data}/log/ddf.log-%d{yyyy-MM-dd-HH}-%i.log.gz">
+                     fileName="${sys:karaf.data}/log/${sys:org.codice.ddf.system.branding}.log"
+                     filePattern="${sys:karaf.data}/log/${sys:org.codice.ddf.system.branding}.log-%d{yyyy-MM-dd-HH}-%i.log.gz">
```
emmberk: Using a property here makes a https://github.com/codice/alliance/commit/9adc0b879e15db4a0f7d292497f86e19066a71bd unecessary. These changes should be removed in downstream projects.

distribution/ddf-common/src/main/resources/etc/system.properties
clockard: Could maybe use maven property replacement/filtering to do that.

gordocanchola:
There was some discussion about the log-pattern to use for log files as well as the karaf logger.
I reverted the karaf logger log-patter changes and created a ticket for any log-pattern changes. 
https://codice.atlassian.net/browse/DDF-3437

gordocanchola:
build now

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -24,6 +24,11 @@
 
 #START DDF SETTINGS
 
+#
+# System branding
+#
+org.codice.ddf.system.branding=${branding-lowercase}
```
emmberk: Could this property name change to not include "ddf" so that it can be reused in downstream projects?

PR #2631, opened 2017-11-07
[2.11.x] DDF-3422 Add SessionIndex to SAML LogoutRequest

ahoffer:
What does this PR do?
Implements SessionIndex for SAML. Specifically:
- SecurityAssertionImpl parses a SecurityToken, it copies the SessionIndex from the Security token into its own copy of the AuthnStatement
- The SessionIndex is extracted from an incoming SAML logout request by the LogoutMessage
- The SamlProtocol class copies the SessionIndex from the LogoutRequest it creates
- The SamlValidator checks for the presence of a SessionIndex when it validates a SAML request.
- Added SessionIndex attribute to XML file used in testing
- Changed how SessionIndex is generated in AuthnStatementProvider. SessionIndex can be any string, but standard recommends small-ish integer. It now generate only positive session indexes and reduces the upper size of a session index
- Added unit test to SessionAssertionTest to check for presence of session index
- Updated LogoutMessageSpecTest.groovy to account for session indexes
- Updated LogoutRequestServiceTest to account for session indexes.  Also added 
  
Other
- Fixes a bug in the toString() method of HandlerResult that caused a null pointer exception (only inside of the IDE)
- Fixed some Sonar Lint findings in LogoutRequestService and LogoutReqestServiceTest

Who is reviewing it?
@mcalcote @vinamartin @lessarderic

@rzwiefel 
@stustison 
@figliold 

How should this be tested? (List steps with links to updated documentation)
Login, logout. There should be no errors. 

https://codice.atlassian.net/browse/DDF-3422

Checklist:
[N/A] Documentation Updated
[X] Update / Add Unit Tests
[N/A Update / Add Integration Tests

Review Comment Legend:

✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist.
❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

### DIFF
--- a/platform/security/common/pom.xml
+++ b/platform/security/common/pom.xml
@@ -176,17 +176,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.27</minimum>
+                                            <minimum>0.40</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.24</minimum>
+                                            <minimum>0.34</minimum>
                                         </limit>
                                         <limit>
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.21</minimum>
+                                            <minimum>0.32</minimum>
                                         </limit>
 
                                     </limits>
--- a/platform/security/common/src/main/java/ddf/security/assertion/impl/SecurityAssertionImpl.java
+++ b/platform/security/common/src/main/java/ddf/security/assertion/impl/SecurityAssertionImpl.java
@@ -204,6 +204,9 @@ public class SecurityAssertionImpl implements SecurityAssertion {
                     if (AuthnStatement.AUTHN_INSTANT_ATTRIB_NAME.equals(name)) {
                       authenticationStatement.setAuthnInstant(DateTime.parse(value));
                     }
+                    if (AuthnStatement.SESSION_INDEX_ATTRIB_NAME.equals(name)) {
+                      authenticationStatement.setSessionIndex(value);
+                    }
                   }
                   break;
                 case AuthnContextClassRef.DEFAULT_ELEMENT_LOCAL_NAME:
@@ -868,6 +871,8 @@ public class SecurityAssertionImpl implements SecurityAssertion {
 
     AuthnContext authnContext;
 
+    String sessionIndex;
+
     boolean isNil;
 
     @Override
@@ -882,11 +887,13 @@ public class SecurityAssertionImpl implements SecurityAssertion {
 
     @Override
     public String getSessionIndex() {
-      return null;
+      return sessionIndex;
     }
 
     @Override
-    public void setSessionIndex(String s) {}
+    public void setSessionIndex(String sessionIndex) {
+      this.sessionIndex = sessionIndex;
+    }
 
     @Override
     public DateTime getSessionNotOnOrAfter() {
--- a/platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
+++ b/platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
@@ -96,6 +96,7 @@ public abstract class SamlValidator {
   protected abstract SAMLVersion getSamlVersion();
 
   void checkPostSignature(SignableSAMLObject samlObject) throws ValidationException {
+    // Signing is optional, so null is OK
     if (samlObject.getSignature() != null) {
       try {
         builder.simpleSign.validateSignature(
@@ -287,6 +288,22 @@ public abstract class SamlValidator {
       logoutRequest = (LogoutRequest) builder.xmlObject;
     }
 
+    /**
+     * A SAML LogoutRequest must include at least one SessionIndex, if the logout is initiated by
+     * Session Participant. However, if the Session Authority initiates the logout, a SessionIndex
+     * is optional. In that case, the absence of a SessionIndex means to logout of all sessions
+     * associated with the Principle. Because the SAMLValidator does not have enough information to
+     * determine who initiated the logout, this method does not throw a ValidationException.
+     * Instead, it logs the condition.
+     */
+    protected void checkSessionIndexes() {
+      if (logoutRequest.getSessionIndexes().isEmpty()) {
+        LOGGER.trace(
+            "Logout request does not contain a session index for name-id {}",
+            logoutRequest.getNameID());
+      }
+    }
+
     @Override
     protected SAMLVersion getSamlVersion() {
       return logoutRequest.getVersion();
@@ -387,6 +404,7 @@ public abstract class SamlValidator {
     @Override
     protected void additionalValidation() throws ValidationException {
       checkPostSignature(logoutRequest);
+      checkSessionIndexes();
     }
   }
 
@@ -415,6 +433,7 @@ public abstract class SamlValidator {
     @Override
     protected void additionalValidation() throws ValidationException {
       checkRedirectSignature("SAMLRequest");
+      checkSessionIndexes();
     }
   }
 
--- a/platform/security/common/src/test/java/ddf/security/assertion/impl/SecurityAssertionImplTest.java
+++ b/platform/security/common/src/test/java/ddf/security/assertion/impl/SecurityAssertionImplTest.java
@@ -25,6 +25,8 @@ import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.List;
+import java.util.Optional;
 import java.util.TimeZone;
 import javax.xml.bind.DatatypeConverter;
 import javax.xml.parsers.DocumentBuilder;
@@ -33,6 +35,7 @@ import javax.xml.parsers.ParserConfigurationException;
 import org.apache.cxf.helpers.DOMUtils;
 import org.apache.cxf.ws.security.tokenstore.SecurityToken;
 import org.junit.Test;
+import org.opensaml.saml.saml2.core.AuthnStatement;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.xml.sax.SAXException;
@@ -55,6 +58,8 @@ public class SecurityAssertionImplTest {
   private static final int NUM_NAUTH = 1;
 
   private static final int NUM_AUTHZ = 0;
+
+  private static final String SESSION_INDEX = "42";
 
   public static Document readXml(InputStream is)
       throws SAXException, IOException, ParserConfigurationException {
@@ -100,7 +105,14 @@ public class SecurityAssertionImplTest {
     assertEquals(PRINCIPAL, assertion.getPrincipal().getName());
     assertEquals(PRINCIPAL, assertion.getPrincipal().toString());
     assertEquals(NUM_ATTRIBUTES, assertion.getAttributeStatements().size());
-    assertEquals(NUM_NAUTH, assertion.getAuthnStatements().size());
+    List<AuthnStatement> authnStatements = assertion.getAuthnStatements();
+    assertEquals(NUM_NAUTH, authnStatements.size());
+    assertEquals(
+        (long) NUM_NAUTH, authnStatements.stream().map(AuthnStatement::getSessionIndex).count());
+    Optional<String> sessionIndex =
+        authnStatements.stream().map(AuthnStatement::getSessionIndex).findFirst();
+    assertTrue(sessionIndex.isPresent());
+    assertEquals(SESSION_INDEX, sessionIndex.get());
     assertEquals(
         DatatypeConverter.parseDateTime(BEFORE).getTimeInMillis(),
         assertion.getNotBefore().getTime());
--- a/platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
+++ b/platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
@@ -22,8 +22,8 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 import java.util.UUID;
-import javax.validation.constraints.NotNull;
 import javax.ws.rs.core.UriBuilder;
 import javax.xml.stream.XMLStreamException;
 import org.apache.cxf.helpers.DOMUtils;
@@ -38,6 +38,7 @@ import org.apache.http.impl.client.HttpClients;
 import org.apache.wss4j.common.ext.WSSecurityException;
 import org.apache.wss4j.common.saml.OpenSAMLUtil;
 import org.apache.wss4j.common.util.DOM2Writer;
+import org.codice.ddf.platform.util.uuidgenerator.UuidGenerator;
 import org.codice.ddf.security.common.jaxrs.RestSecurity;
 import org.opensaml.core.xml.XMLObject;
 import org.opensaml.saml.common.SAMLObject;
@@ -55,13 +56,18 @@ public class LogoutMessageImpl implements LogoutMessage {
 
   public static final String SAML_SOAP_ACTION = "http://www.oasis-open.org/committees/security";
 
-  private SystemCrypto systemCrypto;
-
   static {
     OpenSAMLUtil.initSamlEngine();
   }
 
-  public String getIdpSingleLogoutLocation(@NotNull IDPSSODescriptor descriptor) {
+  private SystemCrypto systemCrypto;
+  private UuidGenerator uuidGenerator;
+
+  public LogoutMessageImpl(UuidGenerator generator) {
+    uuidGenerator = generator;
+  }
+
+  public String getIdpSingleLogoutLocation(IDPSSODescriptor descriptor) {
     return descriptor
         .getSingleLogoutServices()
         .stream()
@@ -71,7 +77,7 @@ public class LogoutMessageImpl implements LogoutMessage {
         .orElse("");
   }
 
-  public SignableSAMLObject extractXmlObject(@NotNull String samlLogoutResponse)
+  public SignableSAMLObject extractXmlObject(String samlLogoutResponse)
       throws WSSecurityException, XMLStreamException {
     Document responseDoc =
         StaxUtils.read(
@@ -83,7 +89,7 @@ public class LogoutMessageImpl implements LogoutMessage {
     return null;
   }
 
-  private <T extends SAMLObject> T extract(@NotNull String samlObject, @NotNull Class<T> clazz)
+  private <T extends SAMLObject> T extract(String samlObject, Class<T> clazz)
       throws WSSecurityException, XMLStreamException {
     Document responseDoc =
         StaxUtils.read(new ByteArrayInputStream(samlObject.getBytes(StandardCharsets.UTF_8)));
@@ -95,13 +101,13 @@ public class LogoutMessageImpl implements LogoutMessage {
   }
 
   @Override
-  public LogoutResponse extractSamlLogoutResponse(@NotNull String samlLogoutResponse)
+  public LogoutResponse extractSamlLogoutResponse(String samlLogoutResponse)
       throws XMLStreamException, WSSecurityException {
     return extract(samlLogoutResponse, LogoutResponse.class);
   }
 
   @Override
-  public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse)
+  public LogoutRequest extractSamlLogoutRequest(String samlLogoutResponse)
       throws XMLStreamException, WSSecurityException {
     return extract(samlLogoutResponse, LogoutRequest.class);
   }
@@ -112,16 +118,18 @@ public class LogoutMessageImpl implements LogoutMessage {
    *
    * @param nameIdString NameId of user to log out
    * @param issuerOrEntityId The Issuer of the LogoutRequest
+   * @param sessionIndexes The list of session indexes selected for log out
    * @return the built <code>LogoutRequest</code>
    */
   @Override
   public LogoutRequest buildLogoutRequest(
-      @NotNull String nameIdString, @NotNull String issuerOrEntityId) {
-    return buildLogoutRequest(nameIdString, issuerOrEntityId, UUID.randomUUID().toString());
+      String nameIdString, String issuerOrEntityId, List<String> sessionIndexes) {
+    return buildLogoutRequest(
+        nameIdString, issuerOrEntityId, getUuidGenerator().generateUuid(), sessionIndexes);
   }
 
   public LogoutRequest buildLogoutRequest(
-      @NotNull String nameIdString, @NotNull String issuerOrEntityId, @NotNull String id) {
+      String nameIdString, String issuerOrEntityId, String id, List<String> sessionIndexes) {
     if (nameIdString == null) {
       throw new IllegalArgumentException("Name ID cannot be null");
     }
@@ -131,9 +139,15 @@ public class LogoutMessageImpl implements LogoutMessage {
     if (id == null) {
       throw new IllegalArgumentException("ID cannot be null");
     }
+    if (sessionIndexes == null) {
+      throw new IllegalArgumentException("Session Index collection can be empty, but not null.");
+    }
 
     return SamlProtocol.createLogoutRequest(
-        SamlProtocol.createIssuer(issuerOrEntityId), SamlProtocol.createNameID(nameIdString), id);
+        SamlProtocol.createIssuer(issuerOrEntityId),
+        SamlProtocol.createNameID(nameIdString),
+        id,
+        sessionIndexes);
   }
 
   /**
@@ -144,8 +158,7 @@ public class LogoutMessageImpl implements LogoutMessage {
    * @return LogoutResponse
    */
   @Override
-  public LogoutResponse buildLogoutResponse(
-      @NotNull String issuerOrEntityId, @NotNull String statusCodeValue) {
+  public LogoutResponse buildLogoutResponse(String issuerOrEntityId, String statusCodeValue) {
     return buildLogoutResponse(issuerOrEntityId, statusCodeValue, null);
   }
 
@@ -157,10 +170,7 @@ public class LogoutMessageImpl implements LogoutMessage {
   }
 
   public LogoutResponse buildLogoutResponse(
-      @NotNull String issuerOrEntityId,
-      @NotNull String statusCodeValue,
-      String inResponseTo,
-      @NotNull String id) {
+      String issuerOrEntityId, String statusCodeValue, String inResponseTo, String id) {
     if (issuerOrEntityId == null) {
       throw new IllegalArgumentException("Issuer cannot be null");
     }
@@ -179,14 +189,14 @@ public class LogoutMessageImpl implements LogoutMessage {
   }
 
   @Override
-  public Element getElementFromSaml(@NotNull XMLObject xmlObject) throws WSSecurityException {
+  public Element getElementFromSaml(XMLObject xmlObject) throws WSSecurityException {
     Document doc = DOMUtils.createDocument();
     doc.appendChild(doc.createElement("root"));
     return OpenSAMLUtil.toDom(xmlObject, doc);
   }
 
   @Override
-  public String sendSamlLogoutRequest(@NotNull LogoutRequest request, @NotNull String targetUri)
+  public String sendSamlLogoutRequest(LogoutRequest request, String targetUri)
       throws IOException, WSSecurityException {
     Element requestElement = getElementFromSaml(request);
     String requestMessage = DOM2Writer.nodeToString(requestElement);
@@ -202,26 +212,20 @@ public class LogoutMessageImpl implements LogoutMessage {
   }
 
   @Override
-  public URI signSamlGetResponse(
-      @NotNull SAMLObject samlObject, @NotNull URI target, String relayState)
+  public URI signSamlGetResponse(SAMLObject samlObject, URI target, String relayState)
       throws WSSecurityException, SimpleSign.SignatureException, IOException {
 
     return signSamlGet(samlObject, target, relayState, SSOConstants.SAML_RESPONSE);
   }
 
   @Override
-  public URI signSamlGetRequest(
-      @NotNull SAMLObject samlObject, @NotNull URI target, String relayState)
+  public URI signSamlGetRequest(SAMLObject samlObject, URI target, String relayState)
       throws WSSecurityException, SimpleSign.SignatureException, IOException {
 
     return signSamlGet(samlObject, target, relayState, SSOConstants.SAML_REQUEST);
   }
 
-  private URI signSamlGet(
-      @NotNull SAMLObject samlObject,
-      @NotNull URI target,
-      String relayState,
-      @NotNull String requestType)
+  private URI signSamlGet(SAMLObject samlObject, URI target, String relayState, String requestType)
       throws WSSecurityException, SimpleSign.SignatureException, IOException {
     Document doc = DOMUtils.createDocument();
     doc.appendChild(doc.createElement("root"));
@@ -242,5 +246,9 @@ public class LogoutMessageImpl implements LogoutMessage {
 
   public void setSystemCrypto(SystemCrypto systemCrypto) {
     this.systemCrypto = systemCrypto;
+  }
+
+  public UuidGenerator getUuidGenerator() {
+    return uuidGenerator;
   }
 }
--- a/platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
+++ b/platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
@@ -9,7 +9,6 @@ import org.opensaml.saml.common.SAMLVersion
 import org.opensaml.saml.saml2.core.LogoutRequest
 import org.opensaml.saml.saml2.core.LogoutResponse
 import org.opensaml.saml.saml2.core.StatusCode
-
 import spock.lang.Specification
 
 import java.time.Instant
@@ -30,6 +29,7 @@ class LogoutMessageSpecTest extends Specification {
     final String ISSUER_ID = "MyIssuerId"
 
     final String IN_RESPONSE_TO = "InResponseToID"
+    final String SESSION_INDEX = "MySessionIndex";
 
     void setup() {
         System.setProperty("org.codice.ddf.system.hostname", "localhost")
@@ -44,8 +44,8 @@ class LogoutMessageSpecTest extends Specification {
         copyResourceToFile(encryptionFile, "/encryption.properties")
 
         EncryptionService encryptionService = Mock(EncryptionService.class) {
-            decrypt(_ as String) >> {String data -> data.substring(ENCRYPT_PREFIX.length())}
-            encrypt(_ as String) >> {String data -> "$ENCRYPT_PREFIX$data"}
+            decrypt(_ as String) >> { String data -> data.substring(ENCRYPT_PREFIX.length()) }
+            encrypt(_ as String) >> { String data -> "$ENCRYPT_PREFIX$data" }
         }
 
         System.setProperty("javax.net.ssl.keyStore", jksFile.getAbsolutePath());
@@ -53,12 +53,15 @@ class LogoutMessageSpecTest extends Specification {
         logoutMessage.systemCrypto = new SystemCrypto(encryptionFile.absolutePath,
                 signatureFile.absolutePath,
                 encryptionService)
+        logoutMessage.uuidGenerator = Mock(org.codice.ddf.platform.util.uuidgenerator.UuidGenerator.class) {
+            generateUuid() >> { UUID.randomUUID().toString() }
+        }
     }
 
     void copyResourceToFile(File file, String resource) throws IOException {
-        file.withOutputStream {fileOutputStream ->
+        file.withOutputStream { fileOutputStream ->
             LogoutMessageSpecTest.class.getResourceAsStream(resource).
-                    withStream {resourceStream -> fileOutputStream << resourceStream
+                    withStream { resourceStream -> fileOutputStream << resourceStream
                     };
         };
     }
@@ -69,34 +72,37 @@ class LogoutMessageSpecTest extends Specification {
 
     def "build valid logout request"() {
         when:
-        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(NAME_ID, ISSUER_ID)
-
+        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(NAME_ID, ISSUER_ID, Collections.singletonList(SESSION_INDEX))
 
         then:
         isNotBlank(logoutRequest.ID)
         NAME_ID.equals(logoutRequest.nameID.value)
         ISSUER_ID.equals(logoutRequest.issuer.value)
         SAMLVersion.VERSION_20.equals(logoutRequest.version)
-        now().
-                isAfter(Instant.ofEpochMilli(logoutRequest.issueInstant.millis))
+        logoutRequest.sessionIndexes.size() == 1
+        SESSION_INDEX.equals(logoutRequest.sessionIndexes.get(0).getSessionIndex());
+        now().isAfter(Instant.ofEpochMilli(logoutRequest.issueInstant.millis))
     }
 
     def "build logout request with invalid info"() {
+
         when:
-        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(nameId, issuerId, id)
+        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(nameId, issuerId, id,
+                sessionIndex == null ? null : Collections.singletonList(sessionIndex))
 
         then:
         thrown(IllegalArgumentException)
 
         where: "any param is null"
-        nameId | issuerId | id
-        null   | null     | "xxx"
-        "xxx"  | null     | "xxx"
-        null   | "xxx"    | "xxx"
-        null   | null     | null
-        "xxx"  | null     | null
-        null   | "xxx"    | null
-        "xxx"  | "xxx"    | null
+        nameId | issuerId | id    | sessionIndex
+        null   | null     | "xxx" | "xxx"
+        "xxx"  | null     | "xxx" | "xxx"
+        null   | "xxx"    | "xxx" | "xxx"
+        null   | null     | null  | null
+        "xxx"  | null     | null  | "xxx"
+        null   | "xxx"    | null  | null
+        "xxx"  | "xxx"    | null  | null
+        "xxx"  | "xxx"    | "xxx" | null
     }
 
     def "build logout response with valid info and inResponseTo"() {
@@ -118,7 +124,7 @@ class LogoutMessageSpecTest extends Specification {
     def "build logout response with no inResponseTo"() {
         when:
         LogoutResponse logoutResponse = logoutMessage.buildLogoutResponse("issuer",
-                StatusCode.SUCCESS )
+                StatusCode.SUCCESS)
 
         then:
         logoutResponse.inResponseTo == null
@@ -126,7 +132,7 @@ class LogoutMessageSpecTest extends Specification {
 
     def "verify signed saml request"() {
         setup:
-        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(NAME_ID, ISSUER_ID)
+        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(NAME_ID, ISSUER_ID, Collections.singletonList(SESSION_INDEX))
 
         def target = new URI("https://mytarget.com")
         def relayState = "MyRelayStateGuid"
--- a/platform/security/handler/security-handler-api/pom.xml
+++ b/platform/security/handler/security-handler-api/pom.xml
@@ -109,17 +109,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.39</minimum>
+                                            <minimum>0.40</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.31</minimum>
+                                            <minimum>0.29</minimum>
                                         </limit>
                                         <limit>
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.34</minimum>
+                                            <minimum>0.35</minimum>
                                         </limit>
 
                                     </limits>
--- a/platform/security/handler/security-handler-api/src/main/java/org/codice/ddf/security/handler/api/HandlerResult.java
+++ b/platform/security/handler/security-handler-api/src/main/java/org/codice/ddf/security/handler/api/HandlerResult.java
@@ -13,6 +13,8 @@
  */
 package org.codice.ddf.security.handler.api;
 
+import java.util.Objects;
+
 /**
  * Encapsulates the return status for each handler. Consists of the status of any action taken by
  * the handler (successfully retrieved desired tokens, responded to a client in order to obtain
@@ -67,7 +69,7 @@ public class HandlerResult {
     sb.append("; Source: ");
     sb.append(source);
     sb.append("; Token: ");
-    sb.append(token.toString());
+    sb.append(Objects.isNull(token) ? "null" : token.toString());
     return sb.toString();
   }
 
--- a/platform/security/idp/security-idp-client/pom.xml
+++ b/platform/security/idp/security-idp-client/pom.xml
@@ -137,6 +137,12 @@
             <version>${org.slf4j.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ddf.platform.util</groupId>
+            <artifactId>util-uuidgenerator-api</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <dependencyManagement>
         <dependencies>
@@ -206,17 +212,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.59</minimum>
+                                            <minimum>0.62</minimum>
                                         </limit>
                                         <limit>
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.47</minimum>
+                                            <minimum>0.51</minimum>
                                         </limit>
                                         <limit>
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.48</minimum>
+                                            <minimum>0.52</minimum>
                                         </limit>
                                     </limits>
                                 </rule>
--- a/platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
+++ b/platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
@@ -23,6 +23,7 @@ import ddf.security.http.SessionFactory;
 import ddf.security.samlp.LogoutMessage;
 import ddf.security.samlp.SamlProtocol;
 import ddf.security.samlp.SimpleSign;
+import ddf.security.samlp.SimpleSign.SignatureException;
 import ddf.security.samlp.ValidationException;
 import ddf.security.samlp.impl.HtmlResponseTemplate;
 import ddf.security.samlp.impl.RelayStates;
@@ -34,8 +35,9 @@ import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.List;
+import java.util.stream.Collectors;
 import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpSession;
 import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -56,6 +58,7 @@ import org.apache.wss4j.common.saml.OpenSAMLUtil;
 import org.apache.wss4j.common.util.DOM2Writer;
 import org.codice.ddf.configuration.SystemBaseUrl;
 import org.codice.ddf.security.common.jaxrs.RestSecurity;
+import org.opensaml.saml.saml2.core.AuthnStatement;
 import org.opensaml.saml.saml2.core.LogoutRequest;
 import org.opensaml.saml.saml2.core.LogoutResponse;
 import org.opensaml.saml.saml2.core.StatusCode;
@@ -67,16 +70,31 @@ import org.w3c.dom.Document;
 @Path("logout")
 public class LogoutRequestService {
 
+  public static final String IDP_REALM_NAME = "idp";
+  public static final String NO_SUPPORT_FOR_POST_OR_REDIRECT_BINDINGS =
+      "The identity provider does not support either POST or Redirect bindings.";
+  public static final String ROOT_NODE_NAME = "root";
+  public static final String UNABLE_TO_CREATE_LOGOUT_REQUEST = "Failed to create logout request";
+  public static final String UNABLE_TO_CREATE_LOGOUT_RESPONSE = "Failed to create logout response";
+  public static final String UNABLE_TO_DECODE_AND_INFLATE_LOGOUT_REQUEST =
+      "Unable to decode and inflate logout request";
+  public static final String UNABLE_TO_DECODE_AND_INFLATE_LOGOUT_RESPONSE =
+      "Unable to decode and inflate logout response.";
+  public static final String UNABLE_TO_DECRYPT_LOGOUT_REQUEST =
+      "Failed to decrypt logout request params. Invalid number of params";
+  public static final String UNABLE_TO_PARSE_LOGOUT_REQUEST = "Unable to parse logout request";
+  public static final String UNABLE_TO_PARSE_LOGOUT_RESPONSE = "Unable to parse logout response";
+  public static final String UNABLE_TO_SIGN_LOGOUT_RESPONSE = "Failed to sign logout response";
+  public static final String UNABLE_TO_VALIDATE_LOGOUT_REQUEST =
+      "Unable to validate logout request";
+  public static final String UNABLE_TO_VALIDATE_LOGOUT_RESPONSE =
+      "Unable to validate logout response";
+  public static final String SECURITY_AUDIT_ROLES = "security.audit.roles";
   private static final Logger LOGGER = LoggerFactory.getLogger(LogoutRequestService.class);
-
   private static final String SAML_REQUEST = "SAMLRequest";
-
   private static final String SAML_RESPONSE = "SAMLResponse";
-
   private static final String RELAY_STATE = "RelayState";
-
   private static final String SIG_ALG = "SigAlg";
-
   private static final String SIGNATURE = "Signature";
 
   static {
@@ -115,9 +133,9 @@ public class LogoutRequestService {
             LogoutRequestService.class.getResourceAsStream("/templates/submitForm.handlebars");
         InputStream redirectStream =
             LogoutRequestService.class.getResourceAsStream("/templates/redirect.handlebars")) {
-      submitForm = IOUtils.toString(submitStream);
-      redirectPage = IOUtils.toString(redirectStream);
-    } catch (Exception e) {
+      submitForm = IOUtils.toString(submitStream, StandardCharsets.UTF_8);
+      redirectPage = IOUtils.toString(redirectStream, StandardCharsets.UTF_8);
+    } catch (IOException | RuntimeException e) {
       LOGGER.debug("Unable to load index page for SP.", e);
     }
   }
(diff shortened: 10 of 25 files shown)


ahoffer:
build now

ahoffer:
build now

ahoffer:
build now

platform/security/common/src/main/java/ddf/security/assertion/impl/SecurityAssertionImpl.java
```diff
@@ -204,6 +204,9 @@ private void parseToken(SecurityToken securityToken) {
                     if (AuthnStatement.AUTHN_INSTANT_ATTRIB_NAME.equals(name)) {
                       authenticationStatement.setAuthnInstant(DateTime.parse(value));
                     }
+                    if (AuthnStatement.SESSION_INDEX_ATTRIB_NAME.equals(name)) {
+                      authenticationStatement.setSessionIndex(value);
+                    }
```
ahoffer: This is where the SecuityAssertion copies the session index out of the SecurityToken

platform/security/common/src/main/java/ddf/security/assertion/impl/SecurityAssertionImpl.java
```diff
@@ -868,6 +871,8 @@ public void setNil(Boolean newNil) {
 
     AuthnContext authnContext;
 
+    String sessionIndex;
+
```
ahoffer: New attribute in the AuthenticationSatement class to hold the session index

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
```diff
@@ -287,6 +288,20 @@ private Request(Builder builder) {
...
+     * associated with the Principle.
+     *
+     * @throws ValidationException
+     */
+    protected void checkSessionIndexes() throws ValidationException {
+      if (logoutRequest.getSessionIndexes().isEmpty()) {
+        throw new ValidationException("Logout request does not contain a session index");
+      }
+    }
+
```
ahoffer: In hindsight, this check might not be the best long-term solution. See the comment at the top of the method. Also, The only kind of requests passed to the SamlValidator are logout requests. If other kinds of requests, like queries or auth, they would incorrectly fail validation.

platform/security/common/src/test/java/ddf/security/assertion/impl/SecurityAssertionImplTest.java
```diff
@@ -102,6 +103,9 @@ public void testSampleAssertion() throws Exception {
     assertEquals(NUM_ATTRIBUTES, assertion.getAttributeStatements().size());
     assertEquals(NUM_NAUTH, assertion.getAuthnStatements().size());
     assertEquals(
+        (long) NUM_NAUTH,
+        assertion.getAuthnStatements().stream().map(AuthnStatement::getSessionIndex).count());
+    assertEquals(
```
ahoffer: Enhance unit test for check for session index

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -112,16 +113,21 @@ public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse
...
+   * @param sessionIndexes
    * @return the built <code>LogoutRequest</code>
    */
   @Override
   public LogoutRequest buildLogoutRequest(
-      @NotNull String nameIdString, @NotNull String issuerOrEntityId) {
-    return buildLogoutRequest(nameIdString, issuerOrEntityId, UUID.randomUUID().toString());
+      @NotNull String nameIdString, @NotNull String issuerOrEntityId, List<String> sessionIndexes) {
+    return buildLogoutRequest(
+        nameIdString, issuerOrEntityId, UUID.randomUUID().toString(), sessionIndexes);
```
ahoffer: LogoutRequest constructors now require a session index collection

platform/security/handler/security-handler-api/src/main/java/org/codice/ddf/security/handler/api/HandlerResult.java
```diff
@@ -67,7 +69,7 @@ public String toString() {
     sb.append("; Source: ");
     sb.append(source);
     sb.append("; Token: ");
-    sb.append(token.toString());
+    sb.append(Objects.isNull(token) ? "null" : token.toString());
```
ahoffer: A null token prevented the object from being displayed in the IDE watches/variables view.

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
```diff
@@ -31,6 +30,9 @@ class LogoutMessageSpecTest extends Specification {
 
     final String IN_RESPONSE_TO = "InResponseToID"
 
+    final List SESSION_INDEXES = Collections.emptyList();
+
+
```
ahoffer: Adapted this class to account for session indexes

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -68,16 +72,30 @@
...
+      "Unable to decode and inflate logout response.";
+  public static final String UNABLE_TO_DECRYPT_LOGOUT_REQUEST =
+      "Failed to decrypt logout request params. Invalid number of params";
+  public static final String UNABLE_TO_PARSE_LOGOUT_REQUEST = "Unable to parse logout request";
+  public static final String UNABLE_TO_PARSE_LOGOUT_RESPONSE = "Unable to parse logout response";
+  public static final String UNABLE_TO_SIGN_LOGOUT_RESPONSE = "Failed to sign logout response";
+  public static final String UNABLE_TO_VALIDATE_LOGOUT_REQUEST =
+      "Unable to validate logout request";
+  public static final String UNABLE_TO_VALIDATE_LOGOUT_RESPONSE =
+      "Unable to validate logout response";
```
ahoffer: I did some Sonar Lint cleanup on this class. Most of the changes were to convert strings into constants. This turned out to be important because the unit tests were testing for the exact strings of error messages generated by this class!

At some point I became concerned that refactoring was going to break something, so I stopped.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -139,28 +157,37 @@ public Response sendLogoutRequest(@QueryParam("EncryptedNameIdTime") String encr
...
+        // Logout removes the SAML assertion. This statement must be called before the SAML
+        // assertion is
+        // removed.
+        List<String> sessionIndexes =
+            getIdpSecurityAssertion()
+                .getAuthnStatements()
+                .stream()
+                .map(AuthnStatement::getSessionIndex)
+                .collect(Collectors.toList());
+
```
ahoffer: New code. Collect the session indexes from the AuthnStatements (FYI, normally there is only one AuthnStatement) and pass them to the factory method that creates the LogoutRequest.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -171,23 +198,15 @@ private Response getLogoutRequest(String relayState, LogoutRequest logoutRequest
...
     LOGGER.debug("Configuring SAML LogoutRequest for POST.");
-    Document doc = DOMUtils.createDocument();
-    doc.appendChild(doc.createElement("root"));
-    LOGGER.debug("Signing SAML POST LogoutRequest.");
-    simpleSign.signSamlObject(logoutRequest);
-    LOGGER.debug("Converting SAML Request to DOM");
-    String assertionResponse = DOM2Writer.nodeToString(OpenSAMLUtil.toDom(logoutRequest, doc));
-    String encodedSamlRequest =
-        Base64.getEncoder().encodeToString(assertionResponse.getBytes(StandardCharsets.UTF_8));
+    String encodedSamlRequest = encodeSaml(logoutRequest);
```
ahoffer: Moved some of the code into an `encodeSaml` helper method

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +389,37 @@ private String getEntityId() {
     return String.format("https://%s:%s%s/saml", hostname, port, rootContext);
   }
 
-  private void logout() {
-    HttpSession session = sessionFactory.getOrCreateSession(request);
+  private SecurityAssertion getIdpSecurityAssertion() {
```
ahoffer: Extracted some code into new helper method `getIdpSecurityAssertion`

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +389,37 @@ private String getEntityId() {
...
-    SecurityTokenHolder tokenHolder =
-        ((SecurityTokenHolder) session.getAttribute(SecurityConstants.SAML_ASSERTION));
+    return new SecurityAssertionImpl(getTokenHolder().getSecurityToken(IDP_REALM_NAME));
+  }
 
-    SecurityAssertion securityAssertion =
-        new SecurityAssertionImpl(tokenHolder.getSecurityToken("idp"));
+  private HttpSession getOrCreateSession() {
+    return sessionFactory.getOrCreateSession(request);
+  }
```
ahoffer: New helper method `getOrCreateSession`. Take advantage of fact that `request` is an object attribute.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +389,37 @@ private String getEntityId() {
...
 
-    if (hasSecurityAuditRole) {
+  private void logSecurityAuditRole() {
+    if (shouldAuditSubject()) {
       SecurityLogger.audit(
           "Subject with admin privileges has logged out: {}",
-          securityAssertion.getPrincipal().getName());
+          getIdpSecurityAssertion().getPrincipal().getName());
     }
+  }
```
ahoffer: New helper method `logSecurityAuditRole`

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +389,37 @@ private String getEntityId() {
...
-    tokenHolder.remove("idp");
+  private boolean shouldAuditSubject() {
+    return Arrays.stream(System.getProperty("security.audit.roles").split(","))
+        .anyMatch(
+            role -> getIdpSecurityAssertion().getPrincipals().contains(new RolePrincipal(role)));
+  }
+
+  private SecurityTokenHolder getTokenHolder() {
+    return (SecurityTokenHolder)
+        getOrCreateSession().getAttribute(SecurityConstants.SAML_ASSERTION);
```
ahoffer: New helper method

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
```diff
@@ -13,11 +13,15 @@
  */
 package org.codice.ddf.security.idp.client;
```
ahoffer: This was a difficult class to work with. I spend day or so refactoring it. Decided I was spending too much time, so I reverted to the original version. 

Fixed up tests that broke because of session index needing to passed into constructors. Also added new tests.

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
```diff
@@ -93,24 +105,8 @@
...
-    protected void buildAndValidateSaml(
-        String samlRequest,
-        String relayState,
-        String signatureAlgorithm,
-        String signature,
-        SignableXMLObject xmlObject)
-        throws ValidationException {
-      return;
-    }
-  }
```
ahoffer: The test was creating  a new subclass of the class-under-test. It did to disable the LogoutRequestService's validation routine. I do not like mocking the class under test, so I removed it.  I update the tests so they passed validation.

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
```diff
@@ -127,16 +123,7 @@ public void setup() throws ParserConfigurationException, SAXException, IOExcepti
...
-
-    logoutRequestService = new MockLogoutRequestService(simpleSign, idpMetadata, relayStates);
-    logoutRequestService.setEncryptionService(encryptionService);
-    logoutRequestService.setLogOutPageTimeOut(LOGOUT_PAGE_TIMEOUT);
-    logoutRequestService.setLogoutMessage(logoutMessage);
-    logoutRequestService.setRequest(request);
-    logoutRequestService.setSessionFactory(sessionFactory);
-
-    logoutRequestService.init();
-
```
ahoffer: Moved inline code into a method so it could be easily called multiple time.

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
```diff
@@ -411,13 +432,35 @@ public void testGetLogoutRequestResponseInvalidSignature() throws Exception {
...
+    doReturn(true).when(simpleSign).validateSignature(anyString(), anyString(), any());
+    initializeLogutRequestService();
+    Response response =
+        logoutRequestService.getLogoutRequest(
+            deflatedSamlResponse, null, relayState, SIGNATURE_ALGORITHM, SIGNATURE);
+    String msg = UNABLE_TO_VALIDATE_LOGOUT_REQUEST.replaceAll(" ", "+");
+    assertTrue(
+        "Expected message containing " + msg, response.getLocation().getQuery().contains(msg));
+  }
+
```
ahoffer: Test validation. A request with no session index is rejected.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
```diff
@@ -33,8 +35,13 @@
 
   private boolean partialLogout = false;
 
-  public LogoutState(Set<String> spDescriptors) {
+  private List<String> sessionIndexes = Collections.emptyList();
```
ahoffer: New state for the `LogoutState`!

platform/security/idp/security-idp-server/src/test/java/org/codice/ddf/security/idp/server/IdpEndpointTest.java
```diff
@@ -261,6 +261,8 @@ public void testDoSoapLogin() throws WSSecurityException {
     assertThat(response.getEntity().toString(), containsString("soapenv:Body"));
     assertThat(response.getEntity().toString(), containsString("saml2p:Response"));
     assertThat(response.getEntity().toString(), containsString("saml2:Assertion"));
+    assertThat(response.getEntity().toString(), containsString("saml2:AuthnStatement"));
+    assertThat(response.getEntity().toString(), containsString("SessionIndex="));
```
ahoffer: Updated this test class to test that session indexes are added

platform/security/handler/security-handler-api/pom.xml
```diff
@@ -109,17 +109,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.39</minimum>
```
vinamartin: ❓  These numbers are awfully close which almost makes me worried about slight environmental differences.

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
bdeining: We can use the UUID Generator service for the UUID generation instead of using `UUID.randomUUID().toString()`.  This will maintain uniformity for UUIDs within the system.
https://github.com/codice/ddf/blob/c4c2554cfa64ff7386cf4aaac6249026fce02d60/catalog/core/catalog-core-metacardgroomerplugin/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L16

platform/security/common/pom.xml
```diff
@@ -176,17 +176,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.27</minimum>
+                                            <minimum>0.22</minimum>
```
bdeining: Why does code coverage go down?

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
```diff
@@ -93,24 +105,8 @@
...
-        String samlRequest,
-        String relayState,
-        String signatureAlgorithm,
-        String signature,
-        SignableXMLObject xmlObject)
-        throws ValidationException {
-      return;
-    }
-  }
+  private String relayState = UUID.randomUUID().toString();;
```
vinamartin: ❗️ Typo?

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
bdeining: nit : extra whitespace

platform/security/common/pom.xml
ahoffer: JaCoCo always seems to short change me. JaCoCo, I want that 5% of coverage back!

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
ahoffer: Is it thread safe?

bdeining: I believe our current implementation is thread-safe (unless i'm missing something), but there is nothing in the API that suggests any implementation needs to be. 
 https://github.com/codice/ddf/blob/c4c2554cfa64ff7386cf4aaac6249026fce02d60/platform/util/platform-util-uuidgenerator/util-uuidgenerator-impl/src/main/java/org/codice/ddf/platform/util/uuidgenerator/impl/UuidGeneratorImpl.java#L37

ahoffer: I grabbed it. If I'll push it up and see if it passes integration tests.

ahoffer:
build now

ahoffer:
build now

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java
```diff
@@ -46,7 +46,10 @@ public AuthNStatementProvider() {
   @Override
   public AuthenticationStatementBean getStatement(TokenProviderParameters providerParameters) {
     AuthenticationStatementBean authBean = new AuthenticationStatementBean();
-    authBean.setSessionIndex(Integer.toString(secureRandom.nextInt()));
+    // "Use small positive integers (or reoccurring constants in a list) for the SessionIndex"
+    // Ref: sstc-saml-core-errata-2.0-wd-07, line 1140
+    int smallPositiveInteger = secureRandom.nextInt(Short.MAX_VALUE);
```
rzwiefel: So `nextInt` will return a number between [0, Short.MAX_VALUE) , since its inclusive of zero we could return zero. Not sure if we want to exclude it since it says small  **positive** integer

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
rzwiefel: ❓ Can this return null?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +389,37 @@ private String getEntityId() {
     return String.format("https://%s:%s%s/saml", hostname, port, rootContext);
   }
 
-  private void logout() {
-    HttpSession session = sessionFactory.getOrCreateSession(request);
+  private SecurityAssertion getIdpSecurityAssertion() {
 
-    SecurityTokenHolder tokenHolder =
-        ((SecurityTokenHolder) session.getAttribute(SecurityConstants.SAML_ASSERTION));
+    return new SecurityAssertionImpl(getTokenHolder().getSecurityToken(IDP_REALM_NAME));
```
rzwiefel: ❓ can `getTokenHolder` or `getSecurityToken` return null? If so this could throw a NPE

ahoffer: I don't know. There were no `null` checks previously, so I didn't add any in. I just used Intellij's "extract method" to create helper methods.

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java
ahoffer: Good point. I'll take care of that.

ahoffer:
Lots of comments, no approvals. :-(

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +389,37 @@ private String getEntityId() {
...
+  }
 
-    tokenHolder.remove("idp");
+  private boolean shouldAuditSubject() {
+    return Arrays.stream(System.getProperty("security.audit.roles").split(","))
+        .anyMatch(
+            role -> getIdpSecurityAssertion().getPrincipals().contains(new RolePrincipal(role)));
+  }
+
+  private SecurityTokenHolder getTokenHolder() {
```
rzwiefel: Looked at the code and looks like getAttribute can return null, so we should annotate this method with `@Nullable`

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
rzwiefel: Checked the code and both `getTokenHolder()` can return null (so we should protect against that) and `getSecurityToken(..)` can return null as well ( which would throw an NPE when passed to `new SecurityAssertionImpl(..)`

mcalcote:
I don't see the problem in the changeset, but was unable to log out successfully when building/testing this PR.

ahoffer:
build now

ahoffer:
build now

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -122,8 +129,15 @@ public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse
   @Override
   public LogoutRequest buildLogoutRequest(
       @NotNull String nameIdString, @NotNull String issuerOrEntityId, List<String> sessionIndexes) {
-    return buildLogoutRequest(
-        nameIdString, issuerOrEntityId, getUuidGenerator().generateUuid(), sessionIndexes);
+    return buildLogoutRequest(nameIdString, issuerOrEntityId, generateUuid(), sessionIndexes);
+  }
+
+  public String generateUuid() {
+    if (Objects.isNull(getUuidGenerator())) {
```
rzwiefel: We shouldn't need to have the runtime null check here. We generally rely on the assumption that blueprint will always give us a non null property (We should always get a proxy object that will eventually resolve to the real service)

ahoffer: I added it because I had originally misconfigured my blueprint file. The ultimate results was that clicking "Logout" on the Logout UI gave you the message. "You have been logged out", but if you clicked "sign in again", it would take you back to your application without asking for a username/password. 

That scared me, so I made sure that the logger would alert an administrator of (potentially) severe security situation. 

Options are: 

1. Take out null check entirely.
2. Take out the RuntimeException, but continue to log the error. (A null pointer exception gets logged at INFO level).
3. Leave as-is.

ahoffer:
build now

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
ahoffer: Talked with Ryan. This is really a more general issue. Created a new ticket  to address the general case. DIB-7658.

platform/security/handler/security-handler-api/pom.xml
ahoffer: Sounds likely to me.

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
```diff
@@ -287,6 +288,20 @@ private Request(Builder builder) {
...
+    /**
+     * A SAML LogoutRequest must include at least one SessionIndex, if the logout is initiated by
+     * Session Participant. However, if the Session Authority initiates the logout, a SessionIndex
+     * is optional. In that case, the absence of a SessionIndex means to logout of all sessions
+     * associated with the Principle.
+     *
+     * @throws ValidationException
+     */
+    protected void checkSessionIndexes() throws ValidationException {
+      if (logoutRequest.getSessionIndexes().isEmpty()) {
```
rzwiefel: ❓ So since we are always throwing the exception if sessionIndexes is empty, Do we never have the case mentioned above where the session authority initiates the logout?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
rzwiefel: I still believe we need to ensure the token returned from `getSecurityToken(IDP_REALM_NAME)` is not null before passing it to the `SecurityAssertionImpl` constructor. Perhaps in this case we just would want to throw a more descriptive error than the NPE that would be thrown otherwise.

Also since getTokenHolder is nullable we need to verify that is not null here either.

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
ahoffer: We do not support session authority initiated logouts. 
At least not yet.
I think I might feel better removing the validation. While it is currently correct, I can imagine thing changing in other parts of security that would make empty session indexes either possible, or at least not a fatal error.

ahoffer: fixed

ahoffer:
build now

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +390,45 @@ private String getEntityId() {
...
       SecurityLogger.audit(
           "Subject with admin privileges has logged out: {}",
-          securityAssertion.getPrincipal().getName());
+          getIdpSecurityAssertion().getPrincipal().getName());
     }
+  }
 
-    tokenHolder.remove("idp");
+  private boolean shouldAuditSubject() {
+    return Arrays.stream(System.getProperty("security.audit.roles").split(","))
```
figliold: :question: Is there any concern that the system property security.audit.roles can be maliciously modified to turn off auditing for roles?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
figliold: :question: Is there any concern with storing state like this in an HA or HS environment?

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -112,16 +116,24 @@ public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse
...
   @Override
   public LogoutRequest buildLogoutRequest(
-      @NotNull String nameIdString, @NotNull String issuerOrEntityId) {
-    return buildLogoutRequest(nameIdString, issuerOrEntityId, UUID.randomUUID().toString());
+      @NotNull String nameIdString, @NotNull String issuerOrEntityId, List<String> sessionIndexes) {
+    return buildLogoutRequest(nameIdString, issuerOrEntityId, generateUuid(), sessionIndexes);
+  }
+
+  private String generateUuid() {
+    return getUuidGenerator().generateUuid();
```
lessarderic: ❓ Should we just use the member variable here instead and get rid of the getter?

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
```diff
@@ -24,12 +23,10 @@ class LogoutMessageSpecTest extends Specification {
     LogoutMessageImpl logoutMessage;
 
     final String ENCRYPT_PREFIX = "ENCRYPTED "
-
     final String NAME_ID = "MyNameId"
-
     final String ISSUER_ID = "MyIssuerId"
-
     final String IN_RESPONSE_TO = "InResponseToID"
+    final List SESSION_INDEXES = Collections.emptyList();
```
lessarderic: ✏️  Nit: Blank line before to be consistent?

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -243,4 +258,12 @@ private URI signSamlGet(
   public void setSystemCrypto(SystemCrypto systemCrypto) {
     this.systemCrypto = systemCrypto;
   }
+
+  public UuidGenerator getUuidGenerator() {
+    return uuidGenerator;
+  }
+
+  public void setUuidGenerator(UuidGenerator uuidGenerator) {
```
lessarderic: ❓ Since it looks like we only need this to mock out the `UuidGenerator`, could we either make this method package-private and annotate with `@VisibleForTesting`, or inject it in the constructor and get rid of the setter?

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
```diff
@@ -69,7 +69,7 @@ class LogoutMessageSpecTest extends Specification {
 
     def "build valid logout request"() {
         when:
-        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(NAME_ID, ISSUER_ID)
+        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(NAME_ID, ISSUER_ID, SESSION_INDEXES)
```
lessarderic: ❓ Shouldn't we also add an assertion for the session indices?

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -112,16 +116,24 @@ public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse
...
+  private String generateUuid() {
+    return getUuidGenerator().generateUuid();
   }
 
   public LogoutRequest buildLogoutRequest(
-      @NotNull String nameIdString, @NotNull String issuerOrEntityId, @NotNull String id) {
+      @NotNull String nameIdString,
+      @NotNull String issuerOrEntityId,
+      @NotNull String id,
+      @NotNull List<String> sessionIndexes) {
```
lessarderic: ❓ Do we need to check that `sessionIndexes` isn't `null` like we do for the other arguments?

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
```diff
@@ -83,7 +83,7 @@ class LogoutMessageSpecTest extends Specification {
 
     def "build logout request with invalid info"() {
         when:
-        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(nameId, issuerId, id)
+        LogoutRequest logoutRequest = logoutMessage.buildLogoutRequest(nameId, issuerId, id, SESSION_INDEXES)
```
lessarderic: ❓ If session indices cannot be `null`, we need to add tests for that here.

platform/security/idp/security-idp-client/pom.xml
```diff
@@ -206,17 +212,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.59</minimum>
+                                            <minimum>0.40</minimum>
```
lessarderic: ❓ Ouch! Anything we can do to bring those numbers back up?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
```diff
@@ -1154,7 +1155,14 @@ Response handleLogoutRequest(
...
     }
 
-    logoutState = new LogoutState(getActiveSps(cookie.getValue()));
+    logoutState =
+        new LogoutState(
+            getActiveSps(cookie.getValue()),
+            logoutRequest
+                .getSessionIndexes()
+                .stream()
+                .map(SessionIndex::getSessionIndex)
```
lessarderic: ❓ Do we need to filter out any `null` or empty session indices? If not, why not just do `new ArrayList(logoutRequest.getSessionIndexes())` instead of using a stream?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
lessarderic: Great question @figliold. Don't think we run the IdP in HA configuration though, do we?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
```diff
@@ -33,8 +35,13 @@
...
+
+  public LogoutState(Set<String> spDescriptors, List<String> sessionIndexes) {
     this.spDescriptors = Collections.synchronizedSet(spDescriptors);
```
lessarderic: ❓ Do we really need this collection to be synchronized? The two methods that access this collection already are.

platform/security/platform-security-core-api/src/main/java/ddf/security/samlp/LogoutMessage.java
```diff
@@ -38,7 +39,10 @@ LogoutResponse extractSamlLogoutResponse(@NotNull String samlLogoutResponse)
   LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutRequest)
       throws XMLStreamException, WSSecurityException;
 
-  LogoutRequest buildLogoutRequest(@NotNull String nameIdString, @NotNull String issuerOrEntityId);
+  LogoutRequest buildLogoutRequest(
```
lessarderic: ❓ Could we remove the `@NotNull` in this file and use `@Nullable` where appropriate instead to follow our convention?

platform/security/idp/security-idp-server/src/test/resources/saml.xml
```diff
@@ -93,7 +93,7 @@
             <saml2:Audience>https://localhost:8993/services</saml2:Audience>
         </saml2:AudienceRestriction>
     </saml2:Conditions>
-    <saml2:AuthnStatement AuthnInstant="2013-04-23T23:39:54.777Z">
+    <saml2:AuthnStatement AuthnInstant="2013-04-23T23:39:54.777Z" SessionIndex="1">
```
lessarderic: ❗️ Based on the schema, it looks like `SessionIndex` is optional, so we should have tests with an `Assertion` that has an `AuthnStatement` without a `SessionIndex`.

platform/security/platform-security-core-api/src/main/java/ddf/security/samlp/SamlProtocol.java
```diff
@@ -485,13 +487,21 @@ public static AttributeQuery createAttributeQuery(Issuer issuer, Subject subject
...
+  public static LogoutRequest createLogoutRequest(
+      Issuer issuer, NameID nameId, String id, List<String> sessionIndexes) {
     LogoutRequest logoutRequest = logoutRequestBuilder.buildObject();
     logoutRequest.setID(id);
     logoutRequest.setIssuer(issuer);
     logoutRequest.setNameID(nameId);
     logoutRequest.setIssueInstant(DateTime.now());
     logoutRequest.setVersion(SAMLVersion.VERSION_20);
+    for (String index : sessionIndexes) {
+      SessionIndexBuilder builder = new SessionIndexBuilder();
```
lessarderic: ❓ Could we create the `SessionIndexBuilder` only once and reuse it instead of creating a new one each time?

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java
```diff
@@ -46,7 +46,13 @@ public AuthNStatementProvider() {
   @Override
   public AuthenticationStatementBean getStatement(TokenProviderParameters providerParameters) {
     AuthenticationStatementBean authBean = new AuthenticationStatementBean();
-    authBean.setSessionIndex(Integer.toString(secureRandom.nextInt()));
+    // "Use small positive integers (or reoccurring constants in a list) for the SessionIndex"
```
lessarderic: ❓ I couldn't quite figure out by reading the spec, but I thought session indices need to be unique for a given Principal? If so, do we have enough resolution here to avoid collisions?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
rzwiefel: Do we have an alternative place to store it if we do need to?

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +390,45 @@ private String getEntityId() {
...
-        ((SecurityTokenHolder) session.getAttribute(SecurityConstants.SAML_ASSERTION));
+    return new SecurityAssertionImpl(getTokenHolder().getSecurityToken(IDP_REALM_NAME));
+  }
 
-    SecurityAssertion securityAssertion =
-        new SecurityAssertionImpl(tokenHolder.getSecurityToken("idp"));
+  private HttpSession getOrCreateSession() {
+    HttpSession session = sessionFactory.getOrCreateSession(request);
+    if (Objects.isNull(session)) {
+      throw new RuntimeException("Unable to get or create HTTP session");
```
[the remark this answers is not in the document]
rzwiefel: getOrCreateSession should never return a null object in the first place so we may not need the null check at all

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +390,45 @@ private String getEntityId() {
...
+  private boolean shouldAuditSubject() {
+    return Arrays.stream(System.getProperty("security.audit.roles").split(","))
+        .anyMatch(
+            role -> getIdpSecurityAssertion().getPrincipals().contains(new RolePrincipal(role)));
+  }
+
+  private SecurityTokenHolder getTokenHolder() {
+    SecurityTokenHolder holder =
+        (SecurityTokenHolder) getOrCreateSession().getAttribute(SecurityConstants.SAML_ASSERTION);
+    if (Objects.isNull(holder)) {
```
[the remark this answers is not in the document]
rzwiefel: Looks like I was incorrect for the getAttribute. Looking at the get or create session, the SAML_ASSERTION will always be set on the session after calling this method

https://github.com/codice/ddf/blob/master/platform/security/core/security-core-impl/src/main/java/ddf/security/http/impl/HttpSessionFactory.java#L38-L44

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java
rzwiefel: We could use a single atomically incremented integer that loops back around to 1 If we're worried about the space being too small for collisions with a random. Although "small" seems ambiguous with out any relative comparison - it could be safest to just  broaden the range a bit.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
rzwiefel: It might actually be best to just move the defensive copying logic into the constructor such that its not the callers responsibility.  Once in there we could simply do a `ImmutableList.copyOf(..)` or similiar.

platform/security/common/pom.xml
ahoffer: Unfortunately, the code is written as a few long methods. Very hard to to test.

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
```diff
@@ -287,6 +288,24 @@ private Request(Builder builder) {
...
+     * A SAML LogoutRequest must include at least one SessionIndex, if the logout is initiated by
+     * Session Participant. However, if the Session Authority initiates the logout, a SessionIndex
+     * is optional. In that case, the absence of a SessionIndex means to logout of all sessions
+     * associated with the Principle. Because the SAMLValidator does not have enough information to
+     * determine who initiated the logout, this method does not throw a ValidationException.
+     * Instead, it logs the condition.
+     *
+     * @throws ValidationException
+     */
+    protected void checkSessionIndexes() throws ValidationException {
```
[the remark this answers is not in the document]
ahoffer: I don't know. There are cases when the lack of a session index could be a legitimate error per the specification. The specification says the logout request MUST have a session index if the logout is initiated by a Session Participant. However, the session index index is OPTIONAL if the logout is initiated by a Session Authority. According to Scott, we only support the former. So, in theory, any logout request without a session index contravenes the specification. However, security isn't really jeopardized if the session index is missing; the IDP logs out the user out of ALL session regardless.

I will remove the `throws ValidationException` and the javadoc comment. Just know that sometime in the future we may to throw that exception.

platform/security/common/src/test/java/ddf/security/assertion/impl/SecurityAssertionImplTest.java
```diff
@@ -102,6 +103,9 @@ public void testSampleAssertion() throws Exception {
     assertEquals(NUM_ATTRIBUTES, assertion.getAttributeStatements().size());
     assertEquals(NUM_NAUTH, assertion.getAuthnStatements().size());
     assertEquals(
```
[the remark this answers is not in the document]
ahoffer: I chose not to assert on value because the value is stored in the file `saml.xml`. I didn't want to tie the test code too tightly to the file. However, I see that the other assertion do use values in the files. If you think it is important, I'll be happy to add an assertion on the value.

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -112,16 +116,24 @@ public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse
    *
    * @param nameIdString NameId of user to log out
    * @param issuerOrEntityId The Issuer of the LogoutRequest
+   * @param sessionIndexes
```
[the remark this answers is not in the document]
ahoffer: fixes

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
```diff
@@ -112,16 +116,24 @@ public LogoutRequest extractSamlLogoutRequest(@NotNull String samlLogoutResponse
...
    * @return the built <code>LogoutRequest</code>
    */
   @Override
   public LogoutRequest buildLogoutRequest(
-      @NotNull String nameIdString, @NotNull String issuerOrEntityId) {
-    return buildLogoutRequest(nameIdString, issuerOrEntityId, UUID.randomUUID().toString());
+      @NotNull String nameIdString, @NotNull String issuerOrEntityId, List<String> sessionIndexes) {
```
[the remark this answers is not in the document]
ahoffer: fixed

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
ahoffer: Yes. I pulled it into a method to make it a little easier to mock. But it looks like I eventually removed that test because I don't find any other callers of that method.

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
ahoffer: I don't know. I'll throw in a null check and see if it breaks anything. There are edge cases when the specification doesn't require a session, but I would rather have callers pass an empty list than a null value.

platform/security/core/security-core-impl/src/main/java/ddf/security/samlp/impl/LogoutMessageImpl.java
ahoffer: I changed it to inject the object as a constructor argument and removed the setter.

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
ahoffer: They are all on the same line. I think github has to format it that way to show that lines were removed from the previous version. This is what it actually looks like;

```  final String ENCRYPT_PREFIX = "ENCRYPTED "
    final String NAME_ID = "MyNameId"
    final String ISSUER_ID = "MyIssuerId"
    final String IN_RESPONSE_TO = "InResponseToID"
    final List SESSION_INDEXES = Collections.emptyList();
```

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
ahoffer: Currently `SESSION_INDEXES` is an empty collection. I'll add a value and assert it.

platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy
ahoffer: Added this line to test for null:
`"xxx"  | "xxx"    | "xxx" | null`

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
```diff
@@ -287,6 +288,24 @@ private Request(Builder builder) {
...
+      if (logoutRequest.getSessionIndexes().isEmpty()) {
```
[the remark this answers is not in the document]
ahoffer: There isn't really way to test it with the existing unit tests.  I can create a new unit test that to will invoke the method on a mocked LogoutRequest, but I don't feel like it will prove anything.

It would ensure this method does not throw an exception if the collection is empty:

```
    protected void checkSessionIndexes() {
      if (logoutRequest.getSessionIndexes().isEmpty()) {
        LOGGER.info(
            "Logout request does not contain a session index for name-id {}",
            logoutRequest.getNameID());
      }
    }
```
The only different between the two conditions is a logging statement.

ahoffer:
Updated PR in response to comments.
build now.

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java
ahoffer: The spec says not to use numbers sequential numbers because it conveys some advantage to attackers. I wasn't so clear on what that advantage could be.

Session indexes do not have to globally unique. Once they just have to be unique for a given Principal. Typically, a Principle only has one SessionIndex. The day may come where they gave 2 or 3, but I don't see a collision as likely.

platform/security/platform-security-core-api/src/main/java/ddf/security/samlp/SamlProtocol.java
ahoffer: Good catch! Fixing it.

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java
ahoffer: There is `LogoutRequestServiceTest.testGetLogoutRequestNoSessionIndex`. It isn't working the way I thought it was, so I'm pushing a fix for it soon.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
ahoffer: If Ryan says it can go, it can go. Removing it.

platform/security/idp/security-idp-server/src/test/resources/saml.xml
ahoffer: There is `LogoutRequestServiceTest.testGetLogoutRequestNoSessionIndex`. It isn't working the way I thought it was, so I'm pushing a fix for it soon.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
ahoffer: Given how `LogoutState` is initialized, it should probably have a constructor that accepts a `LogoutRequest`. But that's more refactoring than I want to attempt right now.

The SAML library wraps the actual session index (string) inside of an object. The `map` in the streaming expression to to extract the `String` from the `SessionIndex` object. 

I reviewed where the constructor was used, and decided the best course of action was to add a setter, `setSessionIndexObjects`, that extracts the strings.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
ahoffer: In a HA configuration you shouldn't need more than one IDP. If you do have multiple IDPs, the SAML Response point to the IDP that issued the Assertion in `<Issuer></Issuer>`

platform/security/idp/security-idp-client/pom.xml
ahoffer: Must have been a JaCoCo run. I just check it again and coverage has gone up, not down. Look for new numbers in the next commit.

ahoffer:
New commit. build now.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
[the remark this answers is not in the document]
ahoffer: @rzwiefel is right. The null check is superfluous;.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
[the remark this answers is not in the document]
ahoffer: I can certainly extract that string to a constant.
I don't feel qualified to answer the second question.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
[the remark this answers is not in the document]
ahoffer: Thanks Ryan.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
ahoffer: Right.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -375,28 +390,45 @@ private String getEntityId() {
...
+      throw new RuntimeException("SAML assertion not present on the HTTP session");
```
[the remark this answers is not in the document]
ahoffer: Removed

ahoffer:
Build now. New commit

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
```diff
@@ -194,10 +196,15 @@ public void testSendLogoutRequestTimeout() throws Exception {
...
+    when(logoutRequest.getIssueInstant()).thenReturn(DateTime.now());
+    SessionIndex sessionIndex = mock(SessionIndex.class);
+    when(sessionIndex.getSessionIndex()).thenReturn(SESSION_INDEX);
+    when(logoutRequest.getSessionIndexes()).thenReturn(Collections.singletonList(sessionIndex));
     Response response = logoutRequestService.sendLogoutRequest(encryptedNameIdWithTime);
     assertEquals(Response.Status.SEE_OTHER.getStatusCode(), response.getStatus());
-    String msg =
-        "Failed to decrypt logout request params. Invalid number of params.".replaceAll(" ", "+");
+    insertLogoutRequest();
+    String msg = LogoutRequestService.UNABLE_TO_DECRYPT_LOGOUT_REQUEST.replaceAll(" ", "+");
```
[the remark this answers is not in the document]
ahoffer: Two error messages can mean the same thing, but not be identical. If the strings are checked for identity then even two messages that mean the same thing will fail the test. In that situation, even a spelling correction will make the test fail. It might be better to catch error messages with the wrong meaning in code review.

platform/security/common/pom.xml
```diff
@@ -176,17 +176,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.27</minimum>
+                                            <minimum>0.40</minimum>
```
lessarderic: 🎉

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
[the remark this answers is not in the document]
lessarderic: I guess my comment regarding the exception was conditional to making the method package-private. If we keep it `protected`, we should keep the exception in the signature otherwise we'll have to break the API down the road.

I personally would prefer to make it package-private for now until we have a good case for doing otherwise. Just my 2 cents.

platform/security/common/src/main/java/ddf/security/samlp/impl/SamlValidator.java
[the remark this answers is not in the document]
lessarderic: I agree that both tests would have the same assertion at the end, i.e., no exception is being thrown, but I think that's something that's worth verifying. Up to you.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/LogoutRequestService.java
```diff
@@ -196,11 +216,21 @@ private Response getSamlpPostLogoutRequest(String relayState, LogoutRequest logo
     return ok.build();
   }
 
+  String encodeSaml(LogoutRequest logoutRequest) throws SignatureException, WSSecurityException {
```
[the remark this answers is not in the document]
lessarderic: Just noticed all the static method calls in there :-(

platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java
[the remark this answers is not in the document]
lessarderic: I guess my point is that if those were integers and not Strings, I assume we wouldn't compare against a common constant, but against a value in the test that is independent of the class we're testing against. In other words, would it be better for tests to fail if the strings aren't exactly the same, or test using common constants and miss the fact that a string got mangled by mistake, not find the problem at test time and ship something with the wrong message?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java
```diff
@@ -1159,8 +1142,8 @@ Response handleLogoutRequest(
     logoutState.setNameId(logoutRequest.getNameID().getValue());
```
lessarderic: ❓ This method re-assigns the `logoutState` parameter on L1140, which is bad practice and can become confusing. This is not something that was changed by this PR, but since re-assigning parameters is considered bad practice and can be confusing, should we consider declaring a new variable instead?

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
```diff
@@ -97,4 +103,22 @@ public String getCurrentRequestId() {
   public void setCurrentRequestId(String currentRequestId) {
     this.currentRequestId = currentRequestId;
   }
+
+  public List<String> getSessionIndexes() {
+    return sessionIndexes;
+  }
+
+  @SuppressWarnings("unused")
```
lessarderic: ❓ If we need this `@SuppressWarnings`, we may want to add a comment explaining why.

platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java
```diff
@@ -97,4 +103,22 @@ public String getCurrentRequestId() {
...
+  public void setSessionIndexes(List<String> sessionIndexes) {
+    if (Objects.nonNull(sessionIndexes)) {
```
lessarderic: ❓ I've noticed `Objects.nonNull` being used instead of simply doing `sessionIndexes != null` for instance. Curious to know why?

ahoffer:
Failed on CI build.  
```java.lang.NoClassDefFoundError: Could not initialize class ddf.security.samlp.SamlProtocol
at ddf.security.samlp.SamlProtocolTest.testCreateResponse(SamlProtocolTest.java:35)```

Rebased because SamlProtocol class has changed. Retrying CI build now.

File: platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java (hunk 42 above)
Comment (lessarderic) [reply platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java]: I guess that's the problem with random numbers. It is _unlikely_ that they will collide but not _impossible_. My worry here is that _when_ it happens, it will be very difficult to know what's going on and almost impossible to reproduce. Increasing the number space would help reduce the collision likelihood and might be good enough, but it all depends on the random generator's number distribution over time and such, which can be tricky.

Comment (lessarderic) [reply platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java]: One thing we could consider here, since we know that we're not expecting hundreds of sessions per Principal, would be to populate a `Set` with _N_ random numbers to guarantee uniqueness, copy those values to a list, use `Collections.shuffle()` to randomize the order, and use Guava's `Iterables.cycle()` to get unique, pseudo-random session indices.

Comment (ahoffer) [reply platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java]: I'm happy to increase the space. I'll double the number of bytes from 4 to 8. Does that sound OK?

**Comment by ahoffer** [comment]
So, funny story. I should have also pulled in `SamlProtocolTest`.
Build now.

**Comment by ahoffer** [comment]
I hadn't pushed the commit. Dang.
Build now.

**Comment by ahoffer** [comment]
Found an issue with how Intellij moved a static init block and ruined everything.
Build now.

**Comment by ahoffer** [comment]
build now

File: platform/security/idp/security-idp-server/src/test/resources/saml.xml (hunk 40 above)
Comment (ahoffer) [reply platform/security/idp/security-idp-server/src/test/resources/saml.xml]: Fixed

File: platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java (hunk 51 above)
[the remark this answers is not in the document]
Comment (ahoffer) [reply platform/security/idp/security-idp-client/src/test/java/org/codice/ddf/security/idp/client/LogoutRequestServiceTest.java]: I see both sides of the argument. I lean toward constants and taking a risk with mangled strings. You have a broader responsibility ensuring the quality of the app. I will copy the strings as you suggest.

File: platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java (hunk 56 above)
Comment (ahoffer) [reply platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java]: I started using those methods when I started using Java streams:
```
    .stream()
        .filter(Objects::nonNull)
        .filter(s -> Objects.nonNull(s.getBinding()))
```
And I have a philosophical preference for method dispatch over build-in operators. But you reminded me it is not idiomatic and looks strange. I replaced them with `!=`

File: platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java (hunk 55 above)
Comment (ahoffer) [reply platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/LogoutState.java]: It is used, but it is called by blueprint and IntelliJ doesn't account for that.

File: platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java (hunk 54 above)
Comment (ahoffer) [reply platform/security/idp/security-idp-server/src/main/java/org/codice/ddf/security/idp/server/IdpEndpoint.java]: Absolutely.

File: platform/security/common/pom.xml (hunk 52 above)
Comment (ahoffer) [reply platform/security/common/pom.xml]: I think I figured out what happened. Some unit tests had errors, which reduced coverage. However, the coverage failure was the last thing in the console output, so it is the "bug" I fixed first. After fixing the unit tests, the coverage went back up.

**Comment by ahoffer** [comment]
Did not see this one in the Jenkins log after 4 hours, so I'm building again. build now.

File: platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java (hunk 42 above)
Comment (lessarderic) [reply platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/AuthNStatementProvider.java]: I'm ok with postponing this change and doing that work in a separate ticket, but I'd recommend creating the ticket now and get it done as part of the SAML Compliance work as soon as we can.

**Comment by ahoffer** [comment]
Build now after rebase

File: platform/security/core/security-core-impl/src/test/groovy/ddf/security/samlp/impl/LogoutMessageSpecTest.groovy [hunk 57]
```diff
@@ -69,34 +69,35 @@ class LogoutMessageSpecTest extends Specification {
...
         where: "any param is null"
-        nameId | issuerId | id
-        null   | null     | "xxx"
-        "xxx"  | null     | "xxx"
-        null   | "xxx"    | "xxx"
-        null   | null     | null
-        "xxx"  | null     | null
-        null   | "xxx"    | null
-        "xxx"  | "xxx"    | null
+        nameId | issuerId | id    | sesionIndex
```
rzwiefel: ❗️ Typo here: should be `sessionIndex`

ahoffer:
Fixed test error. Build now.

ahoffer:
build now

ahoffer:
Update branch. Build now! now!

ahoffer:
build now

ahoffer:
build now

ahoffer:
01:47:58 [ERROR] Failed to execute goal org.owasp:dependency-check-maven:1.4.5:check (default) on project broker-undelivered-messages-ui: One or more exceptions occurred during analysis: Analysis task timed out.

build now

ahoffer:
> 04:35:43 ERROR: Error fetching remote repo 'origin'

Ugh. Time to go back to manual heroing.

Build now.

ahoffer:
git fetch problem
build now

ahoffer:
build now

bdeining:
build now

ahoffer:
build now.

ahoffer:
build now

ahoffer:
build now

ahoffer:
[ERROR] Failed to execute goal org.apache.felix:maven-bundle-plugin:3.3.0:bundle (default-bundle) on project checksum: Execution default-bundle of goal org.apache.felix:maven-bundle-plugin:3.3.0:bundle failed. NullPointerException -> [Help 1]

build now

ahoffer:
Are these failures transient or caused by this PR?
build now

ahoffer:
[ERROR] Failed to execute goal on project video-input-transformer: Could not resolve dependencies for project ddf.catalog.transformer:video-input-transformer:bundle:2.11.4-SNAPSHOT: Could not transfer artifact org.apache.poi:poi-ooxml:jar:3.16 from/to office (http://nexus.phx.connexta.com:8081/nexus/content/groups/public/): /root/.m2/repository/org/apache/poi/poi-ooxml/3.16/poi-ooxml-3.16.jar.part (No such file or directory) -> [Help 1]

build now

ahoffer:
Are these failures transient or caused by this PR?
build now

PR #2638, opened 2017-11-08
Updates PR Template with notes for how to handle porting between branches

coyotesqrl:
#### What does this PR do?
Adds a section on how our process _temporarily_ changes after branching.

#### Who is reviewing it? 
@mcalcote @vinamartin @lessarderic 

@clockard
@pklinef

N/A

This will better centralize documentation of the temporary process change following branching.

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
#### Review Comment Legend:

### DIFF
--- a/PULL_REQUEST_TEMPLATE
+++ b/PULL_REQUEST_TEMPLATE
@@ -32,6 +32,31 @@ https://github.com/orgs/codice/teams
 - [ ] Documentation Updated
 - [ ] Update / Add Unit Tests
 - [ ] Update / Add Integration Tests
+
+#### Notes on Review Process
+For the general case, merging is gated on the following:
+
+* Review by at least two contributors
+* Review by at least two committers
+* Successful CI
+* Successful static analysis (SonarQube/Coverity/etc.)
+* Heroing of change (details vary by ticket)
+
+Certain exceptions may occur. For example, if there is a problem with the CI environment then a manually performed full build and static analysis check may be substituted _at the discretion of the repository maintainers_. Additionally, repository metadata such as this `PULL_REQUEST_TEMPLATE`, the repository `README`, and other non-functional items may be merged without requiring CI and static analysis.
+
+Finally, an abbreviated review process may be in place for forward-port tickets immediately following branching and prior to branch divergence. In that case, approvals, CI, static analysis, and heroing that occur against the PR against the _release branch_ can be considered sufficient for reviewing the PR against the _master branch_.
+
+Once the branches diverge, the process will return to normal.
+
+Insofar as divergence is not a function of time but of change to the parent branch, the repository maintainers will determine when abbreviated review ends.
+
+When abbreviated review is in effect, a note to that effect will be at the top of this template. Following is an example note:
+```
+##### ABBREVIATED REVIEW BETWEEN 2.12.X AND MASTER IS IN EFFECT
+____
+
+```
+
 #### Review Comment Legend:
 - ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
 - ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.


PULL_REQUEST_TEMPLATE
```diff
@@ -32,6 +32,16 @@ https://github.com/orgs/codice/teams
 - [ ] Documentation Updated
 - [ ] Update / Add Unit Tests
 - [ ] Update / Add Integration Tests
+
+#### Note on backporting
+Immediately following branching and prior to branch divergence, the review process for ports
```
lessarderic: ❓ Should we make the review rule a little bit more formal between partial and full review and approval process? I'm not sure people would know just by reading this, and communicating when to switch to full approval might be difficult.

A few suggestions:
* Partial review and approval is only allowed right after a release and up until the first patch version is released
* Partial review and approval is only allowed when there have been less than _N_ commits or PRs against the release branch
* We update the template when a release branch is created to indicate that only partial approval is required, and change it back to the full approval process when we feel that the release branch and master have diverged

coyotesqrl: I don't know that we can make it more defined and not actually be wrong. I could easily foresee a time when we cut a branch for release and work for master that had intentionally been held back is immediately merged causing the branches to diverge immediately. Number of commits or PRs against the release branch isn't the gating factor, it's the commits that only went to its parent branch. And even then, I believe there should be a small degree of judgment applied. Examples:

1. Branch X is created. A brand new feature is immediately committed to master that does not interact with any existing features. I think this case might safely allow for forward-ports with abbreviated review.
2. Branch X is created. A large number of documentation-formatting/layout updates are immediately committed to master. I believe that any bug fixes made to branch X (that involve no updates to documentation) could safely be ported to master with abbreviated review.
3. Branch X is created. No new features are added to master through three subsequent patch releases of X. There is definitely no risk of abbreviated review in this case.
3. Branch X is created. A change is immediately made to an existing feature in master, causing the branches to diverge. I think full review ports from X are now clearly required.

I do like your final suggestion, that we make a template update when we have determined that the branches have diverged and require the abbreviated process to be set aside. The only problem I see with that is that any open port PRs will still have the old template. But this can be managed by committers who'll be aware of the status change and know they're looking at a PR with outdated metadata. 

I'll take a swipe at indicating this on the template and will also add something about metadata-type PRs that can always undergo abbreviated review. This PR, for example, does NOT need static analysis or CI as it only impacts the metadata of the repository.

PULL_REQUEST_TEMPLATE
```diff
@@ -32,6 +35,26 @@ https://github.com/orgs/codice/teams
...
+
+* Review by at least two contributors
+* Review by at least two committers
+* Successful CI
+* Successful static analysis (SonarQube/Coverity/etc.)
+* Heroing of change (details vary by ticket)
+
+Certain exceptions may occur. For example, if there is a problem with the CI environment then a manually performed full build and static analysis check may be substituted _at the discretion of the repository maintainers_. Additionally, repository metadata such as this `PULL_REQUEST_TEMPLATE`, the repository `README`, and other non-functional items may be merged without requiring CI and static analysis.
+
+Finally, an abbreviated review process may be in place for forward-port tickets immediately following branching and prior to branch divergence. In that case, approvals, CI, static analysis, and heroing that occur against the PR against the _release branch_ can be considered sufficient for reviewing the PR against the _master branch_.
```
mcalcote: This is well written and I don't disagree with the process, but I wonder if this is the best location for it to be documented. Should we make any of these items a checkbox (like the bullets above) that the owner/reviewer has to update as they meet the requirements? 

I think I might have missed some of the discussion leading up to the decision to add this.

PULL_REQUEST_TEMPLATE
```diff
@@ -1,3 +1,6 @@
+##### ABBREVIATED REVIEW BETWEEN 2.11.X AND MASTER IS IN EFFECT
```
lessarderic: ❓ How about `ABBREVIATED REVIEW PROCESS` to be consistent with the text in the notes below?
❓ Do we need to mention the branch names?

PULL_REQUEST_TEMPLATE
lessarderic: ❓ Should we bold **abbreviated review process** (or have a header) to make that part easier to find when that review process is in effect?

PULL_REQUEST_TEMPLATE
```diff
@@ -32,6 +35,26 @@ https://github.com/orgs/codice/teams
...
+
+Finally, an abbreviated review process may be in place for forward-port tickets immediately following branching and prior to branch divergence. In that case, approvals, CI, static analysis, and heroing that occur against the PR against the _release branch_ can be considered sufficient for reviewing the PR against the _master branch_.
+
+Once the branches diverge, the process will return to normal.
+
+Insofar as divergence is not a function of time but of change to the parent branch, the repository maintainers will determine when abbreviated review ends.
+
+When abbreviated review is in effect, a note to that effect will be at the top of this template.
+
 #### Review Comment Legend:
```
lessarderic: May want to keep the legend before the long process description otherwise people will never notice it.

PR #2662, opened 2017-11-13
[2.11.x] DDF-3452 Fix Import Command

rzwiefel:
#### What does this PR do? 
- Now when A single metacard fails to process, the import command will continue with the rest of the metacards instead of failing the rest of the transaction

#### Who is reviewing it? 
Anyone

Anyone

Ensure that when a single invalid metacard among a batch is imported, the rest of the valid metacards are imported.

https://codice.atlassian.net/browse/DDF-3452

#### Review Comment Legend:


clockard:
build now

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ImportCommand.java
```diff
@@ -174,9 +173,12 @@ protected Object executeWithSubject() throws Exception {
                         new UncloseableBufferedInputStreamWrapper(zipInputStream), id);
               } catch (IOException | CatalogTransformerException e) {
                 LOGGER.debug("Could not transform metacard: {}", id);
+                entry = zipInputStream.getNextEntry();
+                continue;
               }
               metacard = applyInjectors(metacard, attributeInjectors);
```
lessarderic: ❓ Could we move L179-181 inside the `try` block and just use the normal flow instead of adding  `getNextEntry()` and `continue` on L176-177?

rzwiefel: I like the try scoped to the one statement that it actually handles. And, IMO, the explicit continuing is just as clear than letting it fall through letting implicitly flow back to the start. 

However to stop the duplication of `entry = zipInputStream.getNextEntry()` I did change this to a do-while loop and such there is now only a single `entry = zipInputStream.getNextEntry();` at the beginning of the method instead of manually having to call it before continuing or letting the logic flow till after the switch.

rzwiefel:
Rebased. Build Now

PR #2665, opened 2017-11-13
[2.11.x] DDF-3453 Fixed XML Transformer Issue

rzwiefel:
#### What does this PR do? 
- Previously the XML transformer would allow invalid characters to be
serialized into the XML (via the Escaping print writer). It is now
changed to use the Apache commons escaper which will produce valid XML.

#### Who is reviewing it? 
@stustison @brendan-hofmann @jlcsmith @coyotesqrl @shaundmorris 
@codice/data

Anyone

Contact @rzwiefel for test data that previously broke this and confirm that it no longer produces invalid XML

https://codice.atlassian.net/browse/DDF-3453

#### Checklist:
- [X] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/transformer/catalog-transformer-xml/pom.xml
+++ b/catalog/transformer/catalog-transformer-xml/pom.xml
@@ -85,6 +85,10 @@
             <artifactId>joda-time</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+        <dependency>
             <groupId>ddf.platform.util</groupId>
             <artifactId>platform-util</artifactId>
         </dependency>
@@ -107,6 +111,7 @@
                             catalog-transformer-xml-binding,
                             jaxb2-basics-runtime,
                             gml-v_3_1_1-schema;inline=true,
+                            commons-lang3,
                             platform-util
                         </Embed-Dependency>
                         <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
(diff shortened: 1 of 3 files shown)


catalog/transformer/catalog-transformer-xml/src/test/java/ddf/catalog/transform/xml/TestEscapingPrintWriter.java
```diff
@@ -87,48 +87,6 @@
   }
 
   @Test
```
coyotesqrl: ❓ No new test for this?

rzwiefel: I'm not going to add tests for the Apache commons escaping utility -- this class is merely a proxy for it now.

clockard:
build now

catalog/transformer/catalog-transformer-xml/src/test/java/ddf/catalog/transform/xml/TestEscapingPrintWriter.java
brendan-hofmann: Why did these tests need to be removed? Do they fail against the Apache commons utility?

catalog/transformer/catalog-transformer-xml/pom.xml
```diff
@@ -107,6 +111,7 @@
                             catalog-transformer-xml-binding,
                             jaxb2-basics-runtime,
                             gml-v_3_1_1-schema;inline=true,
+                            commons-lang3,
```
brendan-hofmann: Do we really need the embed or can we use `escapeXml()` from 2.6?

rzwiefel: I don't believe the escapeXml is what we want according to the docs:
https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringEscapeUtils.html#escapeXml(java.lang.String)

It only supports the 5 basic entities, similiar to what ours did previously.

catalog/transformer/catalog-transformer-xml/src/test/java/ddf/catalog/transform/xml/TestEscapingPrintWriter.java
rzwiefel: Correct. They are actually asserting things that are invalid according to the XML spec.

clockard:
Manually verified this works with the provided test data. 👍

PR #2689, opened 2017-11-16
[2.11.x] DDF-3280 Upgrade Karaf to 4.1.3

clockard:
#### What does this PR do?
Initial PR to update Karaf from 4.1.2 to 4.1.3
#### Who is reviewing it? 
@emmberk @peterhuffer @AzGoalie @vinamartin 

@coyotesqrl
@lessarderic
@rzwiefel
@stustison

- [ ] Manually test basic functionality of DDF
- [ ] Drop .cfg and .config files in the /etc directory, and confirm that the configurations are installed successfully.
- [ ] Confirm that the bin scripts work as expected.
- [ ] Walkthrough the Hardening Guide. (Installed with hardening script)
- [ ] Confirm that the waitForReady command still works.
- [ ] Confirm that installing via each install profile works.
- [ ] Confirm that logging still works and is still configurable from org.ops4j.pax.logging.cfg and from the Karaf console. Also, try switching the logging configuration to the xml-style configuration.
 -- Setting log level works from the console unless you are using xml-style configuration
 -- Setting log level via configuration files work
- [ ] Test that all the console commands are still functional.
 -- Tested several of the commonly used commans
- [ ] Confirm SSHing into a running instance still works. Confirm SSHing into itests still works.
 -- SSH works but not restricted to a role when using username/password
- [ ] Confirm debugging a live system
- [ ] Confirm that the Karaf instance can be restarted.

Karaf change set https://github.com/apache/karaf/compare/karaf-4.1.2...karaf-4.1.3
[DDF-3280](https://codice.atlassian.net/browse/DDF-3280)


distribution/ddf-common/src/main/resources/etc/shell.init.script
```diff
@@ -56,6 +56,7 @@ if { %(jlineReader != null) } {
 
   setopt group
   setopt auto-fresh-line
+  unsetopt insert-tab
```
emmberk: Does this conflict with line 78?
https://github.com/codice/ddf/blob/983096dc7cfcc600a34f26780cc4c0ad6edb9e85/distribution/ddf-common/src/main/resources/etc/shell.init.script#L78

paouelle:
❓ Did you confirm that there was no changes to the start and start.bat scripts? These are now checked into DDF.

clockard:
@paouelle Yes, the stop script was updated but not the start. Also verified that `ddf`, `start`, `client` scripts worked as well as running as a service on linux.

clockard:
Closed pull request as we will not be upgrading to karaf 4.1.3 due to issues found with that version

PR #2695, opened 2017-11-17
[2.11.x] DDF-3466 Added an interface to get system high attributes

emmberk:
#### What does this PR do?
This PR adds an interface and impl to get system high attribute values.
#### Who is reviewing it? 
@peterhuffer @AzGoalie 
@codice/security 
@clockard
@lessarderic
1. Start, and confirm that the `security-sts-propertyclaimshandler` bundle installs successfully. Install.
2. Follow the Documentation to update the `User Attribute File` location in the `File Based Claims Handler` configuration in the Admin Console. Confirm that the Handler is updated with the contents of the new file.
The impl will be manually tested in https://github.com/codice/alliance/pull/453.
[DDF-3466](https://codice.atlassian.net/browse/DDF-3466)
#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

- [ ] Review by at least two contributors
- [ ] Review by at least two committers
- [ ] Successful CI
- [ ] Successful static analysis (SonarQube/Coverity/etc.)
- [ ] Heroing of change (details vary by ticket)

#### Review Comment Legend:


platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -158,6 +162,47 @@ public void testRetrieveClaimsValuesNullPrincipal() {
...
+  @Test
+  public void testGetMultiValuedSystemHighAttribute() throws IOException {
+    assertThat(
+        attributeFileClaimsHandler.getValues("test1"),
+        isPresentAnd(containsInAnyOrder("testing1", "testing2", "testing3")));
+  }
+
+  @Test
+  public void testNoMatchGetSystemHighAttribute() throws IOException {
+    System.setProperty(SystemBaseUrl.HOST, "admin");
```
peterhuffer: ❓ Is this needed, or is it to make it more clear? It looks like it's already being set in the `@Before`.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -37,57 +44,35 @@
...
 import org.boon.json.JsonFactory;
 import org.boon.json.ObjectMapper;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-public class AttributeFileClaimsHandler implements ClaimsHandler, RealmSupport {
+import org.codice.ddf.configuration.SystemBaseUrl;
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(AttributeFileClaimsHandler.class);
+public class AttributeFileClaimsHandler implements ClaimsHandler, SystemHighAttributes {
```
peterhuffer: ❓ Why does this no longer implement `RealmSupport`?

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
```diff
@@ -0,0 +1,31 @@
...
+ * This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library.
+ */
+public interface SystemHighAttributes {
+
+  /**
+   * @return an {@link Optional} of the system high attribute {@link Serializable} values, or {@link
+   *     Optional#empty()} if the attribute for the {@code attributeName} is not present
+   */
+  Optional<List<Serializable>> getValues(final String attributeName);
```
clockard: Don't know if it is directly useful right now but it seems like this interface could use a method that returns all system high attributes. Maybe just the attribute names.

platform/security/sts/security-sts-propertyclaimshandler/pom.xml
```diff
@@ -32,55 +32,30 @@
             <groupId>ddf.security.core</groupId>
             <artifactId>security-core-api</artifactId>
         </dependency>
-        <dependency>
-            <groupId>antlr</groupId>
```
clockard: Thanks for cleaning up these dependencies! 👍

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
emmberk: The blueprint did not include a service definition for that interface:
https://github.com/codice/ddf/blob/7dd7f72c536717ed4fb479ef336346c9bf47506c/platform/security/sts/security-sts-propertyclaimshandler/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L45-L47

and the override methods `getSupportedRealms()` and `getHandlerRealm()` just returned null. Looks like it was just some leftover dead code from a previous implementation.

clockard: I am curious about that too

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -148,16 +133,50 @@ public ProcessedClaimCollection retrieveClaimValues(
   }
 
   @Override
-  public List<String> getSupportedRealms() {
-    return supportedRealms;
+  public Optional<List<Serializable>> getValues(final String attributeName) {
+    return Optional.ofNullable(systemHighAttributes.get(attributeName));
```
clockard: Should probably pass back an unmodifiable list.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,22 +48,24 @@
...
 
-  ClaimsParameters unknownClaimsParameters;
+  private ClaimsParameters unknownClaimsParameters;
 
-  File userFile;
+  private File userFile;
 
   @Before
   public void setup() throws IOException {
+    System.setProperty(SystemBaseUrl.HOST, "admin");
```
clockard: 'admin' seems like an odd hostname. Just wanted to make sure that was intended.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
emmberk:
```java
        Collections.unmodifiableList(systemHighAttributes.get(attributeName)));
```
will throw a NPE when `systemHighAttributes` does not contain key `attributeName`.

To keep this method small, I added updated Line 179 instead:
```java
        newSystemHighAttributes.put(
            systemHighAttributeName,
            Collections.unmodifiableList((List<Serializable>) systemHighAttribute.getValue()));
```

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
emmberk: I updated it to `user1` instead.

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
lessarderic: ❓ Since I don't think it's possible for an attribute to exist _and_ have no values (empty list), do we really need to use an `Optional` here? Could we simply return an empty `List` when an attribute doesn't exist?

emmberk: I also added a note in the Documentation.

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
```diff
@@ -0,0 +1,30 @@
...
+ * This code is experimental. While this interface is functional and tested, it may change or be
+ * removed in a future version of the library.
+ */
+public interface SystemHighAttributes {
+
+  /**
+   * @return an {@link List<String>} of the {@link Serializable} values for the system high
+   *     attributeName, or {@List Collections#emptyList()} if the attributeName is not present.
+   */
+  List<Serializable> getValues(final String attributeName);
```
Lambeaux: ❓  If there are no duplicated attributes, would it make sense to return a `Set<Serializable>` instead? Can we make that guarantee?

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -148,16 +130,51 @@ public ProcessedClaimCollection retrieveClaimValues(
...
+      final String systemHighAttributeName = systemHighAttribute.getKey();
+
+      final Object systemHighAttributeValue = systemHighAttribute.getValue();
+      if (systemHighAttributeValue instanceof String) {
+        newSystemHighAttributes.put(
+            systemHighAttributeName, Collections.singletonList((String) systemHighAttributeValue));
+      } else {
+        newSystemHighAttributes.put(
+            systemHighAttributeName, (List<Serializable>) systemHighAttribute.getValue());
+      }
```
Lambeaux: ❓  Should we be throwing an exception if the value is not one of the expected types? If we do, we should inline initialize `this.systemHighAttributes` with an empty map.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -158,6 +160,42 @@ public void testRetrieveClaimsValuesNullPrincipal() {
...
+        attributeFileClaimsHandler.getValues("test1"),
+        containsInAnyOrder("testing1", "testing2", "testing3"));
+  }
+
+  @Test
+  public void testNoMatchGetSystemHighAttribute() throws IOException {
+    assertThat(attributeFileClaimsHandler.getValues("reg"), is(empty()));
+  }
+
+  @Test(expected = Exception.class)
```
Lambeaux: ❓  Are they actually throwing an `Exception` instance? If not we should consider expecting the specific exception being thrown, unless the thrower of the exception is a third party. 

This comment also applies to the below uses of `@Test(expected = Exception.class)` as well.

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
```diff
@@ -0,0 +1,31 @@
...
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.security;
```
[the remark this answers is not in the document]
stustison: Yes, please.

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
lessarderic: Since order could matter, a `List` is probably safer here.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -148,16 +130,51 @@ public ProcessedClaimCollection retrieveClaimValues(
...
+  public List<Serializable> getValues(final String attributeName) {
+    return Optional.ofNullable(systemHighAttributes.get(attributeName))
+        .orElse(Collections.emptyList());
   }
 
-  @Override
-  public String getHandlerRealm() {
-    return realm;
+  public void setAttributeFileLocation(String attributeFileLocation) throws IOException {
+    this.attributeFileLocation =
```
lessarderic: Never noticed these returned the values being validated. Cool!

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -37,57 +44,32 @@
...
+  public void init() throws IOException {
+    Path ddfHomePath = Paths.get(System.getProperty("ddf.home"));
+    Path path = Paths.get(attributeFileLocation);
+    if (!path.isAbsolute()) {
+      path = Paths.get(ddfHomePath.toString(), path.toString());
+    }
+    try (InputStream stream = Files.newInputStream(path)) {
+      String jsonString = IOUtils.toString(stream, Charset.defaultCharset());
+      json = JsonFactory.create().parser().parseMap(jsonString);
     }
```
lessarderic: ❓ Shouldn't we keep logging something like we used to (L85-89) when the file can't be read or parsed?

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -148,16 +130,51 @@ public ProcessedClaimCollection retrieveClaimValues(
...
+      Object value = entry.getValue();
+      if (value instanceof Map) {
+        Set keySet = ((Map) value).keySet();
+        for (Object key : keySet) {
+          claims.add((String) key);
+        }
+      }
+    }
+
+    supportedClaimTypes = claims.stream().map(URI::create).collect(Collectors.toList());
```
lessarderic: ❗️ `supportedClaimTypes` needs to be `volatile` to make sure it's not being changed while `getSupportedClaimTypes()` is called.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -37,57 +44,32 @@
...
-      }
+  public void init() throws IOException {
+    Path ddfHomePath = Paths.get(System.getProperty("ddf.home"));
+    Path path = Paths.get(attributeFileLocation);
+    if (!path.isAbsolute()) {
+      path = Paths.get(ddfHomePath.toString(), path.toString());
+    }
+    try (InputStream stream = Files.newInputStream(path)) {
+      String jsonString = IOUtils.toString(stream, Charset.defaultCharset());
+      json = JsonFactory.create().parser().parseMap(jsonString);
```
lessarderic: ❗️ Since `json` is a member variable being set here and used in two other methods, its access needs to be synchronized somehow, otherwise we're going to be running into issues if it changes while it's being used.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -148,16 +130,51 @@ public ProcessedClaimCollection retrieveClaimValues(
...
+      if (systemHighAttributeValue instanceof String) {
+        newSystemHighAttributes.put(
+            systemHighAttributeName, Collections.singletonList((String) systemHighAttributeValue));
+      } else {
+        newSystemHighAttributes.put(
+            systemHighAttributeName, (List<Serializable>) systemHighAttribute.getValue());
+      }
+    }
+
+    this.systemHighAttributes = newSystemHighAttributes;
```
lessarderic: ❗️ `systemHighAttributes ` needs to be `volatile` to make sure it's not being changed while `getValues()` is called.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
lessarderic: ❓ Or should we log an error and move on to the next attribute? Fail fast or best effort? Is it better to keep all the old values if something's wrong with the new map, or use as many new values as possible?
❗️ And we should also check the type before casting to `List<Serializable>` to avoid getting a `ClassCastException`.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -158,6 +160,42 @@ public void testRetrieveClaimsValuesNullPrincipal() {
...
+  @Test
+  public void testGetSingleValuedSystemHighAttribute() throws IOException {
+    assertThat(attributeFileClaimsHandler.getValues("test"), containsInAnyOrder("testValue"));
+  }
+
+  @Test
+  public void testGetMultiValuedSystemHighAttribute() throws IOException {
+    assertThat(
+        attributeFileClaimsHandler.getValues("test1"),
+        containsInAnyOrder("testing1", "testing2", "testing3"));
```
lessarderic: ❗️ Since we're returning a `List`, we should validate the order as well.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
lessarderic: ❗️ Agreed. We should use specific exceptions, not just `Exception`. Comment applies to the following tests as well.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -158,6 +160,42 @@ public void testRetrieveClaimsValuesNullPrincipal() {
...
+    attributeFileClaimsHandler.setAttributeFileLocation("noFileHere");
+  }
+
+  @Test(expected = Exception.class)
+  public void testExceptionWhenUnexpectedFormatInUsersDotAttributesFile() throws IOException {
+    userFile = temporaryFolder.newFile("notExpectedFormatUsers.attributes");
+    FileUtils.write(userFile, "not json content in expected format", Charset.defaultCharset());
+
+    attributeFileClaimsHandler.setAttributeFileLocation(userFile.getAbsolutePath());
   }
```
lessarderic: ❗️ We should also add the following tests:
* `init()` with an invalid file
* Invalid attribute type in JSON map (e.g., integer) to cover L169-175
* `setAttributeFileLocation(null)`
* `getValues(null)`

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
emmberk: I don't know completely what the DDF convention is, but `Exception` makes more sense to me. This unit test is verifying that the initialization of `AttributeFileClaimsHandler` will fail if the `users.atrtibutes` file doesn't exist. We know when bundles fail to initialize when we go through the installer. I don't care about asserting exactly what kind of exception but rather that there is an exception which indicates that the initialization to fail. That is, if the implementation changed later to throw a different exception when the `users.attribtues` file doesn't exist, this unit test will still pass.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
emmberk: I removed these log statements because I'd like the initialization of `AttributeFileClaimsHandler` to fail (i.e. throw and exception) if there is an issue reading the file. I don't want to catch the exception and report that the `AttributeFileClaimsHandler` was successfully initialized.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
emmberk: Regarding your second comment, the reason I chose to do it this way is the same as my other few replies. I want the initialization of AttributeFileClaimsHandler to fail (i.e. throw and exception) if there is an issue reading the file. In this case, the issue is that the values aren't serializable. Furthermore, a `ClassCastException` should never occur if the `users.attributes` file was set up according to the Documentation guidelines.

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
emmberk: To follow up after our discussion with @stustison and @lessarderic, we're going to keep just this one method for now and iterate in the future once we investigate differentiating between the user claims and the ism attributes defined in the users.attributes file.

platform/security/platform-security-core-api/src/main/java/ddf/security/SystemHighAttributes.java
emmberk: I checked with @stustison that order should not matter. I updated the documentation to clarify this assumption and updated this method to return a `Set<String>` instead.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
emmberk: Added `volatile`

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
emmberk: See my other reply about updating the `getValues` method to return a `Set<String>` instead.

emmberk:
I pushed an update after talking with @stustison and @lessarderic. This PR needs to be re-reviewed.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -37,57 +44,35 @@
...
-      }
+  public void init() throws IOException {
+    Path ddfHomePath = Paths.get(System.getProperty("ddf.home"));
+    Path path = Paths.get(attributeFileLocation);
+    if (!path.isAbsolute()) {
+      path = Paths.get(ddfHomePath.toString(), path.toString());
+    }
+    try (InputStream stream = Files.newInputStream(path)) {
+      String jsonString = IOUtils.toString(stream, Charset.defaultCharset());
+      json = MAPPER.parser().parseMap(jsonString);
```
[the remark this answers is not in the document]
emmberk: I ended up updating this class to use gson instead.

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
emmberk: In the most recent pushed changes, I updated this class to throw `IllegalStateException`s when the attributes file doesn't follow our assumptions.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,120 +44,315 @@
...
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void testNonStringValueInUsersDotAttributesFile() throws IOException {
+    // given
+    System.setProperty(SystemBaseUrl.HOST, "user1");
+
+    final AttributeFileClaimsHandler attributeFileClaimsHandler = new AttributeFileClaimsHandler();
+    attributeFileClaimsHandler.setAttributeFileLocation(
+        createAttributeFilePathFromResourceFileName("users.attributes-nonStringValue"));
```
peterhuffer: ✏️  Make this filename static final variables and re-use.

emmberk: Each of the resource files is only used in one unit test method, except for the valid "users.attributes", which is only specified in one helper method. I don't think moving these filenames to static Strings would provide any value.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,120 +44,315 @@
```
lessarderic: ❗️ Can we rename the class to `AttributeFileClaimsHandlerTest` to match our naming convention for unit test files?

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,120 +44,315 @@
...
-    claim2.setClaimType(URI.create("reg"));
-    claimCollection.add(claim2);
+    // when
+    final ProcessedClaimCollection processedClaims =
+        attributeFileClaimsHandler.retrieveClaimValues(claimCollection, user1ClaimsParameters);
+
+    // then
+    assertThat(
+        processedClaims,
+        containsInAnyOrder(
```
lessarderic: Cool combination of `containsInAnyOrder` and `allOf`. Like it 👍

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,120 +44,315 @@
...
+            allOf(
+                hasProperty("claimType", is(URI.create("test"))),
+                hasProperty("values", containsInAnyOrder("testValue"))),
+            allOf(
+                hasProperty("claimType", is(URI.create("test1"))),
+                hasProperty("values", containsInAnyOrder("testing1", "testing2", "testing3")))));
+  }
+
+  @Test
+  public void testRetrieveClaimValuesLocalhost() throws IOException {
```
lessarderic: ❓Is this testing that no claims are returned for a user that doesn't exist in the config file? If so, could rename the test method to make that clear? And what's the difference between this test and `testNoMatchRetrieveClaimValues`?

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,120 +44,315 @@
...
-        assertThat(processedClaim.getValues().size(), is(1));
-        assertThat(processedClaim.getValues().get(0), is("testValue"));
-      } else if (processedClaim.getClaimType().toString().equals("test1")) {
-        assertThat(processedClaim.getValues().size(), is(3));
-        assertTrue(processedClaim.getValues().contains("testing1"));
-        assertTrue(processedClaim.getValues().contains("testing2"));
-        assertTrue(processedClaim.getValues().contains("testing3"));
-      }
-    }
+  public void testGetSingleValuedSystemHighAttribute() throws IOException {
```
lessarderic: ❓For the tests related to the system high attributes, do you think we should use a separate file that contains a `system` user and security related attributes, e.g., `classification`, etc., instead of using generic user, attribute names and values? Might make it easier to understand what the tests are doing and testing.

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
```diff
@@ -44,120 +44,315 @@
...
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void testNoUsersDotAttributesFile() {
+    // given
+    System.setProperty(SystemBaseUrl.HOST, "user1");
+
+    final AttributeFileClaimsHandler attributeFileClaimsHandler = new AttributeFileClaimsHandler();
+    attributeFileClaimsHandler.setAttributeFileLocation("noFileHere");
+    attributeFileClaimsHandler.init();
```
lessarderic: ❓ Any reason we're calling `init()` twice in this and following tests?

distribution/docs/src/main/resources/content/_configuring/updating-system-users.adoc
```diff
@@ -46,6 +46,19 @@ It is possible for a regex in `users.attributes` to match users as well as a sys
...
+If your data will contain security markings, and these markings are being parsed out into the metacrd security attributes via a PolicyPlugin, then the FQDN user *MUST* be updated with attributes that would grant the privileges to ingest that data.
 Failure to update the FQDN user with sufficient attributes will result in an error being returned for any ingest request.
 ====
+
+[WARNING]
+====
+The following attribute values are not allowed:
+
+* `null`
+* an empty String (`""`)
```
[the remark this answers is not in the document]
emmberk: I think it reads better with the a/an. Updated to
```adoc
[WARNING]
====
The following attribute values are not allowed:

* `null`
* `""`
* a non-String (e.g. `100`, `false`)
* an array including any of the above
* `[]`

Additionally, attribute names should not be repeated, and the order that the attributes are defined and the order of values within an array will be ignored.
====
```

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -110,54 +125,94 @@ public ProcessedClaimCollection retrieveClaimValues(
...
-    return supportedRealms;
+  public Set<String> getValues(final String attributeName) {
+    return Optional.ofNullable(systemHighUserAttributes.get(attributeName))
+        .orElse(Collections.emptySet());
   }
 
-  @Override
-  public String getHandlerRealm() {
-    return realm;
+  public void setAttributeFileLocation(String attributeFileLocation) {
```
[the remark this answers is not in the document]
emmberk: 👍  Oops. I made that update in the original PR but lost it in my last push. I should probably add a unit test for that too.

distribution/docs/src/main/resources/content/_configuring/updating-system-users.adoc
```diff
@@ -46,6 +46,19 @@ It is possible for a regex in `users.attributes` to match users as well as a sys
 
 [WARNING]
 ====
-If your data will contain security markings, and these markings are being parsed out into the METACARD.security attribute via a PolicyPlugin, then the FQDN user *MUST* be updated with attributes that would grant the privileges to ingest that data.
+If your data will contain security markings, and these markings are being parsed out into the metacrd security attributes via a PolicyPlugin, then the FQDN user *MUST* be updated with attributes that would grant the privileges to ingest that data.
 Failure to update the FQDN user with sufficient attributes will result in an error being returned for any ingest request.
 ====
+
+[WARNING]
```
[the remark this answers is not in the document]
emmberk: Do you know of a way to add comments to a json file?

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -13,81 +13,96 @@
...
+    if (!path.isAbsolute()) {
+      path = DDF_HOME_PATH.resolve(path);
+    }
+
+    final Map<String, Map<String, Object>> attributesFileContents;
+    final Type type = new TypeToken<Map<String, Map<String, Object>>>() {}.getType();
+    try (final Reader reader = Files.newBufferedReader(path)) {
+      attributesFileContents = new Gson().fromJson(reader, type);
+    } catch (NoSuchFileException e) {
+      throw new IllegalStateException("Cannot find attributes file at \"" + path + "\"", e);
```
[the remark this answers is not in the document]
emmberk: Do you mean log an error message in addition to the throwing an exception with a message?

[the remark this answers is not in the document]
emmberk: I updated the messages to include the configured "location".

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
```diff
@@ -110,54 +125,94 @@ public ProcessedClaimCollection retrieveClaimValues(
...
+      Set<String> stringValues = new HashSet<>();
+      for (Object multiValuedAttributeValue : multiValuedAttributeValues) {
+        if (!(multiValuedAttributeValue instanceof String)) {
+          throw new IllegalStateException(
+              "Expected the attribute values that are a Collection to only contain String values in the attributes file");
+        }
+
+        stringValues.add((String) multiValuedAttributeValue);
+      }
+      return ImmutableSet.copyOf(stringValues);
```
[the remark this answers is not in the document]
emmberk: Updated to `ImmutableSet.toImmutableSet()`

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/TestAttributeFileClaimsHandler.java
emmberk: `localhost` is included in the users.attributes file being tested here:
```json
{
    "user1" : {
        "test" : "testValue",
        "test1" : [ "testing1", "testing2", "testing3" ]
    },
    "localhost" : {

    },
    ".*host.*" : {
        "reg" : "ex"
    }
}
```
`testNoMatchRetrieveClaimValues` is tested a name that isn't in the users.attributes file. I updated the principal name in `testNoMatchRetrieveClaimValues` to clarify this. Let me know if there is something I can do to make this clearer.

emmberk:
build now

platform/security/sts/security-sts-propertyclaimshandler/src/test/java/org/codice/ddf/security/sts/claims/property/UsersAttributesFileClaimsHandlerTest.java
```diff
@@ -0,0 +1,444 @@
...
+    final UsersAttributesFileClaimsHandler usersAttributesFileClaimsHandler =
+        new UsersAttributesFileClaimsHandler();
+    usersAttributesFileClaimsHandler.setUsersAttributesFileLocation(
+        getPathForValidTestAttributesFile());
+
+    // when
+    final List<URI> supportedClaimTypes = usersAttributesFileClaimsHandler.getSupportedClaimTypes();
+
+    // then
+    // TODO should this be contains instead of containsInAnyOrder?
```
emmberk: @emmberk reminder to address this

platform/security/sts/security-sts-propertyclaimshandler/src/main/java/org/codice/ddf/security/sts/claims/property/AttributeFileClaimsHandler.java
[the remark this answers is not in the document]
emmberk: Updated to 
```java
      ImmutableSet.Builder<String> immutableSetBuilder = ImmutableSet.builder();
      multiValuedAttributeValues.forEach(
          multiValuedAttributeValue -> {
            if (!(multiValuedAttributeValue instanceof String)) {
              final String reason =
                  "Expected the attribute values that are a Collection to only contain String values";
              final String errorMessage = createErrorMessage(reason);
              LOGGER.error(errorMessage);
              throw new IllegalStateException(reason);
            }

            immutableSetBuilder.add((String) multiValuedAttributeValue);
          });

      return immutableSetBuilder.build();
```

emmberk:
build now

emmberk:
build now

emmberk:
build now

emmberk:
build now

emmberk:
I manually confirmed that itests for downstream projects succeed

emmberk:
build now

emmberk:
build now

emmberk:
build now

PR #2696, opened 2017-11-20
DDF-2971 Refactor the Jenkinsfile to match updated Jenkins DSL

mcalcote:
#### What does this PR do?
Removed scripted parallel stage syntax and TODOs related to the DSL.

#### Who is reviewing it? 
@oconnormi @mackncheesiest @vinamartin @blen-desta @LinkMJB 

@codice/build 

@clockard @shaundmorris

These changes were run in the local test jenkins under "matt test multibranch".

[DDF-2971](https://codice.atlassian.net/browse/DDF-2971)

#### Notes on Review Process

* Successful CI

```
____

```

#### Review Comment Legend:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,12 @@
 //"Jenkins Pipeline is a suite of plugins which supports implementing and integrating continuous delivery pipelines into Jenkins. Pipeline provides an extensible set of tools for modeling delivery pipelines "as code" via the Pipeline DSL."
 //More information can be found on the Jenkins Documentation page https://jenkins.io/doc/
 pipeline {
-    agent none
+    agent { label 'linux-large' }
     options {
         buildDiscarder(logRotator(numToKeepStr:'25'))
         disableConcurrentBuilds()
         timestamps()
+        skipDefaultCheckout()
     }
     triggers {
         /*
@@ -20,7 +21,7 @@ pipeline {
         POMFIX = 'libs/libs-pomfix,libs/libs-pomfix-run'
         LARGE_MVN_OPTS = '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC '
         LINUX_MVN_RANDOM = '-Djava.security.egd=file:/dev/./urandom'
-        COVERAGE_EXCLUSIONS = '**/test/**,**/itests/**,**/*Test*,**/sdk/**,**/*.js,**/node_modules/**,**/jaxb/,**/wsdl/,**/nces/sws/**,**/*.adoc,**/*.txt,**/*.xml'
+        COVERAGE_EXCLUSIONS = '**/test/**/*,**/itests/**/*,**/*Test*,**/sdk/**/*,**/*.js,**/node_modules/**/*,**/jaxb/**/*,**/wsdl/**/*,**/nces/sws/**/*,**/*.adoc,**/*.txt,**/*.xml,**/platform-solr-server-standalone/**/*'
     }
     stages {
         stage('Setup') {
@@ -30,7 +31,6 @@ pipeline {
         }
         // Use the pomfix tool to validate that bundle dependencies are properly declared
         stage('Validate Poms') {
-            agent { label 'linux-small' }
             steps {
                 retry(3) {
                     checkout scm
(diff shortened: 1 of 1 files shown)


mackncheesiest:
So, in theory, the single `linux-large` agent that is provisioned at the beginning of the pipeline should be used throughout validate-poms -> linux build -> owasp -> deploy stages. How will this behave with `sonarqube` and `coverity` both not specifying an agent to use?

Will just one of them get to use the same `linux-large` container and the other will spin up a new one? Will they run in parallel on the same container?

oconnormi:
@mackncheesiest I think that we can only have a single parallel branch that doesn't declare an agent (meaning it will use the default one) otherwise two parallel blocks would run at the same time on the same agent. @mcalcote it looks like an agent declaration is missing from one of the final stages parallel branches

Jenkinsfile
```diff
@@ -30,11 +30,7 @@ pipeline {
         }
         // Use the pomfix tool to validate that bundle dependencies are properly declared
         stage('Validate Poms') {
-            agent { label 'linux-small' }
             steps {
-                retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
                     expression { env.CHANGE_TARGET != null }
                 }
             }
-            // TODO DDF-2971 refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
-            steps {
-                parallel(
-                    linux: {
-                        node('linux-large') {
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
+                                sh 'mvn install -B -pl !$DOCS -DskipStatic=true -DskipTests=true -T 1C'
+                                sh 'mvn clean install -B -T 1C -pl !$ITESTS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
+                                sh 'mvn install -B -pl $ITESTS -nsu'
                             }
                         }
-                    },
-                    windows: {
-                        node('proxmox-windows') {
-                            bat 'git config --system core.longpaths true'
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
         }
         // The full build will be run against all regular branches
         stage('Full Build') {
             when { expression { env.CHANGE_ID == null } }
-            // TODO DDF-2971 refactor this stage from scripted syntax to declarative syntax to match the rest of the stages - https://issues.jenkins-ci.org/browse/JENKINS-41334
-            steps{
-                parallel(
-                    linux: {
-                        node('linux-large') {
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
+                            withMaven(maven: 'Maven 3.3.9', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}') {
+                                sh 'mvn clean install -B -T 1C -pl !$ITESTS'
+                                sh 'mvn install -B -pl $ITESTS -nsu'
                             }
                         }
-                    },
-                    windows: {
-                        node('proxmox-windows') {
-                            bat 'git config --system core.longpaths true'
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
-                )
+                }
             }
         }
         stage('Security Analysis') {
-            steps {
-                parallel(
-                    owasp: {
-                        node('linux-large') {
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
+                                    sh 'mvn install -q -B -Powasp -DskipTests=true -DskipStatic=true -pl !$DOCS'
+                                } else {
+                                    sh 'mvn install -q -B -Powasp -DskipTests=true -DskipStatic=true -pl !$DOCS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
                                 }
                             }
                         }
-                    },
-                    nodeJsSecurity: {
-                        node('linux-small') {
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -180,78 +150,63 @@ pipeline {
             }
             steps{
                 withMaven(maven: 'M3', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LINUX_MVN_RANDOM}') {
-                    retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -180,78 +150,63 @@ pipeline {
...
                     sh 'mvn deploy -B -T 1C -DskipStatic=true -DskipTests=true -DretryFailedDeploymentCount=10'
                 }
             }
         }
         stage('Quality Analysis') {
-            steps {
-                parallel(
-                    sonarqube: {
-                        node('linux-large') {
-                            retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -180,78 +150,63 @@ pipeline {
...
                         }
-                    },
-                    // Coverity will be skipped on all PR builds
-                    coverity: {
-                        node('linux-medium') {
-                            script {
-                                if (env.BRANCH_NAME != 'master') {
-                                    echo "Coverity is only run on master"
-                                } else {
-                                    retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -180,78 +150,63 @@ pipeline {
...
                                         }
                                     }
                                 }
                             }
                         }
-                    },
-                    codecov: {
-                        node('linux-large') {
-                            script {
-                                retry(3) {
```
LinkMJB: @mcalcote are we losing the retry functionality by switching away from the scripted syntax? Is there a way we can advise this still?

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
+                            // TODO: Maven downgraded to work around a linux build issue. Falling back to system java to work around a linux build issue. re-investigate upgrading later
+                            withMaven(maven: 'Maven 3.3.9', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}') {
+                                sh 'mvn clean install -B -T 1C -pl !$ITESTS'
+                                sh 'mvn install -B -pl $ITESTS -nsu'
                             }
                         }
-                    },
-                    windows: {
-                        node('proxmox-windows') {
-                            bat 'git config --system core.longpaths true'
```
LinkMJB: @mcalcote any way we can still advise the longpath support? I know that `server-2016-large` supports it by default, but under the current configuration we are still serving that node label using Server 2012 until @oconnormi is done with the Server 2016 Jenkins slaves.

oconnormi: @LinkMJB it's actually now pointing at the real server-2016 nodes 😄

Jenkinsfile
```diff
@@ -180,78 +150,63 @@ pipeline {
...
-                                withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}') {
-                                    withCredentials([string(credentialsId: 'DDF_CodeCov', variable: 'DDF_CODECOV_TOKEN')]) {
-                                        sh 'mvn clean install -B -T 1C -pl !$ITESTS'
-                                        sh 'curl -s https://codecov.io/bash | bash -s - -t ${DDF_CODECOV_TOKEN}'
-                                    }
-                                }
+                    }
+                }
+                stage ('Codecov') {
+                    steps {
```
oconnormi: I think you might need an `agent { label 'linux-medium' }` definition here

Jenkinsfile
```diff
@@ -48,129 +44,103 @@ pipeline {
...
-                                checkout scm
-                            }
-                            timeout(time: 3, unit: 'HOURS') {
-                                // TODO: Maven downgraded to work around a linux build issue. Falling back to system java to work around a linux build issue. re-investigate upgrading later
-                                withMaven(maven: 'Maven 3.3.9', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}', options: [artifactsPublisher(disabled: true), dependenciesFingerprintPublisher(disabled: true, includeScopeCompile: false, includeScopeProvided: false, includeScopeRuntime: false, includeSnapshotVersions: false)]) {
-                                    sh 'mvn install -B -pl !$DOCS -DskipStatic=true -DskipTests=true -T 1C'
-                                    sh 'mvn clean install -B -T 1C -pl !$ITESTS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
-                                    sh 'mvn install -B -pl $ITESTS -nsu'
-                                }
+            parallel {
```
oconnormi: @mcalcote another thing we might want to consider, now that the parallel steps actually fit the declarative syntax, is that we could actually collapse the `Incremental Build` and `Full Build` Stages into a single stage with multiple conditional parallel branches
Since each parallel stage supports the full syntax we could move the conditions from the parent stage into each parallel stage making the pipeline a little simpler. We would then have a single build stage that would run different parallel stages depending on whether it was a PR or a regular branch

```
stage('Build') {
  parallel {
    stage('Linux Full Build') {
      when { expression { env.CHANGE_ID == null } }
      ...
    }
    stage('Windows Full Build') {
      when { expression { env.CHANGE_ID == null } }
      ...
    }
    stage('Linux Incremental Build') {
      when {
        allOf {
          expression { env.CHANGE_ID != null }
          expression { env.CHANGE_TARGET != null }
        }
        ...
      }
      stage('Windows Incremental Build') {
      when {
        allOf {
          expression { env.CHANGE_ID != null }
          expression { env.CHANGE_TARGET != null }
        }
        ...
      }
    }
  }
}
```

Jenkinsfile
mcalcote: Looking through the documentation about it, I didn't see the same options for the functionality we previously had set. https://jenkins.io/doc/book/pipeline/syntax/#options
One advantage to this change is that we significantly reduce the number of checkouts we are performing overall. I feel like, in this case, the benefit would outweigh the loss of retry on a per checkout basis.

Jenkinsfile
mcalcote: One downside of this is that given the current way jenkins reports build progress across nodes it would appear that all 4 stages are being run. In the  current arrangement it at least shows half of the nodes being skipped. I can change it to this, but I want to make sure we are alright with that.

Jenkinsfile
mcalcote: Will add :+1:

Jenkinsfile
mackncheesiest: It looks like `retry` should still be a "basic pipeline step" (https://jenkins.io/doc/pipeline/steps/workflow-basic-steps/#code-retry-code-retry-the-body-up-to-n-times). So, I'm not actually sure that we're unable to just leave that retry block in there. https://github.com/jenkinsci/pipeline-model-definition-plugin/wiki/Running-multiple-steps#timeouts-retries-and-more

oconnormi: yeah we can use the `retry` step with `checkout` but we'll need to add the `skipDefaultCheckout() ` option to the pipeline to avoid automatically running a checkout as part of the `agent` setup

mcalcote: I can add the option and set the checkout with retry on the first stage that uses the `linux-large` agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

Jenkinsfile
mcalcote: Adding the retry logic back into the top level agent as well as each stage where we use an alternate agent.

clockard:
Was playing around with the sonarqube exclusions and I think we might want to update our exclusion list to `**/test/**/*,**/itests/**/*,**/*Test*,**/sdk/**/*,**/*.js,**/node_modules/**/*,**/jaxb/**/*,**/wsdl/**/*,**/nces/sws/**/*,**/*.adoc,**/*.txt,**/*.xml,**/platform-solr-server-standalone/**/*` and add the property `-Dsonar.exclusions=${COVERAGE_EXCLUSIONS}` I did two runs on the 28th with the first one using the current exclusions and second one using the updated exclusions and the new property. https://sonarcloud.io/project/activity?id=ddf&selected_date=2017-11-28T23%3A16%3A01%2B0000

Jenkinsfile
```diff
@@ -20,7 +21,7 @@ pipeline {
         POMFIX = 'libs/libs-pomfix,libs/libs-pomfix-run'
         LARGE_MVN_OPTS = '-Xmx8192M -Xss128M -XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC '
         LINUX_MVN_RANDOM = '-Djava.security.egd=file:/dev/./urandom'
-        COVERAGE_EXCLUSIONS = '**/test/**,**/itests/**,**/*Test*,**/sdk/**,**/*.js,**/node_modules/**,**/jaxb/,**/wsdl/,**/nces/sws/**,**/*.adoc,**/*.txt,**/*.xml'
+        COVERAGE_EXCLUSIONS = '**/test/**/*,**/itests/**/*,**/*Test*,**/sdk/**/*,**/*.js,**/node_modules/**/*,**/jaxb/**/*,**/wsdl/**/*,**/nces/sws/**/*,**/*.adoc,**/*.txt,**/*.xml,**/platform-solr-server-standalone/**/*'
```
shaundmorris: ❓ Platform-solr-server-standalone doesn't have any code donated to Codice, correct?

Jenkinsfile
```diff
@@ -48,129 +48,112 @@ pipeline {
...
-                                        sh 'mvn install -q -B -Powasp -DskipTests=true -DskipStatic=true -pl !$DOCS'
-                                    } else {
-                                        sh 'mvn install -q -B -Powasp -DskipTests=true -DskipStatic=true -pl !$DOCS -Dgib.enabled=true -Dgib.referenceBranch=/refs/remotes/origin/$CHANGE_TARGET'
-                                    }
+            parallel {
+                stage ('Owasp') {
+                    steps {
+                        withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}') {
+                            script {
+                                // If this build is not a pull request, run full owasp scan. Otherwise run incremntal scan
```
shaundmorris: ✏️  `incremental` in the comment

Jenkinsfile
```diff
@@ -180,78 +163,70 @@ pipeline {
...
-                                        }
+            parallel {
+                stage ('SonarCloud') {
+                    steps {
+                        withMaven(maven: 'M35', jdk: 'jdk8-latest', globalMavenSettingsConfig: 'default-global-settings', mavenSettingsConfig: 'codice-maven-settings', mavenOpts: '${LARGE_MVN_OPTS} ${LINUX_MVN_RANDOM}') {
+                            withCredentials([string(credentialsId: 'SonarQubeGithubToken', variable: 'SONARQUBE_GITHUB_TOKEN'), string(credentialsId: 'sonarqube-token', variable: 'SONAR_TOKEN')]) {
+                                script {
+                                    // If this build is not a pull request, run sonar scan. otherwise run incremental scan
+                                    if (env.CHANGE_ID == null) {
+                                        sh 'mvn -q -B -Dfindbugs.skip=true -Dcheckstyle.skip=true org.jacoco:jacoco-maven-plugin:prepare-agent install sonar:sonar -Dsonar.host.url=https://sonarcloud.io -Dsonar.login=$SONAR_TOKEN  -Dsonar.organization=codice -Dsonar.projectKey=ddf -Dsonar.exclusions=${COVERAGE_EXCLUSIONS} -pl !$DOCS,!$ITESTS'
```
shaundmorris: Thanks for updating the sonar URL

shaundmorris: ✏️  We should be able to remove `-Dfindbugs.skip=true` since findbugs was removed. Not a blocker for this PR but would be nice clean-up on the next change

Jenkinsfile
mcalcote: @clockard was there something being analyzed in `platform-solr-server-standalone`? It looks like most of the files in this war would have been ignored based on the existing exclusions.

Jenkinsfile
mcalcote: nice catch 👍

Jenkinsfile
mcalcote: I will pull it out now

Jenkinsfile
clockard: I added that the same time I updated the rest of the path expressions so it might not be doing anything. I just wanted to make sure that nothing from that module was included in the results but didn't look at what was actually in that module.

clockard:
CI build not needed for this pr since the jenkins file isn't used in pr builds right now.

PR #2713, opened 2017-11-28
DDF-3460 Address Sonarqube findings

clockard:
#### What does this PR do?
Addresses many of the sonarqube findings found here https://sonarcloud.io/project/issues?id=ddf&resolved=false&types=BUG
I tried to keep the different types of fixes in there own commit to help with reviewing.
#### Who is reviewing it? 
@peterhuffer @AzGoalie @emmberk 

#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)

@bdeining
@coyotesqrl
@figliold
@rzwiefel
@shaundmorris

#### How should this be tested? (List steps with links to updated documentation)
Install and verify that all basic functionality still works as expected.

#### What are the relevant tickets?
[DDF-3460](https://codice.atlassian.net/browse/DDF-3460)

#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #2724, opened 2017-11-30
DDF-3478 Updates Simple Search to use the same configuration as other UIs

vinamartin:
#### What does this PR do?
Updates Simple Search to use the same configuration as other UIs, which means, if a user configures their system to use header and footers or a system usage modal it will now show up on Simple Search UI without having to configure it separately.

I've also deleted out of date documentation and the ability to configure Simple Search UI separately. 

#### Who is reviewing it? 
@garrettfreibott 
@brjeter 
@tbatie 
@djblue 

@lessarderic 
@coyotesqrl 

Build and run a DDF,

- Verify that without configuring anything, simple search UI works out of the box `/search/simple`
- Verify that after updating Platform UI configuration, the configuration is visible in simple search

screen shot 2017-11-30 at 09 58 07
screen shot 2017-11-30 at 09 58 39
screen shot 2017-11-30 at 09 58 47

[DDF-3478](https://codice.atlassian.net/browse/DDF-3478)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

* Successful CI

```
____

```

#### Review Comment Legend:


catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/ConfigurationStore.java
```diff
@@ -1,98 +0,0 @@
-/**
```
vinamartin: Renamed this file to UiConfigurationPropertiesFactory, see below

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationProperties.java
```diff
@@ -0,0 +1,102 @@
...
+
+public class UiConfigurationProperties {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(UiConfigurationProperties.class);
+
+  private Dictionary<String, Object> properties;
+
+  private Optional<BrandingRegistry> branding = Optional.empty();
+
+  UiConfigurationProperties(
```
vinamartin: Looks up the configuration the other UIs are using for header, footer, modal, etc. and sets it locally to be accessed by getters in the jsp

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,62 @@
...
+  private static ConfigurationAdmin configurationAdmin;
+
+  private UiConfigurationPropertiesFactory() {}
+
+  /**
+   * @return a new UiConfigurationProperties object that contains the properties to build out this
+   *     UI
+   */
+  public static UiConfigurationProperties getProperties() {
+    return new UiConfigurationProperties(configurationAdmin, CONFIG_PID, branding);
```
vinamartin: Creates the properties object each time the page loads.
This is to prevent out of sync configurations and reduce the amount of times we have to fetch the configuration.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /** @return a unique instance of {@link UiConfigurationPropertiesFactory} */
+  public static synchronized UiConfigurationPropertiesFactory getInstance() {
+
+    if (uniqueInstance == null) {
```
vinamartin: This class use to be a singleton, but should it still be at this point?

catalog/ui/search-ui/simple/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -1,45 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
vinamartin: Removes the metatype for this as it won't be used any more.

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -18,10 +18,10 @@
 <head>
 <meta charset="utf-8">
 
-
-
-
-<title><%=ConfigurationStore.getInstance().getProductName()%> Search</title>
+<%
+  UiConfigurationProperties props = UiConfigurationPropertiesFactory.getInstance().getProperties();
```
vinamartin: Creates `props` on page load

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -475,5 +475,29 @@
 
  	<script type="text/javascript" src="js/Search-min.js?bust=${timestamp}"></script>
 
+    <!-- Usage Modal -->
+    <!-- Only show if the modal is enabled -->
+    <% if(props.getSystemUsageEnabled() == true) { %>
```
vinamartin: When `systemUsageEnabled` is selected to `true` in the configuration, then run this script and display this modal, otherwise, keep it hidden. 

This is at the bottom of the page because we need jQuery here.

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -475,5 +475,29 @@
...
+    <!-- Do not allow users to click outside of the modal to dismiss it, the OK button must be pressed -->
+    <script>
+        jQuery(function () {
+            $('#usageModal').modal({
+                backdrop: 'static',
```
vinamartin: Prevents users from clicking outside of the modal or pressing escape to dismiss it. 
Also adds darkens the page behind the modal to bring focus onto it.

distribution/docs/src/main/resources/content/_tables/search.simple.properties.adoc
```diff
@@ -1,48 +0,0 @@
-:title: Simple Search UI
```
vinamartin: Removes outdated documentation.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
coyotesqrl: ✏️ In that you're using dependency injection through blueprint to work with this, I think you should keep this a singleton and change the static `getProperties` method and static fields to be instance method/fields.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationProperties.java
```diff
@@ -0,0 +1,102 @@
...
+      ConfigurationAdmin configurationAdmin,
+      String configPid,
+      Optional<BrandingRegistry> brandingRegistry) {
+    this.branding = brandingRegistry;
+    try {
+      if (configurationAdmin.listConfigurations(String.format("(service.pid=%s)", configPid))
+          != null) {
+        properties = configurationAdmin.getConfiguration(configPid, null).getProperties();
+      } else {
+        properties = new Hashtable<>();
```
coyotesqrl: ✏️ This would benefit from the new `DictionaryMap` class introduced in #2713

paouelle: I would actually say that we should no longer be using Hashtable.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+    return uniqueInstance;
+  }
+
+  public Object clone() throws CloneNotSupportedException {
+    throw new CloneNotSupportedException();
+  }
+
+  public void setBranding(BrandingRegistry branding) {
+    UiConfigurationPropertiesFactory.branding = Optional.ofNullable(branding);
```
paouelle: ✏️  If branding can be `null` then the parameter should be annotated with `@Nullable`. Otherwise, you should be safe to use `Optional.of()`

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public void setConfigurationAdmin(ConfigurationAdmin configurationAdmin) {
```
paouelle: ❓ Why is this setter an instance one if the attribute being changed is static?

vinamartin: I'm going to be refactoring these to not be static.

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -475,5 +475,29 @@
...
+                keyboard: false
+            })
+        })
+    </script>
+    <div id="usageModal" class="modal hide" tabindex="-1" role="dialog"
+        aria-labelledby="myModalLabel" aria-hidden="true">
+        <div class="modal-header">
+            <h3><%=props.getSystemUsageTitle()%></h3>
```
garrettfreibott: ❗ Unless we're already sanitizing it somewhere else, we should make sure to escape html anywhere that user/admin input is being displayed. I think JSP has some built in methods to escape html characters or they can probably be escaped within the get methods themselves.

vinamartin: Great catch, @garrettfreibott ! I'll look into what JSP has as far as escaping stuff like this otherwise I'll see what I can do at the Java level.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationProperties.java
```diff
@@ -0,0 +1,102 @@
...
+      }
+    } catch (IOException | InvalidSyntaxException e) {
+      LOGGER.error(
+          "Failed to retrieve UI configuration for Simple Search, page may not load with the proper configuration.");
+      properties = new Hashtable<>();
+    }
+  }
+
+  private String getStringProperty(String string) {
```
brjeter: ✏️  Could we call this parameter something more descriptive like `propertyKey` or just `key`?

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationProperties.java
```diff
@@ -0,0 +1,102 @@
...
+    String val = (String) properties.get(string);
+    return (val != null) ? val : "";
+  }
+
+  private Boolean getBoolProperty(String string) {
```
brjeter: ✏️  Same as above

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationProperties.java
vinamartin: Certainly, I have no idea why I just called it "string".

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -15,84 +15,48 @@
 
 import java.util.Optional;
 import org.codice.ddf.branding.BrandingRegistry;
+import org.osgi.service.cm.ConfigurationAdmin;
 
 /** Stores external configuration properties. */
```
brjeter: ✏️  This javadoc should probably be updated or removed.

catalog/ui/search-ui/simple/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -17,12 +17,15 @@
            xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0">
 
     <reference id="brandingRegistry" interface="org.codice.ddf.branding.BrandingRegistry" />
+    <reference id="configurationAdmin" interface="org.osgi.service.cm.ConfigurationAdmin"/>
+
     <bean id="configurationStore"
```
brjeter: ✏️  Should this id be updated?

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -15,84 +15,48 @@
...
-    header = "";
-    footer = "";
-    color = "";
-    background = "";
+  /**
+   * @return a new UiConfigurationProperties object that contains the properties to build out this
+   *     UI
+   */
+  public UiConfigurationProperties getProperties() {
+    return new UiConfigurationProperties(configurationAdmin, CONFIG_PID, branding);
```
brjeter: ❓ I'm wondering if it makes more sense to have this factory class be responsible for querying configAdmin for the configuration and just passing the properties into the `UiConfigurationProperties` object. That would simplify the `UiConfigurationProperties` class and keep the configAdmin dependency in this class only.

catalog/ui/search-ui/simple/src/main/resources/OSGI-INF/blueprint/blueprint.xml
vinamartin: Probably should be, great catch.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
vinamartin: My original implementation was actually just having Configuration Store (the original class) do the lookup. I'm not actually sure which is better in this case.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationProperties.java
```diff
@@ -0,0 +1,102 @@
...
+    Boolean val = (Boolean) properties.get(string);
+    return (val != null) ? val : false;
+  }
+
+  public String getProductName() {
+    return branding.map(BrandingRegistry::getProductName).orElse("");
+  }
+
+  public BrandingRegistry getBranding() {
```
lessarderic: ❓ Is this still called? If so, we may want to add `@Nullable` on the return type.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
lessarderic: I vote for updated 😄

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
lessarderic: Agree with @brjeter with getting the `Configuration` object from `ConfigurationAdmin` here. I would let the `UiConfigurationProperties` class take care of getting the values out of the `Configuration` or properties object though. I think it would make for a better separation of concerns.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -15,84 +15,48 @@
...
   }
 
-  /** @return a unique instance of {@link ConfigurationStore} */
-  public static synchronized ConfigurationStore getInstance() {
+  /** @return a unique instance of {@link UiConfigurationPropertiesFactory} */
+  public static synchronized UiConfigurationPropertiesFactory getInstance() {
```
lessarderic: ❓ Would it be easier to simply create the `UiConfigurationPropertiesFactory` where we declare the `uniqueInstance` field (L25) instead of doing lazy creation and having to make the code synchronized?

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -15,84 +15,48 @@
...
-    this.background = background;
-  }
-
-  public String getProductName() {
-    return branding.map(BrandingRegistry::getProductName).orElse("");
-  }
-
-  public BrandingRegistry getBranding() {
-    return branding.orElse(null);
+  public Object clone() throws CloneNotSupportedException {
```
lessarderic: ❓ Can we delete this method?

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -62,23 +62,23 @@
 <body>
 
 	<div class="navbar navbar-inverse navbar-fixed-top">
-		<% String h = ConfigurationStore.getInstance().getHeader();
+		<% String h = props.getHeader();
 			if(h != null && h.trim().length() > 0)
```
lessarderic: ✏️ Don't think any of the `UiConfigutationProperties.get...()` can return `null`, so we probably don't need those `null` checks.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
vinamartin: Yeah, I think so, I'm not sure why it was like this before. I'm guessing it was to make it so this only existed if the configuration existed.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
vinamartin: I think so.

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
vinamartin: Yes, I was actually just about to remove these!

vinamartin: ✏️  Just a note here: I'm going to be getting rid of this whole println thing as well. 
This gets weird as far as sanitizing the inputs goes and I don't think this is the best way to conditionally add HTML to the page.

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -475,5 +475,29 @@
 
  	<script type="text/javascript" src="js/Search-min.js?bust=${timestamp}"></script>
 
+    <!-- Usage Modal -->
+    <!-- Only show if the modal is enabled -->
+    <% if(props.getSystemUsageEnabled() == true) { %>
+    <!-- Do not allow users to click outside of the modal to dismiss it, the OK button must be pressed -->
+    <script>
```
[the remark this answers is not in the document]
vinamartin: Sure, I actually have if blocks indented on my local branch

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
vinamartin: Probably at some point. 
It looks like we created that `PlatformUIConfiguration` to really only be used to get some JSON back for our other UIs so a bit of refactoring would have to be done here.

catalog/ui/search-ui/simple/src/main/java/org/codice/ddf/ui/searchui/simple/properties/UiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * @return a new UiConfigurationProperties object that contains the properties to build out this
+   *     UI
+   */
+  public UiConfigurationProperties getProperties() {
+    Dictionary<String, Object> properties;
+    try {
+      if (configurationAdmin.listConfigurations(String.format("(service.pid=%s)", CONFIG_PID))
+          != null) {
+        properties = configurationAdmin.getConfiguration(CONFIG_PID, null).getProperties();
```
brjeter: ✏️  We have seen weird issues manifest when using this `getConfiguration` method because if it doesn't find a configuration it creates an empty one for you. I'd recommend storing the above `listConfigurations` call to a variable and store it, then reuse it here. I don't like that you have to store the array of `Configuration`s but I think it' might be better than getting it and throwing it away.

vinamartin: If I'm not mistaken, passing the location as `null` should prevent that issue from happening.

brjeter: That's true. I don't think there's any functional issue with this code. I just get really defensive against Config Admin because we've been burned by it so many times. I think whether or not you use the original `listConfigurations` call is semantical. This is nitpicky, so I'm going to approve and leave it up to you 😄

vinamartin: For what it's worth, you being cautious about it is absolutely valid here as when I was testing this I noticed that I was creating that config in my bundle's context when I didn't have that null there.

catalog/ui/search-ui/simple/pom.xml
```diff
@@ -159,17 +164,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0</minimum>
+                                            <minimum>.90</minimum>
```
lessarderic: Nice!

catalog/ui/search-ui/simple/src/test/java/org/codice/ddf/ui/searchui/simple/properties/TestUiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.service.cm.Configuration;
+import org.osgi.service.cm.ConfigurationAdmin;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TestUiConfigurationPropertiesFactory {
```
lessarderic: ❗️ This should be `UiConfigurationPropertiesFactoryTest` to follow our conventions.

catalog/ui/search-ui/simple/src/test/java/org/codice/ddf/ui/searchui/simple/properties/TestUiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,95 @@
...
+    when(configuration.getProperties()).thenReturn(configProps);
+
+    UiConfigurationPropertiesFactory.getInstance().setConfigurationAdmin(configurationAdmin);
+    UiConfigurationProperties props =
+        UiConfigurationPropertiesFactory.getInstance().getProperties();
+
+    assertThat(props.getHeader(), is("Test Header"));
+    assertThat(props.getFooter(), is("Test Footer"));
+    assertThat(props.getSystemUsageEnabled(), is(true));
+  }
```
lessarderic: ❓ Shouldn't we assert the other properties as well?

catalog/ui/search-ui/simple/src/test/java/org/codice/ddf/ui/searchui/simple/properties/TestUiConfigurationPropertiesFactory.java
```diff
@@ -0,0 +1,95 @@
...
+    assertThat(props.getHeader(), is(""));
+    assertThat(props.getFooter(), is(""));
+    assertThat(props.getColor(), is(""));
+    assertThat(props.getBackground(), is(""));
+    assertThat(props.getSystemUsageTitle(), is(""));
+    assertThat(props.getSystemUsageMessage(), is(""));
+    assertThat(props.getSystemUsageEnabled(), is(false));
+    assertThat(props.getSystemUsageOncePerSession(), is(false));
+  }
+}
```
lessarderic: ❗️ We also need tests for the `getProductName()` method, one for when it is set and one for when it isn't.
❗️ We should also make sure that properties that we properly handle properties that have the wrong type, e.g., header isn't a `String` and system usage enabled isn't a `Boolean`.

catalog/ui/search-ui/simple/src/test/java/org/codice/ddf/ui/searchui/simple/properties/TestUiConfigurationPropertiesFactory.java
vinamartin: Ah! Good catch, I couldn't remember which order it went in :)

catalog/ui/search-ui/simple/src/test/java/org/codice/ddf/ui/searchui/simple/properties/TestUiConfigurationPropertiesFactory.java
vinamartin: Fair! I will add these.

catalog/ui/search-ui/simple/src/main/resources/SearchPage.jsp
```diff
@@ -475,5 +476,29 @@
...
+                    keyboard: false
+                })
+            })
+        </script>
+        <div id="usageModal" class="modal hide" tabindex="-1" role="dialog"
+            aria-labelledby="myModalLabel" aria-hidden="true">
+            <div class="modal-header">
+                <h3><%=props.getSystemUsageTitle()%></h3>
+            </div>
+            <div class="modal-body"><%=props.getSystemUsageMessage()%></div>
```
mcalcote: ❗️ If the SystemUsageMessage contains HTML tags for formatting, they will be printed along with the string instead of formatting the text.

vinamartin: Should be fixed with the jsoup change

vinamartin:
build now

PR #2737, opened 2017-12-05
DDF-3484 HitCount-only Queries

dcruver:
#### What does this PR do?
Modifies AtomTransformer to remove any returned entries when the 'count' parameter on the request is 0.

#### Who is reviewing it? 
@mackncheesiest 
@Variadicism 

@bdeining
@jlcsmith

1) ingest a few products
2) execute an opensearch query normally (`https://localhost:8993/services/catalog/query?q=*`)
    - all of the items ingested should be shown
3) execute the same search with `&count=0` (`https://localhost:8993/services/catalog/query?q=*&count=0`)
    - no items will be returned, but the `totalResults` element should show the number of items ingested in step 1.

[DDF-3484](https://codice.atlassian.net/browse/DDF-3484)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

* Successful CI

```
____

```

#### Review Comment Legend:

### DIFF
--- a/catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
+++ b/catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
@@ -40,6 +40,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import javax.activation.MimeType;
 import javax.activation.MimeTypeParseException;
@@ -148,19 +149,8 @@ public class AtomTransformer implements QueryResponseTransformer {
           "Cannot transform null " + SourceResponse.class.getName());
     }
 
-    Date currentDate = new Date();
-
-    ClassLoader tccl = Thread.currentThread().getContextClassLoader();
-    Feed feed = null;
-    try {
-
-      Thread.currentThread().setContextClassLoader(AtomTransformer.class.getClassLoader());
-
-      feed = ABDERA.newFeed();
-
-    } finally {
-      Thread.currentThread().setContextClassLoader(tccl);
-    }
+    final Date currentDate = new Date();
+    final Feed feed = createFeed();
 
     /*
      * Atom spec text (rfc4287) Sect 4.2.14: "The "atom:title" element is a Text construct that
@@ -198,7 +188,6 @@ public class AtomTransformer implements QueryResponseTransformer {
      * required in the atom:feed element.
      */
     if (!StringUtils.isEmpty(SystemInfo.getSiteName())) {
-
       // text is required.
       feed.setGenerator(null, SystemInfo.getVersion(), SystemInfo.getSiteName());
     }
@@ -214,7 +203,6 @@ public class AtomTransformer implements QueryResponseTransformer {
 
     if (sourceResponse.getRequest() != null && sourceResponse.getRequest().getQuery() != null) {
       Element itemsPerPage = feed.addExtension(OpenSearchConstants.ITEMS_PER_PAGE);
-
       Element startIndex = feed.addExtension(OpenSearchConstants.START_INDEX);
 
       /*
@@ -236,11 +224,16 @@ public class AtomTransformer implements QueryResponseTransformer {
     }
 
     if (getCount(sourceResponse) != 0) {
-      addResultElements(sourceResponse, currentDate, feed);
+      sourceResponse.getResults().stream().forEach(r -> addSingleResult(currentDate, feed, r));
     }
 
+    byte[] bytes = createOutputStream(feed);
+    return new BinaryContentImpl(new ByteArrayInputStream(bytes), MIME_TYPE);
+  }
+
+  private byte[] createOutputStream(Feed feed) throws CatalogTransformerException {
     ByteArrayOutputStream baos = new ByteArrayOutputStream();
-    tccl = Thread.currentThread().getContextClassLoader();
+    ClassLoader tccl = Thread.currentThread().getContextClassLoader();
 
     try {
       Thread.currentThread().setContextClassLoader(AtomTransformer.class.getClassLoader());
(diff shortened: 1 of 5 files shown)


catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -473,4 +478,24 @@ private Link addLinkHelper(Action action, Entry entry, String linkType, String m
...
+      return -1;
+    }
+
+    Serializable countSerializable = queryRequest.getProperties().get("count");
+
+    if (countSerializable == null) {
+      return -1;
+    }
+
+    return Integer.valueOf((String) countSerializable);
```
Variadicism: This can throw a `NumberFormatException` if `countSerializable` does not conform to a proper decimal integer format. This should be surrounded by a `try`-`catch` block and that case should be explicitly handled.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -473,4 +478,24 @@ private Link addLinkHelper(Action action, Entry entry, String linkType, String m
...
+    QueryRequest queryRequest = queryResponse.getRequest();
+
+    if (queryRequest == null) {
+      return -1;
+    }
+
+    Serializable countSerializable = queryRequest.getProperties().get("count");
+
+    if (countSerializable == null) {
+      return -1;
```
Variadicism: This method returns -1 in case of error. This method's current usage indicates that if the count cannot be obtained, the results should not be treated as a hit count only query, but returning -1 makes this code unclear if this is the intended behavior. I suggest changing the return type of this method to `Optional<Integer>` and returning `Optional.empty()` for error cases to make the intention clearer.

dcruver: -1 was chosen to emulate the current behavior of this code in cases where an invalid count is supplied. I'd like the use of `Optional` here if this were a public method, but I think switching that would increase the complexity of the calling code. I've added a constant to hopefully make the intent more clear.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
dcruver: Good catch.  I've wrapped this in a try/catch.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -233,6 +235,35 @@ public BinaryContent transform(SourceResponse sourceResponse, Map<String, Serial
...
 
+    if (getCount(sourceResponse) != 0) {
+      addResultElements(sourceResponse, currentDate, feed);
+    }
+
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+    try {
+
+      tccl = Thread.currentThread().getContextClassLoader();
```
paouelle: ✏️  Should be moved outside the first `try`, that way you wouldn't need to embed a second `try/finally` and instead, you could just move the `finally` to the one and only `try` right after the `catch`

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -473,4 +484,28 @@ private Link addLinkHelper(Action action, Entry entry, String linkType, String m
     }
     return georssPositions;
   }
+
+  private int getCount(SourceResponse queryResponse) {
```
paouelle: ✏️  `queryResponse` should be annotated with`@Nullable`

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -233,6 +235,35 @@ public BinaryContent transform(SourceResponse sourceResponse, Map<String, Serial
       startIndex.setText(Integer.toString(sourceResponse.getRequest().getQuery().getStartIndex()));
     }
 
+    if (getCount(sourceResponse) != 0) {
```
paouelle: ❗️ I know the method already had a very high cyclomatic complexity but we are making it worst now. We should take the time and simplify this code while we are here to reduce its complexity.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
bdeining: Do we put `@Nullable` on private methods?

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
bdeining: I agree we should take the time to refactor this to reduce complexity, but it would be ideal to do it in a separate ticket to prevent scope creep.  These additions don't seem to make the class unmanageable or testing impractical.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
millerw8: Looks like this was a preexisting condition, but definitely an easy clean-up

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
millerw8: +1 to Ben, seems like this already took a step towards reducing complexity by separating out the addResultsElements.  This could be better, but incremental changes as it is updated seems reasonable for this rarely used capability.  

I did not get a complexity count, but I am less worried about it since it hasn't been flagged by any of the automated code analysis tools.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
rzwiefel: If we can help another developer down the road by letting them know a response could be null, I'd say we should.

bdeining:
@dcruver I'm wondering if this shouldn't be at the endpoint level.  As it stands right now, we are querying SOLR and getting the results and then dropping them.  Ideally we would pass a query with 0 desired results to SOLR instead.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
dcruver: cleaned up

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
Variadicism: As far as I can tell, `getCount` is only called in one place and that one place just checks that the count `!= 0`. This makes the fact that an invalid count should add all results hidden. With an `Optional`, it is clear that the error is acknowledged and the logic is not much longer:
```
if (getCount(sourceResponse).orElse(-1) != 0) {
```
The calling code is barely more complex, but clearer.

A constant for -1 clarifies the `getCount` method, but since that constant is not used in the `!= 0` check, it still looks suspicious. However, with the constant declaration, I would be comfortable making this :pencil2:.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
mackncheesiest: I would tend to agree that a separate ticket would prevent scope creep, but for the record, both this method and `addResultElements` did actually get flagged by Codacy

[transform](https://www.codacy.com/app/Codice/ddf/file/12919297521/issues/source?bid=4642206&fileBranchId=5986171#l143)

[addResultElements](https://www.codacy.com/app/Codice/ddf/file/12919297521/issues/source?bid=4642206&fileBranchId=5986171#l258)

bdeining: I think the majority of this existed before the proposed changes.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
dcruver: This method returns a simple data type.  It's never null.

Variadicism: No, `queryResponse` can be null. Arguments can be marked with `@Nullable` as well.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
dcruver: To clarify, the contents of `addResultElements()` were previously part of `transform()`. It looks like the code in `addResultElements()` is being added, but it's only being broken out into a new method. The code itself is unmodified except for the try-catch-finally blocks that @paouelle pointed out. Since Codacy is flagging those two methods now, I assume it was also flagging the previous version of `transform()` - which was much longer - as well.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
dcruver: I'm aware of that. My point was in reference to "If we can help another developer down the road by letting them know a response could be null, I'd say we should."

dcruver:
@bdeining - We need this change as well as the one you're proposing because the OpenSearch system at the other end of a Source may not be a DDF instance. I'm putting that change in another PR.

dcruver:
build now

dcruver:
build now

catalog/transformer/catalog-transformer-service-atom/src/test/java/ddf/catalog/transformer/response/query/atom/TestAtomTransformer.java
```diff
@@ -1448,6 +1475,32 @@ protected SourceResponse getSourceResponseStub(String id, String wkt) {
     return response;
   }
 
+  protected SourceResponse getSourceResponseStubZeroCount(String id) {
```
bdeining: Is there a reason why this method is protected vs. private?

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
paouelle: I do not think we should delay this. The fact that the other method was long before doesn't mean we shouldn't do it right now.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
paouelle: I think he meant parameter in this case although the comment applies to both return value and parameters, we should always annotate them with @Nullable even private ones.

bdeining: I'm curious why `we should` be annotating this; is there some guideline that can be referenced?  The fact that the method is annotated with `@nullable` is seemingly of no consequence to other developers (and the null check in the preceding line makes it obvious).  Assuming we are discussing `javax.annotation.Nullable`, this is from the documentation:

`This annotation is useful mostly for overriding a {@link Nonnull} annotation. Static analysis tools should generally treat the annotated items as though they had no annotation, unless they are configured to minimize false negatives. Use {@link CheckForNull} to indicate that the element value should always be checked for a null value.`

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
bdeining: I'm no expert on Codacy, but it looks like the complexity is reduced in this PR : https://www.codacy.com/app/Codice/ddf/pullRequest?prid=1134653.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
paouelle: As I was told by @lessarderic, all methods that returns `null` and all parameters that accepts `null` should now be annotated with @Nullable. This is the convention that was adopted not too long ago. By default, methods do not return `null` and do not accept `null`. So even though looking at the method might tell you, we still want to annotate them to make it clear that this was the intent. Think of it as proper documentation.

bdeining: Where is this convention documented?  This seems to be at odds with the documentation linked above for `@nullable`

Variadicism: I don't see documentation of conventions related to handling `null` for DDF or other Connexta projects.
In this code base and others, the pattern I've noticed is to never pass `null` into something unless it is clear that `null` is an expected value. A method's signature should be sufficient to explain what the method does and how to use it. The `@Nullable` annotation can clarify that `null` as a valid possible value without having to look at the body of the method.
The way that I see it, without `@Nullable`, this method looks like it expects a `SourceResponse` argument. Much of DDF's code seems to assume that arguments are not `null` unless they are marked with `@Nullable`. (See `WorkspaceMetacardImpl` as one such example that I happened upon while working.) Adding `@Nullable` indicates that the first argument can actually be either a `SourceResponse` or `null`.
According to the `Nullable` documentation that you posted, we should be using `@Nonnull` all over the place; however, most things in code are not intended to be null, so using `@Nonnull` on every `Object` expected to not be `null` would result in a high amount of code clutter. Using `@Nullable` when `null` is expected is a less safe, but less cluttered and more common alternative.

paouelle: This is exactly why they switch the convention in DDF from using the @Nonnull annotation to only @Nullable. We are slowly updating code that still use the old @Nonnull annotation as we get into those files.

bdeining: My point is that if this is a DDF convention it should be documented somewhere on the Codice Confluence page if it isn't already.  Right now it seems to be a matter of preference and now a point of contention.  As @paouelle has stated, the drive to use `@Nullable` is recent, and I would argue that this annotation is absent more than it is present in this large code base.  Do we intend on going through the entire code base, adding `@Nullable` where applicable?

paouelle: Find DDF guidelines at https://codice.atlassian.net/wiki/spaces/DDF/pages/6291514/Code+Comment+Guidelines which described this convention

bdeining: Unless i'm mistaken, this describes a convention for annotations for `all externally-facing public and protected methods` and not for private methods.

paouelle: Should we really stop there? At the end of the day, it is meant to clarify the code and the intent.

rzwiefel: I would personally prefer private methods including it as well -- that way if I have to go and update someone elses code somewhere the intents and capability of a method and what should be passed to it are communicated clearly from the beginning.  Having it only on some methods and not others (public / protected and not default / private) seems to add ambiguity to a methods contract.

Also we can use this interface for static analysis benefit, though if we only do it on some of the methods and not others then the benefit of the static analysis is lessened (or wrong even).

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
bdeining: Since this PR does not introduce any more complexity to the AtomTransformer, the following ticket was created to address complexity concerns.
https://codice.atlassian.net/browse/DDF-3528

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
coyotesqrl: I'm with @dcruver. I see zero value in adding that annotation to private methods. Traceability of private methods is trivial for a developer to ensure they don’t break anything, plus static analysis will pick up any potential null accesses missed by humans.

Code clutter is a real concern; Java is overly verbose by nature and adding extra noise for no return on investment makes it more onerous and less desirable as a development language.

dcruver: "to a methods contract": I think you're confusing the scope of the term "contract".  By definition, private methods aren't part of the interface or contract. The intent of @notnull and @nullable is to help clarify the contract of a method that's intentionally being designed for extension. I don't think that blindly adding these annotations to every method parameter and return value is either practical or helpful.

rzwiefel: Regardless of any opinions, they are just 'guidelines' in the end. I don't think anyone would be required to do anything they don't want to. Probably akin to adding javadoc to a private method: Nothing says you have to but nothing also says you can't. 😄

rzwiefel: So those of us that want to be gross and verbose can do as we wish, and those that don't wish to don't have to 😄

bdeining: While this is a style suggestion, I think one of the goals here was to establish some precedent so that it isn't a continuous debate.

rzwiefel: I think in a org of 50 developers working in the same codebase, the fact we've gotten this far harmoniously is a miracle. I don't think a little wiggle room will kill anyone. 

Since it is a guideline, likening it to javadoc on a private method is adequate. It is by no means required but also not disallowed completely. 

The guidelines also state:
```Package private (default visibility), private methods, or otherwise internal code do not have to include JavaDoc comments. If their purpose cannot be clearly conveyed by their name and the name of their parameters, use just as much javadoc as necessary to communicate the behavior.```

Comment (paouelle) [reply catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java]: With this guideline in mine, how does a parameter name say wether it accepts null or not if it ain't annotated and if one doesn't want to start looking at the code to figure out if it should be or if it was meant to be?

Comment (rzwiefel) [reply catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java]: @paouelle I think the point is feel free to use it yourself, but as of now others are not required to use it.

Comment (bdeining) [reply catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java]: I think it would also be situational; in this case, it's obvious that it can be null.

**Comment by bdeining** [comment]
build now

**Comment by bdeining** [comment]
Successfully heroed

**Comment by bdeining** [comment]
build now

File: catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java [hunk 7]
```diff
@@ -233,6 +235,28 @@ public BinaryContent transform(SourceResponse sourceResponse, Map<String, Serial
...
+      throw new CatalogTransformerException("Could not transform into Atom.", e);
+    } finally {
+      Thread.currentThread().setContextClassLoader(tccl);
+    }
+
+    return new BinaryContentImpl(new ByteArrayInputStream(baos.toByteArray()), MIME_TYPE);
+  }
+
+  private void addResultElements(SourceResponse sourceResponse, Date currentDate, Feed feed) {
+    ClassLoader tccl;
```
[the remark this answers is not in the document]
bdeining: https://codice.atlassian.net/browse/DDF-3528

dcruver:
build now

dcruver:
build now

bdeining:
build now

bdeining:
successfully re-hero'd

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -233,156 +223,177 @@ public BinaryContent transform(SourceResponse sourceResponse, Map<String, Serial
...
+    }
 
-      Element source =
-          entry.addExtension(new QName(FEDERATION_EXTENSION_NAMESPACE, "resultSource", "fs"));
+    byte[] bytes = createOutputStream(feed);
+    return new BinaryContentImpl(new ByteArrayInputStream(bytes), MIME_TYPE);
+  }
 
-      /*
-       * According to the os-federation.xsd, the resultSource element text has a max length of
```
brendan-hofmann: ❓ Why was this behavior removed?

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -146,19 +149,8 @@ public BinaryContent transform(SourceResponse sourceResponse, Map<String, Serial
           "Cannot transform null " + SourceResponse.class.getName());
     }
 
-    Date currentDate = new Date();
-
-    ClassLoader tccl = Thread.currentThread().getContextClassLoader();
```
brendan-hofmann: ❓ Why was this behavior removed? It is inconsistent with tccl usage elsewhere in this class.

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
```diff
@@ -233,156 +223,177 @@ public BinaryContent transform(SourceResponse sourceResponse, Map<String, Serial
...
 
-      /*
-       * According to the os-federation.xsd, the resultSource element text has a max length of
-       * 16 and is the shortname of the source id. Previously, we were duplicating the names
-       * in both positions, but since we truly do not have a shortname for our source ids, I
-       * am purposely omitting the shortname text and leaving it as the empty string. The real
-       * source id can still be found in the attribute instead.
-       */
+  private byte[] createOutputStream(Feed feed) throws CatalogTransformerException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
brendan-hofmann: ❓ Shouldn't this be a `FileBackedOutputStream` to help guard against memory usage issues?

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
millerw8: Not removed, just moved below.  many complaints about method complexity resulting in some refactoring

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
millerw8: Not removed, just moved into the createFeed function

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
brendan-hofmann: I did see quite a bit of refactoring, guess I missed some. 😃

catalog/transformer/catalog-transformer-service-atom/src/main/java/ddf/catalog/transformer/response/query/atom/AtomTransformer.java
millerw8: This was a refactoring, old version had a ByteArrayOutputStream, this was not a change.  I think our query restrictions keep this from being a risk with memory issues.  Switching to a filebacked stream would be ok for safety, but I do not expect it would ever actually flip over to use a file.

brendan-hofmann: I am thinking of users with extremely large metacards, but I agree that it is unlikely.

bdeining: Given that this is not a new change, are we comfortable with it as-is?

brendan-hofmann: Yup.

PR #2760, opened 2017-12-14
[2.11.x] DDF-3496 Guest claims profile

ahoffer:
#### What does this PR do?
Guest claims set during the install process were not always reflective of what was presented in the UI. This PR changes the installer to use the guest claims shown by the admin UI.
#### Who is reviewing it? 
@peterhuffer 
@mcalcote 
@rzwiefel 
@garrettfreibott 
@kcrowe01 
@clockard
@figliold
@rzwiefel 
1. Build distro.
2. Unzip distro on Mac and start the application (NOTE:  Must be a configuration that offers installation profiles and the installation profiles must offer choices that affect the guest claims.)
3. Go to Admin UI and start the installation.
4. Pick a profile and observe the change to the default guest claims.
5. Continue through to the end of the installation.
6. After the app restarts, visit the admin UI.
7. Navigate to the Security Add, then the configuration tab, then the STS guest claims. 
8. Observe that the guest claims shown during the installation process match the guest claims that are actually configured in the system.

**Repeat the process on a Windows VM with 2 procs and 4 GB RAM.**

The error was never 100% reproducible, but occurs often during Windows installations.
[DDF-3496](https://codice.atlassian.net/browse/DDF-3496)
#### Checklist:
- [N/A] Documentation Updated
- [None] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an inline code snippet or a gist. 


ahoffer:
First build now

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminConsoleService.java
```diff
@@ -334,7 +334,7 @@ public boolean updateGuestClaimsProfile(String pid, Map<String, Object> configur
       String profile = (String) profileObj;
       guestClaimsHandlerExt.setSelectedClaimsProfileName(profile);
```
figliold: :question: Was the problem that GuestClaimsHandlerExt was not initialized properly before being used here?

ahoffer: I assume that was the problem, though I can't be sure. Probably some timing dependency on parsing the metatype.xml file. From an engineering perspective, it makes sense to use the map that is passed into the method because that is what the user is seeing on the screen. And it is also correct. I feel like something unnecessarily complicated is going on when the method parameter has the values we need but is ignored.

ahoffer:
Clicked "update branch" and it removed by successful build check
build now

clockard:
Do you have an extra commit in this pr from @rzwiefel

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminConsoleService.java
brendan-hofmann: I am inclined to agree, the old way looks wrong on general principle.

PR #2782, opened 2017-12-20
DDF-3489 Address SonarQube findings

bakejeyner:
#### What does this PR do?
Addresses SonarQube major bug findings.

#### Who is reviewing it? 
@blen-desta 
@Lambeaux 
@brjeter 
@idperez 

@coyotesqrl 
@shaundmorris 

Full build and verify basic functionality

[DDF-3489](https://codice.atlassian.net/browse/DDF-3489)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
@@ -43,17 +43,17 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Paths;
 import java.text.ParseException;
-import java.text.SimpleDateFormat;
 import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.TimeZone;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -94,15 +94,14 @@ public class ExportCommand extends CqlCommands {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ExportCommand.class);
 
-  private static final SimpleDateFormat ISO_8601_DATE_FORMAT;
-
-  static {
-    ISO_8601_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH-mm-ss.SSS'Z'");
-    ISO_8601_DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("UTC"));
-  }
+  private static final DateTimeFormatter FORMATTER =
+      DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH-mm-ss.SSS'Z'").withZone(ZoneOffset.UTC);
 
   private static final Supplier<String> FILE_NAMER =
-      () -> "export-" + ISO_8601_DATE_FORMAT.format(Date.from(Instant.now())) + ".zip";
+      () ->
+          String.format(
+              "export-%s.zip",
+              LocalDateTime.ofInstant(Instant.now(), ZoneOffset.UTC).format(FORMATTER));
 
   private static final int PAGE_SIZE = 64;
 
@@ -550,6 +549,7 @@ public class ExportCommand extends CqlCommands {
         revisionFilter, filterBuilder.attribute("metacard.version.id").is().equalTo().text(id));
   }
 
+  @Override
   protected Filter getFilter() throws ParseException, CQLException {
     Filter filter = super.getFilter();
     if (archived) {
--- a/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
+++ b/catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
@@ -303,7 +303,7 @@ public class IngestCommand extends CatalogCommands {
       LOGGER.info("Executor service shutdown was not permitted: {}", e);
     }
 
-    printProgressAndFlush(start, fileCount.get(), ingestCount.get() + ignoreCount.get());
+    printProgressAndFlush(start, fileCount.get(), (long) ingestCount.get() + ignoreCount.get());
     long end = System.currentTimeMillis();
     console.println();
     String elapsedTime = timeFormatter.print(new Period(start, end).withMillis(0));
@@ -503,11 +503,14 @@ public class IngestCommand extends CatalogCommands {
 
   private Metacard generateMetacard(InputStream message) throws IOException {
     try {
-      if (message != null) {
-        return transformer.get().transform(message);
-      } else {
+      if (message == null) {
         throw new IllegalArgumentException("Data file is null.");
       }
+      if (!transformer.isPresent()) {
+        throw new IllegalArgumentException(
+            "Transformation Failed for transformer: " + transformerId);
+      }
+      return transformer.get().transform(message);
 
     } catch (CatalogTransformerException e) {
       throw new IllegalArgumentException(
@@ -614,7 +617,7 @@ public class IngestCommand extends CatalogCommands {
     if (ignoreList != null
         && (ignoreList.contains(extension) || ignoreList.contains(file.getName()))) {
       ignoreCount.incrementAndGet();
-      printProgressAndFlush(start, fileCount.get(), ingestCount.get() + ignoreCount.get());
+      printProgressAndFlush(start, fileCount.get(), (long) ingestCount.get() + ignoreCount.get());
       return;
     }
 
@@ -768,7 +771,8 @@ public class IngestCommand extends CatalogCommands {
                     }
                   });
 
-              printProgressAndFlush(start, fileCount.get(), ingestCount.get() + ignoreCount.get());
+              printProgressAndFlush(
+                  start, fileCount.get(), (long) ingestCount.get() + ignoreCount.get());
             }
           }
         },
--- a/catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
+++ b/catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
@@ -13,6 +13,8 @@
  */
 package ddf.catalog.filter.proxy.adapter;
 
+import static java.lang.Math.abs;
+
 import ddf.catalog.filter.FilterAdapter;
 import ddf.catalog.filter.FilterDelegate;
 import ddf.catalog.impl.filter.FuzzyFunction;
@@ -100,6 +102,8 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
   public static final String CQL_NAUTICAL_MILES = "nautical miles";
 
   public static final String CQL_KILOMETERS = "kilometers";
+
+  private static final double ERROR_THRESHOLD = .000001;
 
   private static final FilterFactory FF = new FilterFactoryImpl();
 
@@ -703,7 +707,7 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
     ExpressionValues filterValues = getExpressions(filter, delegate);
     String wkt = geometryToWkt(filterValues.literal);
 
-    if (distance != 0) {
+    if (abs(distance) > ERROR_THRESHOLD) {
       return ((FilterDelegate<?>) delegate).beyond(filterValues.propertyName, wkt, distance);
     } else {
       return ((FilterDelegate<?>) delegate).nearestNeighbor(filterValues.propertyName, wkt);
--- a/catalog/plugin/catalog-plugin-pointofcontactupdate/src/main/java/org/codice/ddf/catalog/plugin/metacard/PointOfContactUpdatePlugin.java
+++ b/catalog/plugin/catalog-plugin-pointofcontactupdate/src/main/java/org/codice/ddf/catalog/plugin/metacard/PointOfContactUpdatePlugin.java
@@ -25,7 +25,6 @@ import ddf.catalog.operation.UpdateRequest;
 import ddf.catalog.plugin.PreAuthorizationPlugin;
 import ddf.catalog.plugin.StopProcessingException;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -95,16 +94,16 @@ public class PointOfContactUpdatePlugin implements PreAuthorizationPlugin {
 
   private Metacard getPreviousMetacardWithId(String id, Map<String, Metacard> previousMetacards) {
     Metacard previous;
-    try {
-      previous =
-          previousMetacards
-              .entrySet()
-              .stream()
-              .map(e -> e.getValue())
-              .filter((x) -> x.getId().equals(id))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
+    previous =
+        previousMetacards
+            .entrySet()
+            .stream()
+            .map(Map.Entry::getValue)
+            .filter(x -> x.getId().equals(id))
+            .findFirst()
+            .orElse(null);
+
+    if (previous == null) {
       LOGGER.debug("Cannot locate metacard {} for update.", id);
       return null;
     }
--- a/catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
+++ b/catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
@@ -30,8 +30,8 @@ import ddf.security.permission.Permissions;
 import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Set;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -65,24 +65,24 @@ public class PointOfContactPolicyPlugin implements PolicyPlugin {
             .getPreviousStateMetacards();
 
     Metacard previous;
-    try {
-      previous =
-          previousStateMetacards
-              .stream()
-              .filter((x) -> x.getId().equals(newMetacard.getId()))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
-      LOGGER.debug("Cannot locate metacard {} for update.", newMetacard.getId());
-      return new PolicyResponseImpl();
-    }
+    previous =
+        previousStateMetacards
+            .stream()
+            .filter(x -> x.getId().equals(newMetacard.getId()))
+            .findFirst()
+            .orElse(null);
 
     return pointOfContactChanged(newMetacard, previous)
         ? new PolicyResponseImpl(null, PERMISSION_MAP)
         : new PolicyResponseImpl();
   }
 
-  private boolean pointOfContactChanged(Metacard newMetacard, Metacard previousMetacard) {
+  private boolean pointOfContactChanged(Metacard newMetacard, @Nullable Metacard previousMetacard) {
+    if (previousMetacard == null) {
+      LOGGER.debug("Cannot locate metacard {} for update.", newMetacard.getId());
+      return false;
+    }
+
     Attribute newPointOfContact = newMetacard.getAttribute(Metacard.POINT_OF_CONTACT);
     Attribute oldPointOfContact = previousMetacard.getAttribute(Metacard.POINT_OF_CONTACT);
 
--- a/catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
+++ b/catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
@@ -31,7 +31,6 @@ import java.io.Serializable;
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -88,15 +87,14 @@ public class ResourceUriPolicy implements PolicyPlugin {
         ((OperationTransaction) properties.get(OPERATION_TRANSACTION_KEY))
             .getPreviousStateMetacards();
 
-    Metacard previous;
-    try {
-      previous =
-          previousStateMetacards
-              .stream()
-              .filter((x) -> x.getId().equals(input.getId()))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
+    Metacard previous =
+        previousStateMetacards
+            .stream()
+            .filter((x) -> x.getId().equals(input.getId()))
+            .findFirst()
+            .orElse(null);
+
+    if (previous == null) {
       LOGGER.debug(
           "Cannot locate metacard {} for update. Applying permissions to the item", input.getId());
       return policyResponse;
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
@@ -524,6 +524,7 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
     return getFactory(extraData).and(newChildren);
   }
 
+  @Override
   public Object visit(Or filter, Object extraData) {
     List<Filter> children = filter.getChildren();
     List<Filter> newChildren = new ArrayList<>();
@@ -577,6 +578,7 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
     return new Distance(distance, linearUnit).getAs(LinearUnit.METER);
   }
 
+  @SuppressWarnings("squid:S2177" /* Doesn't override parent method */)
   protected Expression visit(Expression expression, Object extraData) {
     if (expression == null) {
       return null;
--- a/catalog/spatial/geocoding/spatial-geocoding-admin-module/src/main/java/org/codice/ddf/spatial/admin/module/service/Geocoding.java
+++ b/catalog/spatial/geocoding/spatial-geocoding-admin-module/src/main/java/org/codice/ddf/spatial/admin/module/service/Geocoding.java
@@ -15,8 +15,11 @@ package org.codice.ddf.spatial.admin.module.service;
 
 import java.lang.management.ManagementFactory;
 import javax.management.InstanceAlreadyExistsException;
+import javax.management.InstanceNotFoundException;
+import javax.management.MBeanRegistrationException;
 import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
+import javax.management.NotCompliantMBeanException;
 import javax.management.ObjectName;
 import org.apache.commons.io.FilenameUtils;
 import org.codice.ddf.spatial.geocoding.GeoEntryExtractionException;
@@ -45,29 +48,32 @@ public class Geocoding implements GeocodingMBean {
   }
 
   private void registerMbean() {
-    ObjectName objectName = null;
-    MBeanServer mBeanServer = null;
     try {
-      objectName = new ObjectName(Geocoding.class.getName() + ":service=geocoding");
-      mBeanServer = ManagementFactory.getPlatformMBeanServer();
-    } catch (MalformedObjectNameException e) {
+      ObjectName objectName = new ObjectName(Geocoding.class.getName() + ":service=geocoding");
+      MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
+      tryRegisterMBean(mBeanServer, objectName);
+    } catch (MalformedObjectNameException
+        | InstanceAlreadyExistsException
+        | MBeanRegistrationException
+        | InstanceNotFoundException
+        | NotCompliantMBeanException e) {
       LOGGER.info("Unable to create Geocoding Configuration MBean.", e);
     }
-    if (mBeanServer != null) {
-      try {
-        try {
-          mBeanServer.registerMBean(this, objectName);
-          LOGGER.debug(
-              "Registered Geocoding Configuration MBean under object name: {}", objectName);
-        } catch (InstanceAlreadyExistsException e) {
-          // Try to remove and re-register
-          mBeanServer.unregisterMBean(objectName);
-          mBeanServer.registerMBean(this, objectName);
-          LOGGER.debug("Re-registered Geocoding Configuration MBean");
-        }
-      } catch (Exception e) {
-        LOGGER.info("Could not register MBean [{}].", objectName);
-      }
+  }
+
+  private void tryRegisterMBean(MBeanServer mBeanServer, ObjectName objectName)
+      throws InstanceAlreadyExistsException, MBeanRegistrationException, InstanceNotFoundException,
+          NotCompliantMBeanException {
+    try {
+      mBeanServer.registerMBean(this, objectName);
+      LOGGER.debug("Registered Geocoding Configuration MBean under object name: {}", objectName);
+    } catch (InstanceAlreadyExistsException e) {
+      // Try to remove and re-register
+      mBeanServer.unregisterMBean(objectName);
+      mBeanServer.registerMBean(this, objectName);
+      LOGGER.debug("Re-registered Geocoding Configuration MBean");
+    } catch (MBeanRegistrationException | NotCompliantMBeanException e) {
+      LOGGER.info("Could not register MBean [{}].", objectName);
     }
   }
 
@@ -101,13 +107,7 @@ public class Geocoding implements GeocodingMBean {
 
   private boolean updateIndex(String resource, boolean createIndex) {
 
-    final ProgressCallback progressCallback =
-        new ProgressCallback() {
-          @Override
-          public void updateProgress(final int progress) {
-            setProgress(progress);
-          }
-        };
+    final ProgressCallback progressCallback = this::setProgress;
 
     LOGGER.trace("Updating GeoNames Index...");
 
--- a/catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/converter/type/ServiceBindingTypeConverter.java
+++ b/catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/converter/type/ServiceBindingTypeConverter.java
@@ -53,6 +53,7 @@ public class ServiceBindingTypeConverter
    *     Optional
    * @return Optional ServiceBindingType created from the values in the map
    */
+  @Override
   public Optional<ServiceBindingType> convert(Map<String, Object> map) {
     Optional<ServiceBindingType> optionalBinding = Optional.empty();
     if (MapUtils.isEmpty(map)) {
@@ -83,7 +84,7 @@ public class ServiceBindingTypeConverter
           (List<Map<String, Object>>) map.get(SPECIFICATION_LINK_KEY)) {
         optionalSpec = slConverter.convert(specMap);
 
-        if (optionalBinding.isPresent()) {
+        if (optionalSpec.isPresent()) {
           if (!optionalBinding.isPresent()) {
             optionalBinding = Optional.of(mapToSchemaElement.getObjectFactory().get());
           }
--- a/catalog/spatial/wfs/1.0.0/spatial-wfs-v1_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v1_0_0/catalog/source/WfsResponseExceptionMapper.java
+++ b/catalog/spatial/wfs/1.0.0/spatial-wfs-v1_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v1_0_0/catalog/source/WfsResponseExceptionMapper.java
@@ -46,7 +46,7 @@ public class WfsResponseExceptionMapper implements ResponseExceptionMapper<WfsEx
           is.reset();
           msg = IOUtils.toString(is);
         } catch (IOException e) {
-          wfsEx = new WfsException("Error reading Response" + (msg != null ? ": " + msg : ""), e);
+          wfsEx = new WfsException("Error reading Response", e);
         }
         if (msg != null) {
           try {
--- a/catalog/spatial/wfs/2.0.0/spatial-wfs-v2_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v2_0_0/catalog/source/WfsFilterDelegate.java
+++ b/catalog/spatial/wfs/2.0.0/spatial-wfs-v2_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v2_0_0/catalog/source/WfsFilterDelegate.java
@@ -117,6 +117,8 @@ public class WfsFilterDelegate extends SimpleFilterDelegate<FilterType> {
 
   private static final String PROPERTY_NOT_QUERYABLE = "'%s' is not a queryable property.";
 
+  private static final String PROPERTY_NOT_QUERYABLE_ARG = "{} is not a queryable property";
+
   // Regex to match coords in WKT
   private static final Pattern COORD_PATTERN =
       Pattern.compile("-?\\.?\\d+(\\.?\\d+)?\\s-?\\.?\\d+(\\.?\\d+)?");
@@ -985,7 +987,7 @@ public class WfsFilterDelegate extends SimpleFilterDelegate<FilterType> {
                 createPropertyIsFilter(attrDesc.getPropertyName(), literal, propertyIsType));
             binaryCompOpsToBeOred.add(filter);
           } else {
-            LOGGER.debug(String.format(PROPERTY_NOT_QUERYABLE, property));
+            LOGGER.debug(PROPERTY_NOT_QUERYABLE_ARG, property);
           }
         }
         if (!binaryCompOpsToBeOred.isEmpty()) {
@@ -1465,7 +1467,7 @@ public class WfsFilterDelegate extends SimpleFilterDelegate<FilterType> {
                 createSpatialOpType(spatialOpType, attrDesc.getPropertyName(), wkt, distance));
             filtersToBeOred.add(filter);
           } else {
-            LOGGER.debug(String.format(PROPERTY_NOT_QUERYABLE, property));
+            LOGGER.debug(PROPERTY_NOT_QUERYABLE_ARG, property);
           }
         }
         if (!filtersToBeOred.isEmpty()) {
@@ -1765,7 +1767,7 @@ public class WfsFilterDelegate extends SimpleFilterDelegate<FilterType> {
     for (SpatialOperatorType spatialOp : spatialOperators.getSpatialOperator()) {
       LOGGER.debug("Adding key [spatialOp Name: {}]", spatialOp.getName());
       spatialOps.put(SPATIAL_OPERATORS.valueOf(spatialOp.getName()), spatialOp);
-      LOGGER.debug("spatialOps Map: {}", spatialOps.toString());
+      LOGGER.debug("spatialOps Map: {}", spatialOps);
     }
   }
 
@@ -1785,7 +1787,7 @@ public class WfsFilterDelegate extends SimpleFilterDelegate<FilterType> {
     for (TemporalOperatorType temporalOp : temporalOperators.getTemporalOperator()) {
       LOGGER.debug("Adding key [temporalOp Name: {}]", temporalOp.getName());
       temporalOps.put(TEMPORAL_OPERATORS.valueOf(temporalOp.getName()), temporalOp);
-      LOGGER.debug("temporalOps Map: {}", temporalOps.toString());
+      LOGGER.debug("temporalOps Map: {}", temporalOps);
     }
   }
 
--- a/catalog/spatial/wfs/2.0.0/spatial-wfs-v2_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v2_0_0/catalog/source/reader/FeatureCollectionMessageBodyReaderWfs20.java
+++ b/catalog/spatial/wfs/2.0.0/spatial-wfs-v2_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v2_0_0/catalog/source/reader/FeatureCollectionMessageBodyReaderWfs20.java
@@ -174,6 +174,8 @@ public class FeatureCollectionMessageBodyReaderWfs20
         responseBuilder.type("text/xml");
         Response response = responseBuilder.build();
         throw new WebApplicationException(e1, response);
+      } catch (VirtualMachineError e) {
+        throw e;
       } catch (JAXBException | XMLStreamException e1) {
         LOGGER.debug("Error in retrieving feature collection.", e1);
       } catch (RuntimeException | Error e) {
--- a/catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventToXmlElementConverter.java
+++ b/catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventToXmlElementConverter.java
@@ -136,21 +136,32 @@ public class SaxEventToXmlElementConverter {
           /*
            * Get the first non-empty prefix that is associated with the URI (the other, non-default prefix)
            */
-          String altNS =
-              namespaceStack
-                  .stream()
-                  .filter(p -> attUri.equals(p.getUri()) && !(p.getPrefix().isEmpty()))
-                  .findFirst()
-                  .get()
-                  .getPrefix();
-          out.writeNamespace(altNS, attUri);
-          addedNamespaces.put(attUri, altNS);
-          out.writeAttribute(altNS, attUri, atts.getLocalName(i), atts.getValue(i));
+          handleNamespaceDupes(atts, addedNamespaces, i, attUri);
         }
       }
     }
 
     return this;
+  }
+
+  private void handleNamespaceDupes(
+      Attributes atts, Multimap<String, String> addedNamespaces, int i, String attUri)
+      throws XMLStreamException {
+    NamespaceMapping altNSMap =
+        namespaceStack
+            .stream()
+            .filter(p -> attUri.equals(p.getUri()) && !(p.getPrefix().isEmpty()))
+            .findFirst()
+            .orElse(null);
+
+    if (altNSMap != null) {
+      String altNS = altNSMap.getPrefix();
+      out.writeNamespace(altNS, attUri);
+      addedNamespaces.put(attUri, altNS);
+      out.writeAttribute(altNS, attUri, atts.getLocalName(i), atts.getValue(i));
+    } else {
+      LOGGER.debug("No prefix matching {}", attUri);
+    }
   }
 
   /**
--- a/distribution/sdk/artemis-sandbox/src/main/java/sdk.example.route/AmqpProducerConsumerExample.java
+++ b/distribution/sdk/artemis-sandbox/src/main/java/sdk.example.route/AmqpProducerConsumerExample.java
@@ -53,6 +53,7 @@ public class AmqpProducerConsumerExample extends RouteBuilder {
   }
 
   @Override
+  @SuppressWarnings("squid:S1854" /* Example code */)
   public void configure() throws Exception {
     onException(Throwable.class).logStackTrace(true);
 
--- a/distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
+++ b/distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
@@ -72,10 +72,13 @@ public class Query extends HttpServlet {
 
   private static final XMLUtils XML_UTILS = XMLUtils.getInstance();
 
+  @SuppressWarnings("squid:S2226" /* Lifecycle managed by blueprint */)
   private transient CatalogFramework catalogFramework;
 
+  @SuppressWarnings("squid:S2226" /* Lifecycle managed by blueprint */)
   private transient SecurityManager securityManager;
 
+  @SuppressWarnings("squid:S2226" /* Lifecycle managed by blueprint */)
   private transient FilterBuilder filterBuilder;
 
   public void setCatalogFramework(CatalogFramework catalogFramework) {
@@ -157,12 +160,16 @@ public class Query extends HttpServlet {
     sb.append("</body>");
     sb.append("</html>");
 
-    StringBuilder message = new StringBuilder();
-    message.append("\n########################################################################\n");
-    message.append(" Query result html:\n");
-    message.append(sb.toString());
-    message.append("\n########################################################################\n");
-    LOGGER.debug(message.toString());
+    if (LOGGER.isDebugEnabled()) {
+      StringBuilder message = new StringBuilder();
+      message.append(
+          "\n########################################################################\n");
+      message.append(" Query result html:\n");
+      message.append(sb.toString());
+      message.append(
+          "\n########################################################################\n");
+      LOGGER.debug(message.toString());
+    }
 
     return sb.toString();
   }
@@ -175,7 +182,7 @@ public class Query extends HttpServlet {
   private String getMetacardForId(String searchPhrase, String proxyTicket) {
 
     Filter filter = filterBuilder.attribute(Metacard.ANY_TEXT).is().like().text(searchPhrase);
-    LOGGER.info("Query filter: {}", filter.toString());
+    LOGGER.info("Query filter: {}", filter);
     String queryError = "Unable to perform query " + filter.toString() + ".";
     QueryRequest request = new QueryRequestImpl(new QueryImpl(filter), true);
     StringBuilder responseString = new StringBuilder();
@@ -193,9 +200,9 @@ public class Query extends HttpServlet {
     }
 
     try {
-      LOGGER.debug("About to query the catalog framework with query {}", filter.toString());
+      LOGGER.debug("About to query the catalog framework with query {}", filter);
       QueryResponse queryResponse = catalogFramework.query(request, null);
-      LOGGER.debug("Got query response from catalog framework for query {}", filter.toString());
+      LOGGER.debug("Got query response from catalog framework for query {}", filter);
       List<Result> results = queryResponse.getResults();
       if (results != null) {
         String message =
--- a/distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/QueryForm.java
+++ b/distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/QueryForm.java
@@ -106,12 +106,16 @@ public class QueryForm extends HttpServlet {
     sb.append("</body>");
     sb.append("</html>");
 
-    StringBuilder message = new StringBuilder();
-    message.append("\n########################################################################\n");
-    message.append(" Query form html:\n");
-    message.append(sb.toString());
-    message.append("\n########################################################################\n");
-    LOGGER.debug(message.toString());
+    if (LOGGER.isDebugEnabled()) {
+      StringBuilder message = new StringBuilder();
+      message.append(
+          "\n########################################################################\n");
+      message.append(" Query form html:\n");
+      message.append(sb.toString());
+      message.append(
+          "\n########################################################################\n");
+      LOGGER.debug(message.toString());
+    }
 
     return sb.toString();
   }
--- a/distribution/sdk/sample-cometd/java-cli-example/src/main/java/org/codice/ddf/cli/CommandExecutor.java
+++ b/distribution/sdk/sample-cometd/java-cli-example/src/main/java/org/codice/ddf/cli/CommandExecutor.java
@@ -20,6 +20,7 @@ import org.codice.ddf.cli.ui.Notify;
 
 public class CommandExecutor {
 
+  @SuppressWarnings("squid:S1181" /*Want to catch throwable*/)
   private static <T extends RunnableCommand> void execute(T cmd) {
     try {
       int exitCode = cmd.run();
@@ -31,6 +32,7 @@ public class CommandExecutor {
     }
   }
 
+  @SuppressWarnings("squid:S1181" /*Want to catch throwable*/)
   public static <T extends RunnableCommand> void executeSingleCommand(Class<T> cls, String[] args) {
     SingleCommand<T> parser = SingleCommand.singleCommand(cls);
     try {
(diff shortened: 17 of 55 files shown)


libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -145,9 +145,11 @@ private static String getMavenHome() throws OwaspDiffRunnerException {
           Arrays.stream(mavenVersionInfo.split(System.getProperty("line.separator")))
               .filter(str -> str.contains("Maven home:"))
               .findFirst()
-              .get()
-              .split("Maven home:")[1]
-              .trim();
+              .orElse(null);
```
blen-desta: ❓I believe the old code used to throw the `OwaspDiffRunnerException` when we did a `.get()` on something that was null so you can use `orElseThrwow()` instead of setting it to null.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
```diff
@@ -54,26 +54,37 @@
 public class LandingPage extends HttpServlet {
   private static final long serialVersionUID = 1L;
 
+  @SuppressWarnings("squid:S2226" /* Lifecycle managed by blueprint. */)
```
blen-desta: I suggest doing a class suppression. Just add one `SuppressWarnings` at line 53.

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
```diff
@@ -172,6 +172,9 @@ private void init() {
     }
   }
 
+  // objectName can remain null after the try block therefore objectName is not guaranteed null or
+  // not null in last ternary operation
+  @SuppressWarnings("squid:S2583")
```
blen-desta: ✏️ Add comment inside suppression parathesis.

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
```diff
@@ -330,6 +334,14 @@ public void addTrustedCertificate(
           e);
     } finally {
       IOUtils.closeQuietly(socket);
+      // close output stream
```
blen-desta: You should use try with resources instead.
Replace the try first try with:
```
try(OutputStream fos = Files.newOutputStream(trustStoreFile)){
```

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
coyotesqrl: ❗️ Same comment as earlier regarding `objectName` and `mBeanServer`: if `mBeanServer` is non-null, then `objectName` must be non-null as well.

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
coyotesqrl: ✏️ Might as well just call `closeQuietly()` on this as well.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -504,6 +504,10 @@ private Metacard readMetacard(File file) throws IngestException {
   private Metacard generateMetacard(InputStream message) throws IOException {
     try {
       if (message != null) {
+        if (!transformer.isPresent()) {
+          throw new IllegalArgumentException(
+              "Transformation Failed for transformer: " + transformerId);
+        }
```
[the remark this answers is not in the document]
bakejeyner: If it is null it throws:
```IllegalArgumentException("Data file is null.")```

But if it is not present it throws:
```IllegalArgumentException("Transformation Failed for transformer: " + transformerId)```

Do you think it is worth combining the two exceptions if you lose that little bit of information?

File: platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java [hunk 6]
```diff
@@ -644,8 +656,9 @@ private boolean isPemParsable(String type, String fileName) {
     // check mime types
     if (PKCS7_TYPE.equals(type) || CERT_TYPE.equals(type) || PEM_TYPE.equals(type)) {
       return true;
-      // check file extensions
-    } else if (StringUtils.endsWithIgnoreCase(fileName, ".crt")
+    }
+    // check file extensions
+    if (StringUtils.endsWithIgnoreCase(fileName, ".crt")
```
coyotesqrl: ✏️ Can simplify the `if(cond1 || cond2...) then return true` to `return (cond1 || cond2...)`

catalog/plugin/catalog-plugin-pointofcontactupdate/src/main/java/org/codice/ddf/catalog/plugin/metacard/PointOfContactUpdatePlugin.java
```diff
@@ -95,19 +94,14 @@ private boolean isResourceMetacard(Metacard metacard) {
...
-      return null;
-    }
+    previous =
+        previousMetacards
+            .entrySet()
+            .stream()
+            .map(e -> e.getValue())
+            .filter((x) -> x.getId().equals(id))
+            .findFirst()
+            .orElse(null);
```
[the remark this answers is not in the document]
bakejeyner: Good catch

platform/security/common/src/main/java/ddf/security/ws/policy/impl/FilePolicyLoader.java
```diff
@@ -41,7 +41,7 @@
    */
   public FilePolicyLoader(BundleContext context, String policyLocation) throws IOException {
```
coyotesqrl: ❓ I believe the only thing that touches this code is its unit test.

@stustison can this file (and its test) be deleted?

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/StaticStsProperties.java
```diff
@@ -183,6 +183,7 @@ public void configureProperties() throws STSException {
     WSSConfig.init();
   }
 
+  @SuppressWarnings("squid:S2177" /* getResourceManager in parent class is private */)
```
coyotesqrl: ✏️ It would be much preferable to simply rename this private method so it does not look like an incorrect override.

platform/security/sts/security-sts-x509validator/src/main/java/org/codice/ddf/security/validator/x509/X509PathTokenValidator.java
```diff
@@ -144,7 +144,8 @@ public boolean canHandleToken(ReceivedToken validateTarget, String realm) {
             || X509TokenValidator.X509_V3_TYPE.equals(
                 ((BinarySecurityTokenType) token).getValueType()))) {
       return true;
-    } else if (token instanceof Element
+    }
+    if (token instanceof Element
```
coyotesqrl: ✏️ Can simplify more by simply returning the conditions.

catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
```diff
@@ -65,24 +65,23 @@ public PolicyResponse processPreUpdate(Metacard newMetacard, Map<String, Seriali
...
-    } catch (NoSuchElementException e) {
-      LOGGER.debug("Cannot locate metacard {} for update.", newMetacard.getId());
-      return new PolicyResponseImpl();
-    }
+    previous =
+        previousStateMetacards
+            .stream()
+            .filter((x) -> x.getId().equals(newMetacard.getId()))
+            .findFirst()
+            .orElse(null);
```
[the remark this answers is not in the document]
bakejeyner: I talked about this one with @rzwiefel.

He suggested moving null checking logic into the pointOfContactChanged method below, so I'm going to put the Logger in that method's null check.

catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/converter/type/ServiceBindingTypeConverter.java
```diff
@@ -83,12 +84,12 @@ public ServiceBindingTypeConverter() {
...
         optionalSpec = slConverter.convert(specMap);
 
-        if (optionalBinding.isPresent()) {
+        if (optionalSpec.isPresent()) {
           if (!optionalBinding.isPresent()) {
             optionalBinding = Optional.of(mapToSchemaElement.getObjectFactory().get());
           }
 
-          optionalBinding.get().getSpecificationLink().add(optionalSpec.get());
+          optionalBinding.get().getSpecificationLink().add(optionalSpec.orElse(null));
```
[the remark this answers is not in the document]
bakejeyner: SonarQube gives an error when you call the get() method without checking isPresent().

In this case it looks like the orElse() method is outside of the scope where the optionalSpec.isPresent() is checked, so it cannot be converted back into get().

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
bakejeyner: Very good idea. Makes the code look a lot cleaner too.

catalog/spatial/registry/registry-schema-bindings/src/main/java/org/codice/ddf/registry/schemabindings/converter/type/ServiceBindingTypeConverter.java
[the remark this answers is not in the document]
blen-desta: So if you have 
```
if (optionalSpec.isPresent()) {
          if (!optionalBinding.isPresent()) {
            optionalBinding = Optional.of(mapToSchemaElement.getObjectFactory().get());
          }

          optionalBinding.get().getSpecificationLink().add(optionalSpec.get());
        }
```
You are verifying that `optionalSpec` is present and within the if block you are calling `optionalSpec.get()`.

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
bakejeyner: It looks like a lot of work goes into getting that `trustStoreFile`. Should I try to move all of that logic before the try?

Or should I make a nested try with resources where the output stream is being initialized?

Or should I keep it as is?

blen-desta: I see what you mean. You can use `IOUtils.closeQuietly(fos);` instead

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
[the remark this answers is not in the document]
bakejeyner: I think it's worth it, I combined the statements.

catalog/plugin/catalog-plugin-pointofcontactupdate/src/main/java/org/codice/ddf/catalog/plugin/metacard/PointOfContactUpdatePlugin.java
```diff
@@ -95,19 +94,14 @@ private boolean isResourceMetacard(Metacard metacard) {
...
-    try {
-      previous =
-          previousMetacards
-              .entrySet()
-              .stream()
-              .map(e -> e.getValue())
-              .filter((x) -> x.getId().equals(id))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
```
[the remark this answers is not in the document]
bakejeyner: The catch was catching the `NoSuchElementException` thrown by the `get()` method.
Sonar wants to check `isPresent()` instead, so there is no need for the exception anymore.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -89,14 +88,14 @@ public PolicyResponse processPreUpdate(Metacard input, Map<String, Serializable>
...
 
     Metacard previous;
-    try {
-      previous =
-          previousStateMetacards
-              .stream()
-              .filter((x) -> x.getId().equals(input.getId()))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
```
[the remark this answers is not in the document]
bakejeyner: Don't need it since we are getting rid of `get()` without a present check. See ☝️

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
```diff
@@ -577,6 +577,9 @@ private double getDistanceInMeters(double distance, String units) {
     return new Distance(distance, linearUnit).getAs(LinearUnit.METER);
   }
 
+  @SuppressWarnings(
+      "squid:S2177" /* This method doesn't and isn't supposed to override it's parent;
+                    it s a copy of a parent function that wasn't inherited */)
   protected Expression visit(Expression expression, Object extraData) {
```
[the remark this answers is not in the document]
bakejeyner: It is missing its `@Override` annotation, I put it there now.

However SonarQube still wrongly thinks the method signature `protected Expression visit(Expression expression, Object extraData)` is trying to overwrite a parent method, when in reality that method is copying a parent method that was not passed down.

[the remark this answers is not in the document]
coyotesqrl: Gotcha. That's...unfortunate and it would be best not to have a method in this class with the same signature as a package-private method in its parent. However...we're probably stuck with it, so the suppression is probably going to have to stay.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -54,8 +54,8 @@
   private static Invoker invoker = new DefaultInvoker();
 
   public OwaspDiffRunner(Runtime runTime, Invoker invoker) {
-    this.runTime = runTime;
-    this.invoker = invoker;
+    setRuntime(runTime);
```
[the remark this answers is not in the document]
bakejeyner: I messaged @tbatie about this  and he recommended: "I think you can create a static setter and getter for both the runtime and invoker".

I'll go ahead and suppress the squid though. I don't think it matters much and it looks a lot cleaner.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -179,4 +181,20 @@ private static String getLocalRepo() throws OwaspDiffRunnerException {
           OwaspDiffRunnerException.UNABLE_TO_RETRIEVE_LOCAL_MAVEN_REPO, e);
     }
   }
+
+  public static Runtime getRuntime() {
```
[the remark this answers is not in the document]
bakejeyner: They were originally for setting the `runTime` and `invoker` class members from the constructor.

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/StaticStsProperties.java
bakejeyner: What would be a good rename? I'm used to getters and setters having a specific naming format.

catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
```diff
@@ -65,24 +65,23 @@ public PolicyResponse processPreUpdate(Metacard newMetacard, Map<String, Seriali
...
 
     Metacard previous;
-    try {
-      previous =
-          previousStateMetacards
-              .stream()
-              .filter((x) -> x.getId().equals(newMetacard.getId()))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
```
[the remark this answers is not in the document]
bakejeyner: Don't need it since we are getting rid of get() without a present check. See ☝️

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -53,9 +54,10 @@
 
   private static Invoker invoker = new DefaultInvoker();
 
+  @SuppressWarnings("squid:S3010" /*  */)
```
blen-desta: Suppression comment?

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
blen-desta: I think you misunderstood my comment. What you had was fine but instead of doing the `orElse(null)` and returning null you should use `.orElseThrow` and throw the `OwaspDiffRunnerException`.

catalog/core/catalog-core-impl/filter-proxy/pom.xml
```diff
@@ -53,6 +53,12 @@
             <groupId>joda-time</groupId>
             <artifactId>joda-time</artifactId>
         </dependency>
+        <dependency>
```
bdeining: might consider moving this to dependency management under the catalog pom.xml since it is used in various places.

catalog/spatial/wfs/1.0.0/spatial-wfs-v1_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v1_0_0/catalog/source/WfsResponseExceptionMapper.java
```diff
@@ -46,7 +46,7 @@ public WfsException fromResponse(Response response) {
           is.reset();
           msg = IOUtils.toString(is);
         } catch (IOException e) {
-          wfsEx = new WfsException("Error reading Response" + (msg != null ? ": " + msg : ""), e);
```
bdeining: What was the finding here?  It might be nice to preserve the message in this exception

bakejeyner: `msg` will always be null, so it will never print out the additional info

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/ExportCommand.java
```diff
@@ -94,15 +94,7 @@
...
-  private static final SimpleDateFormat ISO_8601_DATE_FORMAT;
-
-  static {
-    ISO_8601_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH-mm-ss.SSS'Z'");
-    ISO_8601_DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("UTC"));
-  }
-
-  private static final Supplier<String> FILE_NAMER =
-      () -> "export-" + ISO_8601_DATE_FORMAT.format(Date.from(Instant.now())) + ".zip";
+  private static final Supplier<String> FILE_NAMER;
```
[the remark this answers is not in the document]
bakejeyner: It looks like it is being used in writing a file:
`FILE_NAMER = () -> "export-" + ISO_8601_DATE_FORMAT.format(Date.from(Instant.now())) + ".zip"`
So I believe it is named properly.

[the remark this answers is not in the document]
rzwiefel: Actually, since the Default formatting for an Instant is ISO 8601 -- we can get rid of the simple date format all together and just use 
```java
FILE_NAMER = () -> "export-" + Instant.now().toString();
```

(or sans the toString if you wish to juse have it implicitly called.)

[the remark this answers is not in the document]
rzwiefel: We could also stick it in a String format if so desired.
```java
() -> String.format("export-%s.zip", Instant.now())
```

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
```diff
@@ -503,11 +503,11 @@ private Metacard readMetacard(File file) throws IngestException {
...
     try {
-      if (message != null) {
-        return transformer.get().transform(message);
-      } else {
-        throw new IllegalArgumentException("Data file is null.");
+      if (message == null || !transformer.isPresent()) {
+        throw new IllegalArgumentException(
+            "Transformation Failed for transformer: " + transformerId);
       }
+      return transformer.get().transform(message);
```
rzwiefel: We could also do the following to take advantage of the Optional's provided methods.
```java
Validate.notNull(message); //or a standard (if (message == null) {throw exception})

return transformer
    .map(InputTransformer::transform)
    .orElseThrow(() -> new IllegalArgumentException("Could not get transformer for " + transformerId));
```

platform/security/common/src/main/java/ddf/security/ws/policy/impl/FilePolicyLoader.java
stustison: No, this code is used in downstream projects.

catalog/core/catalog-core-commands/src/main/java/org/codice/ddf/commands/catalog/IngestCommand.java
bakejeyner: Good idea. The whole block would look like:

    if (message == null) {
      throw new IllegalArgumentException("Data file is null.");
    }

    try {
      return transformer
              .map(metacard -> metacard.transform(message))
              .orElseThrow(() -> new IllegalArgumentException("Could not get transformer for " + transformerId));
    } catch (Exception e) {
      throw new IllegalArgumentException("Transformation failed for transformer: " + transformerId, e);
    }

But for some reason error catching on the `metacard.transform(message)` doesn't work. Maybe because its in a return statement.

I think it would be better to avoid the headache of that error catch and just stick with what we had.

rzwiefel: Oh right. Poor support for exceptions within streaming blocks in java. It's because the `Optional::map` takes a Function<T,R> (which does not allow checked exceptions since its not a part of the signature). Oh well ¯\\\_(ツ)\_/¯

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/StaticStsProperties.java
coyotesqrl: As it's private, it doesn't matter too much. Something like `getResourceManagerExtension()` would work okay.

catalog/core/catalog-core-commons/src/main/java/ddf/util/XPathHelper.java
```diff
@@ -228,9 +228,10 @@ public Object evaluate(String xpathExpressionKey, QName returnType)
   public synchronized Object evaluate(
       String xpathExpressionKey, QName returnType, NamespaceContext nsContext)
       throws XPathExpressionException {
-    XPathCache.getXPath().setNamespaceContext(nsContext);
+    XPathCache xpc = new XPathCache();
```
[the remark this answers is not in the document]
bakejeyner: Great catch!

catalog/core/catalog-core-impl/pom.xml
```diff
@@ -22,6 +22,12 @@
     <artifactId>catalog-core-impl</artifactId>
     <name>DDF :: Catalog :: Core :: Impl</name>
     <packaging>pom</packaging>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
```
coyotesqrl: ❓ You're including the dependency here, in a `pom` artifact, when you're using it in the `filter-proxy` sub-module. But `filter-proxy` is a bundle and you'll either need to embed this or get it picked up from the OSGi container. The only other place we're currently using `commons-math3` (at version 2.2) is in `metrics-reporting` where it is embedded. Furthermore, I see no place where we wrap the `commons-math3` module.

So you'll compile and unit tests will pass, but at runtime you should be seeing a failure to install for a large number of bundles.

I would recommend wrapping `commons-math3` and installing it pretty early in the installation process. Then remove it as an embedded bundle from `filter-proxy` so it can import the package from the container.

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -53,6 +54,7 @@
 
   private static Invoker invoker = new DefaultInvoker();
 
+  @SuppressWarnings("squid:S3010" /* Constructor only here for associated unit test */)
   public OwaspDiffRunner(Runtime runTime, Invoker invoker) {
```
coyotesqrl: ✏️ Can probably reduce the visibility of this ctor to be package-private.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
```diff
@@ -577,6 +578,9 @@ private double getDistanceInMeters(double distance, String units) {
     return new Distance(distance, linearUnit).getAs(LinearUnit.METER);
   }
 
+  @SuppressWarnings(
+      "squid:S2177" /* This method doesn't and isn't supposed to override it's parent;
+                    it s a copy of a parent function that wasn't inherited */)
```
bdeining: nit :  `This method doesn't and isn't supposed to override its parent; it's a copy of a parent function that wasn't inherited`

distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
```diff
@@ -162,7 +165,9 @@ private String createPage(HttpServletRequest request) {
     message.append(" Query result html:\n");
     message.append(sb.toString());
     message.append("\n########################################################################\n");
-    LOGGER.debug(message.toString());
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug(message.toString());
```
bdeining: If the `message` is only being used for debugging, it seems wasteful to always build it up and print it conditionally.  Any reason not to build the message itself conditionally as well (move 163-167 inside the LOGGER.isDebugEnabled() check.

distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/QueryForm.java
```diff
@@ -111,7 +111,9 @@ private String createPage(HttpServletRequest request) {
     message.append(" Query form html:\n");
     message.append(sb.toString());
     message.append("\n########################################################################\n");
-    LOGGER.debug(message.toString());
+    if (LOGGER.isDebugEnabled()) {
```
bdeining: Same comment about building `message`

libs/httpproxy/proxy-camel-servlet/src/main/java/org/codice/proxy/http/HttpProxyCamelHttpTransportServlet.java
```diff
@@ -280,7 +281,9 @@ protected HttpConsumer resolve(HttpServletRequest request) {
     HttpConsumer answer = consumers.get(endpointName);
 
     if (answer == null) {
-      log.debug("Consumer Keys: {}", Arrays.toString(consumers.keySet().toArray()));
+      if (LOGGER.isDebugEnabled()) {
+        log.debug("Consumer Keys: {}", Arrays.toString(consumers.keySet().toArray()));
```
bdeining: Is `Arrays.toString(consumers.keySet().toArray()));` necessary?  I think `consumers.keySet().toString()` would print a comma separated list of all the keys.

catalog/spatial/wfs/2.0.0/spatial-wfs-v2_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v2_0_0/catalog/source/reader/FeatureCollectionMessageBodyReaderWfs20.java
```diff
@@ -176,7 +176,7 @@ public Wfs20FeatureCollection readFrom(
         throw new WebApplicationException(e1, response);
       } catch (JAXBException | XMLStreamException e1) {
         LOGGER.debug("Error in retrieving feature collection.", e1);
-      } catch (RuntimeException | Error e) {
+      } catch (Exception e) {
```
Lambeaux: ❓  Curious what the original reason was to catch `Error` and why it's no longer necessary. This change definitely demands justification.

paouelle: ❓ Are we sure that this is equivalent? I know it is not a good idea to catch Errors and RuntimeExceptions but there was most likely a reason for that. Further more we are now throwing back Exception which is not declared. I believe we could keep the original code as long as we had a 'catch (VirtualMachineError)' and throw it back right away without attempting to log

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventToXmlElementConverter.java
```diff
@@ -136,16 +137,18 @@ private SaxEventToXmlElementConverter startConstructingElement(
...
                   .stream()
                   .filter(p -> attUri.equals(p.getUri()) && !(p.getPrefix().isEmpty()))
-                  .findFirst()
-                  .get()
-                  .getPrefix();
-          out.writeNamespace(altNS, attUri);
-          addedNamespaces.put(attUri, altNS);
-          out.writeAttribute(altNS, attUri, atts.getLocalName(i), atts.getValue(i));
+                  .map(NamespaceMapping::getPrefix)
+                  .findFirst();
```
paouelle: ✏️  Add '.orElse(null)' to avoid calling isPresent and get on it. Or again, simply add `.ifPresent(altNs -> { your code }).`

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventToXmlElementConverter.java
```diff
@@ -136,16 +137,18 @@ private SaxEventToXmlElementConverter startConstructingElement(
...
+
+          if (altNs.isPresent()) {
+            out.writeNamespace(altNs.get(), attUri);
+            addedNamespaces.put(attUri, altNs.get());
+            out.writeAttribute(altNs.get(), attUri, atts.getLocalName(i), atts.getValue(i));
+          }
```
Lambeaux: ❓ Is the case where no alternate namespace is found safe to ignore? Should we, at minimum, log to debug?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -862,7 +862,7 @@ private void trySleep(long millis) {
     try {
       Thread.sleep(millis);
     } catch (InterruptedException e) {
-      // Do nothing
+      Thread.currentThread().interrupt();
```
paouelle: ❗️ Although we need to figure out how to propagate interruptions, this will now cause a problem as it this method is called from within `tryUpdate` which is recursive and propagating here means the next attempts will also fail for the same reasons. We need to figure out how to retain the fact that we were interrupted and propagate the interruptions outside the whole thing.

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/SearchController.java
```diff
@@ -176,15 +177,15 @@ public void executeQuery(
...
-              failSource(request, session, search, sourceId, e);
             } catch (TimeoutException e) {
               LOGGER.info("Query timed out for source {}", sourceId, e);
               failSource(request, session, search, sourceId, e);
             } catch (ExecutionException e) {
               LOGGER.info("Query failed for source {}", sourceId, e.getCause());
               failSource(request, session, search, sourceId, e);
+            } catch (InterruptedException e) {
+              LOGGER.info("Query interrupted for source {}", sourceId, e);
+              failSource(request, session, search, sourceId, e);
```
Lambeaux: ❓  Should the thread's interrupted flag be set here as well?

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/AbstractEventController.java
```diff
@@ -256,6 +256,7 @@ public void run() {
                 try {
                   TimeUnit.SECONDS.sleep(1);
                 } catch (InterruptedException e) {
+                  Thread.currentThread().interrupt();
```
paouelle: ❗️ Same issue as before, the interruption must be propagated outside the loop using a finally clause

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/SearchController.java
```diff
@@ -122,6 +122,7 @@ public synchronized void publishResults(
    * @param request {@link SearchRequest} containing a query for 1 or more sources
    * @param session Cometd {@link ServerSession}
    */
+  @SuppressWarnings("squid:S2142" /* Thread interrupt handled */)
```
paouelle: ❗️ That doesn't cut it. Here we need to compute the end time and time remaining in each loop and make sure we wrap the whole thing within a loop where we can propagate the interruption in a finally clause outside of that while loop

distribution/sdk/artemis-sandbox/src/main/java/sdk.example.route/AmqpProducerConsumerExample.java
```diff
@@ -57,7 +57,6 @@ public void configure() throws Exception {
     onException(Throwable.class).logStackTrace(true);
 
     createCamelContext();
-    CamelContext camelContext = getContext();
```
paouelle: ❗️ I don't think you simply want to remove that line. This is an example and although the example is not using it, The idea is that real code would. We need to suppress here I think.

distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
```diff
@@ -175,7 +180,9 @@ private String createPage(HttpServletRequest request) {
   private String getMetacardForId(String searchPhrase, String proxyTicket) {
 
     Filter filter = filterBuilder.attribute(Metacard.ANY_TEXT).is().like().text(searchPhrase);
-    LOGGER.info("Query filter: {}", filter.toString());
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.info("Query filter: {}", filter.toString());
```
paouelle: ✏️  Remove the call to `.toString()` and avoid the `if`

distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
```diff
@@ -193,9 +200,13 @@ private String getMetacardForId(String searchPhrase, String proxyTicket) {
     }
 
     try {
-      LOGGER.debug("About to query the catalog framework with query {}", filter.toString());
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("About to query the catalog framework with query {}", filter.toString());
```
paouelle: ✏️  Remove the call to `.toString()` and avoid the `if`

distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
```diff
@@ -193,9 +200,13 @@ private String getMetacardForId(String searchPhrase, String proxyTicket) {
...
+      }
       QueryResponse queryResponse = catalogFramework.query(request, null);
-      LOGGER.debug("Got query response from catalog framework for query {}", filter.toString());
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Got query response from catalog framework for query {}", filter.toString());
```
paouelle: ✏️  Remove the call to `.toString()` and avoid the `if`

distribution/sdk/sample-cometd/java-cli-example/src/main/java/org/codice/ddf/cli/CommandExecutor.java
```diff
@@ -20,6 +20,7 @@
 
 public class CommandExecutor {
 
+  @SuppressWarnings("squid:S1181" /*Want to catch throwable*/)
```
paouelle: ℹ️  See above reasoning for why it is ok

distribution/sdk/sample-cometd/java-cli-example/src/main/java/org/codice/ddf/cli/CommandExecutor.java
```diff
@@ -31,6 +32,7 @@
     }
   }
 
+  @SuppressWarnings("squid:S1181" /*Want to catch throwable*/)
```
paouelle: ℹ️  See above reason for why it is ok

distribution/sdk/sample-cometd/java-cli-example/src/main/java/org/codice/ddf/cli/CommandExecutor.java
```diff
@@ -45,6 +47,7 @@
...
```
paouelle: ℹ️  See above reason for why it is ok

libs/httpproxy/proxy-camel-servlet/src/main/java/org/codice/proxy/http/HttpProxyCamelHttpTransportServlet.java
paouelle: You can even omit the call to `.toString()`

libs/klv/src/main/java/org/codice/ddf/libs/klv/data/numerical/KlvNumericalDataElement.java
```diff
@@ -44,6 +44,8 @@ public KlvNumericalDataElement(final byte[] key, final String name) {
    * @param errorIndicatorValue value that indicates an encoded error
    * @throws IllegalArgumentException if any arguments are null
    */
+  @SuppressWarnings(
+      "squid:S2789" /* Optional errorIndicatorValue can be null, check for null is ok */)
```
paouelle: ❗️ By design, Optional should never be null. This is bad practice as they are used to avoid dealing with null

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
```diff
@@ -53,6 +54,7 @@
 
   private static Invoker invoker = new DefaultInvoker();
 
+  @SuppressWarnings("squid:S3010" /* Constructor only here for associated unit test */)
```
paouelle: ✏️  Then make it package private and annotate it with `@VisibleForTesting` instead of doing a suppression

platform/metrics/platform-metrics-collector/src/main/java/ddf/metrics/collector/rrd4j/RrdJmxCollector.java
```diff
@@ -353,7 +353,7 @@ private boolean isMbeanAccessible() {
           LOGGER.trace("MBean [{}] not found, sleeping...", mbeanName);
           Thread.sleep(1000);
         } catch (InterruptedException ie) {
-          // Ignore this
+          Thread.currentThread().interrupt();
```
paouelle: ❗️ Flag the interruption and propagate outside of while loop in a finally clause.

platform/notifications/core/platform-core-notificationcommands/src/main/java/org/codice/ddf/notifications/commands/SendCommand.java
```diff
@@ -113,7 +113,7 @@ public Object execute() throws Exception {
         try {
           Thread.sleep(waitTime * 1000L);
         } catch (InterruptedException e) {
-
+          Thread.currentThread().interrupt();
```
paouelle: ❗️ Need to compute end time outside of a while loop for this instance of i and compute the remaining time inside. Interruption should loop back for this particular loop. Outside the overall for(), propagate the interruption in a finally clause

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationServiceImpl.java
```diff
@@ -763,7 +763,9 @@ private void waitForFeatures(Set<Feature> features) throws InterruptedException
                   .map(Object::toString)
                   .collect(Collectors.joining(", ")));
         }
-        this.wait(TimeUnit.SECONDS.toMillis(1));
+        synchronized (this) {
+          this.wait(TimeUnit.SECONDS.toMillis(1));
```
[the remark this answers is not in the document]
Lambeaux: ❗️  Why was synchronization added? Is it even necessary? I would revise to a simple sleep: 
```
TimeUnit.MILLISECONDS.sleep(1);
```

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
```diff
@@ -194,10 +194,7 @@ private void registerMbean() {
         }
       } catch (Exception e) {
         // objectName is not always non-null because new ObjectName(...) can throw an exception
-        LOGGER.info(
-            "Could not register MBean [{}].",
-            objectName != null ? objectName.toString() : KeystoreEditor.class.getName(),
-            e);
+        LOGGER.info("Could not register MBean [{}].", objectName.toString(), e);
```
paouelle: ✏️  Remove the call to `.toString()` and you can even remove the comment

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
```diff
@@ -330,6 +328,7 @@ public void addTrustedCertificate(
           e);
     } finally {
       IOUtils.closeQuietly(socket);
+      IOUtils.closeQuietly(fos);
```
paouelle: ❓ Do we really want to close this output stream quietly? It would mean the file was not properly created.

platform/security/common/src/main/java/ddf/security/ws/policy/impl/FilePolicyLoader.java
```diff
@@ -76,4 +76,8 @@ protected Document loadFromFile(URL policyFileURL) throws IOException {
   public Document getPolicy() {
     return FilePolicyLoader.policyDoc;
   }
+
+  public static void setPolicy(Document document) {
```
paouelle: ❓ Shouldn't `getPolicy()` be static?

platform/solr/solr-factory/src/main/java/org/codice/solr/factory/impl/SolrHttpRequestRetryHandler.java
```diff
@@ -43,14 +43,17 @@ public boolean retryRequest(IOException e, int retryCount, HttpContext httpConte
...
+        LOGGER.debug(
+            "Solr Client: Connection failed, waiting {} before retrying.",
+            DurationFormatUtils.formatDurationWords(waitTime, true, true));
+      }
       synchronized (this) {
         wait(waitTime);
       }
     } catch (InterruptedException ie) {
       LOGGER.debug("Exception while waiting.", ie);
+      Thread.currentThread().interrupt();
```
paouelle: ❗️ Dangerous without checking how this method is called as if it is called within a loop, it will continuously get interrupted. We have to figure out how to properly handle the interruptions and it might turn out that simply propagating it is ok.

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
Lambeaux: ❗️  See above comment with same link. There's a better way to handle this finding along with the embedded `try` blocks. 
https://github.com/codice/ddf/blob/master/platform/security/certificate/security-certificate-generator/src/main/java/org/codice/ddf/security/certificate/generator/CertificateGenerator.java#L63-L83

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationServiceImpl.java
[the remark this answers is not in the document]
rzwiefel: @Lambeaux Synchronization is always necessary where a `wait` is used. 
https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#wait()

catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
```diff
@@ -65,24 +65,24 @@ public PolicyResponse processPreUpdate(Metacard newMetacard, Map<String, Seriali
...
         ? new PolicyResponseImpl(null, PERMISSION_MAP)
         : new PolicyResponseImpl();
   }
 
-  private boolean pointOfContactChanged(Metacard newMetacard, Metacard previousMetacard) {
+  private boolean pointOfContactChanged(Metacard newMetacard, @Nullable Metacard previousMetacard) {
+    if (previousMetacard == null) {
+      LOGGER.debug("Cannot locate metacard {} for update.", newMetacard.getId());
+      return false;
+    }
```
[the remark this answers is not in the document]
rzwiefel: @Lambeaux Why are the less things with `@Nullable` better? (Meaning, I would argue its not **always** better).

Also we often want to reduce complexity of top level functions -- so moving the encompassing logic here isn't a bad idea. If you simply don't like the ternary, then we could just move the policy creation in here instead. This way we reduce the complexity of the top level function and keep all relevant conditions grouped together here. Added explicitness.

```java
  ...
    return updatedPointOfContactPolicy(newMetacard, previous);
  }



  private PolicyResponseImpl getUpdatedPointOfContactPolicy(
      Metacard newMetacard, @Nullable Metacard previousMetacard) {
    if (previousMetacard == null) {
      LOGGER.debug("Cannot locate metacard {} for update.", newMetacard.getId());
      return new PolicyResponseImpl();
    }

    Attribute newPointOfContact = newMetacard.getAttribute(Metacard.POINT_OF_CONTACT);
    Attribute oldPointOfContact = previousMetacard.getAttribute(Metacard.POINT_OF_CONTACT);

    // Don't apply if nothing has changed (Both are still null)
    if (newPointOfContact == null && oldPointOfContact == null) {
      return new PolicyResponseImpl(); 
    }

    // Apply policy if only one or the other is null
    if (newPointOfContact == null ^ oldPointOfContact == null) {
      return new PolicyResponseImpl(null, PERMISSION_MAP);
    }

    // Don't apply if nothing has changed (both values still equal)
    if (newPointOfContact.getValue().equals(oldPointOfContact.getValue())) {
      return new PolicyResponseImpl(); 
    }  
      
    // Something has changed, apply the policy
    return new PolicyResponseImpl(null, PERMISSION_MAP);
  }
```

distribution/sdk/ddf-sso-query-widget/src/main/java/org/codice/ddf/ui/Query.java
rzwiefel: Do we want to keep these kind of ascii line breaks? (At least with the thinking that "if everyone did this ..." then our logs would look even worse). 

 We could probably just leave it as 
```java
LOGGER.debug("Query result html:\n {}", sb);
```
without being much worse for wear. (and we wouldn't need to wrap in `isDebugEnabled` since it will short circuit before evaluating arguments. )

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -703,7 +704,7 @@ public Object visit(Beyond filter, Object delegate) {
     ExpressionValues filterValues = getExpressions(filter, delegate);
     String wkt = geometryToWkt(filterValues.literal);
 
-    if (distance != 0) {
+    if (Precision.compareTo(distance, 0, .000001) == 0) {
```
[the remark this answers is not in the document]
bakejeyner: Good points.

1. I'll add tests for .001 and .0000001 edge cases
2. That fits much better. I'll make that change and then we can decide later whether or not we want to include the `Precision` library at all.
3. Another way to do the fp comparison without the library would be:
`if (abs(distance) < .000001)`

catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
```diff
@@ -65,24 +65,24 @@ public PolicyResponse processPreUpdate(Metacard newMetacard, Map<String, Seriali
...
     Metacard previous;
-    try {
-      previous =
-          previousStateMetacards
-              .stream()
-              .filter((x) -> x.getId().equals(newMetacard.getId()))
-              .findFirst()
-              .get();
-    } catch (NoSuchElementException e) {
-      LOGGER.debug("Cannot locate metacard {} for update.", newMetacard.getId());
```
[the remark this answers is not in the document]
bakejeyner: It isn't removed, it got moved into the `pointOfContactChanged` method below.

catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
```diff
@@ -65,24 +65,24 @@ public PolicyResponse processPreUpdate(Metacard newMetacard, Map<String, Seriali
...
+            .orElse(null);
 
     return pointOfContactChanged(newMetacard, previous)
         ? new PolicyResponseImpl(null, PERMISSION_MAP)
         : new PolicyResponseImpl();
   }
 
-  private boolean pointOfContactChanged(Metacard newMetacard, Metacard previousMetacard) {
+  private boolean pointOfContactChanged(Metacard newMetacard, @Nullable Metacard previousMetacard) {
+    if (previousMetacard == null) {
```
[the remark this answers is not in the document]
bakejeyner: I thought I did:
`@Nullable Metacard previousMetacard`

libs/owasp-diff-runner/src/main/java/ddf/lib/OwaspDiffRunner.java
bakejeyner: SonarQube has an issue with setting static variables in the constructor. After I made my changes I believe @coyotesqrl (Didn't mean to throw you under the bus) mentioned that it was used for testing and I should just suppress instead.

platform/security/certificate/security-certificate-keystoreeditor/src/main/java/org/codice/ddf/security/certificate/keystore/editor/KeystoreEditor.java
bakejeyner: It will catch any IO errors above:

`fos = Files.newOutputStream(trustStoreFile);
 trustStore.store(fos, keyStorePassword.toCharArray());
 } catch (IOException | GeneralSecurityException e) {
 LOGGER.info(
          "Unable to add certificate(s) to trust store from URL: {}",
          (decodedUrl != null) ? decodedUrl : url,
          e);
}`

The close quietly is just to avoid a null fos causing issues.

platform/security/common/src/main/java/ddf/security/ws/policy/impl/FilePolicyLoader.java
bakejeyner: It is implementing the interface method.

catalog/transformer/catalog-transformer-streaming-lib/src/main/java/org/codice/ddf/transformer/xml/streaming/lib/SaxEventToXmlElementConverter.java
bakejeyner: I threw a log in there:
`LOGGER.debug("No prefix matching {}", attUri);`

catalog/ui/search-ui/search-endpoint/src/main/java/org/codice/ddf/ui/searchui/query/controller/SearchController.java
bakejeyner: SonarQube doesn't have a problem with it so it isn't really part of my ticket. If I have time to look into it I'll definitely do that.

libs/klv/src/main/java/org/codice/ddf/libs/klv/data/numerical/KlvNumericalDataElement.java
bakejeyner: When I looked into it all of the classes calling that constructor pass in an optional and never null, so I decided to take the useless check out.

catalog/spatial/wfs/2.0.0/spatial-wfs-v2_0_0-source/src/main/java/org/codice/ddf/spatial/ogc/wfs/v2_0_0/catalog/source/WfsFilterDelegate.java
```diff
@@ -1465,7 +1467,9 @@ private FilterType buildGeospatialFilterType(
                 createSpatialOpType(spatialOpType, attrDesc.getPropertyName(), wkt, distance));
             filtersToBeOred.add(filter);
           } else {
-            LOGGER.debug(String.format(PROPERTY_NOT_QUERYABLE, property));
+            if (LOGGER.isDebugEnabled()) {
+              LOGGER.debug(String.format(PROPERTY_NOT_QUERYABLE, property));
+            }
```
blen-desta: Can't you do the same thing as above?

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
```diff
@@ -93,12 +94,16 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(LandingPage.class);
 
+  @SuppressWarnings("squid:S2226" /* Lifecycle managed by blueprint. */)
```
blen-desta: If you have a suppression on top of the class it will apply to all the methods.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
```diff
@@ -93,12 +94,16 @@
...
   private String logo;
 
+  @SuppressWarnings("squid:S2226" /* Lifecycle managed by blueprint. */)
```
blen-desta: It looks like this is a different problem with a different squid `S1068`. Same for the other two suppressions below.

catalog/spatial/geocoding/spatial-geocoding-admin-module/src/main/java/org/codice/ddf/spatial/admin/module/service/Geocoding.java
```diff
@@ -45,29 +47,23 @@ public Geocoding() {
...
-        LOGGER.info("Could not register MBean [{}].", objectName);
-      }
+  }
+
+  private void tryRegisterMBean(MBeanServer mBeanServer, ObjectName objectName) {
+    try {
+      mBeanServer.registerMBean(this, objectName);
+      LOGGER.debug("Registered Geocoding Configuration MBean under object name: {}", objectName);
+    } catch (InstanceAlreadyExistsException e) {
+      LOGGER.error("Re-registered Geocoding Configuration MBean");
```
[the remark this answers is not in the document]
bakejeyner: I got this code change from @Lambeaux. He said this is the new way we handle this common code problem.

I'll ask him if I should implement the retry or if we shouldn't do that in the new version of this code.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
bakejeyner: It does, but If I didn't suppress the class I would have had to put a suppression on every class variable, and that is super messy. And after taking off that suppression the only bugs were the `Squid:S2226`s. I can always change it to suppress every instance variable if you think that's better.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
bakejeyner: After taking out the class suppression they all have the `Squid:S2226` bug. `Squid:S1068` looks like it is just a warning and out of the scope of my ticket. Another reason I don't think I should mess with the `Squid:S1068` variable not used warnings is that they may be used in blueprint.

catalog/spatial/geocoding/spatial-geocoding-admin-module/src/main/java/org/codice/ddf/spatial/admin/module/service/Geocoding.java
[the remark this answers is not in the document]
Lambeaux: @blen-desta is right -- we can keep the `tryRegisterMBean` convention while still preserving the functionality. Need the re-register code. Also, match log levels.

catalog/security/catalog-security-pointofcontact-policyplugin/src/main/java/org/codice/ddf/catalog/security/policy/metacard/PointOfContactPolicyPlugin.java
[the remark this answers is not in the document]
Lambeaux: Reducing complexity of top level functions is fine. I'm just saying in a function whose purpose is to compare, bailing early on null violates the principle of least astonishment. Unless there's javadoc that says nulls are not valid for comparison.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
blen-desta: Sorry I wasn't clear on what I meant. You have the same suppression on the class and on three variables. You don't need the extra suppressions you have on lines 54, 97 and 100 because of this class suppression.

platform/landing-page/src/main/java/org/codice/ddf/landing/LandingPage.java
blen-desta: What I meant was the issue with `logoToUse`, `background`, and `foreground` is not `Servlets should not have mutable instance fields`, which has a squid of `S2226`. It's "`Unused "private" fields should be removed`, which has a squid of `S1068`. I'm not saying you should address that issue just that the suppression you have isn't doing anything.

catalog/security/catalog-security-resourceplugin/src/main/java/org/codice/ddf/catalog/security/ResourceUriPolicy.java
```diff
@@ -89,14 +88,14 @@ public PolicyResponse processPreUpdate(Metacard input, Map<String, Serializable>
...
+    previous =
+        previousStateMetacards
+            .stream()
+            .filter((x) -> x.getId().equals(input.getId()))
+            .findFirst()
+            .orElse(null);
+
+    if (previous == null) {
```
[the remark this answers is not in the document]
coyotesqrl: Were you going to make this change?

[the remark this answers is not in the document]
bakejeyner: This is what the new code would look like:
`Optional<Metacard> previous =
        previousStateMetacards
            .stream()
            .filter((x) -> x.getId().equals(input.getId()))
            .findFirst();

    if (!previous.isPresent()) {
      LOGGER.debug(
          "Cannot locate metacard {} for update. Applying permissions to the item", input.getId());
      return policyResponse;
    }

    return requiresPermission(input.getResourceURI(), previous.get().getResourceURI())
        ? policyResponse
        : new PolicyResponseImpl();`

I think it does look a little better, but almost everywhere else in the code we are doing `orElse(null)` so in interest of consistency I would like to keep it the way it is.

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -703,7 +705,7 @@ public Object visit(Beyond filter, Object delegate) {
     ExpressionValues filterValues = getExpressions(filter, delegate);
     String wkt = geometryToWkt(filterValues.literal);
 
-    if (distance != 0) {
+    if (abs(distance) > .000001) {
```
shaundmorris: ❓ how did you come up with this number for the distance. Is it worth a comment in the code? Looks like a magic number.

bakejeyner: It kind of is a magic number. I'll go through and add comments describing what it is.

bakejeyner:
build now

bakejeyner:
build now

bakejeyner:
build now

bakejeyner:
build now

bakejeyner:
build now

PR #2787, opened 2017-12-21
DDF-3509 Fixed page size bug with Intrigue queries

rzwiefel:
#### What does this PR do?
Intrigue queries will now enumerate all results for a query up to the page size parameter instead of being limited to the internal catalog maxPageSize

#### Who is reviewing it? 
@pklinef 
anyone

See ticket for reproduction steps
This was simply a missed case when we added the internal catalog maxPageSize to limit the results of an individual query and so the fix was using the ResultIterable to page through all needed results for a query
https://codice.atlassian.net/browse/DDF-3509

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -24,16 +24,24 @@ import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableMap;
 import ddf.action.ActionRegistry;
 import ddf.catalog.CatalogFramework;
+import ddf.catalog.data.Result;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.filter.FilterAdapter;
 import ddf.catalog.filter.FilterBuilder;
 import ddf.catalog.operation.QueryRequest;
 import ddf.catalog.operation.QueryResponse;
+import ddf.catalog.operation.impl.QueryResponseImpl;
 import ddf.catalog.source.SourceUnavailableException;
 import ddf.catalog.source.UnsupportedQueryException;
+import ddf.catalog.util.impl.QueryFunction;
+import ddf.catalog.util.impl.ResultIterable;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.boon.json.JsonParserFactory;
 import org.boon.json.JsonSerializerFactory;
 import org.boon.json.ObjectMapper;
@@ -150,9 +158,39 @@ public class QueryApplication implements SparkApplication {
   private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
       throws UnsupportedQueryException, SourceUnavailableException, FederationException {
     QueryRequest request = cqlRequest.createQueryRequest(catalogFramework.getId(), filterBuilder);
-
     Stopwatch stopwatch = Stopwatch.createStarted();
-    QueryResponse response = catalogFramework.query(request);
+
+    List<QueryResponse> responses = Collections.synchronizedList(new ArrayList<>());
+    QueryFunction queryFunction =
+        (queryRequest) -> {
+          QueryResponse queryResponse = catalogFramework.query(queryRequest);
+          responses.add(queryResponse);
+          return queryResponse;
+        };
+
+    List<Result> results =
+        ResultIterable.resultIterable(queryFunction, request, cqlRequest.getCount())
+            .stream()
+            .collect(Collectors.toList());
+
+    QueryResponse response =
+        new QueryResponseImpl(
+            request,
+            results,
+            true,
+            responses
+                .stream()
+                .filter(Objects::nonNull)
+                .map(QueryResponse::getHits)
+                .findFirst()
+                .orElse(-1l),
+            responses
+                .stream()
+                .filter(Objects::nonNull)
+                .map(QueryResponse::getProperties)
+                .findFirst()
+                .orElse(Collections.emptyMap()));
+
     stopwatch.stop();
 
     return new CqlQueryResponse(


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -189,4 +209,34 @@ public void setFeatureService(FeatureService featureService) {
   public void setEndpointUtil(EndpointUtil util) {
     this.util = util;
   }
+
```
ahoffer: Not sure that we need to union properties and take first non-zero hit count.
The reduction overwrites previous properties with new values if any. Otherwise, the reduction aggregates properties. Unless there is a good reason to aggregate properties, I think we'd be fine with taking the first, last or middle properties and using them.
Why use the first non-zero value of hits? The hit count should be the same for all query result objects. Only the actual results should change.

figliold: Is there an answer for Aaron's question?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -150,9 +156,32 @@ public void init() {
...
       throws UnsupportedQueryException, SourceUnavailableException, FederationException {
     QueryRequest request = cqlRequest.createQueryRequest(catalogFramework.getId(), filterBuilder);
-
     Stopwatch stopwatch = Stopwatch.createStarted();
-    QueryResponse response = catalogFramework.query(request);
+
+    List<QueryResponse> responses = Collections.synchronizedList(new ArrayList<>());
+    List<Result> results =
+        ResultIterable.resultIterable(
+                (queryRequest) -> {
```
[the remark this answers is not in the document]
rzwiefel: no u

brendan-hofmann:
Why not just use https://github.com/codice/ddf/pull/2133 ?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -150,9 +156,32 @@ public void init() {
...
+                request,
+                cqlRequest.getCount())
+            .stream()
+            .collect(Collectors.toList());
+    QueryResponse response =
+        new QueryResponseImpl(
+            request,
+            results,
+            true,
+            responses.stream().map(QueryResponse::getHits).findFirst().orElse(-1l),
```
ahoffer: All the responses should have the same value for hits. Take the first one.
Additionally, `CatalogFramework.query()` always returns at least one response. If those conditions hold, Line 178 collapses to:

`responses.get(0).getHits()`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -150,9 +156,32 @@ public void init() {
...
+            responses
+                .stream()
+                .map(QueryResponse::getProperties)
+                .findFirst()
+                .orElse(Collections.emptyMap()));
```
ahoffer: The responses collection will always have at least one responses (even if there were 0 hits).
Also, the Response will always have properties. If those conditions hold, line 179-183 collapses to:

`responses.get(0).getProperties()`

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
rzwiefel: i prefer not to ever assume size and fetch by an index, this will always be safe no matter how implementations may change

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
rzwiefel: see https://github.com/codice/ddf/pull/2787#discussion_r170010295

rzwiefel:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -150,9 +157,38 @@ public void init() {
...
       throws UnsupportedQueryException, SourceUnavailableException, FederationException {
     QueryRequest request = cqlRequest.createQueryRequest(catalogFramework.getId(), filterBuilder);
-
     Stopwatch stopwatch = Stopwatch.createStarted();
-    QueryResponse response = catalogFramework.query(request);
+
+    List<QueryResponse> responses = Collections.synchronizedList(new ArrayList<>());
+    List<Result> results =
+        ResultIterable.resultIterable(
+                (queryRequest) -> {
```
coyotesqrl: ✏️ Might read a little cleaner if you extract the `QueryFunction` here as a local variable.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -150,9 +157,38 @@ public void init() {
   private CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
       throws UnsupportedQueryException, SourceUnavailableException, FederationException {
     QueryRequest request = cqlRequest.createQueryRequest(catalogFramework.getId(), filterBuilder);
-
     Stopwatch stopwatch = Stopwatch.createStarted();
-    QueryResponse response = catalogFramework.query(request);
+
+    List<QueryResponse> responses = Collections.synchronizedList(new ArrayList<>());
```
coyotesqrl: ❓ Do you really want a synch list here and not something slightly more performant, like a `ConcurrentQueue`, maybe?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -150,9 +157,38 @@ public void init() {
...
+            request,
+            results,
+            true,
+            responses
+                .stream()
+                .filter(Objects::nonNull)
+                .map(QueryResponse::getHits)
+                .findFirst()
+                .orElse(-1l),
+            responses
```
coyotesqrl: ✏️ This double-streaming feels very inefficient but nothing's jumping out at me to really improve it.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
rzwiefel: SynchronizedList should be fine here as the overall number of additions to the list should be relatively small. We can always address in the future if this becomes a point of particular slowdown.

rzwiefel:
build now

rzwiefel:
build now

PR #2805, opened 2018-01-03
DDF-3524 Implement Artemis plugin to add Subject to messages

harrison-tarr:
#### What does this PR do?
Adds a plugin to Artemis that adds a SAML subject to every message
#### Who is reviewing it? 
@roelens8 
@ryeats 
@derekwilhelm 
@jrnorth 
@jaymcnallie 
@stustison
[DDF-3524](https://codice.atlassian.net/browse/DDF-3524)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/SecurityServerPlugin.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apache.activemq.artemis.core.transaction.Transaction;
+
+public interface SecurityServerPlugin {
+
+  void beforeSend(
+      ServerSession session,
+      Transaction tx,
+      Message message,
+      boolean direct,
+      boolean noAutoCreateQueue);
```
jrnorth: ✏️ I find boolean parameters easier to understand when they're framed like "do this thing (yes/no)" vs "don't do this thing (yes/no)". So instead of `noAutoCreateQueue` it would be `autoCreateQueue`.

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Override
+  public void beforeSend(
+      ServerSession session,
+      Transaction tx,
+      Message message,
+      boolean direct,
+      boolean noAutoCreateQueue) {
+
+    LOGGER.trace(
```
jrnorth: Use parameterized logging. Also, do we really want to log a password?

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        String.format("Username: %s, Password: %s", session.getUsername(), session.getPassword()));
+    if (!SUBJECT_CACHE.containsKey(session.getUsername())) {
+      UPAuthenticationToken usernamePasswordToken =
+          new UPAuthenticationToken(session.getUsername(), session.getPassword());
+      try {
+        Subject subject = securityManager.getSubject(usernamePasswordToken);
+        SUBJECT_CACHE.put(session.getUsername(), subject);
+      } catch (SecurityServiceException e) {
+        LOGGER.warn("A bad security thing happened", e);
```
jrnorth: ❓Should we `return` at this point? If this exception is thrown, then `SUBJECT_CACHE.get(session.getUsername())` below will return null and an NPE will be thrown.

harrison-tarr:
There is discussion about making which topics/queues the plugin apply to configurable. Is that a capability we want in this PR or can/should it be an improvement?

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+
+    message.putStringProperty(
+        "subject",
+        DOM2Writer.nodeToString(
```
jrnorth: ✏️ Could save this in a local variable so we don't have to create it again below.

pom.xml
```diff
@@ -170,17 +170,20 @@
...
-        <qpid-jms.version>0.23.0</qpid-jms.version>
+        <proton.version>0.20.0</proton.version>
+        <qpid-jms.version>0.24.0</qpid-jms.version>
         <!--END versions that should be kept in sync between Camel and Artemis features-->
         <jboss-logging.version>3.3.0.Final</jboss-logging.version>
         <jgroups.version>3.6.13.Final</jgroups.version>
         <commons-beanutils.version>1.9.3</commons-beanutils.version>
         <netty.version>4.1.9.Final</netty.version>
+        <commons-beanutils.version>1.9.2</commons-beanutils.version>
+        <netty.version>4.1.14.Final</netty.version>
```
jrnorth: Update the existing `netty.version` instead.

bdeining: And beanutils

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/SecurityServerPlugin.java
harrison-tarr: This is an interface from Artemis, I think we should leave it as `noAutoCreateQueue` for consistency.

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
harrison-tarr: No, this was left over from when I was developing and debugging, thanks for the catch!

pom.xml
harrison-tarr: oh thanks, I think I resolved the conflicts wrong.

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
harrison-tarr: Fixed

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
harrison-tarr: Fixed

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
harrison-tarr: Fixed

pom.xml
harrison-tarr: Fixed

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/SecurityServerPlugin.java
```diff
@@ -0,0 +1,28 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.broker.security.api;
+
+import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.core.server.ServerSession;
+import org.apache.activemq.artemis.core.transaction.Transaction;
+
+public interface SecurityServerPlugin {
```
ryeats: This interface bothers me a bit, I understand why we need it but the name and method aren't very descriptive to what its intended use is.

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -54,7 +57,7 @@
         <network-check-URL-list>http://www.apache.org</network-check-URL-list> -->
 
         <network-check-period>10000</network-check-period>
-        <network-check-timeout>255</network-check-timeout>
+        <network-check-timeout>1000</network-check-timeout>
```
ryeats: revert this back to 255 it is the max that linux supports add a comment to that effect here also for me.

broker/broker-security-impl/src/test/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Test
+  public void testGoodSubjectAmqp() {
+    Message message = new AMQPMessage(1);
+    securityServerPlugin.beforeSend(mockServerSession, null, message, false, false);
+    assertThat(message.getStringProperty("subject"), is("Hello World"));
+  }
+
+  @Test
+  public void testBadSubject() throws SecurityServiceException {
```
roelens8: Even though this is a negative test, I still think it would be nice to assert that the message does indeed not contain a value for the subject after calling `beforeSend`.

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/SecurityServerPlugin.java
harrison-tarr: I left the method "beforeSend" because that's what it is in the Artemis plugin that i'm being wrapped by. I'm certainly open to renaming either the class, method or both though. Do you have any ideas?

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.wss4j.common.util.DOM2Writer;
+import org.codice.ddf.broker.security.api.SecurityServerPlugin;
+import org.codice.ddf.security.handler.api.UPAuthenticationToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SecurityServerPluginImpl implements SecurityServerPlugin {
+  private static final Logger LOGGER = LoggerFactory.getLogger(SecurityServerPluginImpl.class);
+
+  private static final Map<String, Subject> SUBJECT_CACHE = new HashMap<>();
```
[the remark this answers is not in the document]
harrison-tarr: I think we had discussed some kind of expiring cache. That's probably a question for @stustison . In the meantime, I'll use a concurrent hashmap instead

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
```diff
@@ -0,0 +1,87 @@
...
+        LOGGER.warn("Unable to retrieve the Subject from the token", e);
+        return;
+      }
+    }
+    String subjectAsString = getSubjectAsString(session);
+    message.putStringProperty("subject", subjectAsString);
+    if (message instanceof AMQPMessage) {
+      Map applicationPropertiesMap =
+          ((AMQPMessage) message).getProtonMessage().getApplicationProperties().getValue();
+      applicationPropertiesMap.put("subject", subjectAsString);
```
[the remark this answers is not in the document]
harrison-tarr: Not necessarily. I do intend to make the set of queues and topics that this plugin intercepts configurable, but even then, sloppy configuration or coding could leave the subject on a message that someone could see. How would i go about making a new unsigned assertion?

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/SecurityServerPlugin.java
harrison-tarr: Fixed

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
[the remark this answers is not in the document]
harrison-tarr: changed to a concurrent hashmap

distribution/ddf-common/src/main/resources/etc/artemis.xml
harrison-tarr: fixed

broker/broker-security-impl/src/test/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImplTest.java
harrison-tarr: fixed

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/SecurityServerPlugin.java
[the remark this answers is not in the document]
harrison-tarr: fixed

distribution/ddf-common/src/main/resources/etc/artemis.xml
```diff
@@ -27,6 +27,9 @@
                 </anycast>
             </address>
         </addresses>
+        <broker-plugins>
+            <broker-plugin class-name="org.codice.ddf.broker.security.BrokerAuthenticationPlugin" />
```
ryeats: do these same changes need to be made to artemis-backup.xml?

harrison-tarr: fixed

harrison-tarr:
I added functionality to allow filtering which addresses the plugin is applied to.

platform/security/security-util/src/test/java/org/codice/ddf/security/util/SAMLUtilsTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @Before
+  public void setUp() throws SecurityServiceException {}
```
jrnorth: Remove if not needed.

platform/security/security-util/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>ddf.platform.security</groupId>
+        <artifactId>security</artifactId>
+        <version>2.12.0-SNAPSHOT</version>
+    </parent>
+    <groupId>ddf.security</groupId>
+    <artifactId>security-util</artifactId>
```
ryeats: Security-util is not descriptive of what this bundle is for this could end up looking like another ddf-security-common mess.
We have so many utility classes in ddf-security-common that are related SAML it breaks my heart to add another bundle and not pull them into this one or put this into there since they have a common SAML dependency(would prefer the former but it is probably a lot of work)

platform/security/security-util/src/test/java/org/codice/ddf/security/util/SAMLUtilsTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    SecurityToken securityToken = SAML_UTILS.getSecurityTokenFromSAMLAssertion(SAML_ASSERTION);
+    assertThat(securityToken, CoreMatchers.is(instanceOf(SecurityToken.class)));
+  }
+
+  @Test
+  public void testSAMLAssertionParseFail() throws SecurityServiceException {
+
+    SecurityToken securityToken = SAML_UTILS.getSecurityTokenFromSAMLAssertion(BAD_SAML_ASSERTION);
+    assertThat(securityToken, CoreMatchers.is(instanceOf(SecurityToken.class)));
```
jrnorth: Can you assert something else about the security token? This test is performing the same assertion as the above test but this test name suggests that something should be different.

pom.xml
```diff
@@ -170,17 +170,18 @@
         <xalan.bundle.version>2.7.2_3</xalan.bundle.version>
         <!--END versions that MUST exactly match the versions used by the CXF and Camel features-->
         <!-- Artemis dependencies -->
-        <artemis.version>2.1.0</artemis.version>
+        <artemis.version>2.4.0</artemis.version>
```
roelens8: Don't forget to update the other dependencies that go along with upgrading artemis. I believe proton, qpid-jms, and netty all need to be updated to later versions with the upgrade of artemis to 2.4.

harrison-tarr: Are all the versions in that `features.xml` the only versions I need to make sure to bump? Or is there somewhere else I should be looking as well?

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/BrokerMessageInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+   * @param noAutoCreateQueue
+   */
+  void handleMessage(
+      ServerSession session,
+      Transaction tx,
+      Message message,
+      boolean direct,
+      boolean noAutoCreateQueue);
+
+  void setConfiguredAddresses(Set<String> addresses);
```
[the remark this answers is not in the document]
harrison-tarr: I can envision having multiple handlers, each that apply to different message type or address, in which case it would be easiest and simplest for each handler to know which messages it should handle; on the other hand, if you're envisioning a framework where there is a top level handler and it handles all messages and then farms them out to the handlers that need to handle the message based on the address?

platform/security/security-util/pom.xml
harrison-tarr: I started to move this stuff to `ddf-security-common`, but it looks like that's a `jar` vs my `bundle`. I think it's better to use bundles than jars, so I think I should avoid refactoring this code to `ddf-security-common`.

broker/broker-security-impl/src/test/java/org/codice/ddf/broker/security/impl/SubjectInjectorPluginTest.java
```diff
@@ -0,0 +1,126 @@
...
+    securityServerPlugin = new SubjectInjectorPluginTester();
+    securityServerPlugin.setSecurityManager(mockSecurityManager);
+    securityServerPlugin.setConfiguredAddresses(
+        new HashSet<>(Collections.singletonList("test.address")));
+  }
+
+  @Test
+  public void testGoodSubjectNonAmqp() throws SecurityServiceException {
+
+    securityServerPlugin.clearCache();
```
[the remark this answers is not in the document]
harrison-tarr: fixed

platform/security/security-util/src/test/java/org/codice/ddf/security/util/SAMLUtilsTest.java
harrison-tarr: fixed

platform/security/security-util/src/test/java/org/codice/ddf/security/util/SAMLUtilsTest.java
harrison-tarr: fixed

pom.xml
harrison-tarr: fixed

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SecurityServerPluginImpl.java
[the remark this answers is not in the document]
harrison-tarr: Fixed

pom.xml
```diff
@@ -170,17 +170,18 @@
...
         <!--START Camel and Artemis features should use the same versions for these bundles-->
-        <proton.version>0.19.0</proton.version>
```
roelens8: According to the comment above, the proton and qpid-jms versions used here should be kept in sync with the versions used in our Camel features.

harrison-tarr: Does Camel have it's own version declared for this? Or are you saying I may need to upgrade Camel to a version that uses the same Proton version as Artemis?

roelens8: We maintain separate version properties for these dependencies that our Camel features use (in the platform-app pom). No you shouldn't need to upgrade camel itself, just these specific dependencies.

harrison-tarr: It looks like the values for the camel versions correspond to the versions declared here.

platform/security/security-util/pom.xml
stustison: This does seem like it belongs in security-common to me. That jar probably needs a little cleanup anyways, so it is something that can wait. Also to follow our naming guidelines, it would need to be like `security-saml-util` or something similar.

platform/security/security-util/src/main/java/org/codice/ddf/security/util/SAMLUtils.java
```diff
@@ -0,0 +1,115 @@
...
+  private static final SAMLUtils INSTANCE = new SAMLUtils();
+
+  private static final Pattern SAML_PREFIX = Pattern.compile("<(?<prefix>\\w+?):Assertion\\s.*");
+
+  private static final XMLUtils XML_UTILS = XMLUtils.getInstance();
+
+  private static final String EVIDENCE =
+      "<%1$s:Evidence xmlns:%1$s=\"urn:oasis:names:tc:SAML:2.0:assertion\">%2$s</%1$s:Evidence>";
+
+  public static SAMLUtils getInstance() {
```
stustison: If you're planning on this being a singleton then you should also have a private constructor.

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/BrokerMessageInterceptor.java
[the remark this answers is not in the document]
roelens8: Would removing the setConfiguredAddresses() method from the interface and instead just passing in a set of configured addresses through the constructor of the SubjectInjectorPlugin, be sufficient? This way at least the implementation detail isn't exposed through the interface.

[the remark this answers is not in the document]
harrison-tarr: @ryeats @roelens8 I fixed this by moving the set method to the impl. I decided to go with that over using a constructor argument, because it seems to me most blueprint injection uses a setter instead of a constructor argument. If that's wrong I can switch back.

platform/security/security-util/pom.xml
harrison-tarr: fixed

broker/broker-security-api/src/main/java/org/codice/ddf/broker/security/api/BrokerMessageInterceptor.java
```diff
@@ -0,0 +1,47 @@
...
+ * This class abstracts the implementation of the ActiveMQServerPlugin away from the Artemis
+ * Fragment. It is intended to be used in tandem with an implementation of the ActiveMQServerPlugin.
+ *
+ * <p><b> This code is experimental. While this interface is functional and tested, it may change or
+ * be removed in a future version of the library. </b>
+ */
+public interface BrokerMessageInterceptor {
+
+  /**
+   * This message accepts a session, transaction, message, direct boolean and autoCreateQueue
```
jrnorth: :pencil2: noAutoCreateQueue

broker/broker-security-impl/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>.6</minimum>
```
jrnorth: ✏️ If we use .60 instead of .6, any coverage error messages will have two digits of precision instead of just one.

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+  private Element getSubjectAsElement(ServerSession session) {
+    try {
+      return SUBJECT_CACHE
+          .get(session.getUsername(), () -> this.cacheAndReturnSubject(session))
+          .getPrincipals()
+          .oneByType(SecurityAssertion.class)
+          .getSecurityToken()
+          .getToken();
+    } catch (ExecutionException e) {
+      LOGGER.warn("Could not get Subject from token", e);
```
jrnorth: :pencil2: I think you could log `e.getCause()` instead since an `ExecutionException` just wraps the real exception that occurred.

Also, if your value loader function throws an unchecked exception (and I can't tell if it does) then `get()` will throw an `UncheckedExecutionException`, which you may or may not want to handle differently.

broker/broker-security-impl/src/test/java/org/codice/ddf/broker/security/impl/SubjectInjectorPluginTest.java
```diff
@@ -0,0 +1,132 @@
...
+  private Subject mockSubject;
+
+  private SecurityManager mockSecurityManager;
+
+  private ServerSession mockServerSession;
+
+  private SubjectInjectorPlugin securityServerPlugin;
+
+  @Before
+  public void setVariables() throws SecurityServiceException {
```
jrnorth: ✏️ `setup()` (per the DDF unit test standards)

broker/broker-security-impl/src/test/java/org/codice/ddf/broker/security/impl/SubjectInjectorPluginTest.java
```diff
@@ -0,0 +1,132 @@
...
+    assertThat(message.getStringProperty("subject"), is("Hello World"));
+  }
+
+  @Test
+  public void testGoodSubjectAmqp() {
+
+    Message message = new AMQPMessage(1);
+    message.setAddress("test.address");
+    securityServerPlugin.handleMessage(mockServerSession, null, message, false, false);
+    assertThat(message.getStringProperty("subject"), is("Hello World"));
```
jrnorth: :pencil2: Since the plugin adds some extra stuff to the message when it's an AMQP message, it would probably be a good idea to check for it here.

platform/security/security-util/src/test/java/org/codice/ddf/security/util/SAMLUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+  public void testSAMLAssertionParse() throws SecurityServiceException {
+
+    SecurityToken securityToken = SAML_UTILS.getSecurityTokenFromSAMLAssertion(SAML_ASSERTION);
+    assertThat(securityToken, is(instanceOf(SecurityToken.class)));
+  }
+
+  @Test
+  public void testSAMLAssertionParseFail() throws SecurityServiceException {
+
+    SecurityToken securityToken = SAML_UTILS.getSecurityTokenFromSAMLAssertion(BAD_SAML_ASSERTION);
```
jrnorth: ❓This test is identical to the previous test except for the test string. Is there any way to verify that the parser failed?

platform/security/security-util/pom.xml
```diff
@@ -22,8 +22,8 @@
         <version>2.12.0-SNAPSHOT</version>
```
roelens8: We forgot to rename this directory to security-saml-util

broker/broker-adapter/src/main/java/org/codice/ddf/broker/security/BrokerAuthenticationPlugin.java
```diff
@@ -0,0 +1,61 @@
...
+        brokerMessageInterceptor =
+            bundleContext.getService(
+                ((List<ServiceReference<BrokerMessageInterceptor>>)
+                        bundleContext.getServiceReferences(
+                            BrokerMessageInterceptor.class, "(name=subjectInjectorPlugin)"))
+                    .get(0));
+      }
+      brokerMessageInterceptor.handleMessage(session, tx, message, direct, noAutoCreateQueue);
+    } catch (InvalidSyntaxException e) {
+      LOGGER.error(
```
jrnorth: Is `error` the right log level here? Seems like `warn` or even `info` would be more appropriate.

roelens8:
LGTM, successfully hero'd. 👍

broker/broker-security-impl/src/test/java/org/codice/ddf/broker/security/impl/SubjectInjectorPluginTest.java
harrison-tarr: fixed

broker/broker-adapter/src/main/java/org/codice/ddf/broker/security/BrokerAuthenticationPlugin.java
harrison-tarr: I think it's definitely at least `warn`, because it means the message will not be processed as expected. Changed to `warn`.

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+  static Cache<String, Subject> getSubjectCache() {
+    return SUBJECT_CACHE;
+  }
+
+  @VisibleForTesting
+  Subject cacheAndReturnSubject(ServerSession session) throws SecurityServiceException {
+    UPAuthenticationToken usernamePasswordToken =
+        new UPAuthenticationToken(session.getUsername(), session.getPassword());
+    Subject subject = securityManager.getSubject(usernamePasswordToken);
+    SUBJECT_CACHE.put(session.getUsername(), subject);
```
jrnorth: Why is the cache being modified here? This is your value loader function so Guava will update the cache with the value returned here.

platform/security/security-saml-util/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Export-Package>org.codice.ddf.security.util</Export-Package>
```
ryeats: I am on the fence about this now that I think more about it anyone who imports this will have to import or embedded all your dependencies anyway it's only a minor improvement. Another downside this has is it can be an annoyance during development, any time you update this bundle during development it can break anything else using it unless they are rebuilt/deployed also. I think the correct answer may, in the end, be doing http://njbartlett.name/2014/05/26/static-linking.html instead of embedding whole jars or exporting concrete classes.

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
harrison-tarr: fixed

harrison-tarr:
build now

harrison-tarr:
I ran these tests locally and had only 1 (seemingly unrelated) failure. Is it a known issue that you can have so many itests fail?

harrison-tarr:
build now

harrison-tarr:
squashed and rebased commits

harrison-tarr:
build now

harrison-tarr:
build now

harrison-tarr:
build now

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  String getStringSubjectFromSession(ServerSession session) {
+    return SAMLUtils.getInstance().getSubjectAsStringNoSignature(getSubjectAsElement(session));
+  }
+
+  public void setSecurityManager(SecurityManager securityManager) {
+    SubjectInjectorPlugin.securityManager = securityManager;
```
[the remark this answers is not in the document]
harrison-tarr: fixed

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,120 @@
...
+  @VisibleForTesting
+  static Cache<String, Subject> getSubjectCache() {
+    return SUBJECT_CACHE;
+  }
+
+  @VisibleForTesting
+  Subject cacheAndReturnSubject(ServerSession session) throws SecurityServiceException {
+    UPAuthenticationToken usernamePasswordToken =
+        new UPAuthenticationToken(session.getUsername(), session.getPassword());
+    Subject subject = securityManager.getSubject(usernamePasswordToken);
```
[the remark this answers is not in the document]
harrison-tarr: fixed

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,120 @@
...
+          ((AMQPMessage) message).getProtonMessage().getApplicationProperties().getValue();
+      applicationPropertiesMap.put("subject", subjectAsString);
+      ((AMQPMessage) message)
+          .getProtonMessage()
+          .setApplicationProperties(new ApplicationProperties(applicationPropertiesMap));
+    }
+  }
+
+  public void setConfiguredAddresses(Set<String> addresses) {
+    configuredAddresses = new HashSet<>(addresses);
```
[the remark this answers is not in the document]
harrison-tarr: fixed

platform/security/security-saml-util/src/main/java/org/codice/ddf/security/util/SAMLUtils.java
```diff
@@ -0,0 +1,117 @@
...
+    Element result = null;
+
+    Matcher prefix = SAML_PREFIX.matcher(assertion);
+    if (prefix.find()) {
+
+      Thread thread = Thread.currentThread();
+      ClassLoader loader = thread.getContextClassLoader();
+      thread.setContextClassLoader(SAMLUtils.class.getClassLoader());
+
+      try {
```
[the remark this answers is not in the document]
harrison-tarr: fixed

harrison-tarr:
build now

harrison-tarr:
build now

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @VisibleForTesting
+  String getStringSubjectFromSession(ServerSession session) {
+    return SAMLUtils.getInstance().getSubjectAsStringNoSignature(getSubjectAsElement(session));
+  }
+
+  @SuppressWarnings("squid:s2696)")
+  public void setSecurityManager(SecurityManager securityManager) {
+    SubjectInjectorPlugin.securityManager = securityManager;
```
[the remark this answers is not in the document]
harrison-tarr: i thought i suppressed this?

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+      applicationPropertiesMap.put("subject", subjectAsString);
+      ((AMQPMessage) message)
+          .getProtonMessage()
+          .setApplicationProperties(new ApplicationProperties(applicationPropertiesMap));
+    }
+  }
+
+  @SuppressWarnings("squid:s2696)")
+  public void setConfiguredAddresses(Set<String> addresses) {
+    configuredAddresses = new HashSet<>(addresses);
```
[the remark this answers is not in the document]
harrison-tarr: i thought i suppressed this?

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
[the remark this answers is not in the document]
paouelle: remove the `)` at the end of the suppress code … for example use `@SuppressWarnings("squid:s2696")` instead of `@SuppressWarnings("squid:s2696)")`. I would also strongly suggest you put a comment as to why you suppress something

broker/broker-security-impl/src/main/java/org/codice/ddf/broker/security/impl/SubjectInjectorPlugin.java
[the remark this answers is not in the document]
paouelle: remove the `)` at the end of the suppress code … for example use `@SuppressWarnings("squid:s2696")` instead of `@SuppressWarnings("squid:s2696)")`. I would also strongly suggest you put a comment as to why you suppress something

harrison-tarr:
build now

harrison-tarr:
build now

harrison-tarr:
build now

broker/broker-security-impl/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -0,0 +1,29 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<metatype:MetaData xmlns:metatype="http://www.osgi.org/xmlns/metatype/v1.2.0">
+    <OCD name="Subject Injector Plugin Configuration"
+         id="org.codice.ddf.broker.security.impl.SubjectInjectorPlugin">
+
+        <AD
+                description="The addresses that should have the Subject injected into the messages"
```
shaundmorris: ❓ Not a blocker, but is there a plan to update docs with this configuration?

harrison-tarr: I think there should be. I could make a ticket

shaundmorris: This will get tracked on: https://codice.atlassian.net/browse/DDF-3618

PR #2831, opened 2018-01-10
DDF-3529 Fix UI session timeout redirect

troymohl:
#### What does this PR do?
Fixes the session timeout redirect in the UI.
Changes the UI to respect the timeout specified.
Changes the redirect type (HTTP 303 instead of 307). 
Fixes the UI such that non-saved workspaces don't prevent the redirect. 

#### Who is reviewing it? 
@brendan-hofmann 
@adimka 
@brjeter 

@andrewkfiedler
@bdeining

Full build with tests.
1. Install DDF
2. Update the catalog timeout to 120000 (or something reasonable for testing) in 
3. Navigate to /search/catalog
4. Create a new workspace or edit existing workspace. Wait for the time configured in step 2.
5. Verify that the redirect to logout occurs after that time.

[DDF-3529](https://codice.atlassian.net/browse/DDF-3529)

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/session-timeout.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/session-timeout.js
@@ -22,7 +22,7 @@ var invalidateUrl = '/services/internal/session/invalidate?prevurl=';
 var idleNoticeDuration = 60000;
 // Length of inactivity that will trigger user timeout (15 minutes in ms by default)
 // See STIG V-69243
-var idleTimeoutThreshold = parseInt(properties.ui.timeout) > 0 ? parseInt(properties.ui.timeout) : 900000;
+var idleTimeoutThreshold = parseInt(properties.timeout) > 0 ? parseInt(properties.timeout) : 900000;
 
 function getIdleTimeoutDate() {
     return idleTimeoutThreshold + Date.now();
@@ -53,7 +53,9 @@ var sessionTimeoutModel = new (Backbone.Model.extend({
         }
     },
     setPromptTimer: function () {
-        this.promptTimer = setTimeout(this.showPrompt.bind(this), this.get('idleTimeoutDate') - idleNoticeDuration - Date.now());
+        var timeout = this.get('idleTimeoutDate') - idleNoticeDuration - Date.now();
+        timeout = Math.max(0, timeout);
+        this.promptTimer = setTimeout(this.showPrompt.bind(this), timeout);
     },
     showPrompt: function () {
         this.set('showPrompt', true);
@@ -65,7 +67,9 @@ var sessionTimeoutModel = new (Backbone.Model.extend({
         clearTimeout(this.promptTimer);
     },
     setLogoutTimer: function () {
-        this.logoutTimer = setTimeout(this.logout.bind(this), this.get('idleTimeoutDate') - Date.now());
+        var timeout = this.get('idleTimeoutDate') - Date.now();
+        timeout = Math.max(0, timeout);
+        this.logoutTimer = setTimeout(this.logout.bind(this), timeout);
     },
     clearLogoutTimer: function () {
         clearTimeout(this.logoutTimer);
@@ -80,6 +84,9 @@ var sessionTimeoutModel = new (Backbone.Model.extend({
         $(document).off('keydown.sessionTimeout mousedown.sessionTimeout mousemove.sessionTimeout');
     },
     logout: function () {
+        if (window.onbeforeunload != null) {
+          window.onbeforeunload = null;
+        }
         window.location.replace(invalidateUrl + window.location.pathname);
     },
     renew: function () {
--- a/platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
+++ b/platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
@@ -126,7 +126,7 @@ public class SessionManagementService {
   public Response getInvalidate(@Context HttpServletRequest request) {
     StringBuffer requestURL = request.getRequestURL();
     String requestQueryString = request.getQueryString();
-    return Response.temporaryRedirect(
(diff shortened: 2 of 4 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/session-timeout.js
```diff
@@ -53,7 +53,9 @@ var sessionTimeoutModel = new (Backbone.Model.extend({
         }
     },
     setPromptTimer: function () {
-        this.promptTimer = setTimeout(this.showPrompt.bind(this), this.get('idleTimeoutDate') - idleNoticeDuration - Date.now());
+        var timeout = this.get('idleTimeoutDate') - idleNoticeDuration - Date.now();
+        timeout = Math.max(0, timeout);
+        this.promptTimer = setTimeout(this.showPrompt.bind(this), timeout);
```
brjeter: ❓ Does it make sense to do this same thing for `setLogoutTimer` function below? I'm not entirely sure what the difference is.

troymohl: I thought about it, but since this is really just making sure we don't display <0 I thought I'd keep the change minimal.

mackncheesiest: My impression was that `promptTimer` was the timer that, when it reached 0, would warn the user about impending logout and that `logoutTimer` was the timer that actually triggered the logout.

If it works with this change, though, maybe that's wrong.

platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
```diff
@@ -126,7 +126,7 @@ public Response getRenewal(@Context HttpServletRequest request) {
   public Response getInvalidate(@Context HttpServletRequest request) {
     StringBuffer requestURL = request.getRequestURL();
     String requestQueryString = request.getQueryString();
-    return Response.temporaryRedirect(
+    return Response.seeOther(
```
mackncheesiest: Are we sure that we don't want this to be something like a `permanentRedirect` or `movedPermanently`?

troymohl: According to the HTTP status codes, I think See Other is what we want. I don't think we want to send something like a 301 for this. https://en.wikipedia.org/wiki/HTTP_301

catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/session-timeout.js
troymohl: Per this it looks like I should probably change that as well. https://stackoverflow.com/questions/8430966/is-calling-settimeout-with-a-negative-delay-ok

platform/security/servlet/security-servlet-session-expiry/src/main/java/org/codice/ddf/security/servlet/expiry/SessionManagementService.java
brendan-hofmann: I concur. This is not a permanent redirect. It is a forcible redirect under certain conditions (session expiry). This is analogous to what something like a banking website might do.

mackncheesiest: Works for me. Figured I'd bring it up to help make sure that we're not changing it again in a few months.

catalog/ui/catalog-ui-search/src/main/webapp/component/singletons/session-timeout.js
```diff
@@ -22,7 +22,7 @@ var invalidateUrl = '/services/internal/session/invalidate?prevurl=';
 var idleNoticeDuration = 60000;
 // Length of inactivity that will trigger user timeout (15 minutes in ms by default)
 // See STIG V-69243
-var idleTimeoutThreshold = parseInt(properties.ui.timeout) > 0 ? parseInt(properties.ui.timeout) : 900000;
+var idleTimeoutThreshold = parseInt(properties.timeout) > 0 ? parseInt(properties.timeout) : 900000;
```
mackncheesiest: Nice catch. I verified that `timeout` is assigned to `props` and not `props.ui` in https://github.com/codice/ddf/blob/master/catalog/ui/catalog-ui-search/src/main/webapp/properties.js#L69
by looking at responses returned by queries to `/search/catalog/internal/config` and `/services/platform/config/ui`

brendan-hofmann:
When the 307 is returned, I see an unknown syntax error instead of being redirected. 
image

bdeining:
@brendan-hofmann why 307?  Doesn't this PR change it to a 303?

brendan-hofmann:
@bdeining interesting catch. Usually gitflow does the right thing, I wonder if something went wrong this time. I can do a full build of this branch and try again.

troymohl:
@brendan-hofmann I believe that this is a UI bug (where session timeout happens in-between polling in the UI) that the UI needs to handle. There are other ways this can occur like having 2 instances open and logging out of 1 tab but not the other. Although I disagree with adding this to the documentation, I went ahead and added a commit with this documentation.

bdeining:
build now

bdeining:
So we document that two minutes is lower bound for the timeout because we run into syntax error popups; but we also get this same error in other situations?  What's the criteria to document one situation and not others?

bdeining:
build now

brendan-hofmann:
@bdeining It's a `known issue`. If you do a certain thing, this will occur. We recommend doing this other thing instead. I guess the criteria to me is:
1. We know about it.
2. We know what causes it.
3. We can offer advice on avoiding/mitigating it.
@ricklarsen might have better criteria, but that's how I view it.

bdeining:
build now

PR #2840, opened 2018-01-15
DDF-3548 Added configuration to disable unknown errors dialogue

bdeining:
#### What does this PR do?
Added configuration to disable unknown errors dialogue
#### Who is reviewing it? 
@mackncheesiest @mojogitoverhere 
@andrewkfiedler
@jlcsmith
@millerw8
Build / Install / Check `Disable Unknown Error Box` in the Catalog UI Config / Navigate to Intrigue and enter a workspace / turn DDF off / the source poller should eventually return an `unknown error` and it should not show on Intrigue
[DDF-3548](https://codice.atlassian.net/browse/DDF-3548)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationSetup.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationSetup.js
@@ -37,6 +37,7 @@ require([
     'underscore',
     'backbone',
     'marionette',
+    'properties',
     'handlebars/dist/handlebars',
     'component/announcement',
     'js/Marionette.Region',
@@ -45,7 +46,7 @@ require([
     'js/ApplicationHelpers',
     'js/Autocomplete',
     'backbone.customFunctions'
-], function(_, Backbone, Marionette, hbs, announcement) {
+], function(_, Backbone, Marionette, properties, hbs, announcement) {
 
     let getShortErrorMessage = function (error) {
         var extraMessage = error instanceof Error ? error.name : String(error);
@@ -80,6 +81,10 @@ require([
         }
 
         var response = getErrorResponse(event, jqxhr, settings, throwError);
+        if (properties.disableUnknownErrorBox && response.message.substring(0,7) === "Unknown") {
+            return;
+        }
+
         var message;
 
         console.error(event, jqxhr, settings, throwError);
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationSetup.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationSetup.js
@@ -81,10 +81,6 @@ require([
         }
 
         var response = getErrorResponse(event, jqxhr, settings, throwError);
-        if (properties.disableUnknownErrorBox && response.message.substring(0,13) === "Unknown Error") {
-            return;
-        }
-
         var message;
 
         console.error(event, jqxhr, settings, throwError);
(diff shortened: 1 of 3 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationSetup.js
```diff
@@ -80,6 +81,10 @@ require([
         }
 
         var response = getErrorResponse(event, jqxhr, settings, throwError);
+        if (properties.disableUnknownErrorBox && response.message.substring(0,7) === "Unknown") {
```
millerw8: I am torn... don't love the text matching, but I am guessing you did not want put the check up on the default for getErrorResponse then have a magic "null" or some other value to know to return.  I can live with how it is.

mackncheesiest: This code might be able to use a bit of refactoring, but it would be good to do this `response.message.substring` check after the checks in [this section](https://github.com/codice/ddf/pull/2840/files#diff-161be453f112bb95da340c092e44209cR91) as well as the final response that gets displayed can be different from the one returned by `getErrorResponse`

bdeining: This should be independent of what is displayed; the check was done here to immediately return if `disableUnknownErrorBox` is true and it's an unknown error.

mojogitoverhere: I think using the more explicit `Unknown Error` string would be better here. I can imagine other types of unknowns possibly being used down the line like unknown source or something that we don't want to catch in this conditional

mackncheesiest: Yeah, but there could be a `response` that comes back from `getErrorResponse` as an `Unknown Error` and then gets caught in, say, (here)[https://github.com/codice/ddf/pull/2840/files#diff-161be453f112bb95da340c092e44209cR95] and the actual error message that _would_ have been displayed wouldn't have been an "Unknown Error" anymore. So we might be suppressing errors that would have not displayed as an `Unknown Error`.

bdeining: Nice find, moving the logic down

bdeining: changed to check for `Unknown Error`

bdeining:
build now

mojogitoverhere:
Hero build in progress

mojogitoverhere:
:tada: Hero build was successful! 🎉 
First I did a quick build and started DDF with the default admin settings. I manually verified that the `Unknown Error` box pops up and the `sources` request fails in the network tab when DDF is shutdown with Intrigue still open.
screen shot 2018-01-15 at 5 05 30 pm
Then I restarted DDF and verified the `Logged Out` box pops up.
screen shot 2018-01-15 at 5 09 04 pm

Next, I checked `Disable Unknown Error Box` in the admin UI and verified that the `Unknown Error` box does _not_ pop up when the `sources` request fails when DDF is shutdown.
screen shot 2018-01-15 at 5 11 02 pm
Finally, I restarted DDF to verify the `Logged Out` box still pops up but got a `Forbidden` box instead. After a while, the `Logged Out` box also popped up.
screen shot 2018-01-15 at 5 13 00 pm
Hopefully this all makes sense... English is hard.

mackncheesiest:
The "forbidden" likely happened because it attempted to refresh source status while DDF was starting up and received a 403 which caused the error to no longer be "unknown error"

PR #2853, opened 2018-01-17
DDF-2879 Refactor PropertiesLoader and add unit tests

bakejeyner:
#### What does this PR do?
Cleans up PropertiesLoader by making the code cleaner and reducing cognitive complexity. Adds unit testing.
#### Who is reviewing it? 
@Lambeaux 
@blen-desta 
@clockard
@coyotesqrl 
Should have no change in functionality.
[DDF-2879](https://codice.atlassian.net/browse/DDF-2879)
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/libs/common-system/pom.xml
+++ b/libs/common-system/pom.xml
@@ -38,6 +38,10 @@
             <artifactId>slf4j-api</artifactId>
             <version>${org.slf4j.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/libs/common-system/src/main/java/org/codice/ddf/configuration/AbsolutePathResolver.java
+++ b/libs/common-system/src/main/java/org/codice/ddf/configuration/AbsolutePathResolver.java
@@ -16,6 +16,8 @@ package org.codice.ddf.configuration;
 import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import javax.annotation.Nullable;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -25,12 +27,33 @@ public class AbsolutePathResolver {
 
   private final String path;
 
-  public AbsolutePathResolver(String path) {
+  public AbsolutePathResolver(@Nullable String path) {
     this.path = path;
   }
 
+  /**
+   * This method will attempt to convert the already given path into an absolute path: first using
+   * the system property {@code ddf.home} and then using {@code Path}'s {@code toAbsolutePath}'s
+   * method
+   *
+   * @return Absolute path as a String
+   */
   public String getPath() {
+    if (System.getProperty("ddf.home") == null) {
+      LOGGER.warn("System property ddf.home is not set");
+    }
+    return getPath(System.getProperty("ddf.home"));
+  }
 
+  /**
+   * This method will attempt to convert the already given path into an absolute path: first by
+   * appending the given {@code rootPath}, then by using {@code Path}'s {@code toAbsolutePath}'s
+   * method
+   *
+   * @param rootPath String to append to path
+   * @return Absolute path as a String
+   */
+  public String getPath(String rootPath) {
     if (path == null) {
       return null;
     }
@@ -40,11 +63,11 @@ public class AbsolutePathResolver {
     Path absolutePath = Paths.get(path);
 
     if (!absolutePath.isAbsolute()) {
-      if (System.getProperty("ddf.home") != null) {
-        absolutePath = Paths.get(System.getProperty("ddf.home"), path);
+      if (StringUtils.isNotBlank(rootPath)) {
+        absolutePath = Paths.get(rootPath, path);
       } else {
-        LOGGER.warn(
-            "Path {} is relative. System property ddf.home is not set, resolving path to: {} ",
+        LOGGER.debug(
+            "Root path is blank. Resolving relative path [{}] to: {}",
             path,
             absolutePath.toAbsolutePath());
       }
--- a/libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
+++ b/libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
@@ -23,42 +23,66 @@ import org.junit.Test;
 
 public class AbsolutePathResolverTest {
 
+  private static final AbsolutePathResolver TEST_APR =
+      new AbsolutePathResolver(Paths.get("test", "path").toString());
+
+  private static final String USER_DIRECTORY = System.getProperty("user.dir");
+
+  private static final String DEFAULT_ABSOLUTE_PATH =
+      Paths.get(USER_DIRECTORY, "test", "path").toString();
+
   @Test
   public void testNullPath() {
     System.clearProperty("ddf.home");
-    assertThat(new AbsolutePathResolver(null).getPath(), nullValue());
+    AbsolutePathResolver apr = new AbsolutePathResolver(null);
+    assertThat(apr.getPath(), nullValue());
   }
 
   @Test
-  public void ddfHomeNotSet() {
+  public void testNullPathWithRoot() {
+    AbsolutePathResolver apr = new AbsolutePathResolver(null);
+    assertThat(apr.getPath(USER_DIRECTORY), nullValue());
+  }
+
+  @Test
+  public void getPathWhenDdfHomeNotSet() {
     System.clearProperty("ddf.home");
-    assertThat(
-        new AbsolutePathResolver(Paths.get("test", "path").toString()).getPath(),
-        equalTo(Paths.get(System.getProperty("user.dir"), "test", "path").toString()));
+    assertThat(TEST_APR.getPath(), equalTo(DEFAULT_ABSOLUTE_PATH));
   }
 
   @Test
   public void testTransformedPath() {
-    System.setProperty(
-        "ddf.home", Paths.get(System.getProperty("user.dir"), "testTransformedPath").toString());
+    System.setProperty("ddf.home", Paths.get(USER_DIRECTORY, "testTransformedPath").toString());
     assertThat(
-        new AbsolutePathResolver(Paths.get("test", "path").toString()).getPath(),
-        equalTo(
-            Paths.get(System.getProperty("user.dir"), "testTransformedPath", "test", "path")
-                .toString()));
+        TEST_APR.getPath(),
+        equalTo(Paths.get(USER_DIRECTORY, "testTransformedPath", "test", "path").toString()));
   }
 
   @Test
   public void trailingFileSeparator() {
     System.setProperty(
-        "ddf.home",
-        Paths.get(System.getProperty("user.dir"), "trailingFileSeparator").toString()
-            + File.separator);
+        "ddf.home", Paths.get(USER_DIRECTORY, "trailingFileSeparator").toString() + File.separator);
     assertThat(
         new AbsolutePathResolver("test" + File.separator + "path" + File.separator).getPath(),
         equalTo(
-            Paths.get(System.getProperty("user.dir"), "trailingFileSeparator", "test", "path")
-                    .toString()
+            Paths.get(USER_DIRECTORY, "trailingFileSeparator", "test", "path").toString()
                 + File.separator));
+  }
+
+  @Test
+  public void getPathWithNullRootPath() {
+    assertThat(TEST_APR.getPath(null), equalTo(DEFAULT_ABSOLUTE_PATH));
+  }
+
+  @Test
+  public void getPathWithEmptyRootPath() {
+    assertThat(TEST_APR.getPath(""), equalTo(DEFAULT_ABSOLUTE_PATH));
+  }
+
+  @Test
+  public void getPathWithValidRootPath() {
+    assertThat(
+        TEST_APR.getPath(Paths.get(USER_DIRECTORY, "valid").toString()),
+        equalTo(Paths.get(USER_DIRECTORY, "valid", "test", "path").toString()));
   }
 }
--- a/platform/util/platform-util/pom.xml
+++ b/platform/util/platform-util/pom.xml
@@ -82,6 +82,11 @@
             <version>1.16.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ddf.lib</groupId>
+            <artifactId>common-system</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
+++ b/platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
@@ -13,6 +13,8 @@
  */
 package org.codice.ddf.platform.util.properties;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
@@ -20,12 +22,18 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.function.BiFunction;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.text.StrSubstitutor;
+import org.codice.ddf.configuration.AbsolutePathResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.core.io.support.PropertiesLoaderUtils;
@@ -59,7 +67,17 @@ public final class PropertiesLoader {
 
   private static final PropertiesLoader INSTANCE = new PropertiesLoader();
 
-  private PropertiesLoader() {}
+  private static final List<BiFunction<String, ClassLoader, Properties>>
+      PROPERTY_LOADING_STRATEGIES =
+          ImmutableList.of(
+              PropertiesLoader::attemptLoadWithSpring,
+              PropertiesLoader::attemptLoadWithSpringAndClassLoader,
+              PropertiesLoader::attemptLoadWithFileSystem,
+              PropertiesLoader::attemptLoadAsResource);
+
+  private PropertiesLoader() {
+    // Perform operations using the singleton instance.
+  }
 
   public static PropertiesLoader getInstance() {
     return INSTANCE;
@@ -79,14 +97,14 @@ public final class PropertiesLoader {
   public <K, V> Map<K, V> toMap(Properties properties) {
     if (properties != null) {
       final Set<Map.Entry<Object, Object>> entries = properties.entrySet();
-      Map<K, V> map = new HashMap<K, V>(entries.size() * 2);
+      Map<K, V> map = new HashMap<>(entries.size() * 2);
       for (Map.Entry<Object, Object> entry : entries) {
         map.put((K) entry.getKey(), (V) entry.getValue());
       }
 
       return map;
     }
-    return new HashMap<K, V>();
+    return Collections.emptyMap();
   }
 
   /**
@@ -108,108 +126,138 @@ public final class PropertiesLoader {
    * @return Properties deserialized from the specified file, or empty if the load failed
    */
   public Properties loadProperties(String propertiesFile, ClassLoader classLoader) {
-    boolean error = false;
     Properties properties = new Properties();
-    if (propertiesFile != null) {
-      try {
-        LOGGER.debug(
-            "Attempting to load properties from {} with Spring PropertiesLoaderUtils.",
-            propertiesFile);
-        properties = PropertiesLoaderUtils.loadAllProperties(propertiesFile);
-      } catch (IOException e) {
-        error = true;
-        LOGGER.debug("Unable to load properties using default Spring properties loader.", e);
-      }
-      if (error || properties.isEmpty()) {
-        if (classLoader != null) {
-          try {
-            LOGGER.debug(
-                "Attempting to load properties from {} with Spring PropertiesLoaderUtils with class loader.",
-                propertiesFile);
-            properties = PropertiesLoaderUtils.loadAllProperties(propertiesFile, classLoader);
-            error = false;
-          } catch (IOException e) {
-            error = true;
-            LOGGER.debug("Unable to load properties using default Spring properties loader.", e);
-          }
-        } else {
-          try {
-            LOGGER.debug(
-                "Attempting to load properties from {} with Spring PropertiesLoaderUtils with class loader.",
-                propertiesFile);
-            properties =
-                PropertiesLoaderUtils.loadAllProperties(
-                    propertiesFile, PropertiesLoader.class.getClassLoader());
-            error = false;
-          } catch (IOException e) {
-            error = true;
-            LOGGER.debug("Unable to load properties using default Spring properties loader.", e);
-          }
-        }
-      }
-
-      if (error || properties.isEmpty()) {
-        LOGGER.debug("Attempting to load properties from file system: {}", propertiesFile);
-        File propFile = new File(propertiesFile);
-        // If properties file has fully-qualified absolute path (which
-        // the blueprint file specifies) then can load it directly.
-        if (propFile.isAbsolute()) {
-          LOGGER.debug("propertiesFile {} is absolute", propertiesFile);
-          propFile = new File(propertiesFile);
-        } else {
-          String rootDirectory = System.getProperty("karaf.home");
-          if (rootDirectory != null && !rootDirectory.isEmpty()) {
-            propFile = new File(rootDirectory, propertiesFile);
-          } else {
-            rootDirectory = System.getProperty("ddf.home");
-            if (rootDirectory != null && !rootDirectory.isEmpty()) {
-              propFile = new File(rootDirectory, propertiesFile);
-            } else {
-              propFile = new File(propertiesFile);
-            }
-          }
-        }
-        properties = new Properties();
-
-        try (InputStreamReader reader =
-            new InputStreamReader(new FileInputStream(propFile), StandardCharsets.UTF_8)) {
-          properties.load(reader);
-        } catch (FileNotFoundException e) {
-          error = true;
-          LOGGER.debug("Could not find properties file: {}", propFile.getAbsolutePath(), e);
-        } catch (IOException e) {
-          error = true;
-          LOGGER.debug("Error reading properties file: {}", propFile.getAbsolutePath(), e);
-        }
-      }
-      if (error || properties.isEmpty()) {
-        LOGGER.debug("Attempting to load properties as a resource: {}", propertiesFile);
-        InputStream ins = PropertiesLoader.class.getResourceAsStream(propertiesFile);
-        if (ins != null) {
-          try {
-            properties.load(ins);
-            ins.close();
-          } catch (IOException e) {
-            LOGGER.debug("Unable to load properties: {}", propertiesFile, e);
-          } finally {
-            IOUtils.closeQuietly(ins);
-          }
-        }
-      }
-
-      // replace any ${prop} with system properties
-      Properties filtered = new Properties();
-      for (Map.Entry<?, ?> entry : properties.entrySet()) {
-        filtered.put(
-            StrSubstitutor.replaceSystemProperties(entry.getKey()),
-            StrSubstitutor.replaceSystemProperties(entry.getValue()));
-      }
-      properties = filtered;
-
-    } else {
+    if (propertiesFile == null) {
       LOGGER.debug("Properties file must not be null.");
+      return properties;
+    }
+
+    Iterator<BiFunction<String, ClassLoader, Properties>> strategiesIterator =
+        PROPERTY_LOADING_STRATEGIES.iterator();
+    do {
+      properties = strategiesIterator.next().apply(propertiesFile, classLoader);
+    } while (properties.isEmpty() && strategiesIterator.hasNext());
+
+    properties = substituteSystemPropertyPlaceholders(properties);
+    return properties;
+  }
+
+  /** Default property loading strategy. */
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadWithSpring(String propertiesFile, ClassLoader classLoader) {
+    Properties properties = new Properties();
+    try {
+      LOGGER.debug(
+          "Attempting to load properties from {} with Spring PropertiesLoaderUtils.",
+          propertiesFile);
+      properties = PropertiesLoaderUtils.loadAllProperties(propertiesFile);
+    } catch (IOException e) {
+      LOGGER.debug("Unable to load properties using default Spring properties loader.", e);
+    }
+    return properties;
+  }
+
+  /** Try loading properties using Spring and a provided class loader. */
+  @VisibleForTesting
+  static Properties attemptLoadWithSpringAndClassLoader(
+      String propertiesFile, ClassLoader classLoader) {
+    Properties properties = new Properties();
+    try {
+      LOGGER.debug(
+          "Attempting to load properties from {} with Spring PropertiesLoaderUtils with class loader.",
+          propertiesFile);
+      if (classLoader != null) {
+        properties = PropertiesLoaderUtils.loadAllProperties(propertiesFile, classLoader);
+      } else {
+        properties =
+            PropertiesLoaderUtils.loadAllProperties(
+                propertiesFile, PropertiesLoader.class.getClassLoader());
+      }
+    } catch (IOException e) {
+      LOGGER.debug("Unable to load properties using default Spring properties loader.", e);
+    }
+    return properties;
+  }
+
+  /**
+   * Try loading the properties directly from the file system. If the properties file has a
+   * fully-qualified absolute path (which is what the blueprint file should specify) then it can be
+   * loaded directly, using this method. Otherwise the path will be considered relative and attempts
+   * will be made with {@code karaf.home} and {@code ddf.home} property values prepended to the
+   * original path.
+   */
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadWithFileSystem(String propertiesFile, ClassLoader classLoader) {
+    LOGGER.debug("Attempting to load properties from file system: {}", propertiesFile);
+    Properties properties = new Properties();
+
+    String karafHome = System.getProperty("karaf.home");
+    String ddfHome = System.getProperty("ddf.home");
+
+    File propFile;
+    AbsolutePathResolver absPath = new AbsolutePathResolver(propertiesFile);
+    if (StringUtils.isNotBlank(karafHome)) {
+      propFile = new File(absPath.getPath(karafHome));
+    } else if (StringUtils.isNotBlank(ddfHome)) {
+      propFile = new File(absPath.getPath(ddfHome));
+    } else {
+      propFile = new File(propertiesFile);
+    }
+
+    if (propFile.exists()) {
+      try (InputStreamReader reader =
+          new InputStreamReader(new FileInputStream(propFile), StandardCharsets.UTF_8)) {
+        properties.load(reader);
+      } catch (FileNotFoundException e) {
+        LOGGER.debug("Could not find properties file: {}", propFile.getAbsolutePath(), e);
+      } catch (IOException e) {
+        LOGGER.debug("Error reading properties file: {}", propFile.getAbsolutePath(), e);
+        properties.clear();
+      }
+    } else {
+      LOGGER.debug("Could not find properties file: {}", propFile.getAbsolutePath());
     }
 
     return properties;
+  }
+
+  /** Try loading the properties using Java's resource loading facilities. */
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadAsResource(String propertiesFile, ClassLoader classLoader) {
+    LOGGER.debug("Attempting to load properties as a resource: {}", propertiesFile);
+    InputStream ins = PropertiesLoader.class.getResourceAsStream(propertiesFile);
+    Properties properties = new Properties();
+    if (ins != null) {
+      try {
+        properties.load(ins);
+      } catch (IOException e) {
+        LOGGER.debug("Unable to load properties: {}", propertiesFile, e);
+      } finally {
+        IOUtils.closeQuietly(ins);
+      }
+    }
+    return properties;
+  }
+
+  /**
+   * Replace any ${prop} with system properties.
+   *
+   * @param props current state of properties that contain placeholders of the form ${property}.
+   * @return the given property object with system property placeholders switched to their actual
+   *     values.
+   */
+  @VisibleForTesting
+  static Properties substituteSystemPropertyPlaceholders(Properties props) {
+    Properties filtered = new Properties();
+    for (Map.Entry<?, ?> entry : props.entrySet()) {
+      filtered.put(
+          StrSubstitutor.replaceSystemProperties(entry.getKey()),
+          StrSubstitutor.replaceSystemProperties(entry.getValue()));
+      StrSubstitutor.replaceSystemProperties(new Object());
+    }
+    return filtered;
   }
 }
--- a/platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
+++ b/platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
@@ -1,0 +1,189 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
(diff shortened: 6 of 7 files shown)


bakejeyner:
build now

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,206 @@
...
+    propertiesFile = temporaryFolder.newFile(PROPERTIES_FILENAME);
+    Files.write(propertiesFile.toPath(), sb.toString().getBytes());
+  }
+
+  @Test
+  public void testToMapWithProperties() throws Exception {
+    when(propertiesMock.entrySet()).thenReturn(exampleMap.entrySet());
+
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(propertiesMock);
+    Assert.assertEquals("Returns a map of the properties", exampleMap, entries);
```
blen-desta: ✏️  You can import `static org.junit.Assert.assertEquals` and directly call `assertEquals` in all your tests.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,206 @@
...
+        "Returns the correct properties", exampleMap.entrySet(), testProperties.entrySet());
+  }
+
+  @Test
+  public void testAttemptLoadWithSpring() throws Exception {
+    Properties testProperties =
+        PropertiesLoader.attemptLoadWithSpring(
+            PROPERTIES_FILENAME, this.getClass().getClassLoader());
+
+    this.getClass().getClassLoader().getResource("pom.xml");
```
blen-desta: ❓ Why do you have this line?

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +122,140 @@ public Properties loadProperties(String propertiesFile) {
...
-          }
-        }
+    if (propertiesFile == null) {
+      LOGGER.debug("Properties file must not be null.");
+      return properties;
+    }
+
+    Iterator<BiFunction<String, ClassLoader, Properties>> strategiesIterator =
+        PROPERTY_LOADING_STRATEGIES.iterator();
+    do {
```
coyotesqrl: ✏️ An alternative (neither better nor worse) would be something like:
```java
return PROPERTY_LOADING_STRATEGIES.stream()
  .map(f -> f.apply(propertiesFile, classLoader)
  .filter(Objects::nonNull)
  .filter(p -> !properties.isEmpty())
  .map(PropertiesLoader::substituteSystemPropertyPlaceholders);
```

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +122,140 @@ public Properties loadProperties(String propertiesFile) {
...
+   * original path.
+   */
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadWithFileSystem(String propertiesFile, ClassLoader classLoader) {
+    LOGGER.debug("Attempting to load properties from file system: {}", propertiesFile);
+    File propFile = new File(propertiesFile);
+    Properties properties = new Properties();
+
+    if (propFile.isAbsolute()) {
```
coyotesqrl: ❓  If the prop file path is absolute, shouldn't we be confirming that it is beneath `$ddf.home` before trying to read it?

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +122,140 @@ public Properties loadProperties(String propertiesFile) {
...
+      LOGGER.debug("propertiesFile {} is absolute", propertiesFile);
+      propFile = new File(propertiesFile);
+    } else {
+      String rootDirectory = System.getProperty("karaf.home");
+      if (rootDirectory != null && !rootDirectory.isEmpty()) {
```
coyotesqrl: ✏️ StringUtils

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
coyotesqrl: In fact, don't we already have a utility that can normalize a given input path?

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +122,140 @@ public Properties loadProperties(String propertiesFile) {
...
-            StrSubstitutor.replaceSystemProperties(entry.getValue()));
-      }
-      properties = filtered;
-
-    } else {
-      LOGGER.debug("Properties file must not be null.");
+    try (InputStreamReader reader =
+        new InputStreamReader(new FileInputStream(propFile), StandardCharsets.UTF_8)) {
+      properties.load(reader);
+    } catch (FileNotFoundException e) {
```
coyotesqrl: ✏️ We can't _not_ catch this checked exception as it is thrown from the FIS ctor, but we can eliminate the processing hit of it _actually_ occurring by confirming that `propFile` exists prior to entering this try/catch block.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -51,15 +57,23 @@
  * Note that the first successful strategy is the one whose results are non-empty, and no further
  * strategies will be attempted thereafter. Property placeholders are always substituted after the
  * loading has finished.
- *
- * <p>If all strategies fail, then the returned Properties object will be empty.
```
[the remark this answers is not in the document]
bakejeyner: Whoops. We do still return empty properties. Took that out on accident.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
bakejeyner: Whoops

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
blen-desta: `AbsolutePathResolver`?

bakejeyner: We aren't supposed to change this classes functionality at all. Too many things depend on this class including integrators. Message @Lambeaux if you need a more in depth explanation.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
bakejeyner: I'm personally too worried to make that change and have it behave exactly the same as it is now. I don't want some weird stream thing to bite me in the behind. I like the idea of using streams though.

bakejeyner:
build now

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
coyotesqrl: Would not change the functionality to use extant code to do what you're doing here. @Lambeaux would you please take a look and confirm that the `AbsolutePathResolver` would fit the needs here?

bakejeyner:
build now

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+    when(propertiesMock.entrySet()).thenReturn(exampleMap.entrySet());
+
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(propertiesMock);
+    Assert.assertEquals("Returns a map of the properties", exampleMap, entries);
+  }
+
+  @Test
+  public void testToMapWithNull() throws Exception {
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(null);
+    Assert.assertEquals("Returns an empty map", Collections.emptyMap(), entries);
```
Lambeaux: ❗️ Note that all `message` parameters to the assertion functions are really just error messages displayed on test failure. Thus I do not think `"Returns an empty map"` is descriptive enough to assist with troubleshooting. Perhaps something like: 
```
format("Expected the returned map to be empty, but got %s", entries.toString())
```
Also with `assertThat` and Hamcrest matchers, this assertion no longer needs a call to `Collections.emptyMap()` because the test conditions can be better communicated. 
```
assertThat(entries, is(empty()));
``` 
Much more concise! And readable! 👍

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(new Properties());
+    Assert.assertEquals("Returns an empty map", Collections.emptyMap(), entries);
+  }
+
+  @Test
+  public void testLoadPropertiesWithoutPropertiesFile() throws Exception {
+    Properties testProperties = PROPERTIES_LOADER.loadProperties(null, null);
+    Properties emptyProperties = new Properties();
+
+    Assert.assertEquals("Returns empty properties", emptyProperties, testProperties);
```
Lambeaux: Can be simplified using Hamcrest.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  @Test
+  public void testLoadPropertiesWithEmptyPropertiesFile() throws Exception {
+    File emptyFile = temporaryFolder.newFile("tempFile");
+    Properties testProperties = PROPERTIES_LOADER.loadProperties(emptyFile.getPath(), null);
+    Properties emptyProperties = new Properties();
+
+    Assert.assertEquals("Returns empty properties", emptyProperties, testProperties);
```
Lambeaux: Can be simplified using Hamcrest. 
I'll stop posting these comments. All assertions should be revisited using Hamcrest.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+  public void testLoadPropertiesWithoutPropertiesFile() throws Exception {
+    Properties testProperties = PROPERTIES_LOADER.loadProperties(null, null);
+    Properties emptyProperties = new Properties();
+
+    Assert.assertEquals("Returns empty properties", emptyProperties, testProperties);
+  }
+
+  @Test
+  public void testLoadPropertiesWithEmptyPropertiesFile() throws Exception {
+    File emptyFile = temporaryFolder.newFile("tempFile");
```
Lambeaux: ✏️  Can we move this to the `@BeforeClass` so all setup and initialization occurs in the same place?

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    Assert.assertEquals("Returns empty properties", emptyProperties, testProperties);
+  }
+
+  @Test
+  public void testLoadPropertiesWithProperties() throws Exception {
+    Properties testProperties = PROPERTIES_LOADER.loadProperties(propertiesFile.getCanonicalPath());
+
+    Assert.assertEquals(
+        "Returns the correct properties", exampleMap.entrySet(), testProperties.entrySet());
```
Lambeaux: ❗️ Since we're testing each strategy individually, this test is pretty worthless / redundant unless we have a deterministic way to verify strategy ordering. Which we should have. 

Another way to put it: This test already tests the functionality tested by `testAttemptLoadWithSpring` because that's the first strategy and it will always succeed. What we need is a way to verify the order of the strategies list without exposing it to the world. That's what this test should truly be testing. Otherwise, just delete it.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  @Test
+  public void testAttemptLoadWithSpring() throws Exception {
+    Properties testProperties =
+        PropertiesLoader.attemptLoadWithSpring(
+            PROPERTIES_FILENAME, this.getClass().getClassLoader());
```
Lambeaux: ✏️  You could cache the class loader in memory:
```
private static final ClassLoader CLASS_LOADER = PropertiesLoader.class.getClassLoader();
```

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+        PropertiesLoader.attemptLoadWithFileSystem(NON_EXISTENT_FILENAME, null);
+
+    Assert.assertEquals(
+        "Returns empty properties", Collections.emptySet(), testProperties.entrySet());
+  }
+
+  @Test
+  public void testAttemptLoadAsResource() throws Exception {
+    Properties testProperties =
+        PropertiesLoader.attemptLoadAsResource("/" + PROPERTIES_FILENAME, null);
```
Lambeaux: ✏️  Use `String.format(...)` instead of concatenation.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+    Assert.assertEquals(
+        "Returns empty properties", Collections.emptySet(), testProperties.entrySet());
+  }
+
+  @Test
+  public void testSubstituteSystemPropertyPlaceholders() throws Exception {
+    // set up map to replace system properties
+    Map<Object, Object> testMapSystemProperties = new HashMap<>();
+    testMapSystemProperties.put("Java Home", "${java.home}");
+    testMapSystemProperties.put("Java Version", "${java.version}");
```
Lambeaux: ❗️ We probably should not have spaces in the keys. They are usually of the form `org.codice.qualifier.qualifier.propertyname` and we should test as such. Also we should test for system property expansions in the keys as well.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    // set up mock and insert into method
+    when(propertiesMock.entrySet()).thenReturn(testMapSystemProperties.entrySet());
+    Properties testProperties =
+        PropertiesLoader.substituteSystemPropertyPlaceholders(propertiesMock);
+
+    // manually replace system properties
+    testMapSystemProperties.replace("Java Home", System.getProperty("java.home"));
+    testMapSystemProperties.replace("Java Version", System.getProperty("java.version"));
```
Lambeaux: ❗️ Instead of mutating state that you've already passed, which makes this difficult to follow and verify what is being tested, you should have two separate maps. One can be your expected value. One can be the value you pass with the mock.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
coyotesqrl: Clarifying point: we do NOT lose the benefit of short-circuiting with streams as they are lazily processed. Toss in the missing `findFirst()` after the second predicate `filter` (Not sure why I forgot it) and you'll stop processing on the first non-empty.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,200 @@
...
+    propertiesFile = temporaryFolder.newFile(PROPERTIES_FILENAME);
+    Files.write(propertiesFile.toPath(), sb.toString().getBytes());
+  }
+
+  @Test
+  public void testToMapWithProperties() throws Exception {
+    when(propertiesMock.entrySet()).thenReturn(exampleMap.entrySet());
+
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(propertiesMock);
+    Assert.assertEquals("Returns a map of the properties", exampleMap, entries);
```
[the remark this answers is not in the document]
Lambeaux: ❗️ Found it!
https://codice.atlassian.net/wiki/spaces/DDF/pages/6291523/Unit+Testing#UnitTesting-Assertions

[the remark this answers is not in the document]
bakejeyner: That is correct @Lambeaux. I forgot exactly were I found the document but it was on a confluence page somewhere.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
bakejeyner: I would rather keep this in here because this is the only test that uses an extra temporary file. If I was making extra files for all the tests I would see the benefit of moving those initializations into the before class because then that would be the spot where temporary files are made. But If only 1 test is making an extra file I think it makes more sense to put that logic in that test rather than pulling it out into its own section of 1 extra initialized file.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
bakejeyner: Eh. I only use it twice.

bakejeyner:
build now

bakejeyner:
build now

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
rzwiefel: string format isn't necessary for a 2 item concat

libs/common-system/src/main/java/org/codice/ddf/configuration/AbsolutePathResolver.java
```diff
@@ -53,4 +58,58 @@ public String getPath() {
...
+   * systemProperties} in order and returns the first one that isn't null (or empty depending on the
+   * value of {@code allowEmpty})
+   *
+   * @param systemProperties list of strings of system properties to try and append to the path
+   * @param allowEmpty if a system property is an empty string (i.e "") this will be valid and
+   *     appended to the path
+   * @param allowToAbs will call the {@code path.toAbsolutePath} at the end of the function
+   * @return the absolute path
+   */
+  public String getPath(List<String> systemProperties, boolean allowEmpty, boolean allowToAbs) {
```
Lambeaux: ❓ Not sure why this method is needed. I seriously think this is overkill for what we were trying to accomplish. We just needed to extend support for "karaf.home" in `AbsolutePathResolver` without adding too much additional functionality.

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
```diff
@@ -17,30 +17,43 @@
...
 
+import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.nio.file.Paths;
+import java.util.List;
 import org.junit.Test;
 
 public class AbsolutePathResolverTest {
 
+  private static final List<String> SYSTEM_PROPERTIES = ImmutableList.of("ddf.home", "karaf.home");;
```
Lambeaux: ❗️ This is premature optimization. We don't need all this support. I can't think of a case where we would need more than `ddf.home` and `karaf.home`.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +125,133 @@ public Properties loadProperties(String propertiesFile) {
...
+   * will be made with {@code karaf.home} and {@code ddf.home} property values prepended to the
+   * original path.
+   */
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadWithFileSystem(String propertiesFile, ClassLoader classLoader) {
+    LOGGER.debug("Attempting to load properties from file system: {}", propertiesFile);
+    Properties properties = new Properties();
+
+    List<String> systemProperties = ImmutableList.of("karaf.home", "ddf.home");
```
Lambeaux: ❗️ Yeah I understand what you were going for but I can't justify these changes when the base list won't be flexing much, if **at all**, over the course of time.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +125,133 @@ public Properties loadProperties(String propertiesFile) {
...
+
+    AbsolutePathResolver absPath = new AbsolutePathResolver(propertiesFile);
+
+    File propFile = new File(absPath.getPath(systemProperties, false, false));
```
Lambeaux: ❗️ Should change signature to accept a `Path` as the base for resolution instead of taking system property names.

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+public class PropertiesLoaderTest {
+
+  @Mock private Properties propertiesMock;
+
+  @ClassRule public static TemporaryFolder temporaryFolder = new TemporaryFolder();
+  private static final String PROPERTIES_FILENAME = "test.properties";
+  private static final String NON_EXISTENT_FILENAME = "nonExistentFile.properties";
+  private static File propertiesFile;
+
+  private static Map<Object, Object> exampleMap =
```
Lambeaux: ❓ Can this not be final?

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+    when(propertiesMock.entrySet()).thenReturn(exampleMap.entrySet());
+
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(propertiesMock);
+    assertThat(entries, equalTo(exampleMap));
+  }
+
+  @Test
+  public void testToMapWithNull() throws Exception {
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(null);
+    assertThat(entries, equalTo(Collections.emptyMap()));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty map

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  @Test
+  public void testToMapWithEmptyProperties() throws Exception {
+    Map<?, ?> entries = PROPERTIES_LOADER.toMap(new Properties());
+    assertThat(entries, equalTo(Collections.emptyMap()));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty map

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  @Test
+  public void testLoadPropertiesWithoutPropertiesFile() throws Exception {
+    Properties testProperties = PROPERTIES_LOADER.loadProperties(null, null);
+    Properties emptyProperties = new Properties();
+
+    assertThat(testProperties, equalTo(emptyProperties));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create empty props

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  @Test
+  public void testLoadPropertiesWithEmptyPropertiesFile() throws Exception {
+    File emptyFile = temporaryFolder.newFile("tempFile");
+    Properties testProperties = PROPERTIES_LOADER.loadProperties(emptyFile.getPath(), null);
+    Properties emptyProperties = new Properties();
+
+    assertThat(testProperties, equalTo(emptyProperties));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty map

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+            PROPERTIES_FILENAME, this.getClass().getClassLoader());
+
+    assertThat(testProperties.entrySet(), equalTo(exampleMap.entrySet()));
+  }
+
+  @Test
+  public void testAttemptLoadWithSpringFileDNE() throws Exception {
+    Properties testProperties = PropertiesLoader.attemptLoadWithSpring(NON_EXISTENT_FILENAME, null);
+
+    assertThat(testProperties.entrySet(), equalTo(Collections.emptySet()));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty set

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+    assertThat(testProperties.entrySet(), equalTo(exampleMap.entrySet()));
+  }
+
+  @Test
+  public void testAttemptLoadWithSpringAndClassLoaderFileDNE() throws Exception {
+    Properties testProperties =
+        PropertiesLoader.attemptLoadWithSpringAndClassLoader(NON_EXISTENT_FILENAME, null);
+
+    assertThat(testProperties.entrySet(), equalTo(Collections.emptySet()));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty set

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+
+    assertThat(testProperties.entrySet(), equalTo(exampleMap.entrySet()));
+  }
+
+  @Test
+  public void testAttemptLoadWithFileSystemFileDNE() throws Exception {
+    Properties testProperties =
+        PropertiesLoader.attemptLoadWithFileSystem(NON_EXISTENT_FILENAME, null);
+
+    assertThat(testProperties.entrySet(), equalTo(Collections.emptySet()));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty set

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+            String.format("%s%s", File.separator, PROPERTIES_FILENAME), null);
+
+    assertThat(testProperties.entrySet(), equalTo(exampleMap.entrySet()));
+  }
+
+  @Test
+  public void testAttemptLoadAsResourceFileDNE() throws Exception {
+    Properties testProperties = PropertiesLoader.attemptLoadAsResource(NON_EXISTENT_FILENAME, null);
+
+    assertThat(testProperties.entrySet(), equalTo(Collections.emptySet()));
```
Lambeaux: ❗️ `assertThat(entries, is(empty()))` - no need to create an empty set

platform/util/platform-util/src/test/java/org/codice/ddf/platform/util/properties/PropertiesLoaderTest.java
```diff
@@ -0,0 +1,192 @@
...
+  public void testSubstituteSystemPropertyPlaceholders() throws Exception {
+    // set up map to replace system properties
+    Map<Object, Object> testMapSystemProperties = new HashMap<>();
+    testMapSystemProperties.put("Java_Home", "${java.home}");
+    testMapSystemProperties.put("Java_Version", "${java.version}");
+
+    // set up map with replaced system properties
+    Map<Object, Object> testMapSystemPropertiesAfter = new HashMap<>();
+    testMapSystemPropertiesAfter.put("Java_Home", System.getProperty("java.home"));
+    testMapSystemPropertiesAfter.put("Java_Version", System.getProperty("java.version"));
```
Lambeaux: ✅  Much better. Easier to reason about.

bakejeyner:
build now

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
```diff
@@ -23,42 +23,62 @@
...
                 + File.separator));
   }
+
+  @Test
+  public void nullRootPath() {
+    assertThat(TEST_APR.getPath(null), equalTo(DEFAULT_ABSOLUTE_PATH));
+  }
+
+  @Test
+  public void emptyRootPath() {
```
Lambeaux: `testGetPathWithEmptyRootPath`

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
```diff
@@ -23,42 +23,62 @@
...
+    assertThat(TEST_APR.getPath(""), equalTo(DEFAULT_ABSOLUTE_PATH));
+  }
+
+  @Test
+  public void validRootPath() {
```
Lambeaux: `testGetPathWithValidRootPath`

platform/util/platform-util/pom.xml
```diff
@@ -82,6 +82,11 @@
             <version>1.16.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ddf.lib</groupId>
+            <artifactId>common-system</artifactId>
+            <version>2.12.0-SNAPSHOT</version>
```
Lambeaux: ❗️ Replace with `{project.version}` alias.

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
```diff
@@ -23,42 +23,62 @@
...
-        new AbsolutePathResolver(Paths.get("test", "path").toString()).getPath(),
-        equalTo(
-            Paths.get(System.getProperty("user.dir"), "testTransformedPath", "test", "path")
-                .toString()));
+        TEST_APR.getPath(),
+        equalTo(Paths.get(USER_DIRECTORY, "testTransformedPath", "test", "path").toString()));
   }
 
   @Test
   public void trailingFileSeparator() {
```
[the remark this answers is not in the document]
coyotesqrl: I'm going to be honest and say that I personally don't care whether test methods are prefixed with 'test'; the annotation and the fact we're in a test class is sufficient to me. Others may care more. That said, a more descriptive name is always to be preferred.

libs/common-system/pom.xml
```diff
@@ -38,6 +38,11 @@
             <artifactId>slf4j-api</artifactId>
             <version>${org.slf4j.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
+            <scope>provided</scope>
```
coyotesqrl: ✏️ It's fine to include the scope here, but as this artifact is dependencyManaged, it would only really matter if you were overriding the scope; just as you inherit the version, you inherit the scope.

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
```diff
@@ -23,42 +23,62 @@
 
 public class AbsolutePathResolverTest {
 
+  private static final AbsolutePathResolver TEST_APR =
+      new AbsolutePathResolver(Paths.get("test", "path").toString());;
+
+  private static final String USER_DIRECTORY = System.getProperty("user.dir");
```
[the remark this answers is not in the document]
coyotesqrl: I would create a temporary folder, otherwise you risk polluting `$HOME` and potentially failing on a future run of the test if an orphan file is left there.

libs/common-system/pom.xml
```diff
@@ -38,6 +38,10 @@
             <artifactId>slf4j-api</artifactId>
             <version>${org.slf4j.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
```
Lambeaux: ❗️ Need `<scope>provided</scope>`

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
```diff
@@ -23,42 +23,66 @@
...
     System.clearProperty("ddf.home");
-    assertThat(new AbsolutePathResolver(null).getPath(), nullValue());
+    AbsolutePathResolver apr = new AbsolutePathResolver(null);
+    assertThat(apr.getPath(), nullValue());
+  }
+
+  @Test
+  public void testNullPathWithRoot() {
+    AbsolutePathResolver apr = new AbsolutePathResolver(null);
+    assertThat(apr.getPath("USER_DIRECTORY"), nullValue());
```
Lambeaux: ❓ Did you mean to reference the variable **USER_DIRECTORY** and not a String with the value of `"USER_DIRECTORY"`?

libs/common-system/pom.xml
bakejeyner: @coyotesqrl mentioned the scope is inherited from the parent's dependency manager

libs/common-system/src/test/java/org/codice/ddf/configuration/AbsolutePathResolverTest.java
bakejeyner: Yes. Yes I did.

libs/common-system/pom.xml
Lambeaux: Okay. I see where that is happening 👍  We only benefit from that on master, not release branches.

bakejeyner:
build now
(CI download failure)

libs/common-system/src/main/java/org/codice/ddf/configuration/AbsolutePathResolver.java
```diff
@@ -25,10 +27,17 @@
...
   }
 
+  /**
+   * This method will attempt to convert the already given path into an absolute path: first using
+   * the system property {@code ddf.home} and then using {@code Path}'s {@code toAbsolutePath}'s
+   * method
+   *
+   * @return Absolute path as a String
+   */
   public String getPath() {
```
coyotesqrl: ❓ Why wouldn't use just update this method to be `return getPath(System.getProperty("ddf.home"));`

You could keep a little extra baggage from this older implementation - logging specifically that `ddf.home` is not set would be worthwhile, I suppose - but otherwise, why duplicate logic?

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +126,140 @@ public Properties loadProperties(String propertiesFile) {
...
+    } while (properties.isEmpty() && strategiesIterator.hasNext());
+
+    properties = substituteSystemPropertyPlaceholders(properties);
+    return properties;
+  }
+
+  /** Default property loading strategy. */
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadWithSpring(String propertiesFile, ClassLoader classLoader) {
```
coyotesqrl: ✏️ This method can simply return `attemptLoadWithSpringAndClassLoader(propertiesFile, null)`. You'd want to update the log message in that method then to indicate whether the CL passed in was null or not.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +126,140 @@ public Properties loadProperties(String propertiesFile) {
...
+
+    if (propFile.exists()) {
+      try (InputStreamReader reader =
+          new InputStreamReader(new FileInputStream(propFile), StandardCharsets.UTF_8)) {
+        properties.load(reader);
+      } catch (FileNotFoundException e) {
+        LOGGER.debug("Could not find properties file: {}", propFile.getAbsolutePath(), e);
+      } catch (IOException e) {
+        LOGGER.debug("Error reading properties file: {}", propFile.getAbsolutePath(), e);
+        properties.clear();
```
coyotesqrl: ❓ Is this what we want to do if we fail to close? Clear out the read properties? I can't see why we wouldn't just log the failed close, if that's what happened, and then return the read contents.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +126,140 @@ public Properties loadProperties(String propertiesFile) {
...
+  @SuppressWarnings("squid:S1172" /* Used in bi-function */)
+  @VisibleForTesting
+  static Properties attemptLoadAsResource(String propertiesFile, ClassLoader classLoader) {
+    LOGGER.debug("Attempting to load properties as a resource: {}", propertiesFile);
+    InputStream ins = PropertiesLoader.class.getResourceAsStream(propertiesFile);
+    Properties properties = new Properties();
+    if (ins != null) {
+      try {
+        properties.load(ins);
+        ins.close();
```
coyotesqrl: ❗️ No need to close here; the finally block will manage this for you.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
```diff
@@ -108,108 +126,140 @@ public Properties loadProperties(String propertiesFile) {
...
+      } catch (IOException e) {
+        LOGGER.debug("Unable to load properties: {}", propertiesFile, e);
+        properties.clear();
```
coyotesqrl: ❗️ Once you remove the call to `close` above, you won't need to clear the properties any longer.

libs/common-system/src/main/java/org/codice/ddf/configuration/AbsolutePathResolver.java
bakejeyner: @Lambeaux and I felt preserving the log statement was more important.

coyotesqrl: You can still do that by adding the baggage as I suggested.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
bakejeyner: The logic would be slightly different in your solution.
When `attemptLoadWithSpring` is passed `null`, down the line the class loader used is `ClassUtils.getDefaultClassLoader()`.
When `attemptLoadWithSpringAndClassLoader` is passed `null`, down the line the class loader used is `PropertiesLoader.class.getClassLoader()`.

platform/util/platform-util/src/main/java/org/codice/ddf/platform/util/properties/PropertiesLoader.java
bakejeyner: The goal of this ticket is to refactor and preserve functionality, whether that functionality be good or bad. Changing functionality could result in not only our code but integrator's code breaking as well.

libs/common-system/src/main/java/org/codice/ddf/configuration/AbsolutePathResolver.java
```diff
@@ -40,13 +60,20 @@ public String getPath() {
...
-      if (System.getProperty("ddf.home") != null) {
-        absolutePath = Paths.get(System.getProperty("ddf.home"), path);
+      if (StringUtils.isNotBlank(rootPath)) {
+        absolutePath = Paths.get(rootPath, path);
       } else {
-        LOGGER.warn(
-            "Path {} is relative. System property ddf.home is not set, resolving path to: {} ",
-            path,
-            absolutePath.toAbsolutePath());
+        if (System.getProperty("ddf.home") == null) {
```
coyotesqrl: ❗️ No. The log message for the property being null belongs in the overloaded method, not here.

bakejeyner:
build now

PR #2859, opened 2018-01-17
DDF-3513 Adds user pref for previewing images by hover

andrewkfiedler:
#### What does this PR do?
 - Adds user preference to preview images by hovering.

#### Who is reviewing it? 
@bdeining
@jlcsmith
@mackncheesiest 
@mojogitoverhere 

Turn on full image preview on hover.  Then go to the result previews and hover over an image and hover over the zoom icon.  This will pop open a full view of the image.  Click to see the image in a new window.

Turn off full image preview on hover.  Then go to the result previews and hover over an image and hover over the zoom icon.  Verify nothing pops up.  Verify you can still click the icon to see the image in a new window.

https://codice.atlassian.net/browse/DDF-3513

screen shot 2018-01-17 at 11 58 41 am
screen shot 2018-01-17 at 11 58 33 am
screen shot 2018-01-17 at 11 58 20 am

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/component.less
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/component.less
@@ -212,3 +212,5 @@
 @import 'dropdown/map-settings/dropdown.map-settings.less';
 @import 'about/about';
 @import 'about-menu/about-menu';
+@import 'hover-preview/hover-preview';
+@import 'dropdown/hover-preview/dropdown.hover-preview.less';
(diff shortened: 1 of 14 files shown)


bdeining:
Verified that on hover, the image is shown in a pop out view from the Results Tab and the Table View.  It appears that this is not the full sized image upon inspection (see screenshots).  It appears that the hover capability is not tied into the inspector view.  Was this intentional?  

screen shot 2018-01-18 at 12 59 45 pm
screen shot 2018-01-18 at 12 15 27 pm

andrewkfiedler:
The hover is displaying the full size of the thumbnail, so it'll depend on whether or not that exceeded the normal thumbnail size.  We could add a new action to try and get the full image rather than the thumbnail though.  

I didn't hook it into the inspector view, wasn't sure if we wanted it there as well.

bdeining:
build now

PR #2879, opened 2018-01-23
DDF-3297 add security manager to DDF

stustison:
Unit tests for added components will be completed in one of the follow on tickets, of which there will be many.

#### Who is reviewing it? 
@SmithJosh @ahoffer @Lambeaux @lessarderic 
@coyotesqrl
@pklinef
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


stustison:
build now

stustison:
build now

distribution/ddf-common/src/main/resources/bin/setenv
```diff
@@ -75,6 +75,8 @@ export DDF_ON_ERROR="bin/ddf_on_error.sh script "
 # Defines the special on-error Java options
 JAVA_ERROR_OPTS="-XX:OnOutOfMemoryError=\${DDF_ON_ERROR}%p -XX:OnError=\${DDF_ON_ERROR}%p"
 
+export KARAF_SYSTEM_OPTS="-Dpolicy.provider=net.sourceforge.prograde.policy.ProGradePolicy -Djava.security.manager=net.sourceforge.prograde.sm.ProGradeJSM -Djava.security.policy==${DDF_HOME}etc/all.policy -DproGrade.getPermissions.override=sun.rmi.server.LoaderHandler:loadClass,org.apache.jasper.compiler.JspRuntimeContext:initSecurity"
```
oconnormi: Fyi, we had moved the definition of pretty much all java properties into the `system.properties` file a while back. I had done some quick testing back then to make sure that the security manager properties would still be picked up and used correctly that way. Also things like `ddf.home` were set in there in order to avoid some things that were happening when the karaf scripts reference all these variables. Might be worth a quick try since it doesn't require platform specific computation of paths and such.

stustison:
build now

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,293 @@ grant
...
+    permission java.security.SecurityPermission "insertProvider";
+}
+
+grant codeBase "file:/platform-parser-xml" {
+    permission java.lang.RuntimePermission "setContextClassLoader";
+}
 
 //
-// Commons-io includes functionality to monitor directories asynchronously
+// itest specific perms
```
paouelle: ❓ Doesn't this particular one allows someone to inject these test libraries at runtime and gain access? Or are you confident that they wouldn't even be able to do that?

platform/osgi/platform-osgi-configadmin/pom.xml
```diff
@@ -195,17 +205,17 @@
                                         <limit>
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.75</minimum>
+                                            <minimum>0.25</minimum>
```
paouelle: ✏️  Ouch!

distribution/ddf-common/src/main/resources/bin/karaf
```diff
@@ -0,0 +1,331 @@
+#!/bin/sh
```
paouelle: ❓ Is it possible to add comments as to what needed to be changed in this file? This would help the next team that needs to perform a karaf upgrade.

distribution/ddf-common/src/main/resources/bin/karaf.bat
```diff
@@ -0,0 +1,455 @@
+@echo off
```
paouelle: ❓ Same comment as for `karaf` about adding comments

paouelle:
In addition to system re-validation, let's not forget to retest starting DDF in all 3 ways: ddf, start, service wrapper on both Windows and Linux. 
We should also re-test export/import is still functional with changes to keystore and policy files under pdp.

distribution/ddf-common/src/main/resources-filtered/etc/startup.properties
```diff
@@ -1,12 +1,15 @@
...
+mvn\:ddf.platform.osgi/platform-osgi-condpermadmin/${project.version} = 2
+mvn\:ddf.platform.osgi/platform-osgi-conditions/${project.version} = 2
 mvn\:org.apache.felix/org.apache.felix.metatype/1.1.2 = 5
 mvn\:org.apache.karaf.services/org.apache.karaf.services.eventadmin/4.1.2 = 5
 mvn\:org.ops4j.pax.url/pax-url-aether/2.5.2 = 5
 mvn\:org.ops4j.pax.logging/pax-logging-api/1.10.1 = 8
 mvn\:org.ops4j.pax.logging/pax-logging-log4j2/1.10.1 = 8
 # Using a boot feature won't be good enough. We need to actually replace Felix with our own impl.
 # The following ddf.platform.osgi lines (for the currently used karaf version) used to be:
+# The following ddf.platform.osgi line (for the currently used karaf version) used to be:
```
paouelle: ✏️  duplicated line

distribution/ddf-common/src/main/resources/bin/setenv.bat
```diff
@@ -62,6 +62,11 @@ rem SET KARAF_OPTS=
 rem Uncomment out the line below to enable cxf logging interceptors
 rem set EXTRA_JAVA_OPTS="-Dcom.sun.xml.ws.transport.http.HttpAdapter.dump=true"
 
+set DDF_HOME_PERM=%DDF_HOME:/=\%
+set DDF_HOME_PERM=%DDF_HOME_PERM:\bin\..=\%
+set DDF_HOME_POLICY=/%DDF_HOME%
+set DDF_HOME_POLICY=%DDF_HOME_POLICY:c:=C:%
```
paouelle: ❓ Is this assuming DDF_HOME is always in the C drive? What if it ends up in a different drive?

distribution/ddf-common/src/main/resources/bin/setenv.bat
```diff
@@ -72,13 +77,11 @@ rem As a work around, we will have the ddf_on_error.bat script use the karaf.pid
 rem and it will default to being invoked via the 'script'
 rem The space would be important at the end of the DDF_ON_ERROR value to separate the last argument from
 rem the %p used by the JVM to represent the pid of the JVM
-rem set DDF_ON_ERROR=bin\ddf_on_error.bat script^ 
+rem set DDF_ON_ERROR=bin\ddf_on_error.bat script^
```
paouelle: ❗️ the space at the end of that line shouldn't be removed

platform/osgi/platform-osgi-conditions/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.0</minimum>
```
paouelle: ❓ Why 0?

platform/osgi/platform-osgi-configadmin/pom.xml
```diff
@@ -62,6 +62,11 @@
             <artifactId>gson</artifactId>
             <version>${gson.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.platform</groupId>
+            <artifactId>org.eclipse.osgi</artifactId>
+            <version>3.11.3</version>
```
paouelle: ✏️  Should we not move this version in the parent pom along with all others?

distribution/ddf-common/src/main/resources/bin/karaf
stustison: They had a bug in their file, it looks like they fixed the ticket I wrote.

distribution/ddf-common/src/main/resources/common-bin.xml
```diff
@@ -13,268 +13,247 @@
...
+            </directory>
+            <outputDirectory>/</outputDirectory>
+            <includes>
+                <include>bin/admin.bat</include>
+                <include>bin/stop.bat</include>
+                <include>bin/shell.bat</include>
+                <include>bin/client.bat</include>
+                <include>bin/inc.bat</include>
+            </includes>
+            <lineEnding>dos</lineEnding>
```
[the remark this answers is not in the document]
stustison: Because they are included in our code now.

distribution/ddf-common/src/main/resources/security/default.policy
stustison: You'd need access to the OS anyways.

distribution/ddf-common/src/main/resources/bin/setenv.bat
stustison: It is a comment.

platform/osgi/platform-osgi-conditions/pom.xml
stustison: Because I haven't written any tests yet and likely won't for this PR.

stustison:
build now

distribution/sdk/sample-soap-endpoint/src/main/resources/OSGI-INF/permissions.perm
```diff
@@ -0,0 +1,17 @@
+(java.io.FilePermission "-" "read,write,delete")
```
coyotesqrl: ✏️ Nit: It would be good to group these permissions together. In fact in general, if permissions could be sorted where possible, it would probably help when reviewing and editing them.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -0,0 +1,79 @@
...
+  public boolean isPostponed() {
+    return false;
+  }
+
+  @Override
+  public boolean isSatisfied() {
+    if (args.length == 0) {
+      return false;
+    }
+    String bundleName = bundle.getHeaders().get("Bundle-SymbolicName");
```
coyotesqrl: ❓ How much do we care that symbolic name isn't unique? We do a pretty good job of following a naming convention that limits collisions, but there's nothing preventing them.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -0,0 +1,79 @@
...
+package org.codice.ddf.condition;
+
+import java.util.Dictionary;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+import org.osgi.framework.Bundle;
+import org.osgi.service.condpermadmin.Condition;
+import org.osgi.service.condpermadmin.ConditionInfo;
+
+public class BundleNameCondition implements Condition {
```
coyotesqrl: ❗️ Javadocs to explain the operation/function of this class.

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,203 @@
...
+public class PermissionActivator implements BundleActivator {
+
+  private static final String BUNDLE_NAME_CONDITION =
+      "org.codice.ddf.condition.BundleNameCondition";
+
+  private static final String PRINCIPAL_CONDITION = "org.codice.ddf.condition.PrincipalCondition";
+
+  private static final String SIGNER_CONDITION = "org.codice.ddf.condition.SignerCondition";
+
+  private static final String OSGI_CODEBASE = "file:/";
```
coyotesqrl: ✏️ Back to my comment in the policy file, I suggest we change this to `osgi:`.

platform/osgi/platform-osgi-configadmin/src/main/java/org/apache/felix/cm/impl/helper/BaseTracker.java
```diff
@@ -0,0 +1,290 @@
...
+import org.osgi.service.cm.ManagedServiceFactory;
+import org.osgi.service.log.LogService;
+import org.osgi.util.tracker.ServiceTracker;
+
+/**
+ * The <code>BaseTracker</code> is the base class for tracking <code>ManagedService</code> and
+ * <code>ManagedServiceFactory</code> services. It maps their <code>ServiceRegistration</code> to
+ * the {@link ConfigurationMap} mapping their service PIDs to provided configuration.
+ */
+public abstract class BaseTracker<S> extends ServiceTracker<S, ConfigurationMap<?>> {
```
coyotesqrl: ✏️ It would help in reviewing and for future upgrades if the changes to this file from its base state were commented.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,36 +31,23 @@ deny {
...
-    permission java.util.PropertyPermission "javax.net.ssl.*", "read, write";
-};
-
-//
-// DDF Migration Framework and Config Admin Migratable
+// DDF Platform Migratable
 //
 grant
-  codeBase "file:${ddf.home.policy}system/ddf/platform/platform-migratable/-" {
+  codeBase "file:/platform-migratable" {
```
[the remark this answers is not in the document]
stustison: File is the only one allowed by the security manager code.

distribution/ddf-common/src/main/resources/common-bin.xml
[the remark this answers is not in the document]
paouelle: Don't we still need to exclude those from the original karaf distribution to be safe?

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,296 @@ grant
...
-    permission java.io.FilePermission "${ddf.home}${/}deploy${/}-", "read, execute";
+grant codeBase "file:/org.eclipse.osgi" {
+    permission java.security.AllPermission;
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}org.ops4j.pax.url.war.cfg", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}jmx.acl.org.apache.camel.cfg", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}org.apache.karaf.decanter.sla.checker.cfg", "read,write";
+}
 
-    permission java.io.FilePermission "${ddf.home}${/}etc${/}-", "read, write, execute";
+grant codeBase "file:/registry-api-impl/catalog-plugin-metacardbackup-filestorage/org.apache.servicemix.bundles.xstream/spatial-csw-endpoint/platform-http-proxy/broker-route-manager/spatial-csw-transformer/spatial-csw-source/catalog-core-camelcontext/org.ops4j.pax.web.pax-web-jsp/org.apache.aries.transaction.blueprint/org.apache.aries.proxy/org.apache.camel.camel-blueprint/org.ops4j.pax.swissbox.core/org.apache.camel.camel-core" {
```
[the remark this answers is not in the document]
stustison: Our code isn't what is reading this, it is pro-grade and it won't process the white-space.

distribution/ddf-common/src/main/resources/bin/setenv.bat
paouelle: I know it is, but it is part of anther solution to the problem where there is a need for a space at the end of this line as per the comment 2 lines above. 
Either we remove all the comments or we leave them intact

distribution/ddf-common/src/main/resources/security/default.policy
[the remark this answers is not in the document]
coyotesqrl: We do, technically speaking _own_ the security manager code. ¯\_(ツ)_/¯

distribution/ddf-common/src/main/resources/security/default.policy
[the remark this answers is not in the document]
stustison: Things work fine if you have multiple conditions with permissions for a single bundle. It does slow down the entire container for every single one that you add, however. That is why I had to group these together. Unless people want itests that take 2.5 hours to run that is :)

platform/osgi/platform-osgi-conditions/pom.xml
paouelle: Why not? Should we not be increasing tech debt?

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,296 @@ grant
...
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keys.properties", "read";
+}
+
+grant codeBase "file:/pax-logging-log4j2" {
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}ddf.log", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}ingest_error.log", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}solr.log", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}artemis.log", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}security.log", "read,write";
+    permission java.net.SocketPermission "*", "listen,resolve";
```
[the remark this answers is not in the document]
stustison: Not sure, this is all that was needed for the itests to run.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,296 @@ grant
...
+}
+
+grant codeBase "file:/catalog-transformer-zip" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
+}
+
+grant codeBase "file:/sample-soap-endpoint" {
+    permission java.io.FilePermission "<<ALL FILES>>", "read,write,delete";
+    permission java.io.FilePermission "-", "read,write,delete";
```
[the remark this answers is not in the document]
stustison: It grants permission to resources within the bundle. And, yes you would think it would be.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,296 @@ grant
...
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}issuer${/}-", "read,write";
+}
+
+grant codeBase "file:/solr-banana-provisioner" {
+    permission java.util.PropertyPermission "solr.client", "read,write";
+}
+
+grant codeBase "file:/io.netty.common" {
+    permission java.io.FilePermission "${/}proc${/}sys${/}net${/}core${/}somaxconn", "read";
```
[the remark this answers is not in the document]
stustison: No, but it still gets looked up on Windows machines.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,296 @@ grant
...
+}
+
+grant codeBase "file:/security-certificate-generator" {
+    permission java.security.SecurityPermission "insertProvider";
+}
+
+grant codeBase "file:/platform-parser-xml" {
+    permission java.lang.RuntimePermission "setContextClassLoader";
```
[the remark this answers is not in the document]
stustison: Didn't seem to.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -69,115 +56,296 @@ grant
...
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}definitions${/}-", "read";
+}
+
+grant codeBase "file:/org.apache.karaf.jaas.modules" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keys.properties", "read";
+}
+
+grant codeBase "file:/pax-logging-log4j2" {
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}ddf.log", "read,write";
```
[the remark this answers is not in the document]
paouelle: I have to concur with Richard here. We would also be missing the log files that can be created by the service wrapper too.

platform/osgi/platform-osgi-condpermadmin/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
```
paouelle: ❓ Are we still using findbugs? I thought we got rid of it.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
stustison: It is all we have so it is going to have to be good enough.

distribution/ddf-common/src/main/resources/security/default.policy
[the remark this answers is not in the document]
paouelle: Should we add a comment in the file to that regard?

platform/osgi/platform-osgi-configadmin/src/main/java/org/apache/felix/cm/impl/helper/BaseTracker.java
pklinef: ❓ What did change?

distribution/ddf-common/src/main/resources/security/default.policy
pklinef: ❓ Should we still put test specific grants in a different policy file that is only loaded by the itests?  I worry about someone adding a grant here for itests that could also impact the operational system.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -188,29 +374,78 @@ grant {
...
+
+    permission java.io.FilePermission "${ddf.home.perm}karaf.pid", "read, write";
+
+    permission java.io.FilePermission "${ddf.home.perm}data", "read, write";
+    permission java.io.FilePermission "${ddf.home.perm}instances", "read, write";
+    permission java.io.FilePermission "${ddf.home.perm}solr", "read, write";
+    permission java.io.FilePermission "${ddf.home.perm}system", "read, write";
+    permission java.io.FilePermission "${ddf.home.perm}workspace", "read, write";
+
+    permission java.io.FilePermission "${ddf.home.perm}data${/}-", "read, write";
```
pklinef: ❗️ We need a ticket to lock the data folder down better.

stustison:
build now

distribution/ddf-common/src/main/resources/common-bin.xml
```diff
@@ -293,6 +274,20 @@
...
+    <dependencySet>
+      <outputDirectory>system/ddf/platform/osgi/platform-osgi-conditions/${project.version}
+      </outputDirectory>
+      <includes>
+        <include>ddf.platform.osgi:platform-osgi-conditions:jar:${project.version}</include>
+      </includes>
+    </dependencySet>
   </dependencySets>
 
 </component>
```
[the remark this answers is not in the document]
stustison: Not at the moment. It should be fixed on the latest blueprint soon.

distribution/ddf-common/src/main/resources/security/default.policy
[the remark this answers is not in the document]
stustison: I'll make a ticket to look into making pro-grade more osgi friendly

distribution/ddf-common/src/main/resources/security/default.policy
stustison: I'll make a ticket to look into separating into multiple files.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -189,28 +357,77 @@ grant {
...
+    permission org.osgi.framework.BundlePermission "*", "provide, require, host, fragment";
+    permission org.osgi.framework.CapabilityPermission "*", "require, provide";
+    permission org.osgi.framework.ServicePermission "*", "get, register";
+
+    permission java.lang.RuntimePermission "accessDeclaredMembers";
+    permission java.lang.reflect.ReflectPermission "suppressAccessChecks";
+
+    permission org.osgi.service.event.TopicPermission "*", "publish, subscribe";
+
+    permission java.lang.RuntimePermission "setContextClassLoader";
```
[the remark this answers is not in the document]
stustison: Haven't so far!

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
stustison: You don't have access to the maven coordinates from the bundle.

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
stustison: I think it would be a little more work than this because the policy objects would also need to support the osgi stuff. I can make a ticket to look into this.

platform/osgi/platform-osgi-condpermadmin/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>osgi</artifactId>
+        <groupId>ddf.platform</groupId>
+        <version>2.12.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>bundle</packaging>
+    <artifactId>platform-osgi-condpermadmin</artifactId>
```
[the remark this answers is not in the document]
stustison: This is literally what OSGi names it so I copied it.

catalog/transformer/catalog-transformer-xml/src/main/java/ddf/catalog/transformer/xml/XmlResponseQueueTransformer.java
```diff
@@ -242,7 +251,6 @@ public BinaryContent transform(SourceResponse response, Map<String, Serializable
             fjp.invoke(
                 new MetacardForkTask(
                     ImmutableList.copyOf(response.getResults()),
-                    fjp,
```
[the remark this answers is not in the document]
coyotesqrl: This still looks wrong to me.

distribution/ddf-common/src/main/resources-filtered/etc/custom.properties
```diff
@@ -65,6 +65,8 @@ felix.fileinstall.enableConfigSave=false
 org.osgi.framework.system.packages.extra = \
     org.apache.karaf.branding, \
     sun.misc, \
+    javax.crypto \
```
coyotesqrl: ❗️ Missing commas

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +31,340 @@ deny {
...
+}
+
+grant codeBase "file:/org.apache.karaf.jaas.modules" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keys.properties", "read";
+}
+
+grant codeBase "file:/pax-logging-log4j2" {
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}-", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}data${/}log${/}ingest_error.log", "read,write";
```
coyotesqrl: ✏️ Should be able to remove the specific log file entries here now that `log/-` is covered.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +31,340 @@ deny {
...
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}org.ops4j.pax.web.cfg", "read";
+}
+
+grant codeBase "file:/org.apache.cxf.cxf-rt-transports-http/org.ops4j.pax.web.pax-web-runtime/org.ops4j.pax.web.pax-web-jetty/org.ops4j.pax.web.pax-web-extender-whiteboard/org.ops4j.pax.web.pax-web-extender-war/org.eclipse.jetty.util/org.eclipse.jetty.server/org.eclipse.jetty.servlet/org.eclipse.jetty.security/org.eclipse.jetty.io/javax.servlet-api/org.apache.shiro.core/security-filter-web-sso/security-filter-login/security-filter-authorization/platform-filter-response/platform-filter-clientinfo/platform-filter-delegate/security-interceptor-guest-wrapper/security-interceptor-guest/security-handler-api" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}metadata", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}metadata${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
+    permission java.io.FilePermission "etc${/}keystores${/}-", "read";
```
coyotesqrl: ❗️ Missing root of `etc`.

stustison: It needs to match exactly this permission or it doesn't work. There are perms for both with and without the root.

distribution/ddf-common/src/main/resources/security/default.policy
stustison: Going to leave for now since I had weird behavior with this bundle.

distribution/ddf-common/src/main/resources/security/default.policy
coyotesqrl: Might want to leave a comment to that effect. Oy. :rofl:

platform/osgi/platform-osgi-conditions/pom.xml
stustison: There will be a follow-on ticket.

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
```diff
@@ -35,13 +36,29 @@ public UrlResourceReaderConfigurator(ConfigurationAdmin configAdmin) {
     this.configAdmin = configAdmin;
   }
 
+  @SuppressWarnings("squid:S2142")
```
paouelle: ✏️  Add a comment as to why we are suppressing

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
```diff
@@ -35,13 +36,29 @@ public UrlResourceReaderConfigurator(ConfigurationAdmin configAdmin) {
...
   public void setUrlResourceReaderRootDirs(String... rootResourceDirs) throws IOException {
     Configuration configuration = configAdmin.getConfiguration(PID, null);
     Dictionary<String, Object> properties = new DictionaryMap<>();
     Set<String> rootResourceDirectories =
         ImmutableSet.<String>builder().add(rootResourceDirs).build();
     properties.put("rootResourceDirectories", rootResourceDirectories);
     configuration.update(properties);
+    for (int i = 0; i < 5; i++) {
```
paouelle: ❓ Why `5`?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
```diff
@@ -35,13 +36,29 @@ public UrlResourceReaderConfigurator(ConfigurationAdmin configAdmin) {
...
+      if (updatedConfig
+          .getProperties()
+          .get("rootResourceDirectories")
+          .equals(rootResourceDirectories)) {
+        break;
+      } else {
+        try {
+          TimeUnit.SECONDS.sleep(5);
+        } catch (InterruptedException e) {
+          // ignore
```
paouelle: ❗️ Interruptions should not be simply ignored. We need to remember that an interruption occurred from within and propagate it as part of a try/finally outside of the loop.

For example: ```try {
  for (...) {
    try {
      TimeUnits.SECONDS.sleep(5);
     interrupted = true;
    }
  }
} finally { 
  if (interrupted) {
    Thread.currentThread().interrupt();
  }
}```

platform/osgi/platform-osgi-conditions/pom.xml
paouelle: Not a fan of increasing tech debt or creating code without test cases. We typically do not want to do that.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -0,0 +1,131 @@
...
+   *     invoked multiple times in the same permission check, the same Dictionary will be passed
+   *     multiple times. The SecurityManager treats this Dictionary as an opaque object and simply
+   *     creates an empty dictionary and passes it to subsequent invocations if multiple invocations
+   *     are needed.
+   * @return {@code true} if all the Condition objects are satisfied. Otherwise, {@code false} if
+   *     one of the Condition objects is not satisfied.
+   */
+  @Override
+  public boolean isSatisfied(Condition[] conditions, Dictionary<Object, Object> context) {
+    for (Condition condition : conditions) {
```
paouelle: :pencil: could be rewritten as `return Stream.of(conditions).allMatch(Condition::isSatisfied)`

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/PrincipalCondition.java
```diff
@@ -0,0 +1,117 @@
...
+ * Permission Info. The Permissions of a ConditionalPermission Info can only be used if the
+ * associated Conditions are satisfied.
+ *
+ * <p>Selects bundles based on the identity of the principal executing the call. @ThreadSafe
+ */
+public class PrincipalCondition implements Condition {
+
+  private String[] args;
+
+  @SuppressWarnings("squid:S1172")
```
paouelle: ❗️ Please add a comment as to why we are suppressing

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/PrincipalCondition.java
```diff
@@ -0,0 +1,117 @@
...
+  @Override
+  public boolean isSatisfied() {
+    javax.security.auth.Subject subject =
+        javax.security.auth.Subject.getSubject(AccessController.getContext());
+    return subject
+        .getPrincipals()
+        .stream()
+        .map(Principal::getName)
+        .collect(Collectors.toList())
+        .containsAll(Arrays.asList(args));
```
paouelle: ❗️ Avoid creating a new list each time `containsAll()` is called

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/PrincipalCondition.java
```diff
@@ -0,0 +1,117 @@
...
+   *     invoked multiple times in the same permission check, the same Dictionary will be passed
+   *     multiple times. The SecurityManager treats this Dictionary as an opaque object and simply
+   *     creates an empty dictionary and passes it to subsequent invocations if multiple invocations
+   *     are needed.
+   * @return {@code true} if all the Condition objects are satisfied. Otherwise, {@code false} if
+   *     one of the Condition objects is not satisfied.
+   */
+  @Override
+  public boolean isSatisfied(Condition[] conditions, Dictionary<Object, Object> context) {
+    for (Condition condition : conditions) {
```
paouelle: :pencil2: could be rewritten as `return Stream.of(conditions).allMatch(Condition::isSatisfied)`

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
```diff
@@ -0,0 +1,198 @@
...
+   *     invoked multiple times in the same permission check, the same Dictionary will be passed
+   *     multiple times. The SecurityManager treats this Dictionary as an opaque object and simply
+   *     creates an empty dictionary and passes it to subsequent invocations if multiple invocations
+   *     are needed.
+   * @return {@code true} if all the Condition objects are satisfied. Otherwise, {@code false} if
+   *     one of the Condition objects is not satisfied.
+   */
+  @Override
+  public boolean isSatisfied(Condition[] conditions, Dictionary<Object, Object> context) {
+    for (Condition condition : conditions) {
```
paouelle: :pencil2: could be rewritten as `return Stream.of(conditions).allMatch(Condition::isSatisfied)`

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
```diff
@@ -0,0 +1,198 @@
...
+   * @param context A Dictionary object that implementors can use to track state. If this method is
+   *     invoked multiple times in the same permission check, the same Dictionary will be passed
+   *     multiple times. The SecurityManager treats this Dictionary as an opaque object and simply
+   *     creates an empty dictionary and passes it to subsequent invocations if multiple invocations
+   *     are needed.
+   * @return {@code true} if all the Condition objects are satisfied. Otherwise, {@code false} if
+   *     one of the Condition objects is not satisfied.
+   */
+  @Override
+  public boolean isSatisfied(Condition[] conditions, Dictionary<Object, Object> context) {
```
paouelle: Identical implementation in multiple subclass, that would be a good candidate to be defined as a default interface method

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+    return identities;
+  }
+
+  private void collectAltNames(Collection<List<?>> altNames, List<String> identities) {
+    for (List item : altNames) {
+      Integer type = (Integer) item.get(0);
+      if (type == 0 || type == 2) {
+        try {
+          if (item.toArray()[1] instanceof byte[]) {
```
paouelle: ✏️  Can we avoid calling `toArray[]` multiple times on `item`?

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
```diff
@@ -0,0 +1,198 @@
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private static String getExtendedCertAttribute(
+      X500Principal principal, ASN1ObjectIdentifier identifier) {
+    RDN[] rdNs = new X500Name(principal.getName()).getRDNs(identifier);
+    if (rdNs != null && rdNs.length > 0) {
```
paouelle: ✏️  Can be replaced with `if (ArrayUtils.isNotEmpty(rDNs)`

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,205 @@
...
+      if (parsedPrincipal.hasAlias()) {
+        principals.add(parsedPrincipal.getAlias());
+      } else {
+        principals.add(
+            parsedPrincipal.getPrincipalClass() + "/" + parsedPrincipal.getPrincipalName());
+      }
+    }
+
+    if (!principals.isEmpty()) {
+      conditionInfos.add(new ConditionInfo(PRINCIPAL_CONDITION, principals.toArray(new String[0])));
```
paouelle: ✏️  More optional to do `new String[principals.size()]` to avoid creating an array for nothing

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,205 @@
...
+                parsedPermission.getActions());
+      }
+      List<ConditionInfo> conditionInfos = new ArrayList<>();
+      addCodebase(parsedPolicyEntry, conditionInfos);
+      addSignedBy(parsedPolicyEntry, conditionInfos);
+      addPrincipals(parsedPolicyEntry, conditionInfos);
+      infos.add(
+          conditionalPermissionAdmin.newConditionalPermissionInfo(
+              null,
+              (conditionInfos.isEmpty()) ? null : conditionInfos.toArray(new ConditionInfo[0]),
```
paouelle: ✏️ More optional to do new ConditionInfo[conditionInfos.size()] to avoid creating an array for nothing

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,205 @@
...
+    ConditionalPermissionUpdate conditionalPermissionUpdate =
+        conditionalPermissionAdmin.newConditionalPermissionUpdate();
+    Priority priorityResult = null;
+    List<ConditionalPermissionInfo> allGrantInfos = new ArrayList<>();
+    List<ConditionalPermissionInfo> allDenyInfos = new ArrayList<>();
+    for (ParsedPolicy parsedPolicy : parsedPolicies) {
+      List<ParsedPolicyEntry> grantEntries = parsedPolicy.getGrantEntries();
+      List<ParsedPolicyEntry> denyEntries = parsedPolicy.getDenyEntries();
+      List<ConditionalPermissionInfo> grantInfos = new ArrayList<>();
+      List<ConditionalPermissionInfo> denyInfos = new ArrayList<>();
```
paouelle: ❓ Why create temporary lists just to add all entries computed in them back to the `allXXXInfos` list after? Why not just pass the `allXXXInfos` list directly?

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  private void buildConditionalPermissionInfo(
+      ConditionalPermissionAdmin conditionalPermissionAdmin,
+      List<ParsedPolicyEntry> entries,
+      List<ConditionalPermissionInfo> infos,
+      String type) {
+    for (ParsedPolicyEntry parsedPolicyEntry : entries) {
+      List<ParsedPermission> permissions = parsedPolicyEntry.getPermissions();
+      PermissionInfo[] permissionInfos = new PermissionInfo[permissions.size()];
```
paouelle: ✏️  Can be replaced with ```PermissionInfo[] permissionInfos = permissions.stream()
  .map(p -> new PermissionInfo(
        p.getPermissionType(),
       replaceSystemProperties(parsedPermission.getPermissionName()),
       parsedPermission.getActions()))
  .toArray(PermissionInfo[]::new);

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,205 @@
...
+
+    ConditionalPermissionAdmin conditionalPermissionAdmin = permAdminTracker.getService();
+    String policyDir = SecurityActions.getSystemProperty("ddf.home") + File.separator + "security";
+    if (policyDir.startsWith("=")) {
+      policyDir = policyDir.substring(1);
+    }
+    File policyDirFile = new File(policyDir);
+    List<ParsedPolicy> parsedPolicies = new ArrayList<>();
+    for (File file : Objects.requireNonNull(policyDirFile.listFiles())) {
+      parsedPolicies.add(new Parser(false).parse(new FileReader(new File(file.getAbsolutePath()))));
```
paouelle: ❓ What is the point in doing `new File(file.getAbsolutePath())`?
❗️ The FileReader should be properly closed or why not use `Parser.parse(File)`?

platform/osgi/platform-osgi-configadmin/pom.xml
paouelle: ❗️ There is a lot of new code and not one unit test case for it 😞

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -0,0 +1,131 @@
...
+   *     Condition is not satisfied.
+   */
+  @Override
+  public boolean isSatisfied() {
+    if (args.length == 0) {
+      return false;
+    }
+    String bundleName = bundle.getHeaders().get("Bundle-SymbolicName");
+    String key = bundleName + args[0];
+    Boolean storedResult = DECISION_MAP.get(key);
```
paouelle: ✏️  Could be replaced with ```return DECISION_MAP.computeIfAbsent(key, k ->
  Stream.of(REGEX.split(args[0])).anyMatch(bundleName::contains));```

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
```diff
@@ -35,13 +36,29 @@ public UrlResourceReaderConfigurator(ConfigurationAdmin configAdmin) {
...
+  @SuppressWarnings("squid:S2142")
   public void setUrlResourceReaderRootDirs(String... rootResourceDirs) throws IOException {
     Configuration configuration = configAdmin.getConfiguration(PID, null);
     Dictionary<String, Object> properties = new DictionaryMap<>();
     Set<String> rootResourceDirectories =
         ImmutableSet.<String>builder().add(rootResourceDirs).build();
     properties.put("rootResourceDirectories", rootResourceDirectories);
     configuration.update(properties);
+    for (int i = 0; i < 5; i++) {
+      Configuration updatedConfig = configAdmin.getConfiguration(PID, null);
```
paouelle: ❓ Do we really want to create the config object if it doesn't exist already? `getConfiguration()` wil do that.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
```diff
@@ -35,13 +36,29 @@ public UrlResourceReaderConfigurator(ConfigurationAdmin configAdmin) {
...
+      if (updatedConfig
+          .getProperties()
```
paouelle: ❗️ `getProperties()  can be returned as `null`

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
```diff
@@ -35,13 +36,29 @@ public UrlResourceReaderConfigurator(ConfigurationAdmin configAdmin) {
...
+          .get("rootResourceDirectories")
```
paouelle: ❓ What if the `rootResourceDirectories` is not defined in the config object?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/KarafConsole.java
```diff
@@ -83,7 +83,12 @@ public String runCommand(String command, Principal... principals) {
    * @param command command to execute. Cannot be {@code null}.
    * @return command output. Can be empty but not {@code null}.
    */
+  @SuppressWarnings("squid:S00112")
```
paouelle: ❗️ Add a comment to explain why you are suppressing

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
```diff
@@ -867,6 +874,95 @@ public String getOption() {
...
+      return new HomePermVmOption(option);
+    }
+
+    private HomePermVmOption(String option) {
+      super(option);
+    }
+
+    @Override
+    @SuppressWarnings({
+      "squid:S00112" /* A generic RuntimeException is perfectly reasonable in this case. */
```
paouelle: ❓ Is it? What about using `AssertionError` instead?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
```diff
@@ -867,6 +874,95 @@ public String getOption() {
...
+      return new HomePolicyVmOption(option);
+    }
+
+    private HomePolicyVmOption(String option) {
+      super(option);
+    }
+
+    @Override
+    @SuppressWarnings({
+      "squid:S00112" /* A generic RuntimeException is perfectly reasonable in this case. */
```
paouelle: ❓ Is it? What about using AssertionError instead?

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
```diff
@@ -867,6 +874,95 @@ public String getOption() {
             .map(Path::toAbsolutePath)
             .map(Path::toString)
             .map(s -> StringUtils.replace(super.getOption(), "{karaf.home}", s))
+            .map(s -> s.replace('\\', '/'))
+            .map(s -> s.replace("/bin/..", "/"))
+            .orElseGet(super::getOption);
+      } catch (IOException e) {
+        throw new RuntimeException("Unable to determine current exam directory", e);
```
paouelle: ❓ What about using AssertionError instead?

distribution/ddf-common/src/main/resources/bin/setenv
paouelle: Was this re-tested?

stustison:
build now

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +30,345 @@ deny {
...
+// DDF Platform Migratable
 //
 grant
-  codeBase "file:${ddf.home.policy}system/ddf/platform/platform-migratable/-" {
+  codeBase "file:/platform-migratable" {
     // some permissions might be repeated here to ensure that they are granted to the migration
     // framework if we ever tighten the permissions above
 
-    permission javax.management.MBeanServerPermission "createMBeanServer";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}pdp${/}-", "read, delete";
```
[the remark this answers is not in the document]
stustison: Sometimes 3rd party code directly checks for these permissions whether they use them or not.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +30,345 @@ deny {
...
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}-", "read, delete";
```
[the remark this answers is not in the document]
stustison: Yes.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +30,345 @@ deny {
...
+grant codeBase "file:/persistence-core-impl/persistence-core-listeners/org.apache.aries.blueprint.core/org.apache.aries.blueprint.cm/org.ops4j.pax.web.pax-web-spi/catalog-ui-search/standard/catalog-solr-provider/catalog-solr-solrclient" {
+    permission java.lang.RuntimePermission "createClassLoader";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}-", "read";
+}
+
+grant codeBase "file:/jetty-server/catalog-core-directorymonitor/catalog-security-filter/spatial-geocoding-plugin/spatial-geocoding-geocoder/spatial-geocoding-offline-catalog/registry-identification-plugin/registry-federation-admin-service-impl" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}system.properties", "read,write";
```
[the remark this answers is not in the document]
stustison: Still need the perm.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +30,345 @@ deny {
...
+}
+
+grant codeBase "file:/org.apache.karaf.shell.ssh/org.apache.sshd.core" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}host.key", "read,write";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}branding-ssh.properties", "read";
```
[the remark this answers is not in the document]
stustison: They are created on the fly.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +30,345 @@ deny {
...
+grant codeBase "file:/security-sts-server/security-sts-realm/catalog-opensearch-source/catalog-opensearch-endpoint" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}issuer${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
+    permission java.lang.RuntimePermission "createClassLoader";
+}
+
+grant codeBase "file:/catalog-rest-endpoint/catalog-transformer-resource/ftp/org.apache.tika.core/org.codice.thirdparty.tika-bundle/tika-input-transformer/video-input-transformer" {
+    permission java.lang.RuntimePermission "createClassLoader";
+    permission java.io.FilePermission "<<ALL FILES>>", "execute";
```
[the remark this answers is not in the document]
stustison: They directly check for this permission.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
```diff
@@ -0,0 +1,198 @@
...
+import org.osgi.service.condpermadmin.Condition;
+import org.osgi.service.condpermadmin.ConditionInfo;
+
+/**
+ * The interface implemented by a Condition. Conditions are bound to Permissions using Conditional
+ * Permission Info. The Permissions of a ConditionalPermission Info can only be used if the
+ * associated Conditions are satisfied.
+ *
+ * <p>Selects bundles based on the identity of the signer. @ThreadSafe
+ */
```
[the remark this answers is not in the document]
stustison: Like what? I'm not sure what else I'd add to be honest.

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
```diff
@@ -0,0 +1,205 @@
...
+
+    if (priorityResult == Priority.GRANT) {
+      conditionalPermissionUpdate.getConditionalPermissionInfos().addAll(allGrantInfos);
+      conditionalPermissionUpdate.getConditionalPermissionInfos().addAll(allDenyInfos);
+      conditionalPermissionUpdate
+          .getConditionalPermissionInfos()
+          .add(getAllPermission(conditionalPermissionAdmin, ConditionalPermissionInfo.ALLOW));
+    } else if (priorityResult == Priority.DENY) {
+      conditionalPermissionUpdate.getConditionalPermissionInfos().addAll(allDenyInfos);
+      conditionalPermissionUpdate.getConditionalPermissionInfos().addAll(allGrantInfos);
```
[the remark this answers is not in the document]
stustison: Yes.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
stustison: Going to leave Streaming out of this stuff since traditional loops are usually faster.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
stustison: No access to those libraries at this level.

platform/osgi/platform-osgi-condpermadmin/src/main/java/org/codice/ddf/condpermadmin/PermissionActivator.java
stustison: Just made it easier to debug. I can get rid of it now.

platform/osgi/platform-osgi-configadmin/src/main/resources/OSGI-INF/permissions.perm
```diff
@@ -61,4 +61,29 @@
...
+(java.util.PropertyPermission "org.ops4j.pax.url.mvn.localRepository" "read")
+(java.util.PropertyPermission "felix.fileinstall.log.level" "read")
+(java.util.PropertyPermission "org.apache.felix.eventadmin.AsyncToSyncThreadRatio" "read")
+(java.util.PropertyPermission "user.home" "read")
+(java.util.PropertyPermission "felix.fileinstall.log.level" "read")
+(java.lang.RuntimePermission "getClassLoader")
+
+(java.util.PropertyPermission "org.apache.felix.eventadmin.Timeout" "read")
+(java.util.PropertyPermission "org.ops4j.pax.url.mvn.settings" "read")
+(java.lang.RuntimePermission "setContextClassLoader")
```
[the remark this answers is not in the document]
stustison: Nope.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/SignerCondition.java
stustison: I don't have any control over the interface. I didn't really want to make a parent class because I don't even know if this implementation is correct or correct for all of them because the spec is completely ambiguous on what to do with this. Just going to leave it for now until we figure out what to do with it or what the right approach is.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
stustison: Then it isn't a config for this anyways.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
stustison: It can but it won't for this.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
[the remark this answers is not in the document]
stustison: Was seeing a race condition and this check fixed it.

stustison: Seemed like a nice round number and this is test code so I don't really care.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
stustison: Yes.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/config/UrlResourceReaderConfigurator.java
stustison: I was ignoring it because this is test code and I don't really care about them.

distribution/test/itests/test-itests-common/src/main/java/org/codice/ddf/itests/common/AbstractIntegrationTest.java
stustison: This is all test code. I don't care what exception is used as long as you're able to see the error.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -31,153 +30,345 @@ deny {
...
+grant codeBase "file:/admin-configuration-configupdater" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}-", "read";
+}
 
+grant codeBase "file:/catalog-core-standardframework" {
     permission java.lang.RuntimePermission "createClassLoader";
-    permission java.util.PropertyPermission "javax.net.ssl.*", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
+    permission java.io.FilePermission "<<ALL FILES>>", "execute";
```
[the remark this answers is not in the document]
stustison: Same answer as above.

distribution/ddf-common/src/main/resources/bin/setenv.bat
stustison: I'll add a comment because the IDE wants to remove the space and I don't want to not be able to edit this file in the IDE.

stustison:
build now

stustison:
build now

platform/osgi/platform-osgi-conditions/pom.xml
lessarderic: Are you planning to work on the tests next @stustison? Not just putting a ticket in the backlog and hope it gets done at some point?

stustison: I am planning to do tests immediately after this, yes. I may also need to work on making the itests less flakey for that next round as well to help with stability.

stustison:
build now

bdeining:
do we plan on addressing the SonarQube findings in this PR?

stustison:
I'll fix or suppress any issues that I created, but I'm not going to fix the already existing stuff. It is all in test code anyways that I really don't care about.

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

stustison:
build now

PR #2886, opened 2018-01-25
DDF-3515 Adds list capability to workspaces

andrewkfiedler:
#### What does this PR do?
 - Adds list capability to workspaces.  The list is composed of a title,
 an optional cql filter for limiting acceptance to the list, an icon,
 and a list of bookmarks (which are the content of the list).
 - Users can create a list from the Lists tab or from any result.
 - At the moment, the filter is only enforced upon entrance to the list.
 So if a result is already in the list and a new filter is applied that
it wouldn't be able to pass, it will remain in the list.  In the future
we could put in some logic to inform the user when this happens.
 - Users can assign an icon to lists.
 - At the moment, the number of lists the user can create per workspace
 is not limited.

#### Who is reviewing it? 
@bdeining 
@millerw8
@mackncheesiest 
@adimka 

Play around with the list functionality and try to break it or find pain points.

[DDF-3515](https://codice.atlassian.net/browse/DDF-3515)


millerw8:
So, I think this is gonna be able to meet the goals intended and provide some nice general capability.  I do think in the effort to reduce clicks, it might be nice if the administrator could provide default CQL filters to the name/icons.  Even if the user could go view and change it afterwards, and ootb even if nothing has a filter... being able to add a default one at the administrative level will definite offer a reduced-click experience.

millerw8:
This seems to be broken in firefox.  Worked well in Chrome, but I cannot create a new list from either mechanism in firefox.

andrewkfiedler:
@millerw8 thanks, I'll check it out in Firefox and IE as well.  

I was actually wondering if Administrators would want to set up default lists that are included in each workspace, that way users don't have to create them at all unless they're doing something custom.

millerw8:
I would be good with the administrator controlling the lists or not.  Limiting the list types would actually help me show this off better, but does limit the capability some.

bdeining:
screen shot 2018-01-30 at 12 41 05 pm
screen shot 2018-01-30 at 12 41 15 pm

bdeining:
screen shot 2018-01-30 at 12 47 22 pm

bdeining:
screen shot 2018-01-30 at 12 57 17 pm
❗️ in chrome, the option to create a new list appears in an odd location and is inconsistent with the view in chrome

catalog/ui/catalog-ui-search/src/main/webapp/component/result-add/result-add.view.js
```diff
@@ -0,0 +1,93 @@
...
+ **/
+/*global define, require, module*/
+var Marionette = require('marionette');
+var $ = require('jquery');
+var template = require('./result-add.hbs');
+var CustomElements = require('js/CustomElements');
+var store = require('js/store');
+var ListCreateView = require('component/list-create/list-create.view');
+var lightboxInstance = require('component/lightbox/lightbox.view.instance');
+var List = require('js/model/list');
```
millerw8: should be js/model/List (capital L).... breaks on RHEL deployments otherwise

andrewkfiedler: Good catch.

bdeining:
build now

bdeining:
sucessfully rehero'd; lgtm

catalog/ui/catalog-ui-search/src/main/webapp/component/editor/editor.hbs
```diff
@@ -30,15 +30,15 @@
         <span>
             Edit
         </span>
-    </button>
-    <button class="editor-cancel is-negative">
+    </button><!--
```
Schachte: Remove empty HTML comment

catalog/ui/catalog-ui-search/src/main/webapp/component/editor/editor.hbs
```diff
@@ -30,15 +30,15 @@
...
+    --><button class="editor-cancel is-negative">
          <span class="fa fa-times">
          </span>
         <span>
             Cancel
         </span>
-    </button>
-    <button class="editor-save is-positive">
+    </button><!--
+    --><button class="editor-save is-positive">
```
Schachte: Remove empty HTML comment

catalog/ui/catalog-ui-search/src/main/webapp/component/ingest-details/ingest-details.hbs
```diff
@@ -27,23 +27,23 @@
         <span>
             Clear
         </span>
-    </button>
-    <button class="footer-start is-positive">
+    </button><!--
+    --><button class="footer-start is-positive">
```
Schachte: Remove empty HTML comment

adimka: These actually serve a purpose. The handlebars template inserts whitespace if you put the button on a new line and don't do this.

Schachte: @adimka Ah, cool, good to know.

bdeining:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/workspace/ListMetacardTypeImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public static final String LIST_METACARD_TYPE_NAME = "metacard.list";
+
+  public static final String LIST_CQL = "list.cql";
+
+  public static final String LIST_ICON = "list.icon";
+
+  public static final String LIST_BOOKMARKS = "list.bookmarks";
+
+  private static final Set<AttributeDescriptor> DESCRIPTORS;
```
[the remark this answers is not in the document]
bdeining: We should address this finding

catalog/ui/catalog-ui-search/src/main/webapp/component/list-editor/list-editor.view.js
```diff
@@ -0,0 +1,155 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+/*global define, require, module*/
+var Marionette = require('marionette');
+var _ = require('underscore');
```
mackncheesiest: Should we try to standardize whether `_` is `lodash` or `underscore` moving forward in new files?

catalog/ui/catalog-ui-search/src/main/webapp/component/list-editor/list-editor.view.js
```diff
@@ -0,0 +1,155 @@
...
+var $ = require('jquery');
+var template = require('./list-editor.hbs');
+var CustomElements = require('js/CustomElements');
+require('behaviors/button.behavior');
```
mackncheesiest: Should we be assigning this to a var?

catalog/ui/catalog-ui-search/src/main/webapp/js/filter.js
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+module.exports = {
+    matchesFilters: matchesFilters,
+    matchesCql: function(metacardJSON, cqlString) {
+        if (cqlString === "") {
+            return true;
+        }
+        return this.matchesFilters(metacardJSON, cql.read(cqlString));
```
Variadicism: The body of this function can be simplified to `return cqlString === "" || this.matchesFilters(metacardJSON, cql.read(cqlString));`.

catalog/ui/catalog-ui-search/src/main/webapp/component/workspace-lists/workspace-lists.hbs
```diff
@@ -0,0 +1,31 @@
...
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+ --}}
+<div align="center" class="lists-empty">
+    <div class="is-header">
+        You don't have any lists.  <button class="quick-search is-primary in-text">Search</button> 
+        for something and add it to a list or create a 
+        <div class="quick-create is-primary in-text composed-button">
```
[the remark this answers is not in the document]
mackncheesiest: It's because that div is populated with a `PopoutView.createSimpleDropdown` that needs to be passed the text to display as a parameter
https://github.com/codice/ddf/pull/2886/files/eb4c5fd615daf78bdd4ed2ddf87daae7974e9804#diff-1cdf0ea194aec512d1275398e6544c45R57

catalog/ui/catalog-ui-search/src/main/webapp/js/model/List.js
```diff
@@ -0,0 +1,118 @@
...
+
+function getRelevantIcon(iconName) {
+    return iconMap[iconName];
+}
+
+function generateCql(bookmarks) {
+    if (bookmarks.length === 0) {
+        return '';
+    }
+    return cql.write({
```
glenhein: Is there a way to do a geospatial search for a list? I'm thinking about the target list. If it contains several metacards, I would want to do a search with those location fields OR'd together.

catalog/ui/catalog-ui-search/src/main/webapp/component/list-item/list-item.view.js
```diff
@@ -0,0 +1,124 @@
...
+ *
+ **/
+/*global define, require, module*/
+var Marionette = require('marionette');
+var _ = require('underscore');
+var _merge = require('lodash/merge');
+var $ = require('jquery');
+var template = require('./list-item.hbs');
+var CustomElements = require('js/CustomElements');
+require('behaviors/button.behavior');
```
glenhein: no var?

catalog/ui/catalog-ui-search/src/main/webapp/js/model/List.js
glenhein: What if instead of the javascript building the CQL, you called into the Java code, which could then see if any CqlBuilder instances are registered for the list type?

glenhein:
Hero: I built and installed the PR. The lists seem to work as expected. As far as usability, at one point I accidently added a filter to the list, but the filter was empty, causing the system to refuse to add any other metacards to the list. It was a bit confusing until I figured out the problem. 

Also, as I mentioned in a PR comment, it doesn't look like we can perform a geospatial query based on an entire target list. These seems like a serious limitation.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/workspace/WorkspaceTransformer.java
```diff
@@ -80,6 +80,17 @@ private void setupMetacardMappers() {
                   .map(this::toMetacardXml)
                   .collect(Collectors.toList());
             }));
+    metacardToJsonEntryMapper.put(
+        WorkspaceAttributes.WORKSPACE_LISTS,
+        remapValue(
+            value -> {
+              List<Map<String, Object>> content = (List) value;
```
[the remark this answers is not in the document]
andrewkfiedler: I don't think we have to worry, we do something similar for casting the query list.  I could update both to account for possible casting issues though.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/workspace/WorkspaceTransformer.java
```diff
@@ -115,6 +126,18 @@ private void setupJsonMappers() {
                   .map(this::transform)
                   .collect(Collectors.toList());
             }));
+    jsonToMetacardEntryMapper.put(
+        WorkspaceAttributes.WORKSPACE_LISTS,
+        remapValue(
+            value -> {
+              List<String> lists = (List) value;
```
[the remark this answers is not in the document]
andrewkfiedler: I don't think we have to worry, we do something similar for casting the query list.  I could update both to account for possible casting issues though.

catalog/ui/catalog-ui-search/src/main/webapp/component/dropdown/popout/dropdown.popout.view.js
```diff
@@ -0,0 +1,25 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+ * even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details. A copy of the GNU Lesser General Public License
+ * is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+/*global define*/
+var Marionette = require('marionette');
+var _ = require('underscore');
```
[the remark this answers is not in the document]
andrewkfiedler: Nope, I must have forgotten to drop this.  It won't really bloat the UI at all in this case since underscore is required very early on in the UI tree.  But if it was a library that no other component used, it definitely would cause it to be packaged and delivered too early in webpack.

catalog/ui/catalog-ui-search/src/main/webapp/component/ingest-details/ingest-details.hbs
andrewkfiedler: Basically exactly what Adam said, there are other ways to get around this but this is the best I've found.  A good portion of this PR I went through and removed the other way of doing it (search for whitespaceFix).

catalog/ui/catalog-ui-search/src/main/webapp/component/list-editor/list-editor.view.js
andrewkfiedler: I think there should be a tech debt ticket to go through and try and move us completely to lodash, but I think we've run into issues removing underscore in the past.  Lodash is more maintained and developed, and also supports per method packages which helps out with reducing UI size.

catalog/ui/catalog-ui-search/src/main/webapp/component/list-editor/list-editor.view.js
andrewkfiedler: The behaviors are a bit strange, where the act of requiring them is what causes a side effect.  They're like backbone associations (they don't get assigned) in that they're mixed into the Marionette framework.  All this does is make sure behaviors have been mixed in so that the current view can use them.

catalog/ui/catalog-ui-search/src/main/webapp/component/list-editor/list-editor.view.js
```diff
@@ -0,0 +1,155 @@
...
+    });
+  },
+  edit: function() {
+    this.$el.addClass('is-editing');
+    this.regionManager.forEach(function(region) {
+      if (region.currentView && region.currentView.turnOnEditing) {
+        region.currentView.turnOnEditing();
+      }
+    });
+    var tabable = _.filter(this.$el.find('[tabindex], input, button'), function(
```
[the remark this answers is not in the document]
andrewkfiedler: Ah, it's a misspelling of tabbable (https://en.wiktionary.org/wiki/tabbable).

catalog/ui/catalog-ui-search/src/main/webapp/component/list-interactions/list-interactions.view.js
```diff
@@ -0,0 +1,67 @@
...
+    handleDelete: function(){
+        this.model.collection.remove(this.model);
+    },
+    handleDuplicate: function(){
+        var copyAttributes = JSON.parse(JSON.stringify(this.model.attributes));
+        delete copyAttributes.id;
+        delete copyAttributes.query;
+        var newList = new this.model.constructor(copyAttributes);
+        this.model.collection.add(newList);
+    },
```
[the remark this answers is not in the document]
andrewkfiedler: For the views I typically prefix event based methods with either `handle` or `trigger`.  

Lately I've been leaning towards exclusively using `trigger` for events that originate from users, that way we can start using`handle` as a prefix for events that originate from the model.  So I'll probably go back and update this to use the `trigger` prefix, except for the `handleResult` method which will keep the `handle` prefix.

catalog/ui/catalog-ui-search/src/main/webapp/component/list-item/list-item.view.js
andrewkfiedler: The behaviors are a bit strange, where the act of requiring them is what causes a side effect. They're like backbone associations (they don't get assigned) in that they're mixed into the Marionette framework. All this does is make sure behaviors have been mixed in so that the current view can use them.

catalog/ui/catalog-ui-search/src/main/webapp/component/result-add/result-add.less
```diff
@@ -0,0 +1,31 @@
...
+        text-align: left;
+        height: @minimumButtonSize;
+        line-height: @minimumButtonSize;
+
+        span:first-of-type {
+            margin-right: @mediumSpacing;
+        }
+    }
+
+    .is-existing-list.fails-filter::after {
```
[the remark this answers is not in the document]
andrewkfiedler: Here's a good explanation of it: https://stackoverflow.com/questions/17684797/should-i-use-single-colons-or-double-colons-for-before-after-first-le

The short of it is to stick with `::`.

catalog/ui/catalog-ui-search/src/main/webapp/component/tabs/workspace-content/tabs-workspace-content.js
```diff
@@ -16,16 +16,18 @@ define([
...
+], function (_, Tabs, workspaceExploreView, workspaceSavedView, resultsView, WorkspaceSearchView, WorkspaceListsView) {
     
     var WorkspaceContentTabs = Tabs.extend({
         defaults: {
             tabs: {
                 'Search': WorkspaceSearchView,
            //     'Searches': workspaceExploreView,
              //   'Results': resultsView,
-                'Bookmarks': workspaceSavedView
+              //  'Bookmarks': workspaceSavedView,
```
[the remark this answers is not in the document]
andrewkfiedler: At this point we can remove them.  They're older components for workspaces.  There are probably a few components we can delete as well.

catalog/ui/catalog-ui-search/src/main/webapp/decorator/menu-navigation.decorator.js
```diff
@@ -13,52 +13,82 @@
...
+        return expandedItems;
+    } else {
+        return expandComposedMenus(expandedItems);
+    }
+};
+
+const handleArrowKey = (componentView, up) => {
+    let menuItems = componentView.handleTabIndexes();
+    let currentActive = menuItems.filter(element => $(element).hasClass('is-active'))[0];
+    var potentialNext = menuItems[menuItems.indexOf(currentActive) + (up === true ? -1 : 1)];
```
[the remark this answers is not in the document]
andrewkfiedler: It's typically best to stick with `===` in javascript in order to avoid unintended side effects from type coersion, things like empty strings or empty arrays or even the number 0 turning out to be false.

catalog/ui/catalog-ui-search/src/main/webapp/js/filter.js
andrewkfiedler: While entirely possible, I think it makes the code less readable.

catalog/ui/catalog-ui-search/src/main/webapp/js/model/List.js
andrewkfiedler: I think that'd be a great idea for extending list interactions in the future.  Then we could have a lot of different options for doing searches based off a list.

catalog/ui/catalog-ui-search/src/main/webapp/component/workspace-lists/workspace-lists.view.js
```diff
@@ -0,0 +1,137 @@
...
+var template = require('./workspace-lists.hbs');
+var store = require('js/store');
+var ListSelectorView = require('component/dropdown/list-select/dropdown.list-select.view');
+var DropdownModel = require('component/dropdown/dropdown');
+var ResultSelectorView = require('component/result-selector/result-selector.view');
+var $ = require('jquery');
+var ListCreateView = require('component/list-create/list-create.view');
+var PopoutView = require('component/dropdown/popout/dropdown.popout.view');
+
+let selectedListId;
```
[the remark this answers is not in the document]
rzwiefel: But if we never transition then we'll be stuck in the past forever.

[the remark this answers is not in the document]
rzwiefel: With that same logic we should not be using Arrow functions either since most of the existing codebase uses old style logic.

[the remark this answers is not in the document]
andrewkfiedler: We should look at transitioning to `let` and `const` though.  No reason to not do so in new files like this one.

https://stackoverflow.com/questions/40058571/is-there-a-reason-not-to-replace-var-with-let-in-javascript

There are some sticky points with doing a massive find / replace unfortunately, like defining a `var` inside of an `if` statement.  I don't think we are doing that a whole lot, if at all.  It would be worth trying to do a massive find / replace and seeing if things just work.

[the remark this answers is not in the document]
Variadicism: I would very much like to use more ES6; I just suggested changing that `let` since the decision seemed to be to use `var` generally for consistency and I wasn't sure where we stood on ES6 compatibility. Using `let` and `const` everywhere would be even better, though. _Down with `var`!_

If we have full ES6 capabilities, then at the very least, all the new code in this pull request should use `let` or `const` rather than `var`. Dealing with preexisting ES5 code with `var` is something we should discuss... preferably sooner rather than later.

catalog/ui/catalog-ui-search/src/main/webapp/decorator/menu-navigation.decorator.js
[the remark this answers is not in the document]
Variadicism: I get that, but this `=== true` is not stopping type coercion; it's just making it more confusing. If an empty string or 0 or other falsy values are passed in here, they will still be treated as false because there is no check for `=== false`. Type coercion is still affecting this statement; all that's changed with `=== true` is that truthy values besides `true` will now be treated like `false`, which is unintuitive.

bdeining:
build now

jlcsmith:
Small change in the latest commit on the "add-to-list" capability:
capture

andrewkfiedler:
build now

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/workspace/ListMetacardTypeImpl.java
```diff
@@ -0,0 +1,94 @@
...
+  public static final String LIST_BOOKMARKS = "list.bookmarks";
+
+  private static final Set<AttributeDescriptor> LIST_DESCRIPTORS;
+
+  static {
+    LIST_DESCRIPTORS = new HashSet<>();
+
+    LIST_DESCRIPTORS.add(
+        new AttributeDescriptorImpl(
+            Metacard.ID,
```
rzwiefel: We could mix these in with the Core types instead of having to redefine them, though this should still work I believe (via the constructor on metacard type https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/data/impl/MetacardTypeImpl.java#L112-L116)

PR #2887, opened 2018-01-25
DDF-3566: Update Intrigue to use CQL query syntax with the CQL endpoint

garrettfreibott:
#### What does this PR do?
Intrigue now exclusively uses the characters `* ? \` to represent wildcard, single character wildcard, and escape respectively. These characters are translated to their CQL-equivalent syntax before querying the CQL endpoint and workspaces are saved with their queries in CQL format. Any user-inputted CQL special characters are escaped before sending. If a user wants to utilize these CQL characters, they must the `* ? \` syntax mentioned above.

#### Who is reviewing it? 
@djblue 
@vinamartin 
@rzwiefel

@codice/ui 
@codice/io 

@andrewkfiedler
@pklinef 

Install DDF, create a workspace with a query that uses CQL and Intrigue special characters, and ensure that when the query is run and the workspace is created, all outgoing requests have the following character transformations:
`* `=> `%`
`?` => `_`
`%` =>`\%`
`_` => `\_`
And that the queries displayed to the user within Intrigue use the original-inputted search text and not the CQL-translated versions. Also make sure that other searches behave normally.

Previously, Intrigue would send a CQL query to the CQL endpoint and put the user-inputted search string into the query verbatim. This caused issues when the CQL endpoint tried to incorrectly interpret underscores and percentages. The solution decided on was to exclusively enforce the use of the already-commonly used `* _ \` on the frontend so it can be translated accurately to and from CQL syntax.

[DDF-3566](https://codice.atlassian.net/browse/DDF-3566)

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/cql.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/cql.js
@@ -187,6 +187,28 @@ define([
         return results;
     }
 
+    // Mapping of Intrigue's query language syntax to CQL syntax
+    const userqlToCql = {
+        '*': '%',
+        '?': '_',
+        '%': '\\%',
+        '_': '\\_'
+    };
+
+    const translateUserqlToCql =
+        (str) => str.replace( /([^*?%_])?([*?%_])/g,
+            (_, a = '', b) => a + (a === '\\' ? b : userqlToCql[b]));
+
+    //Mapping of CQL syntax to Intrigue's query language syntax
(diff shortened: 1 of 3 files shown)


garrettfreibott:
build now

garrettfreibott:
build now

garrettfreibott:
build now

garrettfreibott:
build now

rzwiefel:
Heroed successfully

PR #2901, opened 2018-01-30
DDF-3538 Resolve cognitive complexity finding in Application Config Installer

bakejeyner:
#### What does this PR do?
Resolves cognitive complexity finding in ApplicationConfigInstaller

#### Who is reviewing it? 
@Lambeaux 
@blen-desta 

@coyotesqrl 
@shaundmorris

[3538](https://codice.atlassian.net/browse/3538)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
+++ b/platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
@@ -14,22 +14,17 @@
 package org.codice.ddf.admin.application.service.impl;
 
 import ddf.security.Subject;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.EnumSet;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.concurrent.Callable;
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.karaf.features.FeaturesService;
 import org.codice.ddf.admin.application.service.ApplicationService;
 import org.codice.ddf.admin.application.service.ApplicationServiceException;
+import org.codice.ddf.platform.util.properties.PropertiesLoader;
 import org.codice.ddf.security.common.Security;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -77,64 +72,56 @@ public class ApplicationConfigInstaller extends Thread {
 
   @Override
   public void run() {
+    // use PropertiesLoader to load properties, returns empty properties if there is an issue
+    Properties props = PropertiesLoader.getInstance().loadProperties(fileName, null);
 
-    File configFile = new File(fileName);
-    if (!configFile.exists()) {
-      LOGGER.debug("No config file located, cannot load from it.");
+    if (props.isEmpty()) {
+      LOGGER.debug("No applications were found in the configuration file.");
       return;
     }
-    InputStream is = null;
+
+    LOGGER.trace("Found applications to install from config.");
+    for (Entry<Object, Object> curApp : props.entrySet()) {
+      String appName = (String) curApp.getKey();
+      String appLocation = (String) curApp.getValue();
+      LOGGER.debug("Starting app {} at location: {}", appName, appLocation);
+
+      startApp(appName, appLocation);
+    }
+    LOGGER.trace("Finished installing applications, uninstalling installer module...");
+
+    installAndUninstallFeatures();
+  }
+
+  private void startApp(String appName, String appLocation) {
     try {
-      is = new FileInputStream(configFile);
-      Properties props = new Properties();
-      props.load(is);
-      if (!props.isEmpty()) {
-        LOGGER.debug("Found applications to install from config.");
-        for (Entry<Object, Object> curApp : props.entrySet()) {
-          String appName = curApp.getKey().toString();
-          String appLocation = curApp.getValue().toString();
-          LOGGER.debug("Starting app {} at location: {}", appName, appLocation);
-
-          try {
-            if (StringUtils.isNotEmpty(appLocation)) {
-              appService.addApplication(new URI(appLocation));
-            }
-            executeAsSystem(
-                () -> {
-                  appService.startApplication(appName);
-                  return true;
-                });
-          } catch (ApplicationServiceException ase) {
-            LOGGER.warn("Could not start " + appName, ase);
-          } catch (URISyntaxException use) {
-            LOGGER.warn(
-                "Could not install application, location is not a valid URI " + appLocation, use);
-          }
-        }
-        LOGGER.debug("Finished installing applications, uninstalling installer module...");
-
-        try {
-          if (!StringUtils.isBlank(postInstallFeatureStart)) {
-            featuresService.installFeature(
-                postInstallFeatureStart, EnumSet.of(FeaturesService.Option.NoAutoRefreshBundles));
-          }
-          if (!StringUtils.isBlank(postInstallFeatureStop)) {
-            featuresService.uninstallFeature(postInstallFeatureStop);
-          }
-        } catch (Exception e) {
-          LOGGER.debug("Error while trying to run the post-install start and stop operations.", e);
-        }
-
-      } else {
-        LOGGER.debug("No applications were found in the configuration file.");
+      if (!appLocation.isEmpty()) {
+        appService.addApplication(new URI(appLocation));
       }
-      is.close();
-    } catch (FileNotFoundException fnfe) {
-      LOGGER.warn("Could not file the configuration file at " + configFile.getAbsolutePath(), fnfe);
-    } catch (IOException ioe) {
-      LOGGER.warn("Could not load file as property list.", ioe);
-    } finally {
-      IOUtils.closeQuietly(is);
+      executeAsSystem(
+          () -> {
+            appService.startApplication(appName);
+            return true;
+          });
+    } catch (ApplicationServiceException ase) {
+      LOGGER.warn("Could not start {}", appName, ase);
+    } catch (URISyntaxException use) {
+      LOGGER.warn(
+          "Could not install application, location is not a valid URI {}", appLocation, use);
+    }
+  }
+
+  private void installAndUninstallFeatures() {
+    try {
+      if (StringUtils.isNotBlank(postInstallFeatureStart)) {
+        featuresService.installFeature(
+            postInstallFeatureStart, EnumSet.of(FeaturesService.Option.NoAutoRefreshBundles));
+      }
+      if (StringUtils.isNotBlank(postInstallFeatureStop)) {
+        featuresService.uninstallFeature(postInstallFeatureStop);
+      }
+    } catch (Exception e) {
+      LOGGER.debug("Error while trying to run the post-install start and stop operations.", e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -83,58 +82,63 @@ public void run() {
       LOGGER.debug("No config file located, cannot load from it.");
       return;
     }
-    InputStream is = null;
-    try {
-      is = new FileInputStream(configFile);
+
+    try (InputStream is = new FileInputStream(configFile)) {
```
paouelle: ❗️ We shouldn't be using try-with-resources on input stream as it will throw an exception if it fails to close the stream at the end but at that point, we already got everything we needed from it. This is why the original code was using `IOUtils.closeQuietly()`.  Although, just putting it in a finaly clause should be good enough.

This is a good candidate for a suppression.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -83,58 +82,63 @@ public void run() {
...
-      } else {
+
+      if (props.isEmpty()) {
         LOGGER.debug("No applications were found in the configuration file.");
+        return;
+      }
+
+      LOGGER.debug("Found applications to install from config.");
+      for (Entry<Object, Object> curApp : props.entrySet()) {
+        String appName = curApp.getKey().toString();
```
paouelle: ✏️  Technically calling `toString() ` won't hurt here but it should also be safe to blindly typecast to String since we are dealing with a Properties object where both keys and values are strings by design.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -83,58 +82,63 @@ public void run() {
...
     } catch (IOException ioe) {
       LOGGER.warn("Could not load file as property list.", ioe);
-    } finally {
-      IOUtils.closeQuietly(is);
+    }
+  }
+
+  private void startApp(String appName, String appLocation) {
+    try {
+      if (StringUtils.isNotEmpty(appLocation)) {
```
paouelle: ❗️ I don't believe `appLocation` can be `null` here as if it was we would have gotten a NullPointerException above so calling `appLocation.isEmpty()` should be good enough.

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -83,58 +82,63 @@ public void run() {
...
+    } catch (ApplicationServiceException ase) {
+      LOGGER.warn("Could not start " + appName, ase);
+    } catch (URISyntaxException use) {
+      LOGGER.warn("Could not install application, location is not a valid URI " + appLocation, use);
+    }
+  }
+
+  private void installAndUninstallFeatures() {
+    try {
+      if (!StringUtils.isBlank(postInstallFeatureStart)) {
```
paouelle: ✏️  `isNotBlank()` instead

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -83,58 +82,63 @@ public void run() {
...
+        featuresService.installFeature(
+            postInstallFeatureStart, EnumSet.of(FeaturesService.Option.NoAutoRefreshBundles));
+      }
+      if (!StringUtils.isBlank(postInstallFeatureStop)) {
```
paouelle: ✏️  `isNotBlank()` instead

bakejeyner:
build now

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -76,59 +76,39 @@ public ApplicationConfigInstaller(
   }
 
   @Override
+  @SuppressWarnings("squid:S2093" /* Intentionally closing InputStream in finally clause */)
```
clockard: Why are we closing this in a finally? Looks like we could just wrap only the properties section in a try with resource and then just return on error and avoid this suppression.

bakejeyner: After reviewing this with @paouelle we decided the finally block is the way to go.

1. It is an input stream so we don't care about if it cannot close or not
2. If everything works well in the `try with resources` and the `InputStream` has an error closing, an `IOException` is thrown and a misleading logger is logged: `LOGGER.warn("Could not load file as property list.", ioe);`

clockard: OK. I would still recommend only wrapping the creation of the properties in the try/catch/finally

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -138,6 +118,37 @@ public void run() {
...
+  private void installAndUninstallFeatures() {
+    try {
+      if (StringUtils.isNotBlank(postInstallFeatureStart)) {
+        featuresService.installFeature(
+            postInstallFeatureStart, EnumSet.of(FeaturesService.Option.NoAutoRefreshBundles));
+      }
+      if (StringUtils.isNotBlank(postInstallFeatureStop)) {
+        featuresService.uninstallFeature(postInstallFeatureStop);
+      }
+    } catch (Exception e) {
```
clockard: Is there a more specific exception we can catch here?

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
clockard: Better yet replace all of the property loading stuff with `PropertiesLoader.getInstance().loadProperties(filename)`

paouelle: Doesn't that one do a lot more then just loading properties? It also performs system variable substitution

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
paouelle: Unfortunately, not with Karaf. They declare they throw Exception 😞

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
clockard: I don't think that should be a problem with this properties file. It does provide a bit more functionality though. All in all I think it is a positive change.

paouelle: Then I am good with it too.

bakejeyner:
build now

bakejeyner:
build now

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -77,64 +72,55 @@ public ApplicationConfigInstaller(
...
+      if (!appLocation.isEmpty()) {
+        appService.addApplication(new URI(appLocation));
+      }
+      executeAsSystem(
+          () -> {
+            appService.startApplication(appName);
+            return true;
+          });
+    } catch (ApplicationServiceException ase) {
+      LOGGER.warn("Could not start " + appName, ase);
```
bdeining: Should be parameterized

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -77,64 +72,55 @@ public ApplicationConfigInstaller(
...
+    } catch (URISyntaxException use) {
+      LOGGER.warn("Could not install application, location is not a valid URI " + appLocation, use);
```
bdeining: Should be parameterized

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -77,64 +72,55 @@ public ApplicationConfigInstaller(
...
+
+    LOGGER.debug("Found applications to install from config.");
+    for (Entry<Object, Object> curApp : props.entrySet()) {
+      String appName = (String) curApp.getKey();
+      String appLocation = (String) curApp.getValue();
+      LOGGER.debug("Starting app {} at location: {}", appName, appLocation);
+
+      startApp(appName, appLocation);
+    }
+    LOGGER.debug("Finished installing applications, uninstalling installer module...");
```
bdeining: This might be more appropriate at the `trace` level

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -77,64 +72,55 @@ public ApplicationConfigInstaller(
...
-    File configFile = new File(fileName);
-    if (!configFile.exists()) {
-      LOGGER.debug("No config file located, cannot load from it.");
+    if (props.isEmpty()) {
+      LOGGER.debug("No applications were found in the configuration file.");
       return;
     }
-    InputStream is = null;
+
+    LOGGER.debug("Found applications to install from config.");
```
bdeining: This might be more appropriate at the `trace` level.

bakejeyner:
build now

bakejeyner:
build now

bakejeyner:
build now
(yarn error)

idperez:
hero success

platform/admin/core/admin-core-appservice/src/main/java/org/codice/ddf/admin/application/service/impl/ApplicationConfigInstaller.java
```diff
@@ -77,64 +72,56 @@ public ApplicationConfigInstaller(
...
+      executeAsSystem(
+          () -> {
+            appService.startApplication(appName);
+            return true;
+          });
+    } catch (ApplicationServiceException ase) {
+      LOGGER.warn("Could not start {}", appName, ase);
+    } catch (URISyntaxException use) {
+      LOGGER.warn(
+          "Could not install application, location is not a valid URI {}", appLocation, use);
```
shaundmorris: ✏️  looks like a typo here `ase` instead of `use`

bakejeyner: I believe it was an acronym for the exception: `catch (ApplicationServiceException ase)`

PR #2909, opened 2018-01-31
DDF-3582 Adds auto-redirect / workspace creation

andrewkfiedler:
#### What does this PR do?
 - Updates the UI to auto-redirect to a new workspace if a User comes in to the start page (no hash -> /search/catalog/) without any existing workspaces.
 - Updates the UI to auto-redirect to a workspace if the User comes in to the start page (no hash -> /search/catalog/) and only has one workspace.
 - No redirects / creates will happen if there is a hash (#workspaces, #upload, etc.).

#### Who is reviewing it? 
@bdeining
@jlcsmith
@millerw8 
@adimka 
@mackncheesiest

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationStart.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationStart.js
@@ -26,17 +26,34 @@ require([
 
     var workspaces = store.get('workspaces');
 
+    function hasEmptyHashAndNoWorkspaces() {
+        return workspaces.length === 0 && location.hash === "";
+    }
+
+    function checkForEmptyHashAndOneWorkspace() {
+        if (location.hash === "" && workspaces.fetched && workspaces.length === 1) {
+            location.hash = '#workspaces/'+workspaces.first().id;
+        }
+    }
+
     function attemptToStart() {
-        if (workspaces.fetched && user.fetched){
+        checkForEmptyHashAndOneWorkspace();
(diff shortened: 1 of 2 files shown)


andrewkfiedler:
It'll work fine, this only modifies the hash.  `location.search` would update the param.

millerw8:
So the scary scenario here is one person one creates a public workspace... all new users automatically get redirected.  Is it possible to make it so users only get redirected to a single workspace if it they own none or one?

andrewkfiedler:
I had talked about that with @mackncheesiest, @Variadicism , and @adimka .  I agree that it should be user owned workspaces only, but wanted to get some more on board before I went that route.

catalog/ui/catalog-ui-search/src/main/webapp/js/ApplicationStart.js
```diff
@@ -26,17 +26,38 @@ require([
 
     var workspaces = store.get('workspaces');
 
+    function getWorkspacesOwnedByUser() {
+        return workspaces.filter(workspace => user.isGuest() ? workspace.get('localStorage') === true : workspace.get('metacard.owner') === user.get('user').get('email'));
```
rzwiefel: ✏️ just a little nit you can ignore if you don't care, but this line is quite long and could very well benefit from that bountiful vertical screen real estate that is so easily scrollable. It might even, perhaps, add just a small token of readability to this line of code.

bdeining:
successfully hero'd, looks good

bdeining:
build now

PR #2918, opened 2018-02-01
DDF-3590 Update jackson libraries to 2.9.4

bdeining:
#### What does this PR do?
Upgrates the jackson libraries to 2.9.4 Port to master
#### Who is reviewing it? 
@emmberk @peterhuffer 
@jlcsmith 
@millerw8 

CI Build passes

[DDF-3590](https://codice.atlassian.net/browse/DDF-3590)


bdeining:
build now

platform/solr/platform-solr-server-standalone/dependency-check-maven-config.xml
```diff
@@ -16,7 +16,16 @@
    ]]></notes>
         <cve>CVE-2016-3720</cve>
     </suppress>
-
+    <suppress>
```
shaundmorris: ❓ Does the dependency-check-maven-config.xml suppression not cover this module?

bdeining: I don't believe it does; @clockard might know the answer better than I

bdeining:
build now

catalog/ui/search-ui/search-endpoint/pom.xml
```diff
@@ -184,6 +184,27 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson2-version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jdk8</artifactId>
+            <version>${jackson2-version}</version>
+            <scope>compile</scope>
```
paouelle: ❓ Why do we need to specify `compile` as the scope? Isn't it the default one with maven?

platform/solr/platform-solr-server-standalone/dependency-check-maven-config.xml
clockard: This module supplies its own so we generally need to put a copy of the suppression here as well. Just looked into this a little more and I am wondering if this is going to be a separate issue. We are shading in solr-core in solr-dependencies which include jackson 2.5.4 and 1.9.13

clockard: Looks like we don't actually shade in databind only core and dataformat-smile so this shouldn't be an issue. I think that cve applies to all versions up through 2.8.10

PR #2921, opened 2018-02-02
[2.11.x]DDF-3600 Use all registered attribute desciptors to build filters in the CSW Endpoint

brjeter:
#### What does this PR do?
- Includes all registered attribute descriptors to determine the datatype of filter literal expressions. Without this, some expressions were being left as String instead of their actual data type and the query fails in Solr.
- No longer passing in the CSW metacard type into the `CswQueryFilterTransformer` and `CswRecordMapperFilterVisitor` since we are using all metacard types instead (already passed in).
#### Who is reviewing it? 
@emmberk @emanns95 @blen-desta 
@bdeining 
@clockard 
- Setup a loopback CSW Federation Profile source
- Query the CSW source by several different types of attributes and verify they work as expected (specifically query by `effective` and `datetime.start`, which were broken before this change). Also verify there are no exceptions in the logs similar to this
```
ddf.catalog.source.UnsupportedQueryException: propertyIsLessThan(String,String) not supported by org.opengis.filter.Filter Delegate.
```
The `CswRecordMapperFilterVisitor` uses the CSW metacard type's attribute descriptors to determine a property's data type in some places. In others, **all** registered attribute descriptors are used. As a result some filter literals are left as a String instead of being mapped to their proper data type. This primarily affects any Date attributes that are not in the CSW metacard type.
[DDF-3600](https://codice.atlassian.net/browse/DDF-3600)
#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
+++ b/catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
@@ -84,15 +84,11 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
 
   private final Map<String, AttributeType> attributeTypes;
 
-  private final MetacardType metacardType;
-
   private final CswRecordMap cswRecordMap;
 
   private Filter visitedFilter;
 
-  public CswRecordMapperFilterVisitor(
-      CswRecordMap cswRecordMap, MetacardType metacardType, List<MetacardType> metacardTypes) {
-    this.metacardType = metacardType;
+  public CswRecordMapperFilterVisitor(CswRecordMap cswRecordMap, List<MetacardType> metacardTypes) {
     this.cswRecordMap = cswRecordMap;
 
     attributeTypes = new HashMap<>();
@@ -242,10 +238,9 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
       name = cswRecordMap.getProperty(propertyName, namespaceContext);
 
       if (SPATIAL_QUERY_TAG.equals(extraData)) {
-        AttributeDescriptor attrDesc = metacardType.getAttributeDescriptor(name);
-        if (attrDesc != null
-            && !AttributeType.AttributeFormat.GEOMETRY.equals(
-                attrDesc.getType().getAttributeFormat())) {
+        AttributeType attributeType = attributeTypes.get(name);
+        if (attributeType != null
+            && !AttributeType.AttributeFormat.GEOMETRY.equals(attributeType.getAttributeFormat())) {
           throw new UnsupportedOperationException(
               "Attempted a spatial query on a non-geometry-valued attribute ("
                   + propertyName
@@ -262,17 +257,8 @@ public class CswRecordMapperFilterVisitor extends DuplicatingFilterVisitor {
   public Object visit(PropertyIsBetween filter, Object extraData) {
     Expression expr = visit(filter.getExpression(), extraData);
 
-    AttributeType type =
-        attributeTypes.get(((PropertyName) filter.getExpression()).getPropertyName());
-
-    LiteralExpressionImpl typedLowerExpression = (LiteralExpressionImpl) filter.getLowerBoundary();
-    setExpressionType(type, typedLowerExpression);
-
-    LiteralExpressionImpl typedUpperExpression = (LiteralExpressionImpl) filter.getUpperBoundary();
-    setExpressionType(type, typedUpperExpression);
-
-    Expression lower = visit((Expression) typedLowerExpression, expr);
-    Expression upper = visit((Expression) typedUpperExpression, expr);
+    Expression lower = visit(filter.getLowerBoundary(), expr);
+    Expression upper = visit(filter.getUpperBoundary(), expr);
     return getFactory(extraData).between(expr, lower, upper);
   }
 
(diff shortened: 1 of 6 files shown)


catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
```diff
@@ -452,34 +406,9 @@ public Object visit(PropertyIsLessThanOrEqualTo filter, Object extraData) {
...
-      } else if (type.getBinding() == Long.class) {
-        typedExpression.setValue(Long.valueOf((String) typedExpression.getValue()));
-      } else if (type.getBinding() == Float.class) {
-        typedExpression.setValue(Float.valueOf((String) typedExpression.getValue()));
-      } else if (type.getBinding() == Double.class) {
-        typedExpression.setValue(Double.valueOf((String) typedExpression.getValue()));
-      } else if (type.getBinding() == Boolean.class) {
-        typedExpression.setValue(Boolean.valueOf((String) typedExpression.getValue()));
-      }
-    }
```
brjeter: It's not overly clear from just looking at the code, but every time we visit a LiteralExpression we end up on line 424 which does the exact same data conversion at this method that I've removed.

catalog/spatial/csw/spatial-csw-endpoint/src/main/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/transformer/CswRecordMapperFilterVisitor.java
```diff
@@ -262,17 +257,8 @@ public Object visit(PropertyName expression, Object extraData) {
...
     Expression expr = visit(filter.getExpression(), extraData);
 
-    AttributeType type =
-        attributeTypes.get(((PropertyName) filter.getExpression()).getPropertyName());
-
-    LiteralExpressionImpl typedLowerExpression = (LiteralExpressionImpl) filter.getLowerBoundary();
-    setExpressionType(type, typedLowerExpression);
-
-    LiteralExpressionImpl typedUpperExpression = (LiteralExpressionImpl) filter.getUpperBoundary();
-    setExpressionType(type, typedUpperExpression);
```
brjeter: Several of these similar code blocks were removed because the data type is now correctly determined when visiting the literal expression initially in the `visit(Literal expression, Object extraData)` method.

clockard:
I think the attributeTypes in CswRecordMapperFilterVisitor could get out of date. Would recommend using the AttributeRegistry service instead.

brjeter:
I added a new commit to use the Attribute Registry.

catalog/spatial/csw/spatial-csw-endpoint/src/test/java/org/codice/ddf/spatial/ogc/csw/catalog/endpoint/CswQueryFactoryTest.java
```diff
@@ -254,19 +253,17 @@ public void setUp()
 
     queryFactory = new CswQueryFactory(cswRecordMap, filterBuilder, filterAdapter);
 
-    AttributeRegistry mockAttributeRegistry = mock(AttributeRegistry.class);
-    when(mockAttributeRegistry.lookup(TITLE_TEST_ATTRIBUTE))
-        .thenReturn(Optional.of(mock(AttributeDescriptor.class)));
-    queryFactory.setAttributeRegistry(mockAttributeRegistry);
+    AttributeRegistryImpl attributeRegistry = new AttributeRegistryImpl();
+    attributeRegistry.registerMetacardType(getCswMetacardType());
```
emmberk: Why is this getCswMetacardType() method still there? I thought you removed the CswMetacardType

brjeter: These unit tests are testing against those particular attributes in the `CswMetacardType`. Initially, I was going to mock the `AttributeRegistry`, but it was going to be a pain since I would need when statements for many attributes. I thought it made more sense to make a real `AttributeRegistry` and initialize it with the `CswMetacardType` that this test is already setting up and asserting against.

brjeter: This test probably shouldn't be making up its own version of the `CswMetacardType` to assert against, but I thought that was out of the scope of this PR. Honestly I think we could have a whole epic for cleaning up the CSW code because it's a mess in a lot of ways.

bdeining: I agree with this.  We've added new dynamics like the QueryFilterTransformer that will render a bunch of code obsolete.  Ideally we would move GMD to that dynamic anyway.

emmberk: Ok. Might be worth leaving a comment or renaming that method, even though it's just in the tests, to not say CswMetacardType because that's confusing since there isn't a CswMetacardType at all.

blen-desta:
build now

blen-desta:
Manually verified that dates are not String but Date objects (Instant). 👍

PR #2971, opened 2018-02-14
DDF-3515 Updates List search federation to always be enterprise

andrewkfiedler:
#### What does this PR do?
 - If we don't do this, it would be possible for a list to be created using defaults that might restrict its sources.

#### Who is reviewing it? 
@rzwiefel 
@pklinef
@mackncheesiest 
@adimka 

Create a list after setting your defaults to something that doesn't include a particular source.  Verify that you can add a result from that source to your list and still retrieve the result when looking at the list.

[3515](https://codice.atlassian.net/browse/DDF-3515)


andrewkfiedler:
I'm going to close this in favor of a larger ticket to address more areas similar to this.

PR #2972, opened 2018-02-15
DDF-3629 Updates instantiations of new Searches to be more specific

andrewkfiedler:
#### What does this PR do?
 - Previously these areas would get the defaults for the unspecified attributes.  This could possibly cause issues with attributes like federation being left blank.  Whether or not we should do so for the "create search from geo" is hard to determine.
 - It might be worth exploring doing this in alternative ways.  We could go back and make it so user defaults need to be explicitly mixed in rather than having them as part of the Query defaults for all new queries.  Or we could make a method to apply user defaults and use that where we need it.  Or we could have two different query models, with one that extends the other and simply mixes in the user defaults.  There are a lot of options.

#### Who is reviewing it? 
@djblue 
@rzwiefel 
@birda 
@pklinef

[3629](https://codice.atlassian.net/browse/DDF-3629)

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
@@ -81,7 +81,8 @@ module.exports = Backbone.Collection.extend({
         }
         var queryForWorkspace = new Query.Model({
             title: title,
-            cql: cqlQuery
+            cql: cqlQuery,
+            type: 'text'
         });
         this.create({
             title: title,
@@ -95,7 +96,8 @@ module.exports = Backbone.Collection.extend({
             title: 'Example Local',
             federation: 'local',
             excludeUnnecessaryAttributes: false,
-            cql: "anyText ILIKE '%'"
+            cql: "anyText ILIKE '%'",
+            type: 'basic'
         });
         this.create({
             title: 'Template Local',
@@ -109,7 +111,8 @@ module.exports = Backbone.Collection.extend({
             title: 'Example Federated',
             federation: 'enterprise',
             excludeUnnecessaryAttributes: false,
-            cql: "anyText ILIKE '%'"
+            cql: "anyText ILIKE '%'",
+            type: 'basic'
         });
         this.create({
             title: 'Template Federated',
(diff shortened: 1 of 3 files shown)


andrewkfiedler:
I mentioned some alternatives in the commit message (What does this PR do) above.  I'm interested to know what people would lean towards for this, as it isn't the only way we can go about handling it.

catalog/ui/catalog-ui-search/src/main/webapp/component/metacard-interactions/metacard-interactions.view.js
```diff
@@ -146,7 +146,10 @@ define([
             }, '');
             let cqlString = "(" + combinedCqlString + ")";
             var newQuery = new Query.Model({
-                type: locations.length > 1 ? 'advanced' : 'basic'
+                type: locations.length > 1 ? 'advanced' : 'basic',
```
andrewkfiedler: This is the one case where I'm not sure if we should actually be using the user's defaults for queries instead.

andrewkfiedler: This is one case where I'm not sure if we should actually be using the user's defaults for queries instead.

catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
```diff
@@ -81,7 +81,8 @@ module.exports = Backbone.Collection.extend({
         }
         var queryForWorkspace = new Query.Model({
             title: title,
-            cql: cqlQuery
+            cql: cqlQuery,
```
andrewkfiedler: Similarly, should we be using the user's defaults for searches here?

catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
```diff
@@ -122,7 +125,8 @@ module.exports = Backbone.Collection.extend({
         var queryForWorkspace = new Query.Model({
             title: 'Example Location',
```
andrewkfiedler: Similarly, should we be using the user's defaults for searches here?

catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
```diff
@@ -135,7 +139,8 @@ module.exports = Backbone.Collection.extend({
         var queryForWorkspace = new Query.Model({
             title: 'Example Temporal',
```
andrewkfiedler: Similarly, should we be using the user's defaults for searches here?

catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
mackncheesiest: Hmm...maybe we should be using user defaults here. Will that end up impacting the underlying list queries negatively?

catalog/ui/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
mackncheesiest: I'm not sure that we need user defaults in the example searches, though. I'd expect those to be static across all users regardless of their search settings.

PR #3009, opened 2018-02-26
DDF-3506 Turn off unnecessary owasp checks and reduce log level

mcalcote:
#### What does this PR do?
Update the Jenkinsfile to reduce unnecessary owasp checks according to the Maven Dependency Check Plugin https://jeremylong.github.io/DependencyCheck/dependency-check-maven/configuration.html

Also set logging level to FATAL to reduce noise.

#### Who is reviewing it? 
@oconnormi @mackncheesiest 

@codice/build 

@clockard @shaundmorris

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -828,7 +828,7 @@
                     <plugin>
                         <groupId>org.owasp</groupId>
                         <artifactId>dependency-check-maven</artifactId>
-                        <version>1.4.5</version>
+                        <version>3.1.1</version>
                         <dependencies>
                             <dependency>
                                 <groupId>ddf.support</groupId>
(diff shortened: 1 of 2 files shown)


mcalcote:
build now

mackncheesiest:
Why are we muting all of the logs during the incremental and full builds?

pom.xml
```diff
@@ -828,7 +828,7 @@
                     <plugin>
                         <groupId>org.owasp</groupId>
                         <artifactId>dependency-check-maven</artifactId>
-                        <version>1.4.5</version>
+                        <version>3.1.1</version>
```
mackncheesiest: Now since we're upgrading/modifying the plugin anyway, we might as well disable the analyzers in the plugin config so that we don't have a maintenance issue between pipeline jobs and legacy jobs

i.e. [see alliance's config](https://github.com/codice/alliance/blob/d5fa4f3e7be250ea51cf39d1eafc0224d9b60c06/pom.xml#L671-L684)

mcalcote:
This changeset is somewhat experimental and intended to start this conversation. 
The advantage of muting those specific steps is that the job may be quicker and reduce a large amount of unused output. 
Great suggestion on the config of the dependency-plugin, will update as we move forward with these changes.

mcalcote:
This work has been picked up a under another ticket. Discussion here will be applied to that work.

PR #3060, opened 2018-03-14
DDF-3612 Adds access to export actions from outside inspector

andrewkfiedler:
#### What does this PR do?
 - Adds access to export actions from outside the inspector.  They can now be accessed from a dropdown in the actions for result items.
 - Updates actions view to be more user friendly by separating export actions from map actions and removing common words (pushing them to the header "Export as").
 - Moves navigation decorator to be done through a behavior.  This will be safer than the decoration route since decorating runs the risk of overwriting view properties on the views they're applied to.  It also makes listening to lifecycle events and attaching events simpler.
 - Updates theming to handle a wider range of colors.  In particular, now when button colors are primarily white, the text will flip to contrast properly.  So users can use buttons with lighter colors now.
 - The theming for links has been updated to provide a better experience as well by not relying solely on color to differentiate them.  Since users can choose their own background color, it was increasingly complicated to try and provide a contrasting link color.  Now links use the contrast color for the background (white or black), and are constantly underlined and bold to differentiate them from surrounding text.  Also, on hover their backgrounds will now highlight to further differentiate them.
 - Fixes issue where the datepicker was using the wrong color for its background.
 - Fixes issue with workspace items where the click handler was attached to the wrong element.  This caused it to be impossible to select by tabbing to the workspace and pressing enter.
 - Adds sorting to actions so they come in a more predictable order.

#### Who is reviewing it? 
@adimka 
@mackncheesiest 
@Variadicism 
@jlcsmith

Verify that export actions are 

Users wanted access to actions outside the inspector.

[3612](https://codice.atlassian.net/browse/DDF-3612)

screen shot 2018-03-14 at 12 26 13 pm
screen shot 2018-03-14 at 12 26 07 pm

#### Review Comment Legend:

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/component/metacard-actions/metacard-actions.view.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/component/metacard-actions/metacard-actions.view.js
@@ -15,7 +15,7 @@
 /*global define, setTimeout*/
 define([
     'marionette',
-    'underscore',
+    'lodash',
     'jquery',
     './metacard-actions.hbs',
     'js/CustomElements',
@@ -27,7 +27,6 @@ define([
         setDefaultModel: function(){
             this.model = this.selectionInterface.getSelectedResults().first();
         },
-        className: 'is-list',
         template: template,
         tagName: CustomElements.register('metacard-actions'),
         regions: {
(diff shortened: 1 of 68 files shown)


catalog/ui/catalog-ui-search/src/main/webapp/component/metacard-actions/metacard-actions.view.js
```diff
@@ -44,6 +43,18 @@ define([
                 this.setDefaultModel();
             }
         },
+        serializeData: function() {
+            return {
+                exportActions: _.sortBy(this.model.getExportActions().map(action => ({
+                    url: action.get('url'),
+                    title: action.get('title').replace('Export as', '').replace('Export','')
```
Variadicism: This replacement should only remove `"Export"` or `"Export as"` from the beginning of titles and should also ensure that whitespace after the `"Export"` or `"Export as"` is removed so as to not leave leading white space on the resulting text. On top of that, if `"as"` begins an export format, e.g., "Export asterisks file", then the `"as"` will be dropped accidentally. I suggest using this instead:
```
title: action.get('title').replace(/^Export( as)?\s+\b/, '')
```

andrewkfiedler:
build now

PR #3073, opened 2018-03-20
DDF-3297 added unit tests for security manager init code and conditions

stustison:
#### Who is reviewing it? 
@SmithJosh @tbatie 
@lessarderic
@stustison
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


stustison:
build now

stustison:
build now

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    BundleNameCondition bundleNameCondition =
+        new BundleNameCondition(
+            bundle,
+            new ConditionInfo(
+                BundleNameCondition.class.getName(), new String[] {"bundle1/bundle2/bundle3"}));
+    boolean satisfied = bundleNameCondition.isSatisfied();
+    // no cache this time around
+    assertThat(satisfied, is(true));
+    satisfied = bundleNameCondition.isSatisfied();
+    // cached
```
lessarderic: ❓ Does this prove that the condition result was cached, or just that the class returns consistent results? Should we consider adding a package-private getter to get the cache `Map`, or a way to inject it, so we can actually verify that the entry was properly cached?
Same question for all methods.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    headers.put("Bundle-SymbolicName", "bundle4");
+    when(bundle.getHeaders()).thenReturn(headers);
+    BundleNameCondition bundleNameCondition =
+        new BundleNameCondition(
+            bundle, new ConditionInfo(BundleNameCondition.class.getName(), new String[0]));
+    boolean satisfied = bundleNameCondition.isSatisfied();
+    // no cache this time around
+    assertThat(satisfied, is(false));
+    satisfied = bundleNameCondition.isSatisfied();
+    // cached
```
lessarderic: ❗️ We're not caching anything if `conditionInfo.getArgs()` is empty, so this assertion is a little misleading.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
```diff
@@ -0,0 +1,140 @@
...
+        new BundleNameCondition(
+            bundle,
+            new ConditionInfo(
+                BundleNameCondition.class.getName(), new String[] {"bundle1/bundle2"}));
+
+    boolean satisfied =
+        bundleNameCondition1.isSatisfied(
+            new Condition[] {bundleNameCondition1, bundleNameCondition2}, null);
+    assertThat(satisfied, is(true));
+  }
```
lessarderic: ❓ Shouldn't we check that the result is cached?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Test
+  public void testIsAllNotSatisfied() {
```
lessarderic: ❗️ We should add a test where one `BundleNameCondition.isSatisfied()` returns `true` and the other returns `false`.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
```diff
@@ -0,0 +1,140 @@
...
+    boolean satisfied = bundleNameCondition.isSatisfied();
+    // no cache this time around
+    assertThat(satisfied, is(false));
+    satisfied = bundleNameCondition.isSatisfied();
+    // cached
+    assertThat(satisfied, is(false));
+  }
+
+  @Test
+  public void testIsAllSatisfied() {
```
lessarderic: ❓ Since `isSatisfied(Condition[], Dictionary)` is defined in `AbstractCondition` and doesn't really care about the actual `Condition` implementation, should we put those tests in a separate `Test` class instead of making them look like they're `BundleNameCondition` specific?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/PrincipalConditionTest.java
```diff
@@ -0,0 +1,79 @@
...
+        new PrincipalCondition(
+            bundle,
+            new ConditionInfo(
+                PrincipalCondition.class.getName(), new String[] {"role1", "role2", "role3"}));
+    boolean satisfied = principalCondition.isSatisfied();
+    assertThat(satisfied, is(false));
+  }
+
+  @Test
+  public void testIsNotSatisfied() {
```
lessarderic: ✏️ `testIsSatisfiedWithPrincipalMissing()`?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/PrincipalConditionTest.java
```diff
@@ -0,0 +1,79 @@
...
+            new ConditionInfo(
+                PrincipalCondition.class.getName(), new String[] {"role1", "role2", "role3"}));
+
+    boolean satisfied =
+        Subject.doAs(javaSubject, (PrivilegedAction<Boolean>) principalCondition::isSatisfied);
+    assertThat(satisfied, is(false));
+  }
+
+  @Test
+  public void testIsSatisfied() {
```
lessarderic: ✏️ `testIsSatisfiedWithAllPrincipals()`?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/PrincipalConditionTest.java
```diff
@@ -0,0 +1,79 @@
...
+        new PrincipalCondition(
+            bundle,
+            new ConditionInfo(
+                PrincipalCondition.class.getName(), new String[] {"role1", "role2", "role3"}));
+
+    boolean satisfied =
+        Subject.doAs(javaSubject, (PrivilegedAction<Boolean>) principalCondition::isSatisfied);
+    assertThat(satisfied, is(true));
+  }
+}
```
lessarderic: ❓ Don't we need a test where the `Subject` has more principals than needed, i.e., `testIsSatisfiedWithExtraPrincipals()`?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/SignerConditionTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.junit.Test;
+import org.osgi.framework.Bundle;
+import org.osgi.service.condpermadmin.ConditionInfo;
+import sun.security.x509.X509CertImpl;
+
+public class SignerConditionTest {
```
lessarderic: ❗️ We're missing many conditions and paths here that would be really difficult to test higher up. We should mock out the value returned by `bundle.getSignerCertificates(Bundle.SIGNERS_TRUSTED)` and test all the paths and cases in `isSatisfied()`, `getExtendedCertAttribute()`, `getSubjectAlternativeNames()` and `collectAltNames()`, otherwise we won't know for sure that this code always works as expected and may cause a hard to find security regression down the road.

platform/osgi/platform-osgi-conditions/pom.xml
```diff
@@ -48,6 +48,14 @@
     </dependencies>
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
+                    <!--Plugin wants to remove imports that should be there-->
```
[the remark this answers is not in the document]
stustison: It fails because checkstyle is broken for imports for comments and always has been. I'm not going to fix it for this PR, but we should probably just disable everything but the header check since it is already being done by the Google formatter.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -27,7 +27,7 @@
  *
  * <p>Selects bundles based on the name. @ThreadSafe
  */
-public class BundleNameCondition implements Condition {
+public class BundleNameCondition extends AbstractCondition implements Condition {
```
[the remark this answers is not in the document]
stustison: The first paragraph is the same, there is a description under that which tells what the specific implementation does.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -27,7 +27,7 @@
...
 
   private static final ConcurrentHashMap<String, Boolean> DECISION_MAP =
```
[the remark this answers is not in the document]
stustison: Yes. We end up with over 50,000 objects in the map.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/AbstractCondition.java
```diff
@@ -0,0 +1,76 @@
...
+   * @param context A Dictionary object that implementors can use to track state. If this method is
+   *     invoked multiple times in the same permission check, the same Dictionary will be passed
+   *     multiple times. The SecurityManager treats this Dictionary as an opaque object and simply
+   *     creates an empty dictionary and passes it to subsequent invocations if multiple invocations
+   *     are needed.
+   * @return {@code true} if all the Condition objects are satisfied. Otherwise, {@code false} if
+   *     one of the Condition objects is not satisfied.
+   */
+  @Override
+  public boolean isSatisfied(Condition[] conditions, Dictionary<Object, Object> context) {
```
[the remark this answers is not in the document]
stustison: Can't change this, it is part of the spec.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/AbstractCondition.java
```diff
@@ -0,0 +1,76 @@
...
+    for (Condition condition : conditions) {
```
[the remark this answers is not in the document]
stustison: I'm trying to avoid the streaming stuff because it is really easy to get things that aren't as performant as traditional loops with that and this code is extremely sensitive to performance.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/PrincipalCondition.java
```diff
@@ -72,50 +55,14 @@ public boolean isPostponed() {
...
-   *     one of the Condition objects is not satisfied.
-   */
-  @Override
-  public boolean isSatisfied(Condition[] conditions, Dictionary<Object, Object> context) {
-    for (Condition condition : conditions) {
-      if (!condition.isSatisfied()) {
-        return false;
+    if (subject != null) {
+      Set<Principal> principals = subject.getPrincipals();
+      List<String> principalNameList = new ArrayList<>();
```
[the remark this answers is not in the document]
stustison: Same thing, I'm avoiding streams here on purpose.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
stustison: Just that there are consistent results.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
stustison: Fixed.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
stustison: Fixed.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
stustison: Wouldn't that just be the same thing as the first 2 tests?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/PrincipalConditionTest.java
stustison: Sure.

stustison:
Going to also tack on a performance fix in this area of the code with this PR

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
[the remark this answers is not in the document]
lessarderic: Just wondering why we keep repeating the same thing in all of them. Especially since they are _not_ interfaces but classes. We could just remove the first part and keep the part that explains what the implementation does.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/SignerConditionTest.java
stustison: I added a test for subject alt name, but I'm not sure how to populate it with binary data. I know it is possible and the code there should work.

stustison:
build now

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
lessarderic: If we're not testing that results are cached, we shouldn't have comments about caching. Testing for consistency isn't really something we should be testing for and is not something we test for anywhere else. Either caching is an implementation detail and we don't care from a test perspective, or the caching part is critical and we need to make sure we're testing it properly.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
lessarderic: I don't think so. Right now we're testing when they all match and when none match, but not when there's a partial match. Unless I'm missing something...

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/SignerConditionTest.java
lessarderic: Can't we just use a mock instead of reading in a real `.der` file? Wouldn't that make it possible to test those cases? I'm just worried that if we don't have good unit tests here, someone could introduce a security related bug in the future and we won't know about it.

stustison: I'd just be testing the mock instead of reading a real der sequence that is actually saved in a cert at that point. Someone could definitely introduce something here. However, we don't really have any reason to use this code in DDF and it is mainly meant for downstream projects if they ever have a need for adding conditions against bundle signers. I could also remove the code that deals with bytes, I'd just rather not do that.

lessarderic: Ok, let's not worry about that part then.

Can we add a test for this exception though: https://github.com/codice/ddf/pull/3073/files#diff-48a09791a0210f51250e2c0cdfce43ebR108?

We should also make https://github.com/codice/ddf/pull/3073/files#diff-48a09791a0210f51250e2c0cdfce43ebR125 a `RuntimeException` instead of an `Exception` and see if we can add a unit test for it as well.

platform/osgi/platform-osgi-condpermadmin/src/test/resources/deny.policy
```diff
@@ -0,0 +1,11 @@
+priority "deny";
```
SmithJosh: ❓ Can we include examples of principal and signer conditions in these test policies?

platform/osgi/platform-osgi-condpermadmin/src/test/resources/grant.policy
```diff
@@ -0,0 +1,17 @@
+priority "grant";
```
SmithJosh: ❓ Can we include examples of principal and signer conditions in these test policies?

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/SignerConditionTest.java
stustison: Yep, added tests for the exceptions.

platform/osgi/platform-osgi-condpermadmin/src/test/resources/grant.policy
stustison: Sure.

platform/osgi/platform-osgi-conditions/src/test/java/org/codice/ddf/condition/BundleNameConditionTest.java
stustison: I could I guess, or I could put the same test in all of them, or I can just leave it as is. I don't really feel strongly about it either way as long as the code is tested.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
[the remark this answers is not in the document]
stustison: I dunno, that is just how I did it :)

stustison:
build now

stustison:
build now

stustison:
build now

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/AbstractCondition.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Returns whether the specified set of Condition objects are satisfied. Although this method is
+   * not static, it must be implemented as if it were static. All of the passed Condition objects
+   * will be of the same type and will correspond to the class type of the object on which this
+   * method is invoked. This method must be called inside a permission check only.
+   *
+   * @param conditions The array of Condition objects, which must all be of the same class and
+   *     mutable. The receiver must be one of those Condition objects.
```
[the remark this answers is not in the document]
stustison: No, I don't think it is our responsibility. If it is, no one is doing it. Equinox ensures that the array is built correctly.

platform/osgi/platform-osgi-conditions/src/main/java/org/codice/ddf/condition/BundleNameCondition.java
```diff
@@ -76,12 +55,12 @@ public boolean isSatisfied() {
       return false;
     }
     String bundleName = bundle.getHeaders().get("Bundle-SymbolicName");
```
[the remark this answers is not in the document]
stustison: The constructor and isSatisfied call actually end up happening on the same thread since these objects are created on the fly using reflection, so it yields no performance benefit. That was why I ripped out the REGEX call and moved it back to the bundle activator to change the data that the condition is handed, unfortunately there is no way to know the bundles ahead of time.

stustison:
build now

PR #3091, opened 2018-03-27
[WIP] DDF-3700 OpenSearch refactoring and bug fixes

emmberk:
##### DO NOT MERGE. WIP.
____
#### What does this PR do?
This PR includes a few commits to refactor and fix bugs in the OpenSearch Endpoint/Source.

- `Cleaned up catalog-opensearch-source pom` - This fixes fixes the issue of not being able to run unit tests inside the IDE.
- `Moved OpenSearchParser to be a static util class in opensearch-source` - `OpenSearchParser` was originally an interface so that downstream projects could provide their own implementation. However, it makes more sense as a static class instead of a service because all of the methods can be static.
- `Addressed some minor static analysis findings in catalog-opensearch`
- `Added After and Before implementations to OpenSearchFilterVisitor` - This commit adds `Before` and `After` implementations to `OpenSearchFilterVisitor`, instead of dropping those filters from the query. This fixes modified `BEFORE` and `AFTER` temporal queries in Intrigue.
- `Refactored OpenSearchSource so that it does not fail when no keyword is specified -- STILL NEEDS UNIT TESTS` This fixes the issue of failing queries when the keyword was not specified. This commit still needs unit tests.
- `Moved OpenSearch parameter names to new constants class` `opensearch-endpoint` and `opensearch-source` use many of the same constants for query parameters. This commit moves these to a new constants class and also improves the documentation for the source/endpoint.
- `Updated OpenSearchVisitorObject to have getPolygonSearch and getPointRadiusSearch methods instead of getContextualSearch` 
- `Updated polygon to bounding box method in the OpenSearch Source to use com.vividsolutions.jts.geom.Geometry#getEnvelopeInternal` The custom method to construct a bounding box from a polygon was replaced with a third-party method.
- `Added logic in the OpenSearch Source to ignore point-radius filters when the radius is not greater than 0`
- `Added logic in the OpenSearch Source to ignore temporal filters that are not for "modified"`
- `Added logic in the OpenSearch Source to ignore spatial filters that are not for "location"`

#### Who is reviewing it? 
@codice/io 
@bdeining
@clockard
[DDF-3700](https://codice.atlassian.net/browse/DDF-3700)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


clockard:
Lots of good things in this pr. Thanks!

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/OpenSearchQuery.java
```diff
@@ -334,11 +342,11 @@ public void addTypeFilter(String type, String versions) {
 
     if (versions != null && !versions.isEmpty()) {
       LOGGER.debug("Received versions from client.");
-      String[] typeVersions = versions.split(",");
-      List<Filter> typeVersionPairsFilters = new ArrayList<Filter>();
+      String[] typeVersions = versions.split(OpenSearchParameters.VERSION_DELIMINATOR);
```
stustison: Might be more performant to compile the regex from `OpenSearchParameters.VERSION_DELIMINATOR` and then call split from the regex instead.

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/PolygonSpatialFilter.java
```diff
@@ -14,14 +14,15 @@
...
+import org.codice.ddf.opensearch.OpenSearchParameters;
 
 public class PolygonSpatialFilter extends SpatialFilter {
   private String[] latLonAry;
 
   public PolygonSpatialFilter(String lonLatStr) {
     super();
 
-    latLonAry = lonLatStr.split(" |,\\p{Space}?");
+    latLonAry = lonLatStr.split(OpenSearchParameters.POLYGON_LON_LAT_DELIMINATOR);
```
stustison: Same comment here about the regex.

catalog/opensearch/catalog-opensearch-endpoint/src/test/java/org/codice/ddf/opensearch/query/OsNamespaceContext.java
```diff
@@ -54,7 +52,7 @@ public String getPrefix(String namespaceURI) {
 
   @Override
   public Iterator<String> getPrefixes(String namespaceURI) {
-    ArrayList<String> prefixList = new ArrayList<String>(1);
+    ArrayList<String> prefixList = new ArrayList<>(1);
```
stustison: Having the size at `1` is going to cause the add to resize the list every single time.
How about just `return Collections.singletonList(getPrefix(namespaceURI)).iterator()`?

catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchParser.java
```diff
@@ -41,11 +41,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class OpenSearchParserImpl implements OpenSearchParser {
+public final class OpenSearchParser {
```
bdeining: What's the rationale behind making this final?  What if downstream projects wish to extend this class?

catalog/opensearch/catalog-opensearch-endpoint/src/test/java/org/codice/ddf/endpoints/OpenSearchEndpointTest.java
```diff
@@ -114,19 +110,15 @@ public void testProcessQueryForProperHandlingOfSiteNameLOCAL()
...
-                        .contains(testSiteName));
-
-                return new QueryResponseImpl(queryRequest);
-              }
+            invocation -> {
+              QueryRequest queryRequest = (QueryRequest) invocation.getArguments()[0];
+
+              // ***Test verification***
+              // This assert is the whole point of this unit test
+              Assert.assertTrue(
```
bdeining: ideally we use `assertThat` here.  not sure the comment is useful above

catalog/opensearch/catalog-opensearch-common/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+        </dependency>
+    </dependencies>
+    <parent>
```
bdeining: ✏️  The parent seems to be in a weird location, i would have expected it to be at the top of the pom

catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchParser.java
emmberk: I'm open to discussion about this, but my motivation was `OpenSearchParser` only contains some static methods to add some query parameters in a `WebClient`, and it should never be instantiated.

I would think that a new class
```java
public class AnotherProjectOpenSearchParser {

    public static void populateSomething(WebClient client,
```
and using the static methods from each class directly 
```java
AnotherProjectOpenSearchParser.populateSomething(webClient
```
```java
OpenSearchParser.populateSearchOptions(webClient
```
would work well.

catalog/opensearch/catalog-opensearch-endpoint/src/test/java/org/codice/ddf/endpoints/OpenSearchEndpointTest.java
emmberk: There are many of these `Assert` methods in the unit tests in these opensearch bundles that I didn't touch, and it's out of the scope of this ticket to fix all of those.

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/BBoxSpatialFilter.java
```diff
@@ -54,14 +61,29 @@ public BBoxSpatialFilter(String bbox) {
...
    */
   private String createWKT() {
-    StringBuilder wktBuilder = new StringBuilder("POLYGON((");
-    wktBuilder.append(minX + " " + minY);
-    wktBuilder.append("," + minX + " " + maxY);
-    wktBuilder.append("," + maxX + " " + maxY);
-    wktBuilder.append("," + maxX + " " + minY);
-    wktBuilder.append("," + minX + " " + minY);
-    wktBuilder.append("))");
+    String wktBuilder =
```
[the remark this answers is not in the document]
emmberk: It was a static analysis finding to use string concatenation instead of StringBuilder. For readability, I replaced this with 
```java
String.format("POLYGON((%1$s %2$s,%1$s %4$s,%3$s %4$s,%3$s %2$s,%1$s %2$s))", minX, minY, maxX, maxY)
```

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/BBoxSpatialFilter.java
```diff
@@ -30,15 +29,35 @@
    * @param bbox comma-delimited list of lat/lon (deg) bounding box coordinates
    *     (West,South,East,North)
    */
-  public BBoxSpatialFilter(String bbox) {
+  public BBoxSpatialFilter(final String bbox) {
     super();
 
-    String[] bboxArY = bbox.split(" |,\\p{Space}?");
+    String[] bboxArY = bbox.split(OpenSearchConstants.BBOX_DELIMITER);
```
mojogitoverhere: ❓  Is there a reason the `bbox` string format was made more restrictive?

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/BBoxSpatialFilter.java
```diff
@@ -30,15 +29,35 @@
...
+      this.maxY = Double.parseDouble(bboxArY[OpenSearchConstants.BBOX_LIST_SIZE - 1]);
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException(
+          "The "
+              + OpenSearchConstants.BBOX
+              + " OpenSearch Endpoint parameter must have four "
+              + OpenSearchConstants.BBOX_DELIMITER
+              + "-delimited double values but is \""
+              + bbox
+              + "\".",
```
mojogitoverhere: ✏️ `String.format` might be more readable here

catalog/opensearch/catalog-opensearch-source/src/test/java/ddf/catalog/source/opensearch/impl/OpenSearchFilterVisitorTest.java
```diff
@@ -112,372 +133,548 @@ public void testAndFilter() {
...
+        allOf(
+            is(notNullValue()),
+            hasProperty("startDate", is(START_DATE)),
+            hasProperty("endDate", is(MAXIMUM_DATE))));
+  }
+
+  @Test
+  public void testBefore() {
+    org.opengis.filter.temporal.Before beforeFilter =
+        (org.opengis.filter.temporal.Before)
```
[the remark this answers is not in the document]
emmberk: It conflicts with the `@org.junit.Before` annotation if it's not qualified.

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/PolygonSpatialFilter.java
```diff
@@ -14,14 +14,15 @@
...
+import org.codice.ddf.opensearch.OpenSearchConstants;
 
 public class PolygonSpatialFilter extends SpatialFilter {
   private String[] latLonAry;
 
   public PolygonSpatialFilter(String lonLatStr) {
     super();
 
-    latLonAry = lonLatStr.split(" |,\\p{Space}?");
+    latLonAry = lonLatStr.split(OpenSearchConstants.POLYGON_LON_LAT_DELIMITER);
```
mojogitoverhere: ❓Is there a reason the lonLatStr string format was made more restrictive?

catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java
```diff
@@ -301,26 +292,81 @@ public SourceResponse query(QueryRequest queryRequest) throws UnsupportedQueryEx
...
+        openSearchFilterVisitorObject.getPointRadiusSearch();
+    final Polygon polygonSearch = openSearchFilterVisitorObject.getPolygonSearch();
+    final TemporalFilter temporalSearch = openSearchFilterVisitorObject.getTemporalSearch();
+
+    Map<String, String> searchPhraseMap = null;
+    if (contextualSearch != null) {
+      searchPhraseMap = contextualSearch.getSearchPhraseMap();
+    }
+
+    if (!MapUtils.isEmpty(searchPhraseMap)
```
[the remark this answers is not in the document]
emmberk: I also fixed a few more that I found.

catalog/opensearch/catalog-opensearch-api/pom.xml
```diff
@@ -16,54 +16,27 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
     <parent>
-        <artifactId>catalog-opensearch-pom</artifactId>
+        <artifactId>opensearch</artifactId>
```
mackncheesiest: ✏️ Total nit, but can we make the ordering of groupid : artifactid : version consistent?

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/BBoxSpatialFilter.java
emmberk: The spec says that the bbox query parameter must be in an exact format:
https://github.com/codice/ddf/blob/3d8784a8d77590b5aaed142d80ea8a68ae4caad0/catalog/opensearch/catalog-opensearch-endpoint/src/main/resources/ddf-os.xml#L59-L60

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/query/filter/PolygonSpatialFilter.java
emmberk: The spec says that the `polygon` query parameter must be in an exact format:
https://github.com/codice/ddf/blob/3d8784a8d77590b5aaed142d80ea8a68ae4caad0/catalog/opensearch/catalog-opensearch-endpoint/src/main/resources/ddf-os.xml#L62-L64

emmberk:
I think that I've addressed all comments so far. I still need to complete some thorough manual testing, but hopefully that doesn't result in any more code changes.

catalog/opensearch/catalog-opensearch-endpoint/src/main/java/org/codice/ddf/opensearch/endpoint/query/OpenSearchQuery.java
```diff
@@ -255,11 +273,10 @@ public void addTemporalFilter(String dateStart, String dateEnd, String dateOffse
 
     // If either start date OR end date is specified and non-empty, then
     // a temporal filter can be created
-    if ((dateStart != null && !dateStart.trim().isEmpty())
-        || (dateEnd != null && !dateEnd.trim().isEmpty())) {
-      temporalFilter = new TemporalFilter(dateStart, dateEnd);
-    } else if (dateOffset != null && !dateOffset.trim().isEmpty()) {
-      temporalFilter = new TemporalFilter(Long.parseLong(dateOffset));
+    if (StringUtils.isNotBlank(dateStart) || StringUtils.isNotBlank(dateEnd)) {
+      temporalFilter = new TemporalFilter(dateStart.trim(), dateEnd.trim());
```
clockard: can dateStart or dateEnd be null here? Just thinking that previously we were checking for null and if one of them isn't null and one is then we would get a NPE on one of the trim calls.

emmberk:
build now

catalog/opensearch/catalog-opensearch-api/src/main/java/org/codice/ddf/opensearch/source/PointRadiusSearch.java
```diff
@@ -38,4 +40,25 @@ public double getLat() {
...
     return radius;
   }
+
+  @Override
+  public boolean equals(Object object) {
+    if (this == object) {
+      return true;
+    }
+
+    if (!(object instanceof PointRadiusSearch)) {
```
peterhuffer: ❓ Should we check for null input before doing any other comparisons or does this take care of null here?

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchFilterVisitor.java
```diff
@@ -271,16 +269,20 @@ public Object visit(PropertyIsEqualTo filter, Object data) {
...
-          openSearchFilterVisitorObject.setId(isEqualsTo.getExpression2().toString());
-        }
-      } else if (filter instanceof PropertyIsEqualToLiteral) {
-        PropertyIsEqualToLiteral isEqualsTo = (PropertyIsEqualToLiteral) filter;
-        if (Metacard.ID.equals(isEqualsTo.getExpression1().toString())) {
-          openSearchFilterVisitorObject.setId(isEqualsTo.getExpression2().toString());
+      final Expression expression1 = filter.getExpression1();
+      if (expression1 instanceof PropertyName) {
+        final String propertyName = ((PropertyName) expression1).getPropertyName();
+        final String expectedPropertyIsEqualToTerm = Metacard.ID;
```
clockard: Is this local variable just for clarity?

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchFilterVisitor.java
```diff
@@ -271,16 +269,20 @@ public Object visit(PropertyIsEqualTo filter, Object data) {
...
+        if (StringUtils.equals(expectedPropertyIsEqualToTerm, propertyName)) {
```
clockard: Is there a reason we don't just do `expectedPropertyIsEqualToTerm.equals(propertyName)`?

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
```diff
@@ -957,4 +968,78 @@ private WebClient newOpenSearchClient(String url, Subject subj) {
...
+    }
+  }
+
+  /**
+   * Method to combine spatial searches into either a polygon or a point-radius search. OpenSearch
+   * endpoints will ignore multiple spatial query parameters. This method is protected so that
+   * downstream projects may try to implement another algorithm (e.g. best-effort) to combine
+   * searches.
+   *
+   * @return null if the searches cannot be combined, or a {@link SpatialSearch} with one {@link
```
clockard: Looks like we return optional and not null

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
```diff
@@ -957,4 +968,78 @@ private WebClient newOpenSearchClient(String url, Subject subj) {
...
+   */
+  protected Optional<SpatialSearch> createCombinedSpatialSearch(
+      final Queue<PointRadiusSearch> pointRadiusSearches, final Queue<Polygon> polygonSearches) {
+    final boolean filterContainedSomePointRadiusSearchCriteria =
+        CollectionUtils.isNotEmpty(pointRadiusSearches);
+    final boolean filterContainedSomePolygonSearchCriteria =
+        CollectionUtils.isNotEmpty(polygonSearches);
+
+    if (filterContainedSomePointRadiusSearchCriteria && filterContainedSomePolygonSearchCriteria) {
+      LOGGER.debug(
```
clockard: Hope this doesn't surprise anyone since we are dropping all spatial terms and not taking the last one

catalog/opensearch/catalog-opensearch-api/src/main/java/org/codice/ddf/opensearch/source/OpenSearchParser.java
```diff
@@ -41,9 +41,10 @@ void populateSearchOptions(
       WebClient client, QueryRequest queryRequest, Subject subject, List<String> parameters);
 
   /**
-   * Fills in the OpenSearch query URL with contextual information (Note: Section 2.2 - Query: The
-   * OpenSearch specification does not define a syntax for its primary query parameter, searchTerms,
-   * but it is generally used to support simple keyword queries.)
+   * Fills in the OpenSearch query URL with the contextual information is contained in the search
+   * phrase link Map<String, String>}. (Note: Section 2.2 - Query: The OpenSearch specification does
```
kcover: ✏️ Did you mean to put "information that is contained in the search..." or maybe just "information contained in the search..."?

catalog/opensearch/catalog-opensearch-api/src/main/java/org/codice/ddf/opensearch/source/PointRadiusSearch.java
emmberk: `instanceOf` already checks for `null`.

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchFilterVisitor.java
emmberk: Yes. I also use it down below in the log message.

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
emmberk: Right. Before this PR, the last point-radius filter in the query would be used to populate the query parameters. If there were no point-radius filters, then the  last polygon filter in the query would be used to populate the query parameters.
https://github.com/codice/ddf/blob/b279c781e127c5fc9eb73956b9b79609f6b5a5cd/catalog/opensearch/catalog-opensearch-source/src/main/java/ddf/catalog/source/opensearch/impl/OpenSearchSource.java#L979-L996
The OpenSearch Source was dropping spatial filters without any indication and populating the query parameters from the filters in an unexpected way. This logic was updated to not support any filters with multiple spatial filters. This implementation could be updated in the future or in downstream projects to, for example, combine all spatial criteria into a single `Geometry`.

emmberk:
build now

emmberk:
build now

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchParserImpl.java
```diff
@@ -167,41 +144,25 @@ public void populateTemporal(WebClient client, TemporalFilter temporal, List<Str
...
-      return;
-    }
-
-    if (shouldConvertToBBox) {
-      final BoundingBox boundingBox = createBoundingBoxFromPolygon(polygon);
-      checkAndReplace(client, boundingBox.toBboxString(), OpenSearchConstants.BBOX, parameters);
-    } else {
+    if (geometry != null) {
+      //      checkAndReplace(client, , OpenSearchConstants.GEOMETRY, parameters);
+      // TODO
```
emmberk: @emmberk reminder to fix this

emmberk:
build now

clockard:
Successfully ran manual test steps. 👍 :shipit:

emmberk:
build now

emmberk:
build now

PR #3102, opened 2018-03-30
DDF-3711 Adds loading indicators, fixes various bugs, refactors things into components

andrewkfiedler:
#### What does this PR do?
 - Adds loading indicators to areas of the installer that needed them.  One step that needed it was the guest claims page, and the other step was the system properties page.
 - Adds loading indicators to configurations, applications, features, and the system information page.
 - Fixes various styling issues such as: removing the margins on bootstrap panel classes which caused an extra 20px at the bottom of certain panels (and caused the shutdown / restart buttons of the installer to overlap in a weirdly partial way with the panel above them), removes the padding on the left of the alerts so they line up with the main well below it, adds a bit of spacing to the multivalued fields of the configuration editing view, removed the media query that added strange padding / margin to the main content panel.
 - Refactors quite a few things into components to make them easier to maintain and update going forward.
 - Removes a lot of unnecessary composite views and controllers.  At the same time I updated the collection views to properly utilize Marionette and Backbone so that we aren't constantly destroying and rerendering collections unnecessarily (we were basically doing this on all the collections, so things should be much faster now).
 - Removes some now unnecessary css.
 - Fixes how we parse descriptions for apps.  Previously we parsed them in such a way that only the last sentence of the description was preserved, which made them make absolutely no sense.  - Given that users apparently didn't notice the issue with descriptions for 4 years, I removed them from being omnipresent on the applications view.  Now only the name is shown, and on hover a popover shows the description.  I also updated the view so that users can keyboard navigate to the apps.  I also changed the app view so that it responded better to the users page size (in other words made it responsive).
 - Updated the configurations view so that everything is alphabetically sorted.
 - Updated the features view so that everything is alphabetically sorted.
 - Added a filter by feature name ability to the features view to allow users to easily find what they need.  Also added a filter by install status.
 - Updated the features view to remove the status column.  I merged it with the actions column since they both indicated the same thing (install or uninstall).
 - Updated the features view to add a tooltip on the action since it's possible the feature name is far from eyesight given we display them in a table (maybe we should rethink that), so now users know what they're installing or uninstalling on hover.
 - Updated the features view to add opacity as an indicator that a feature is uninstalled.
 - Updated the features view to add some indication of the currently hovered over row.
 - Updated the system information view to display the information in a more useful way (similar to the features view it was a table previously and labels were far from their respective values).

#### Who is reviewing it? 
@djblue 
@rzwiefel 
@tbatie 
@clockard

Try out all the things you used to be able to do and they should work.  I moved a lot of code around but the funtionality hasn't changed much.  Now that things are in components though it will be easy to start improving this UI.

A lot of this code hasn't been touched in 4 years because people were scared of it.  Hopefully, it will seem a little less intimidating now.

[DDF-3711](https://codice.atlassian.net/browse/DDF-3711)

#### Screenshots
screen shot 2018-03-30 at 12 58 56 pm
screen shot 2018-03-30 at 12 55 47 pm
screen shot 2018-03-30 at 12 56 00 pm
screen shot 2018-03-30 at 12 58 18 pm
screen shot 2018-03-30 at 12 58 27 pm
screen shot 2018-03-30 at 12 58 05 pm
screen shot 2018-03-30 at 12 57 54 pm
screen shot 2018-03-30 at 12 57 34 pm
screen shot 2018-03-30 at 12 57 12 pm
screen shot 2018-03-30 at 12 56 28 pm
screen shot 2018-03-30 at 12 56 12 pm


andrewkfiedler:
build now

clockard:
Manually checked out changes on DDF and downstream project. LGTM :shipit:

PR #3106, opened 2018-03-31
DDF-3715 Updates map views to properly clean up their drawing tools when destroyed, refactors the drawing tools by pulling out common methods

andrewkfiedler:
#### What does this PR do?
 - Updates the map views to properly clean up their drawing tools when destroyed.
 - Starts the process of pulling out common drawing tool functionality, starting with the controllers.  Now there is a base drawing controller that each can extend as necessary.

#### Who is reviewing it? 
@adimka 
@mackncheesiest 
@jlcsmith
@rzwiefel 

Test out each of the drawing methods and verify they work as expected.  Make sure to have both types of map open (2d and 3d), and test closing each and reopening each and then drawing with each type.  Verify no errors are thrown.

It looks like this started being an issue when we updated cesium to destroy itself properly, but regardless both maps should be disposing of their drawing tools when they get destroyed.

[DDF-3715](https://codice.atlassian.net/browse/DDF-3715)

Before fix when 3d map is closed and drawing is attempted

### DIFF
--- a/ui/packages/catalog-ui-search/src/main/webapp/js/widgets/drawing.controller.js
+++ b/ui/packages/catalog-ui-search/src/main/webapp/js/widgets/drawing.controller.js
@@ -18,8 +18,11 @@ var wreqr = require('wreqr');
 
 module.exports = Marionette.Controller.extend({
     enabled: true,
-    drawingType: 'extendme!',
+    drawingType: undefined,
     initialize: function() {
+        if (typeof this.drawingType === "undefined"){
+            throw "drawingType needs to be overwritten";
+        }
         this.listenTo(wreqr.vent, `search:${this.drawingType}display`, function(model) {
             this.show(model);
         });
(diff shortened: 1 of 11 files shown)


clockard:
Nice clean up! Lots of duplication removed 👍

jlcsmith:
build now

ui/packages/catalog-ui-search/src/main/webapp/js/widgets/drawing.controller.js
```diff
@@ -0,0 +1,147 @@
...
+                    model: bboxModel
+                });
+                view.updatePrimitive(model);
+                this.addView(view);
+            }
+
+            return bboxModel;
+        }
+    },
+    draw: function(model) {
```
mackncheesiest: ❓ This just looks like the draw function from `cesium.bbox.js`/`openlayers.bbox.js`.

Is the idea that this falls back to a bbox draw tool if nothing else is specified? Is there anything else we can do to make it more apparent that this should be overridden? Or can we at least have `cesium.bbox.js` and `openlayers.bbox.js` properly inherit this instead of redefining the function again?

andrewkfiedler: Yeah, I'm thinking we should update this to throw an error that says to implement the method.

mackncheesiest:
build now

PR #3083, opened 2018-03-26
DDF-3693 Adding DMS coordinates

Bdthomson:
#### What does this PR do?
Adds "Copy DMS Coordinates" option to map context-menu.
#### Who is reviewing it? 
(please choose AT LEAST two reviewers that need to approve the PR before it can get merged)

@Schachte 
@Variadicism 

#### Select relevant component teams: 
https://github.com/orgs/codice/teams

@codice/ui 
#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)

@andrewkfiedler
@bdeining
#### How should this be tested? (List steps with links to updated documentation)
1. Right click map.
2. Copy DMS coordinates (verify value in popup matches DMS coordinates in bottom left-hand corner).
#### What are the relevant tickets?
[DDF-3693](https://codice.atlassian.net/browse/DDF-3693)
#### Screenshots (if appropriate)
image

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

#### Notes on Review Process
Please see [Notes on Review Process](https://codice.atlassian.net/wiki/spaces/DDF/pages/71946981/Pull+Request+Guidelines) for further guidance on requirements for merging and abbreviated reviews. 

#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #3107, opened 2018-04-03
DDF-3693 Updates map-context-menu to use existing conversion library

andrewkfiedler:
#### What does this PR do?
 - Updates map-context-menu to use the existing conversion library rather than openlayers.  Depending on openlayers here is a bit heavy handed since it will pull in the entire openlayers library, which would be strange (especially if you're on the cesium map).  In the future we could update it to pull in exactly the conversion portion of openlayers if we want, but we would need to be using the es2015 modules version of openlayers.
 - Removes `clickDms` from the map model since it's unused.

#### Who is reviewing it? 
@Schachte 
@Variadicism 
@jlcsmith 
@bdeining 
@glenhein 

1. Right click map.
2. Copy DMS coordinates (verify value in popup matches DMS coordinates in bottom left-hand corner).

[DDF-3693](https://codice.atlassian.net/browse/DDF-3693)

#### Review Comment Legend:

### DIFF
--- a/ui/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/map.model.js
+++ b/ui/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/map.model.js
@@ -15,7 +15,7 @@
 /*global require*/
 var Backbone = require('backbone');
 var MetacardModel = require('js/model/Metacard');
-var ol = require('openlayers');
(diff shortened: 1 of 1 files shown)


andrewkfiedler:
small follow-on to: https://github.com/codice/ddf/pull/3083

andrewkfiedler:
build now

PR #3110, opened 2018-04-03
DDF-3718 Updated docs unit test and fixed broken anchors

mcalcote:
#### What does this PR do?
Updates the unit tests in the documentation to check for broken anchors (internal links), freemarker stack traces and missing images.

The primary changes take place in DocumentationTest.java and the rest of the changes are to fix broken links found using the new unit test.

#### Who is reviewing it? 
@vinamartin @garrettfreibott @ahoffer @sblackmore 

@codice/docs 

@ricklarsen @rzwiefel 

Inspect doc changes, review test design, run full build

[DDF-3718](https://codice.atlassian.net/browse/DDF-3718)

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/distribution/docs/src/main/java/ddf/docs/NoOp.java
+++ b/distribution/docs/src/main/java/ddf/docs/NoOp.java
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package ddf.docs;
+
+public class NoOp {}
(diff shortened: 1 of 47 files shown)


distribution/docs/src/test/java/ddf/docs/DocumentationTest.java
```diff
@@ -20,36 +20,93 @@
...
+                            || s.toLowerCase().contains(FREEMARKER_MSG)
+                            || s.contains(BASE64_MISSING));
+              } catch (IOException e) {
+                throw new RuntimeException(e);
+              }
+            }));
+  }
+
+  @Test
+  public void testBrokenAnchorsPresent() throws IOException, URISyntaxException {
```
ricklarsen: ❤️

vinamartin:
❓ I am fine with this change and I'm happy that we've fixed all those broken links. Awesome stuff! The only thing I'm a little confused about is how other projects typically deal with this. I found this: http://www.writethedocs.org/guide/tools/testing/#testing-your-documentation . 
I'm fine with the unit test approach, but to me, this feels like something that should break during compile. I'm not entirely sure how that would work, it seems to me like most tools really should have that handled. Is that something that isn't possible with the current way we build docs?

mcalcote:
@ricklarsen what happens when the document can't be rendered during the build? Does the build fail?

If it does, then I would argue we are alright, though those linked tools may help reduce time to diagnosis. 

I think that these tests provide a pretty solid net since we ensure the docs were actually rendered in the first place in the expected location and that they do not contain any of the the known issues they have had in the past (or may have in the future).

I am not against investigating those additional libraries for the benefits they offer. Maybe we should create a spike?

distribution/docs/src/main/java/ddf/docs/NoOp.java
```diff
@@ -0,0 +1,16 @@
+/**
```
mcalcote: Apparently this class is needed to have surefire run the tests in the docs module.

mcalcote:
After rebasing I found another broken anchor, updated. Build now

PR #3112, opened 2018-04-03
DDF-3706 Adds better support for text selection

andrewkfiedler:
#### What does this PR do?
 - Previously text selection was not allowed in the result sidebar or
 the table.
 - Now it is supported by taking into account how long the mouse has
 been held down by the user, or whether the alt key is being used.  This
 helps prevent weird interactions between the result selection behavior
 and the text selection behavior.  Text selection behavior through
 double / triple click is supported only through the alt key.
 - Better support for double click behavior has been added as well.
 Double clicks (or any sustained sequence of clicks under 500ms apart)
 are now seen as a single click.  This is accomplished by looking at
 time between clicks and the target of the clicks (so that users can
 still select multiple results in a short amount of time).
 - Converts the result-selection.decorator to a behavior for
 ease of use and simpler maintainability.
 - Deletes the decorators since they are no longer utilized.

#### Who is reviewing it? 
@adimka 
@mackncheesiest 
@Variadicism 
@jlcsmith
@bdeining 

Test selecting test from the sidebar results view or the table view.  Verify it doesn't select or deselect the result.  Verify that you can depress the alt key and perform the other text selection actions like double and triple click to select text.

Verify that double clicking a result will only toggle the selection of that result once (not back and forth).

[DDF-3706](https://codice.atlassian.net/browse/DDF-3706)

#### Video
See ticket or https://codice.atlassian.net/secure/attachment/18229/ddf-3706-text-selection.mp4


ui/packages/catalog-ui-search/src/main/webapp/behaviors/selection.behavior.js
```diff
@@ -0,0 +1,127 @@
...
+    interpretClick: function (event) {
+        if (event.altKey || this.isTextSelection() || this.isDoubleClick(event)) {
+            return;
+        }
+        const resultid = event.currentTarget.getAttribute('data-resultid');
+        const alreadySelected = this.options.selectionInterface.getSelectedResults().get(resultid) !== undefined;
+        //shift key wins over all else
+        if (event.shiftKey) {
+            this.handleShiftClick(resultid, alreadySelected);
+        } else if (event.ctrlKey || event.metaKey) {
```
Variadicism: :question: Why is the use of a `Meta` key considered equivalent to a `Ctrl` here?

ui/packages/catalog-ui-search/src/main/webapp/behaviors/selection.behavior.js
```diff
@@ -0,0 +1,127 @@
...
+            this.options.selectionInterface.clearSelectedResults();
+            this.handleControlClick(resultid, alreadySelected);
+        }
+    },
+    handleShiftClick: function (resultid, alreadySelected) {
+        const selectedResults = this.options.selectionInterface.getSelectedResults();
+        const indexClicked = this.options.selectionInterface.getCompleteActiveSearchResults().indexOfId(resultid);
+        const firstIndex = getMinIndex(this.options.selectionInterface);
+        const lastIndex = getMaxIndex(this.options.selectionInterface);
+        if (firstIndex === -1 && lastIndex === -1) {
```
Variadicism: This seems to be equivalent to `selectedResults.length === 0`, which would be easier to read and understand and not dependent on specialized return values.

ui/packages/catalog-ui-search/src/main/webapp/behaviors/selection.behavior.js
andrewkfiedler: For mac support.  On mac, by default it uses the `meta` key like the `ctrl` key.

ui/packages/catalog-ui-search/src/main/webapp/behaviors/selection.behavior.js
```diff
@@ -0,0 +1,127 @@
...
+const Marionette = require('marionette');
+
+function getMaxIndex(selectionInterface) {
+    const selectedResults = selectionInterface.getSelectedResults();
+    const completeResults = selectionInterface.getCompleteActiveSearchResults();
+    let maxIndex = -1;
+    selectedResults.forEach(function (result) {
+        maxIndex = Math.max(maxIndex, completeResults.indexOf(result));
+    });
+    return maxIndex;
```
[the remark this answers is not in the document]
andrewkfiedler: 👍 I'll update it to be like getMinIndex.

jlcsmith:
build now

jlcsmith:
build now

PR #3119, opened 2018-04-05
DDF-3712 Allow resource requests with a derived URI

kcover:
#### What does this PR do?

Adds new functionality to resource requests allowing you to make a request with a derived URI. It also adds code to help find derived resources when using in place monitoring.

#### Who is reviewing it? 

@peterhuffer @emmberk @oconnormi 

https://github.com/orgs/codice/teams/io

@clockard
@pklinef

Verify various types of products ingest and download properly. Message me for details.

[DDF-3712](https://codice.atlassian.net/browse/DDF-3712)
[CAL-431](https://codice.atlassian.net/browse/CAL-431)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/ResourceRequestByDerivedUri.java
+++ b/catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/ResourceRequestByDerivedUri.java
@@ -13,36 +13,37 @@
  */
 package ddf.catalog.operation.impl;
 
+import ddf.catalog.data.types.Core;
 import ddf.catalog.operation.ResourceRequest;
 import java.io.Serializable;
 import java.net.URI;
 import java.util.Map;
 
-public class ResourceRequestByProductUri extends OperationImpl implements ResourceRequest {
+public class ResourceRequestByDerivedUri extends OperationImpl implements ResourceRequest {
 
   protected String name;
 
   protected URI uri;
 
   /**
-   * Implements a ResourceRequestByProductUri and specifies the {@link URI}
+   * Implements a ResourceRequestByDerivedUri and specifies the {@link URI}
    *
    * @param uri the URI
    */
-  public ResourceRequestByProductUri(URI uri) {
+  public ResourceRequestByDerivedUri(URI uri) {
     this(uri, null);
   }
 
   /**
-   * Implements a ResourceRequestByProductUri and specifies the {@link URI} and a {@link Map} of
+   * Implements a ResourceRequestByDerivedUri and specifies the {@link URI} and a {@link Map} of
    * properties
    *
    * @param uri the URI
    * @param properties the properties
    */
-  public ResourceRequestByProductUri(URI uri, Map<String, Serializable> properties) {
+  public ResourceRequestByDerivedUri(URI uri, Map<String, Serializable> properties) {
     super(properties);
-    this.name = GET_RESOURCE_BY_PRODUCT_URI;
+    this.name = Core.DERIVED_RESOURCE_URI;
     this.uri = uri;
   }
 
--- a/catalog/core/catalog-core-contentresourcereader/src/main/java/org/codice/ddf/catalog/content/resource/reader/ContentResourceReader.java
+++ b/catalog/core/catalog-core-contentresourcereader/src/main/java/org/codice/ddf/catalog/content/resource/reader/ContentResourceReader.java
@@ -156,20 +156,6 @@ public class ContentResourceReader implements ResourceReader {
       LOGGER.debug("Resource URI is content scheme");
       String contentId = resourceUri.getSchemeSpecificPart();
       if (contentId != null && !contentId.isEmpty()) {
-        if (arguments != null
-            && arguments.get(ContentItem.QUALIFIER_KEYWORD) instanceof String
-            && StringUtils.isNotBlank((String) arguments.get(ContentItem.QUALIFIER_KEYWORD))) {
-          try {
-            resourceUri =
-                new URI(
-                    resourceUri.getScheme(),
-                    resourceUri.getSchemeSpecificPart(),
-                    (String) arguments.get(ContentItem.QUALIFIER_KEYWORD));
-          } catch (URISyntaxException e) {
-            throw new ResourceNotFoundException("Unable to create with qualifier", e);
-          }
-        }
-
         ReadStorageRequest readRequest = new ReadStorageRequestImpl(resourceUri, arguments);
         try {
           ReadStorageResponse readResponse = storage.read(readRequest);
@@ -184,6 +170,9 @@ public class ContentResourceReader implements ResourceReader {
         } catch (StorageException e) {
           throw new ResourceNotFoundException(e);
         }
+      } else {
+        throw new ResourceNotSupportedException(
+            String.format("Invalid URI %s received", resourceUri));
       }
     }
 
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
@@ -15,8 +15,10 @@ package ddf.catalog.impl.operations;
 
 import ddf.catalog.Constants;
 import ddf.catalog.content.data.ContentItem;
+import ddf.catalog.data.Attribute;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
+import ddf.catalog.data.types.Core;
 import ddf.catalog.federation.FederationException;
 import ddf.catalog.filter.FilterDelegate;
 import ddf.catalog.filter.impl.LiteralImpl;
@@ -536,44 +538,13 @@ public class ResourceOperations extends DescribableImpl {
       Map<String, Serializable> requestProperties,
       boolean fanoutEnabled)
       throws ResourceNotSupportedException, ResourceNotFoundException {
-
     ResourceInfo resourceInfo;
-    Query query = null;
-    URI resourceUri = null;
-    String name = resourceRequest.getAttributeName();
-    Object value = resourceRequest.getAttributeValue();
-
-    validateResourceInfoRequest(name, value);
+    Query query;
+    String attributeName = resourceRequest.getAttributeName();
+    Object attributeValue = resourceRequest.getAttributeValue();
 
     try {
-      if (ResourceRequest.GET_RESOURCE_BY_PRODUCT_URI.equals(name)) {
-        // because this is a get resource by product uri, we already
-        // have the product uri to return
-        LOGGER.debug("get resource by product uri");
-        resourceUri = (URI) value;
-        URI truncatedUri = resourceUri;
-        if (StringUtils.isNotBlank(resourceUri.getFragment())) {
-          resourceRequest
-              .getProperties()
-              .put(ContentItem.QUALIFIER_KEYWORD, resourceUri.getFragment());
-          try {
-            // Creating the truncated URL this way is important to preserve encoding!!
-            String uriString = resourceUri.toString();
-            truncatedUri = new URI(uriString.substring(0, uriString.lastIndexOf('#')));
-          } catch (URISyntaxException e) {
-            throw new ResourceNotFoundException(
-                "Could not resolve URI by doing a URI based query: " + value);
-          }
-        }
-        query = createPropertyStartsWithQuery(Metacard.RESOURCE_URI, truncatedUri.toString());
-      } else if (ResourceRequest.GET_RESOURCE_BY_ID.equals(name)) {
-        // since this is a get resource by id, we need to obtain the
-        // product URI
-        LOGGER.debug("get resource by id");
-        String metacardId = (String) value;
-        LOGGER.debug("metacardId = {},   site = {}", metacardId, site);
-        query = createMetacardIdQuery(metacardId);
-      }
+      query = createPropertyIsEqualToQuery(attributeName, attributeValue.toString());
 
       QueryRequest queryRequest =
           new QueryRequestImpl(
@@ -582,11 +553,23 @@ public class ResourceOperations extends DescribableImpl {
               Collections.singletonList(site == null ? this.getId() : site),
               resourceRequest.getProperties());
 
-      resourceInfo =
-          getResourceInfo(
-              queryRequest, resourceUri, requestProperties, federatedSite, fanoutEnabled);
-    } catch (UnsupportedQueryException | FederationException e) {
+      QueryResponse queryResponse = queryOperations.query(queryRequest, null, true, fanoutEnabled);
+      if (queryResponse.getResults().isEmpty()) {
+        throw new ResourceNotFoundException(
+            String.format(
+                "Could not find a metacard with a %s of %s.",
+                attributeName, attributeValue.toString()));
+      }
+      if (queryResponse.getResults().size() > 1) {
+        throw new UnsupportedQueryException(
+            String.format(
+                "found multiple metacards with a %s of %s.",
+                attributeName, attributeValue.toString()));
+      }
 
+      resourceInfo =
+          getResourceInfo(queryResponse, attributeValue, requestProperties, federatedSite);
+    } catch (UnsupportedQueryException | FederationException e) {
       throw new ResourceNotFoundException(DEFAULT_RESOURCE_NOT_FOUND_MESSAGE, e);
     }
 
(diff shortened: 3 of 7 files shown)


clockard:
I think you accidentally picked up @peterhuffer changes in this pr

catalog/core/catalog-core-contentresourcereader/src/main/java/org/codice/ddf/catalog/content/resource/reader/ContentResourceReader.java
```diff
@@ -181,14 +169,51 @@ public ResourceResponse retrieveResource(URI resourceUri, Map<String, Serializab
               new ResourceResponseImpl(
                   new ResourceImpl(
                       new BufferedInputStream(is), contentItem.getMimeType(), fileName));
-        } catch (StorageException e) {
+        } catch (StorageException | NullPointerException e) {
```
clockard: We shouldn't be catching a NPE

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/ResourceRequestByDerivedUri.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package ddf.catalog.operation.impl;
+
+import ddf.catalog.data.types.Core;
+import ddf.catalog.operation.ResourceRequest;
+import java.io.Serializable;
+import java.net.URI;
+import java.util.Map;
+
+public class ResourceRequestByDerivedUri extends OperationImpl implements ResourceRequest {
```
clockard: Where is this class used? Could use some javadoc as well

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -637,9 +615,11 @@ private ResourceInfo getResourceInfo(
     return new ResourceInfo(metacard, resourceUri);
   }
 
+  // TODO: remove probably
```
clockard: TODOs need to be addressed

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/ResourceRequestByDerivedUri.java
peterhuffer: Will be used in Alliance by the chipping service, who needs to download the derived overview product from NITFs.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -639,7 +628,8 @@ private ResourceInfo getResourceInfo(
 
   private void validateResourceInfoRequest(String name, Object value)
       throws ResourceNotSupportedException {
-    if (!(ResourceRequest.GET_RESOURCE_BY_PRODUCT_URI.equals(name) && value instanceof URI)
+    if (!(ResourceRequest.GET_RESOURCE_BY_PRODUCT_URI.equals(name)
+            || name.equals(Core.DERIVED_RESOURCE_URI) && value instanceof URI)
```
kcover: Just noticed the logic isn't what I intended here. I'll fix it on my next run through.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -817,4 +797,26 @@ protected void validateGetResourceRequest(ResourceRequest getResourceRequest)
...
+        if (ser instanceof String) {
+          URI uri = new URI((String) ser);
+
+          if (uri.getFragment().equals(qualifier)) {
+            resourceUri = uri;
+            return resourceUri;
+          }
+        }
+      } catch (URISyntaxException e) {
+        LOGGER.debug("A metacard has a derived uri without the proper syntax.");
```
kcover: I assume derived uris are checked for the right format when they're saved so this could just be a no-op but if not then I should do something more attention grabbing than a debug log. Like maybe throw an UnsupportedResourceException or something?

kcover: I think I'll just make it an error level message and include the url.

kcover:
build now

kcover:
fixed formatting
build now

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/operation/impl/ResourceRequestByDerivedUri.java
```diff
@@ -0,0 +1,59 @@
...
+
+  protected String name;
+
+  protected URI uri;
+
+  /**
+   * Implements a ResourceRequestByDerivedUri and specifies the ${@link URI}
```
peterhuffer: ❓ `$` sign?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -639,7 +629,9 @@ private ResourceInfo getResourceInfo(
 
   private void validateResourceInfoRequest(String name, Object value)
       throws ResourceNotSupportedException {
-    if (!(ResourceRequest.GET_RESOURCE_BY_PRODUCT_URI.equals(name) && value instanceof URI)
+    if (!((ResourceRequest.GET_RESOURCE_BY_PRODUCT_URI.equals(name)
```
peterhuffer: ❓ I'm curious if there is some way we can get rid of this `validateResourceInfoRequest` method. Currently we would only allow resource requests by these 3 attributes we check here. We may be able to do a catalog query based on the incoming request attribute/value and then fetching that metacards resource.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -545,35 +546,14 @@ protected ResourceInfo getResourceInfo(
 
     validateResourceInfoRequest(name, value);
 
+    if (value instanceof URI) {
```
peterhuffer: ✏️ I think this check could go away to if we determine the URI from the queried metacard.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/ResourceOperationsTest.java
```diff
@@ -374,8 +361,6 @@ public void testGetResourceInfoWithFragment() throws Exception {
     ResourceOperations resourceOperationsSpy = spy(resourceOperations);
     resourceOperationsSpy.getResourceInfo(
         resourceRequestMock, "site", isEnterprise, federatedSite, requestProperties, false);
-    verify(resourceOperationsSpy)
-        .createPropertyStartsWithQuery(Metacard.RESOURCE_URI, "bobUri:test");
   }
```
kcover: I just realized this doesn't seem to be testing anything . Can I change this to do something meaningful or should I just delete it?

clockard: can you just change this to verify createPropertyEqualToQuery?

clockard: Same for the two cases below as well

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -306,6 +300,28 @@ public ResourceResponse retrieveResource(URI resourceURI, Map<String, Serializab
     }
   }
 
+  private URI getQualifiedUri(URI resourceURI, String qualifier) {
+    String[] queryParams = resourceURI.getQuery().split("&");
+    boolean qualifierSet = false;
+
+    for (String param : queryParams) {
+      if (param.split("=")[0].equals(ContentItem.QUALIFIER_KEYWORD)) {
```
clockard: Could we just do a `param.startsWith(ContentItem.QUALIFIER_KEYWORD+"=")`

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/ResourceOperationsTest.java
```diff
@@ -434,11 +417,34 @@ public void testGetResourceInfoWithFragmentAndMultipleResults() throws Exception
     ResourceOperations.ResourceInfo resourceInfo =
         resourceOperationsSpy.getResourceInfo(
             resourceRequestMock, "site", isEnterprise, federatedSite, requestProperties, false);
-    verify(resourceOperationsSpy)
-        .createPropertyStartsWithQuery(Metacard.RESOURCE_URI, "bobUri:test");
```
kcover: These verify statements were being used to make sure that the fragment was removed from the uri first before making the query. Since we don't do that anymore these statements are no longer needed.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -612,14 +589,30 @@ private ResourceInfo getResourceInfo(
...
+      if (uri != null) {
+        for (Result result : queryResponse.getResults()) {
+          if (uri.equals(result.getMetacard().getResourceURI())) {
+            metacard = result.getMetacard();
+            break;
+          }
         }
+      } else {
+        throw new UnsupportedQueryException(
+            "A resource request resulted in multiple metacards being returned.");
```
emmberk: This else doesn't look to me like it's lined up in the right place. Should this be 
```java
    if (queryResponse.getResults().size() > 1) {
      if (uri != null) {
        for (Result result : queryResponse.getResults()) {
          if (uri.equals(result.getMetacard().getResourceURI())) {
            metacard = result.getMetacard();
            break;
          }
        }
      }
    } else {
      throw new UnsupportedQueryException(
              "A resource request resulted in multiple metacards being returned.");
    }
```
 instead?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -543,37 +544,14 @@ protected ResourceInfo getResourceInfo(
     String name = resourceRequest.getAttributeName();
     Object value = resourceRequest.getAttributeValue();
 
-    validateResourceInfoRequest(name, value);
+    if (value instanceof URI) {
+      resourceUri = (URI) value;
+    }
```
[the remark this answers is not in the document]
kcover: part of this change is making this method able to perform a query with any metacard attribute, so we don't want to type check the values anymore.

catalog/core/catalog-core-api-impl/src/main/java/ddf/catalog/content/data/impl/ContentItemImpl.java
```diff
@@ -141,11 +141,9 @@ public ContentItemImpl(
     }
     try {
       this.mimeType = new MimeType(this.mimeTypeRawData);
+      uri = new URI(CONTENT_SCHEME, this.id, this.qualifier);
     } catch (MimeTypeParseException e) {
       LOGGER.debug("Unable to create MimeType from raw data {}", mimeTypeRawData);
-    }
-    try {
-      uri = new URI(CONTENT_SCHEME, this.id, this.qualifier);
```
emmberk: Why was this logic changed? If we are unable to set `this.mimeType` do we still want to continue to set `this.uri`?

catalog/core/catalog-core-contentresourcereader/src/main/java/org/codice/ddf/catalog/content/resource/reader/ContentResourceReader.java
```diff
@@ -156,19 +156,6 @@ public ResourceResponse retrieveResource(URI resourceUri, Map<String, Serializab
...
-          try {
-            resourceUri =
-                new URI(
-                    resourceUri.getScheme(),
-                    resourceUri.getSchemeSpecificPart(),
-                    (String) arguments.get(ContentItem.QUALIFIER_KEYWORD));
-          } catch (URISyntaxException e) {
-            throw new ResourceNotFoundException("Unable to create with qualifier", e);
-          }
-        }
```
emmberk: ✏️ extra line?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -817,4 +778,41 @@ protected void validateGetResourceRequest(ResourceRequest getResourceRequest)
...
+        }
+      }
+    }
+    if (urisWithMatchingQualifier.isEmpty()) {
+      throw new ResourceNotFoundException(
+          String.format(
+              "Couldn't find a derived URI with the qualifier \"%s\" on metacard \"%s\"",
+              qualifier, metacard.getId()));
+    }
+    if (urisWithMatchingQualifier.size() > 1) {
```
emmberk: Could you throw the exception immediately in the loop when you find multiple URIs with the qualifier? 
```java
  private URI getDerivedResourceUriForQualifier(Metacard metacard, String qualifier) throws ResourceNotFoundException {
    URI derivedResourceUriWithMatchingQualifier = null;
    final Attribute derivedResourceUriAttribute = metacard.getAttribute(Core.DERIVED_RESOURCE_URI);

    if (derivedResourceUriAttribute == null) {
      for (Serializable value : derivedResourceUriAttribute.getValues()) {
        if (value instanceof String) {
          try {
            URI uri = new URI((String) value);
            if (uri.getFragment().equals(qualifier)) {
              if (derivedResourceUriWithMatchingQualifier != null) {
                throw new ResourceNotFoundException(
                        String.format(
                                "Found multiple URIs with the qualifier \"%s\" on metacard \"%s\"",
                                qualifier, metacard.getId()));
              }
              derivedResourceUriWithMatchingQualifier = uri;
            }
          } catch (URISyntaxException e) {
            LOGGER.debug(
                    String.format(
                            "Found derived URI without the proper syntax \"%s\" on metacard \"%s\"",
                            value, metacard.getId()));
          }
        }
      }
    }

    if (derivedResourceUriWithMatchingQualifier == null) {
      throw new ResourceNotFoundException(
          String.format(
              "Could not find a derived URI with the qualifier \"%s\" on metacard \"%s\"",
              qualifier, metacard.getId()));
    }

    return derivedResourceUriWithMatchingQualifier;
  }
```

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -817,4 +778,41 @@ protected void validateGetResourceRequest(ResourceRequest getResourceRequest)
...
+      for (Serializable ser : derivedResourcesAttribute.getValues()) {
+        if (ser instanceof String) {
+          try {
+            URI uri = new URI((String) ser);
+
+            if (uri.getFragment().equals(qualifier)) {
+              urisWithMatchingQualifier.add(uri);
+            }
+          } catch (URISyntaxException e) {
+            LOGGER.error(
```
emmberk: This should probably be at the debug level: https://codice.atlassian.net/wiki/spaces/DDF/pages/7045131/Log+Levels

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/ResourceOperationsTest.java
```diff
@@ -434,11 +421,85 @@ public void testGetResourceInfoWithFragmentAndMultipleResults() throws Exception
...
     assertThat(resourceInfo.getResourceUri(), equalTo(testUri));
```
emmberk: Is this `assertThat` still needed?

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/ResourceOperationsTest.java
```diff
@@ -401,14 +392,10 @@ public void testGetResourceInfoWithEscapedCharacters() throws Exception {
...
-    verify(resourceOperationsSpy)
-        .createPropertyStartsWithQuery(
-            Metacard.RESOURCE_URI,
-            "scheme:///" + URLEncoder.encode("part1?file://some/pathto/file"));
     assertThat(resourceInfo.getResourceUri(), equalTo(uri));
   }
 
-  @Test
+  @Test(expected = ResourceNotFoundException.class)
   public void testGetResourceInfoWithFragmentAndMultipleResults() throws Exception {
```
emmberk: ❓ Why does this unit test result in a `ResourceNotFoundException`?

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
```diff
@@ -306,6 +303,26 @@ public ResourceResponse retrieveResource(URI resourceURI, Map<String, Serializab
     }
   }
 
+  private URI setQualifierParamForUri(URI resourceURI, String qualifier)
```
emmberk: ✏️ This might be a little clearer if it was just a void method that set the qualifier on the `resourceURI` argument. Or, maybe renaming the method to something like `createUriWithQualifier` could work too.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/ResourceOperationsTest.java
kcover: Before if we got more than one resource back we would just take the first one in the array. Now, We want to throw an exception so this test will make sure that happens.

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
kcover: The URI class provides no way for me to modify the URI otherwise I would just append the qualifier. I think the name works here though, createUriWithQualifier might sound like we're making an entirely different URI.

kcover:
build now

kcover:
rebased with master
build now

kcover:
added dependency
build now

kcover:
embedded dependency
build now

catalog/spatial/ogc/spatial-ogc-urlresourcereader/pom.xml
```diff
@@ -106,7 +114,9 @@
                             ${project.artifactId};blueprint.graceperiod:=true;blueprint.timeout:=604800000
                         </Bundle-SymbolicName>
                         <Embed-Dependency>
-                            platform-util
+                            platform-util,
+                            httpcore,
+                            httpclient
```
oconnormi: Hopefully this causes the ci build to work now. I wonder if it would actually make more sense to embed the catalog-core-urlresourcereader instead of embedding its embedded transitive dependencies

oconnormi:
I think this was affected by an outage the other day. Trying again.

Build Now

kcover:
Changed code to avoid troublesome dependencies
build now

kcover:
Fixed sonarQube issues
build now

kcover:
fixed formatting
build now

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -598,56 +574,53 @@ protected ResourceInfo getResourceInfo(
...
-    URI resourceUri = uri;
-    QueryResponse queryResponse = queryOperations.query(queryRequest, null, true, fanoutEnabled);
-    if (queryResponse.getResults().isEmpty()) {
-      throw new ResourceNotFoundException(
-          "Could not resolve source id for URI by doing a URI based query.");
+    URI resourceUri;
+
+    if (queryResponse.getResults().size() > 1) {
+      throw new UnsupportedQueryException(
+          "A resource request resulted in multiple metacards being returned.");
```
pklinef: ✏️ It would be useful to include or log which attribute and value was not unique.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -817,4 +780,44 @@ protected void validateGetResourceRequest(ResourceRequest getResourceRequest)
           "Cannot perform getResource with null attribute value or null attributeName, either passed in from endpoint, or as output from PreResourcePlugin");
     }
   }
+
+  // Ignoring cognitive complexity to maintain error catching
```
pklinef: ✏️ I think you can reduce the cognitive complexity while still catching errors.  Here are some articles that I think might help.

https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html
https://chrismm.com/blog/writing-good-code-reduce-the-cognitive-load/

catalog/core/catalog-core-urlresourcereader/src/main/java/ddf/catalog/resource/impl/URLResourceReader.java
pklinef: I am a fan of using better verbs over `get` or `set` since they have a meaning for Java beans.  It is also misleading with this function since nothing is actually set in the parent class and could be static.

kcover:
made improvements
build now

shaundmorris:
@kcover is this PR blocked on anything? Seems like a lot of reviewing and effort went into it.

kcover:
@shaundmorris I think it's done but it goes with the CDM changes @peterhuffer is making so I think he's gonna take over and make sure it works with what he's done. After that it should be good to merge.

shaundmorris:
@peterhuffer, @kcover  just verifying the decision was to hold off on this PR since I believe the CDM changes that were blocking this PR.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/ResourceOperations.java
```diff
@@ -817,4 +782,50 @@ protected void validateGetResourceRequest(ResourceRequest getResourceRequest)
...
+        LOGGER.debug(
+            String.format(
+                "Found derived URI without the proper syntax \"%s\" on metacard \"%s\"",
+                ser, metacard.getId()));
+        continue;
+      }
+      if (!uri.getFragment().equals(qualifier)) {
+        continue;
+      }
+      if (uriWithMatchingQualifier != null) {
```
bakejeyner: ❓ Will `uriWithMatchingQualifier` always be null at this point? It looks like it is assigned null at the beginning of this method, and is untouched until this check.

kcover: It will be null the first time through but we're in a loop here so if we find a URI that contains the given qualifier we save it in uriWithMatchingQualifier. If we then come back through and find another one, then somehow the Uri got added to the metacard twice which is a no-no, and we'll then run through this if-statement. Looking back on it I should've put more effort into making this method readable.

rzwiefel:
@kcover @peterhuffer Any update on this PR? I'm not sure which PR this was waiting for since it doesn't looked like it was linked. Can this be updated and pushed through or are there any other concerns currently blocking this?

rzwiefel:
CDM changes have been merged but looking to get this prioritized still. Tagged with `Needs Champion` and closing for now until its reprioritized.

PR #3126, opened 2018-04-06
DDF-3735 Upgrades Cesium to fix zoom issue

andrewkfiedler:
#### What does this PR do?
 - Cesium was having issues maintaining the correct coordinate frame of reference during certain zooms, particularly when the user zoomed too far out.  It appears to be fixed in the latest version.

#### Who is reviewing it? 
@jlcsmith 
@djblue 
@Variadicism 
@bdeining 

Before using the version with upgraded cesium, verify that on the old cesium zooming far out will cause results to become unclickable.  Then use the upgraded cesium and verify that this behavior disappears.

[DDF-3735](https://codice.atlassian.net/browse/DDF-3735)

### DIFF
--- a/ui/packages/catalog-ui-search/package.json
+++ b/ui/packages/catalog-ui-search/package.json
@@ -45,7 +45,7 @@
     "backbone.modelbinder": "1.1.0",
     "backbone.paginator": "2.0.5",
     "bootstrap": "3.3.7",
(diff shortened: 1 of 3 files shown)


bdeining:
verified that results are clickable when zoomed out in IE11 / Chrome / Firefox

PR #3132, opened 2018-04-10
DDF-3743 Updates dev server to handle ws

andrewkfiedler:
#### What does this PR do?
 - The websockets were failing in certain browsers (Firefox, Edge, not Chrome though) when auth wasn't passed.

#### Who is reviewing it? 
@brendan-hofmann
@djblue 
@mackncheesiest 
@Variadicism 

Run the dev server and verify it defaults to admin:admin for the basic auth.  Verify you can also pass your own auth through the auth variable.

Looks like this issue was introduced when we upgraded to the latest webpack, so it's possible it's a bug in one of the newer versions of the webpack dev server or the http-proxy-middleware, etc.  Or they've simply updated how they expect it to be used.

[DDF-3743](https://codice.atlassian.net/browse/DDF-3743)

### DIFF
--- a/ui/packages/ace/lib/start.js
+++ b/ui/packages/ace/lib/start.js
@@ -1,6 +1,7 @@
 const open = require('opn')
 const webpack = require('webpack')
 const WebpackDevServer = require('webpack-dev-server')
+const chalk = require('chalk')
 
 const webpackConfig = require('./webpack.config')
 
@@ -8,7 +9,10 @@ module.exports = ({ args, pkg }) => {
   const config = webpackConfig({
     env: process.env.NODE_ENV || args.env || 'development',
     main: pkg.main,
-    alias: pkg.alias
+    alias: pkg.alias,
+    auth: args.auth ||
+      console.log(chalk.yellow('WARNING: using default basic auth (admin:admin)! See options for how to override this.')) ||
+      'admin:admin'
   })
 
   WebpackDevServer.addDevServerEntrypoints(config, config.devServer)
--- a/ui/packages/ace/package.json
+++ b/ui/packages/ace/package.json
@@ -24,6 +24,7 @@
     "babel-preset-latest": "6.24.1",
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-0": "6.24.1",
+    "chalk": "2.3.2",
     "cheerio": "0.22.0",
     "commander": "2.15.1",
     "console-polyfill": "0.3.0",
--- a/ui/packages/ace/bin.js
+++ b/ui/packages/ace/bin.js
@@ -76,6 +76,7 @@ program
 program
   .command('start')
   .description('start the dev server')
+  .option('-a, --auth <auth>', 'auth <username:password> (default: admin:admin)')
   .option('-e, --env <env>', 'build environment <development|test|production>')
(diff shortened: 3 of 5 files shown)


ui/packages/ace/lib/start.js
```diff
@@ -1,14 +1,19 @@
...
+const warn = clc.yellowBright.blink
 
 module.exports = ({ args, pkg }) => {
   const config = webpackConfig({
     env: process.env.NODE_ENV || args.env || 'development',
     main: pkg.main,
-    alias: pkg.alias
+    alias: pkg.alias,
+    auth: args.auth ||
+      console.log(warn('WARNING: using default basic auth (admin:admin)! See options for how to override this.')) ||
```
brendan-hofmann: ❓ This syntax is new to me. I am guessing `console.log` or `warn` evaluates/returns a falsey value, so it falls through to the third case?

mackncheesiest: I don't know if it has a special name, but I've had success searching for `Javascript double pipe` and getting [results](https://stackoverflow.com/questions/2802055/what-does-the-construct-x-x-y-mean/2802064#2802064) about it in the past

ui/packages/ace/package.json
```diff
@@ -63,6 +63,7 @@
     "worker-loader": "1.1.1"
   },
   "devDependencies": {
+    "cli-color": "1.2.0",
```
djblue: How would you feel about using [chalk](https://www.npmjs.com/package/chalk)? It's already part of our dependency graph, so we wouldn't be bringing in new dependencies.

ui/packages/ace/lib/start.js
```diff
@@ -1,14 +1,19 @@
 const open = require('opn')
 const webpack = require('webpack')
 const WebpackDevServer = require('webpack-dev-server')
+const clc = require('cli-color')
 
 const webpackConfig = require('./webpack.config')
+const warn = clc.yellowBright.blink
```
Variadicism: `yellowBright` corresponds to the color code for bright yellow as opposed to "bright/bold" + "yellow". The bright yellow color code is not widely used and not supported by some terminals; `yellow.bold` is more commonly used and better supported.
`blink` is rarely supported and would be rather annoying even if it is supported.

ui/packages/ace/lib/start.js
mackncheesiest: I think it evaluates to `undefined`

mackncheesiest: image

ui/packages/ace/package.json
```diff
@@ -24,6 +24,7 @@
     "babel-preset-latest": "6.24.1",
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-0": "6.24.1",
+    "chalk": "2.3.2",
```
shaundmorris: might be worth an nsp check on this version, if it is new.

brendan-hofmann: I didn't find any known vulnerabilities, other than hoek.

brendan-hofmann:
build now

ui/packages/ace/package.json
shaundmorris: cool thanks

PR #3140, opened 2018-04-11
DDF-3748 Updates how cesium coordinates are retrieved

andrewkfiedler:
#### What does this PR do?
- The previous method did not work in IE11.
- Based on https://groups.google.com/forum/#!topic/cesium-dev/p-Rcg0JeH7Y it would
appear using pickEllipsoid is better and will result in more accurate
positions.

#### Who is reviewing it? 
@jlcsmith
@brendan-hofmann 
@mackncheesiest 
@adimka 

Verify that coordinates are displayed when mousing over the map in IE11.

[DDF-3748](https://codice.atlassian.net/browse/DDF-3748)


brendan-hofmann:
❓ The referenced thread indicates that `scene.globe.pick` is preferable when using a terrain provider. Won't always using `pickEllipsoid` produce inaccurate results if a terrain provider is configured? Should we switch between the two depending on the availability of terrain information?

bdeining:
build now

PR #3159, opened 2018-04-13
DDF-3660: Move in place content monitoring to CreateRequests

peterhuffer:
#### What does this PR do?
- Introduce simple Harvester API
- Creates Directory and WebDav harvesters
- Current Harvest Manager will be used to manage content monitoring services
- Remove CamelContext and rely on Camel's OSGi dynamic component discovery

Still need to add tests and updated documentation

#### Who is reviewing it? 
@emmberk @kcover @oconnormi @brendan-hofmann @rzwiefel 

@clockard
@lessarderic

- Verify restart behavior with CDMs (Move, Delete, and In Place). Creates, updates and deletes.
- Verify that the catalog accurately reflects the remote content after shutting down, making changes in remote source, and starting back up.

[DDF-3660](https://codice.atlassian.net/browse/DDF3660)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -201,7 +181,7 @@ public void init() {
   private Object configure() {
     if (StringUtils.isEmpty(monitoredDirectory)) {
       LOGGER.warn("Cannot setup camel route - must specify a directory to be monitored");
-      return null;
+      throw new IllegalArgumentException("Property {monitoredDirectory} cannot be empty");
```
kcover: Seems like just the exception would be enough, do we need the warning too?

bdeining: Agreed; I don't think we should log AND throw here

rzwiefel:
Should this be split into separate commits? EG, the new harvesting api as an isolated, independent commit

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/PollingHarvester.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Starts the {@link ScheduledExecutorService} which will call the {@link #poll()} method for
+   * every interval defined by {@code pollInterval}.
+   */
+  public void init() {
+    executorService =
+        Executors.newSingleThreadScheduledExecutor(
+            StandardThreadFactoryBuilder.newThreadFactory(
+                "org.codice.ddf.catalog.harvest.pollingHarvester"));
+    executorService.scheduleAtFixedRate(this::poll, 5, pollInterval, timeUnit);
```
rzwiefel: ❓ Why the 5 initial delay?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,39 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.harvest;
+
+/**
+ * A {@code Harvester} is responsible for fetching remote resources and notifying any registered
+ * {@link Listener}s of create, update and delete events.
+ *
+ * <p>In the event that a {@code Harvester} goes down, it must remember the last state of the
+ * external server it was harvesting from. Upon a {@code Harvester} coming back up, it should only
```
lessarderic: ❓I’m not sure how strict we can be here. There’s always a chance that an `Harvester` ends up sending the same event twice to a its listeners no matter what, so we may want to change this to say that it is best effort and put a constraint on the listeners that they need to be idempotent. Thoughts?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,39 @@
...
+ * notify registered listener of new events.
+ */
+public interface Harvester {
+
+  /**
+   * Registers a new {@link Listener} to receive events from the remote server being harvested from.
```
lessarderic: ✏️ Source instead of server?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A {@code Harvester} is responsible for fetching remote resources and notifying any registered
+ * {@link Listener}s of create, update and delete events.
+ *
+ * <p>In the event that a {@code Harvester} goes down, it must remember the last state of the
+ * external server it was harvesting from. Upon a {@code Harvester} coming back up, it should only
+ * notify registered listener of new events.
+ */
+public interface Harvester {
```
lessarderic: ❓Should we assign a unique ID to each harvester and pass it down to the listeners? This might be useful down the road when we start supporting rules, i.e., allow users to write rules based on what harvester the resource came from. This could also be handled by the listeners if we prefer, which may be a better abstraction.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Registers a new {@link Listener} to receive events from the remote server being harvested from.
+   *
+   * @param listener the {@link Listener} to register
+   */
+  void registerListener(Listener listener);
```
lessarderic: ❓Should we throw an `InvalidArgumentException` if the same listener is added twice?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Unregisters a {@link Listener}.
+   *
+   * @param listener the {@link Listener} to unregister
```
lessarderic: ❓ Exception if listener doesn’t exist, or no-op? We should document the behavior.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
lessarderic: ❓What happens when a listener registers the first time? Does it receive events for everything that already exists? Should that be an option the listeners provide when they register?
❓Does a harvester need to remember what events were sent to what listener? For instance if an event is sent to listener A but not B and the system goes down, is the expectation that only B will be sent the event upon restart?
❓How are listeners re-registered after a restart?

lessarderic: ❓To avoid the above problems, should we just allow a single listener and assume it’s always the same one that registers and unregisters?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Listener.java
```diff
@@ -0,0 +1,42 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.harvest;
+
+/** Responds to events from a {@link Harvester}. */
+public interface Listener {
+
+  /**
+   * Called when a {@link Harvester} receives a newly created file on a remote server.
```
lessarderic: ✏️ Source instead of server? Also in following comments.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Listener.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Called when a {@link Harvester} receives a deleted file on a remote server. On a delete event,
+   * the resource will not be available, as it has already been deleted. {@link Listener}s should
+   * use the {@link HarvestedResource#getCorrelationId()} method to identify the current resource
+   * being processed.
+   *
+   * @param resource the {@link HarvestedResource} deleted on the remote server
+   */
+  void onDelete(HarvestedResource resource);
```
lessarderic: ❗️ Since most of the `HarvestedResource` methods will not return anything useful (`null` most likely, which would force us to add `@Nullable` on a lot of its methods just for this case), we should consider passing in only the unique ID and/or the URI of the delete resource.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/HarvestedResource.java
```diff
@@ -11,37 +11,35 @@
...
+public interface HarvestedResource {
+
+  Map<String, Object> getProperties();
+
+  /**
+   * A globally unique identifier for this {@code HarvestedResource}.
+   *
+   * @return a globally unique identifier
+   */
+  String getCorrelationId();
```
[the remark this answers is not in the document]
peterhuffer: Yes, I can use the URI. 👍 I added these at different points.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
peterhuffer: I like the idempotent idea.

Currently each harvester has its own listener. A listener would only receive events upon registration, no past events yet. I ended up having to register listeners before first harvest to avoid this right now. For the first pass I think we should allow single listener like you suggested and build on it later.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
peterhuffer: Since we are going with a 1:1 harvester and listener I don't think this is necessary right now, but definitely needs to be done in the future.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
peterhuffer: I can throw an IllegalArgumentException from the current implementations and document expected behavior of adding the same listener twice. I see the `InvalidArgumentException` in a sun package, which intellij says not to use.

lessarderic: Sorry, I meant `IllegalArgumentException`.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
lessarderic: ❓Should we add a note in the comments about the fact that `Harvester` implementations may thread out calls to the listeners if needed and that listeners don't need to?

catalog/harvest/catalog-harvest-common/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
```
lessarderic: ❗️We should update the coverage when all the unit tests have been updated, or add a `TODO` and refer to another PR or DDF ticket.

catalog/harvest/catalog-harvest-manager/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>harvest</artifactId>
+        <groupId>ddf.catalog.harvest</groupId>
+        <version>2.12.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <name>DDF :: Catalog :: Harvest :: Manager</name>
+    <artifactId>catalog-harvest-manager</artifactId>
```
lessarderic: ❓To be consistent with my comment about the `HarvesterManager` class, we should consider renaming the Maven module `catalog-source-harvester` or `catalog-harvest-source`.

catalog/harvest/catalog-harvest-manager/src/main/java/org/codice/ddf/catalog/harvest/manager/HarvesterManager.java
```diff
@@ -0,0 +1,223 @@
...
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.harvest.manager;
```
lessarderic: ❓To be consistent, should we consider renaming the java package to `org.codice.ddf.catalog.harvest.source`?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -54,20 +55,22 @@
...
+
+  public FileSystemPersistenceProvider(String mapName, String persistencePath) {
     this.mapName = mapName;
+    this.persistencePath = persistencePath;
     File dir = new File(getPersistencePath());
-    if (!dir.exists()) {
-      if (!dir.mkdir()) {
-        LOGGER.debug("Unable to create directory: {}", dir.getAbsolutePath());
-      }
+    if (!dir.exists() && !dir.mkdir()) {
```
lessarderic: ❗️Shouldn't we throw an exception if we fail instead of just logging?
❓What would happen if the directory gets created, nothing gets written to it, the system shuts down and this method gets called again when the system restarts? Would this prevent things from being persisted?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -93,16 +96,22 @@ String getMapStorePath() {
...
+      boolean created = dir.mkdirs();
+      if (created || dir.exists()) {
+        LOGGER.trace("Entering: store - key: {}", key);
+        try (OutputStream file =
+                new FileOutputStream(getMapStorePath() + key + PERSISTED_FILE_SUFFIX);
+            OutputStream buffer = new BufferedOutputStream(file);
+            ObjectOutputStream output = new ObjectOutputStream(buffer)) {
+          output.writeObject(value);
+        } catch (IOException e) {
+          LOGGER.debug("IOException storing value in cache with key = " + key, e);
```
lessarderic: ❗️Failing to write to the file system is pretty bad and there may be something an admin can do about it so we should consider logging something higher than `debug` here and on L113.

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -93,16 +96,22 @@ String getMapStorePath() {
...
-    LOGGER.trace("Entering: store - key: {}", key);
-    try (OutputStream file = new FileOutputStream(getMapStorePath() + key + PERSISTED_FILE_SUFFIX);
-        OutputStream buffer = new BufferedOutputStream(file);
-        ObjectOutputStream output = new ObjectOutputStream(buffer)) {
-      output.writeObject(value);
-    } catch (IOException e) {
-      LOGGER.debug("IOException storing value in cache with key = " + key, e);
+
+    if (new File(getPersistencePath()).canWrite()) {
+      boolean created = dir.mkdirs();
```
lessarderic: ❓Any reason we can't create this directory once in the constructor instead of making this call every time we write a file to disk?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -93,16 +96,22 @@ String getMapStorePath() {
...
+      if (created || dir.exists()) {
+        LOGGER.trace("Entering: store - key: {}", key);
+        try (OutputStream file =
+                new FileOutputStream(getMapStorePath() + key + PERSISTED_FILE_SUFFIX);
```
lessarderic: ❓Should we have a common `private` method that takes in the `key` and suffix and returns the file path instead of repeating `getMapStorePath() + key + PERSISTED_FILE_SUFFIX` multiple times?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -117,7 +126,9 @@ public void storeAll(Map<String, Object> keyValueMap) {
   public void delete(String key) {
     File file = new File(getMapStorePath() + key + PERSISTED_FILE_SUFFIX);
     if (file.exists()) {
-      if (!file.delete()) {
+      try {
+        Files.delete(file.toPath());
+      } catch (IOException e) {
         LOGGER.debug("File was unable to be deleted: {}", file.getAbsolutePath());
```
lessarderic: ❓Is this going to cause a problem upon restart? Should we log something higher than `debug`?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -204,7 +209,9 @@ public void clear() {
     File[] files = new File(getMapStorePath()).listFiles(getFilenameFilter());
     if (files != null) {
       for (File file : files) {
-        if (!file.delete()) {
+        try {
+          Files.delete(file.toPath());
+        } catch (IOException e) {
           LOGGER.debug("File was unable to be deleted: {}", file.getAbsolutePath());
```
lessarderic: ❓Is this going to cause a problem upon restart? Should we log something higher than `debug`?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
peterhuffer: I copied this class over from the CDM module - I'm not sure why it was originally like this, but I would guess this would cover the case where a user deletes the directory and then a caller tries to call store.

catalog/core/catalog-core-directorymonitor/pom.xml
```diff
@@ -25,7 +25,7 @@
 
     <artifactId>catalog-core-directorymonitor</artifactId>
     <name>DDF :: Catalog :: Core :: Content Directory Monitor</name>
-    <packaging>bundle</packaging>
+    <packaging>jar</packaging>
```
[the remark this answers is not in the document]
brendan-hofmann: Is there a reason not to make an experimental interface for them to register under? We can iterate on the harvester<->listener aspect, as long as we don't start down this path of embedding jars. It's an OSGi antipattern that we'd be better off avoiding.

catalog/core/catalog-core-directorymonitor/src/main/java/org/codice/ddf/catalog/content/monitor/ContentDirectoryMonitor.java
```diff
@@ -366,32 +324,33 @@ private boolean isMyRoute(String routeId) {
...
-          stringBuilder.append("&readLockTimeout=" + (2 * readLockIntervalMilliseconds));
-          stringBuilder.append("&readLockCheckInterval=" + readLockIntervalMilliseconds);
-
-          /* File Exclusions */
-          String exclusions = getBlackListAsRegex();
-          if (StringUtils.isNotBlank(exclusions)) {
-            stringBuilder.append("&exclude=" + exclusions);
-          }
+
+        stringBuilder.append("file:");
```
brendan-hofmann: ✏️ These blocks could be chained. e.g.
```java
stringBuilder
  .append("something")
  .append("other")
  .append(things);
```
or whatever google-format would do to that.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,46 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.harvest;
+
+/**
+ * A {@code Harvester} is responsible for fetching remote resources and notifying any registered
+ * {@link Listener}s of create, update and delete events.
+ *
+ * <p>In the event that a {@code Harvester} goes down, it should make a best effort to remember the
```
brendan-hofmann: ✏️ I know it's just javadoc, but I'd prefer we close the `<p>` tags here and elsewhere.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,46 @@
...
+ * up, it should notify registered {@link Listener}s of new events.
+ *
+ * <p><b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface Harvester {
+
+  /**
+   * Registers a new {@link Listener} to receive events from the remote source being harvested from.
+   * Newly {@link Listener}s will only receive new events from this {@code Harvester}.
```
brendan-hofmann: ✏️ Newly...registered?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param listener the {@link Listener} to register
+   * @throws IllegalArgumentException if the {@link Listener} is already registered or cannot be
```
brendan-hofmann: ❓ Can registering be idempotent like unregistering is below?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
brendan-hofmann: If the directories already exist, this is equivalent to `dir.exists()`. If any nodes do not exist, this creates them. We could pull up the `dir` itself though. https://docs.oracle.com/javase/7/docs/api/java/io/File.html#mkdirs()

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
```diff
@@ -0,0 +1,131 @@
...
+      return (FileAlterationObserver) fileSystemPersistenceProvider.loadFromPersistence(key);
+    }
+
+    LOGGER.trace("no existing file observer for persistence key [{}], creating new observer", key);
+    return new FileAlterationObserver(dir);
+  }
+
+  @Override
+  public void poll() {
+    fileAlterationObserver.addListener(fileListener);
```
brendan-hofmann: ❓ Do we still need this pattern? It was used because the listener is not serializable, and the observer was being used as the backing store.

catalog/harvest/catalog-harvest-manager/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+                        <Export-Package/>
+                        <Import-Package>
+                            !org.apache.tools.ant,
+                            !org.apache.tools.ant.types,
+                            org.apache.commons.io.comparator,
+                            *
+                        </Import-Package>
+                        <Embed-Dependency>
+                            catalog-harvest-persistentlistener,
+                            catalog-harvest-file,
```
brendan-hofmann: ❗️ Same as @lessarderic's question about the jars, I don't think the impls should be embedded. They should export services instead.

catalog/harvest/catalog-harvest-manager/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -1,40 +1,39 @@
...
+    <reference id="metacardOnlyAdapter" interface="org.codice.ddf.catalog.harvest.Adapter"/>
 
     <cm:managed-service-factory
-            id="contentDirectoryMonitor_MSF"
-            factory-pid="org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor"
-            interface="org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor">
-        <cm:managed-component class="org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor"
+        id="contentMonitor_MSF"
+        factory-pid="org.codice.ddf.catalog.harvest.manager.HarvesterManager"
+        interface="org.codice.ddf.catalog.harvest.manager.HarvesterManager">
```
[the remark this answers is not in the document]
brendan-hofmann: While technically you can do this (the interface is inferred), I agree that it's not the best idea.

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -138,14 +155,13 @@ public Object load(String key) {
     return null;
   }
 
-  Object loadFromPersistence(String key) {
-    File file = new File(getMapStorePath() + key + PERSISTED_FILE_SUFFIX);
+  public Object loadFromPersistence(String key) {
```
emmberk: ✏️ `@Nullable`

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
```diff
@@ -0,0 +1,131 @@
...
+    inititialListeners.forEach(this::registerListener);
+    fileSystemPersistenceProvider = new FileSystemPersistenceProvider("harvest/directory");
+    fileAlterationObserver = getCachedObserverOrCreate(persistenceKey, dir);
+    fileListener = new DirectoryHarvesterListenerAdaptor(listeners);
+
+    super.init();
+  }
+
+  private FileAlterationObserver getCachedObserverOrCreate(String key, String dir) {
+    if (fileSystemPersistenceProvider.loadAllKeys().contains(key)) {
```
emmberk: ❓  Can `fileSystemPersistenceProvider.loadAllKeys()` return `null`?

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvesterListenerAdaptor.java
```diff
@@ -0,0 +1,70 @@
...
+      listeners.forEach(listener -> listener.onUpdate(harvestedResource));
+    }
+  }
+
+  @Override
+  public void onFileDelete(final File file) {
+    listeners.forEach(listener -> listener.onDelete(file.toURI().toASCIIString()));
+  }
+
+  private HarvestedResource createHarvestedResource(File file) {
```
emmberk: ✏️ `@Nullable`

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
emmberk: ✏️ `LOGGER.debug("IOException storing value in cache with key = {}", key, e);`

catalog/harvest/catalog-harvest-webdav/src/main/java/org/codice/ddf/catalog/harvest/webdav/DavAlterationObserver.java
```diff
@@ -11,7 +11,7 @@
  * License is distributed along with this program and can be found at
  * <http://www.gnu.org/licenses/lgpl.html>.
  */
-package org.codice.ddf.catalog.content.monitor;
+package org.codice.ddf.catalog.harvest.webdav;
```
emmberk: ✏️ `initialize` on line 107

catalog/harvest/catalog-harvest-manager/src/main/java/org/codice/ddf/catalog/harvest/manager/HarvesterManager.java
```diff
@@ -0,0 +1,225 @@
...
+        String[] incomingAttrOverrides = (String[]) o;
+        setAttributeOverrides(Arrays.asList(incomingAttrOverrides));
+      }
+
+      destroy(0);
+      init();
+    }
+  }
+
+  public void destroy(int code) {
```
emmberk: ✏️ Can the `code` argument be removed?

peterhuffer: Unfortunately no. The MSF `destroy-method` takes this as an argument. If it is not here, then there are odd behaviors on when this method is called. So far I know that code=1 when the configuration is deleted, and code=2 when the system is being shutdown.

bdeining: I think requiring a code is actually a bug : https://issues.apache.org/jira/browse/ARIES-1436.  Upgrading CM would eliminate the need for this argument

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Listener.java
```diff
@@ -0,0 +1,49 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.harvest;
+
+/**
+ * Responds to events from a {@link Harvester}. <b>{@code Listener}s are idempotent, receiving the
+ * same create, update, or delete event multiple times has no effect.</b>
+ *
+ * <p>{@link Harvester}s may thread out calls to {@code Listener}s if needed, so {@code Listener}s
+ * don't have to.
```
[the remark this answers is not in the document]
peterhuffer: Listeners don't have to make thread out their calls since they may already be in a separate thread. Added clarification

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/PollingHarvester.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.codice.ddf.catalog.harvest.Harvester;
+import org.codice.ddf.platform.util.StandardThreadFactoryBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A {@link Harvester} who is capable of polling a source for content. */
```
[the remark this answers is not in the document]
peterhuffer: Some external context store.

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
peterhuffer: The observer is still keeping the state of the directory being monitored and needs to be persisted.

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/PollingHarvester.java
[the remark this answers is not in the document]
emmberk: Might be worthwhile to change it to say `an external context store` so that it won't be confused with a DDF Source.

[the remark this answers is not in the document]
peterhuffer: Typo'd. Meant `content` not `context`

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
peterhuffer: No, it will return an empty set if there is nothing cached in the provider.

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
stustison: You need to write an audit whenever you create a new file / directory or read an existing file / directory. You might also want to catch the AccessControlException that `dir.exists()` can throw and give a more helpful error message.

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
stustison: This needs to be audited.

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -93,16 +97,24 @@ String getMapStorePath() {
...
-        ObjectOutputStream output = new ObjectOutputStream(buffer)) {
-      output.writeObject(value);
-    } catch (IOException e) {
-      LOGGER.debug("IOException storing value in cache with key = " + key, e);
+
+    if (new File(getPersistencePath()).canWrite()) {
+      boolean created = dir.mkdirs();
+      if (created || dir.exists()) {
+        LOGGER.trace("Entering: store - key: {}", key);
+        try (OutputStream file = new FileOutputStream(getPersistedFilePathForKey(key));
```
stustison: Audits?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -115,10 +127,15 @@ public void storeAll(Map<String, Object> keyValueMap) {
 
   @Override
   public void delete(String key) {
-    File file = new File(getMapStorePath() + key + PERSISTED_FILE_SUFFIX);
+    File file = new File(getPersistedFilePathForKey(key));
     if (file.exists()) {
-      if (!file.delete()) {
-        LOGGER.debug("File was unable to be deleted: {}", file.getAbsolutePath());
+      try {
+        Files.delete(file.toPath());
```
stustison: Audits?

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/FileSystemPersistenceProvider.java
```diff
@@ -200,12 +210,21 @@ public boolean accept(File file, String name) {
...
+  }
+
   public void clear() {
     File[] files = new File(getMapStorePath()).listFiles(getFilenameFilter());
     if (files != null) {
       for (File file : files) {
-        if (!file.delete()) {
-          LOGGER.debug("File was unable to be deleted: {}", file.getAbsolutePath());
+        try {
+          Files.delete(file.toPath());
```
stustison: Audits? You may also want to catch the AccessControlExceptions that can be thrown for `.listFiles` or `.delete`

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
```diff
@@ -0,0 +1,131 @@
...
+    listeners.add(listener);
+  }
+
+  @Override
+  public void unregisterListener(Listener listener) {
+    listeners.remove(listener);
+  }
+
+  private void validateDirectory(String dir) {
+    if (!harvestFile.exists()) {
```
stustison: Catch AccessControlExceptions?

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvesterListenerAdaptor.java
```diff
@@ -0,0 +1,70 @@
...
+    try {
+      return new HarvestedFile(
+          new FileInputStream(file), file.getName(), file.toURI().toASCIIString());
```
stustison: Audit.

catalog/harvest/catalog-harvest-webdav/src/main/java/org/codice/ddf/catalog/harvest/webdav/WebDavHarvester.java
```diff
@@ -0,0 +1,163 @@
...
+   */
+  public WebDavHarvester(String address, Set<Listener> initialListeners) {
+    super(5L);
+    Validate.notEmpty(address, "address {location} cannot be empty");
+
+    persistenceProvider = new FileSystemPersistenceProvider("harvest/webdav");
+    webdavListener = new HarvestedResourceListener();
+
+    initialListeners.forEach(this::registerListener);
+    persistentKey = DigestUtils.sha1Hex(address);
```
stustison: Should avoid sha1 in any new code. Use `Hashing.sha256().hashString(...)` from the google libraries instead.

catalog/harvest/catalog-harvest-webdav/src/main/java/org/codice/ddf/catalog/harvest/webdav/WebDavHarvester.java
```diff
@@ -0,0 +1,163 @@
...
+
+    @Override
+    public void onFileDelete(DavEntry entry) {
+      listeners.forEach(listener -> listener.onDelete(entry.getLocation()));
+    }
+
+    private HarvestedResource createHarvestedResource(DavEntry entry) {
+      File file;
+      try {
+        file = entry.getFile(SardineFactory.begin());
```
stustison: Audit.

catalog/harvest/catalog-harvest-api/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
```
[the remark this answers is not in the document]
lessarderic: Good point @emmberk. Didn't notice this was an API bundle. I like the idea of disabling the plugin altogether instead of setting everything to 0.0.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Listener.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.harvest;
+
+/**
+ * Responds to events from a {@link Harvester}. <b>{@code Listener}s are idempotent, receiving the
```
lessarderic: ✏️ Suggestion: _`Listener` implementations must be idempotent, i.e., receiving the same create, update or delete event multiple times for the same `HarvestedResource` must produce the same result_

catalog/harvest/catalog-harvest-common/src/main/java/org/codice/ddf/catalog/harvest/common/HarvestedFile.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  public HarvestedFile(InputStream is, String name, String uri, Map<String, Object> properties) {
+    super(is, name);
+    try {
+      this.uri = new URI(uri);
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException(String.format("Invalid URI [%s] received.", uri));
+    }
+    this.properties = properties;
```
lessarderic: ❗️We should remove the properties map (see my comment above), but if we don't, we should at least make sure the `properties` argument isn't `null`.

catalog/harvest/catalog-harvest-file/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
```
lessarderic: ❗️Should add comment/TODO to increase low coverage.

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
```diff
@@ -0,0 +1,131 @@
...
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.io.monitor.FileAlterationListenerAdaptor;
+import org.apache.commons.io.monitor.FileAlterationObserver;
+import org.codice.ddf.catalog.harvest.Listener;
+import org.codice.ddf.catalog.harvest.common.FileSystemPersistenceProvider;
+import org.codice.ddf.catalog.harvest.common.PollingHarvester;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DirectoryHarvester extends PollingHarvester {
```
lessarderic: ❗️JavaDoc.

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DirectoryHarvester.class);
+
+  private static final long DEFAULT_POLL_INTERVAL = 5;
+
+  private final Set<Listener> listeners = new HashSet<>();
```
lessarderic: ❗️This `Set` needs to be synchronized to prevent possible threading issues when listeners are being added or removed while notifications are being sent by the `DirectoryHarvesterListenerAdaptor`. Recommend taking a look at https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/event/EventListenerSupport.html.

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
```diff
@@ -0,0 +1,131 @@
...
+  public void poll() {
+    fileAlterationObserver.addListener(fileListener);
+    fileAlterationObserver.checkAndNotify();
+    fileAlterationObserver.removeListener(fileListener);
+    fileSystemPersistenceProvider.store(persistenceKey, fileAlterationObserver);
+  }
+
+  @Override
+  public void registerListener(Listener listener) {
+    if (!listeners.isEmpty()) {
```
lessarderic: ❓If we can use the Apache Commons' `EventListenerSupport` class, should we just ignore this check and allow multiple listeners?

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvesterListenerAdaptor.java
```diff
@@ -0,0 +1,70 @@
...
+import org.codice.ddf.catalog.harvest.common.HarvestedFile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DirectoryHarvesterListenerAdaptor extends FileAlterationListenerAdaptor {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(DirectoryHarvesterListenerAdaptor.class);
+
+  private final Set<Listener> listeners;
```
lessarderic: ❓Replace with Apache Commons' `EventListenerSupport`?

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvester.java
lessarderic: ❓Instead of adding and removing listeners here and doing the notifications in the `DirectoryHarvesterListenerAdaptor` class, would it make more sense to move the `registerListener` and `unregisterListener` methods to the `DirectoryHarvesterListenerAdaptor` class and simply delegate down to that class here? That way all the listener management code would be in one class and we wouldn't need to pass the listeners' list around.

catalog/harvest/catalog-harvest-manager/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit>
```
lessarderic: ❗️Add coverage comment/TODO.

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/HarvestedResourceTransformer.java
```diff
@@ -0,0 +1,189 @@
...
+        return metacardOptional.get();
+      }
+    }
+    return null;
+  }
+
+  private Optional<Metacard> doTransform(
+      InputTransformer inputTransformer, String metacardId, InputStream is, String resourceName) {
+    try {
+      if (StringUtils.isNotEmpty(metacardId)) {
```
lessarderic: ❓ Kind of bugs me that we pass down the `metacardId` all the way down here even when it's `null`. Could we pass a `BiFunction<InputTransformer, InputStream, Metacard>` instead and just call it here?

In other words, the `transform(HarvestedResource, String)` method signature could look something like:
```
private Metacard transform(HarvestedResource harvestedResource,
    BiFunction<InputTransformer, InputStream, Metacard> transformFunction)
```
we could have the `public Metacard transform...()` methods call it like this:
```
public Metacard transformHarvestedResource(HarvestedResource harvestedResource) {
  return transform(harvestedResource, (it, is) -> it.transform(is));
}
```
or
```
public Metacard transformHarvestedResource(HarvestedResource resource, String metacardId) {
  return transform(harvestedResource, (it, is) -> it.transform(is, metacardId));
}
```
and have this method simply do:
```
private Optional<Metacard> doTransform(
     InputTransformer inputTransformer,
     InputStream is,
     BiFunction<InputTransformer, InputStream, Metacard> transformFunction,
     String resourceName) {
  try {
    return Optional.of(transformFunction.apply(inputTransformer, is));
  }
```
?

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/HarvestedResourceTransformer.java
```diff
@@ -0,0 +1,189 @@
...
+          "Failed to transform resource [{}] with [{}] transformer. Trying next one.",
+          resourceName,
+          inputTransformer);
+    }
+
+    return Optional.empty();
+  }
+
+  private MimeType guessMimeTypeFor(InputStream is, String fileExt) {
+    try (TemporaryFileBackedOutputStream tfbos = new TemporaryFileBackedOutputStream()) {
```
lessarderic: ❓ Do we really need to create another `TemporaryFileBackedOutputStream` here and make a copy again? We've already done that before calling this method (L86).

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/MetacardOnlyAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+
+    this.catalogFramework = catalogFramework;
+    this.harvestedResourceTransformer = harvestedResourceTransformer;
+  }
+
+  @Override
+  public String create(HarvestedResource resource) throws HarvestException {
+    Metacard metacard = harvestedResourceTransformer.transformHarvestedResource(resource);
+
+    if (metacard == null) {
```
lessarderic: ❓ Should we throw the exception straight from `harvestedResourceTransformer.transformHarvestedResource(resource)` instead of checking for `null` and throwing here?

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/MetacardOnlyAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+          try {
+            CreateResponse response;
+            response = catalogFramework.create(createRequest);
+
+            List<Metacard> createdMetacards = response.getCreatedMetacards();
+            if (createdMetacards.size() == 1) {
+              return createdMetacards.get(0).getId();
+            } else {
+              LOGGER.debug(
+                  "Received multiple metacards for a single created resource. [{}].",
```
lessarderic: ❓ What if we didn't get anything, i.e., `createdMetacards.size() == 0`?

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/MetacardOnlyAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+                String.format("Failed to create resource [%s]", resource.getName()), e);
+          }
+        });
+  }
+
+  @Override
+  public void update(HarvestedResource resource, String updateId) throws HarvestException {
+    Metacard metacard = harvestedResourceTransformer.transformHarvestedResource(resource, updateId);
+
+    if (metacard == null) {
```
lessarderic: ❓ Same question as above.

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/MetacardOnlyAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+            List<String> updatedMetacardIds =
+                response
+                    .getUpdatedMetacards()
+                    .stream()
+                    .map(update -> update.getNewMetacard().getId())
+                    .collect(Collectors.toList());
+
+            if (updatedMetacardIds.size() == 1) {
+              return updatedMetacardIds.get(0);
+            } else {
```
lessarderic: ❓ `updatedMetacardIds.size() == 0`?

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/MetacardOnlyAdapter.java
```diff
@@ -0,0 +1,202 @@
...
+            return SECURITY.runWithSubjectOrElevate(callable);
+          } catch (SecurityServiceException | InvocationTargetException e) {
+            LOGGER.debug("Error executing code with subject", e);
+            return null;
+          }
+        });
+  }
+
+  private Map<String, Serializable> getSecurityMap() {
+    Subject subject = SECURITY.runAsAdmin(SECURITY::getSystemSubject);
```
lessarderic: ❓ Could we use Guava's `ImmutableMap.of(SecurityConstants.SECURITY_SUBJECT, SECURITY.runAsAdmin(SECURITY::getSystemSubject))`?

catalog/harvest/catalog-harvest-metacardadapter/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,35 @@
...
+        <argument ref="mimeTypeToTransformer"/>
+        <argument ref="mimeTypeMapper"/>
+    </bean>
+
+    <bean id="metacardOnlyAdapter" class="org.codice.ddf.catalog.harvest.adapter.MetacardOnlyAdapter">
+        <argument ref="catalogFramework"/>
+        <argument ref="harvestedResourceTransformer"/>
+    </bean>
+
+    <service ref="metacardOnlyAdapter" interface="org.codice.ddf.catalog.harvest.StorageAdapter"/>
```
lessarderic: ❓ Use filter for service lookup?

catalog/harvest/catalog-harvest-persistentlistener/src/main/java/org/codice/ddf/catalog/harvest/listeners/PersistentListener.java
```diff
@@ -0,0 +1,104 @@
...
+        new FileSystemPersistenceProvider("harvest/persistent/" + DigestUtils.sha1Hex(pid));
+  }
+
+  @Override
+  public void onCreate(HarvestedResource resource) {
+    String key = DigestUtils.sha1Hex(resource.getUri().toASCIIString());
+    if (resourceNotCreated(key)) {
+      String resourceId = null;
+      try {
+        resourceId = adapter.create(resource);
```
lessarderic: ❓What should we do when these operations fail? If we just log a debug message and move on, we will be missing events and nobody will even know that resource events failed to be processed. Should we retry? Log an error? @clockard, could we generate some kind of alert here?

catalog/harvest/catalog-harvest-webdav/src/main/java/org/codice/ddf/catalog/harvest/webdav/WebDavHarvester.java
```diff
@@ -0,0 +1,163 @@
...
+import org.codice.ddf.catalog.harvest.common.PollingHarvester;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Polls an HTTP WebDav address. */
+public class WebDavHarvester extends PollingHarvester {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(WebDavHarvester.class);
+
+  private final Set<Listener> listeners = new HashSet<>();
```
lessarderic: ❗️This `Set` needs to be synchronized to prevent possible threading issues when listeners are being added or removed while notifications are being sent. Recommend taking a look at https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/event/EventListenerSupport.html.

catalog/harvest/catalog-harvest-webdav/src/main/java/org/codice/ddf/catalog/harvest/webdav/WebDavHarvester.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    LOGGER.trace(
+        "no existing webdav observer for persistence key [{}], creating new observer", key);
+    return new DavAlterationObserver(new DavEntry(rootEntryLocation));
+  }
+
+  @Override
+  public void poll() {
+    observer.addListener(webdavListener);
```
lessarderic: ❓Not sure I understand why we have to add and then remove the `webdavListener` here. We may want to add a comment to explain why.

catalog/harvest/catalog-harvest-manager/src/main/java/org/codice/ddf/catalog/harvest/manager/HarvesterManager.java
```diff
@@ -0,0 +1,223 @@
...
+import org.apache.commons.collections.MapUtils;
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.Validate;
+import org.codice.ddf.catalog.content.monitor.ContentDirectoryMonitor;
+import org.codice.ddf.catalog.harvest.Adapter;
+import org.codice.ddf.catalog.harvest.file.DirectoryHarvester;
+import org.codice.ddf.catalog.harvest.listeners.PersistentListener;
+import org.codice.ddf.catalog.harvest.webdav.WebDavHarvester;
+
+public class HarvesterManager {
```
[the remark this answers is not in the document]
emmberk: 👍 Renamed to `SourceHarvester`

catalog/harvest/catalog-harvest-manager/src/main/java/org/codice/ddf/catalog/harvest/manager/HarvesterManager.java
emmberk: I added a suppression for the SonarQube code smell.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
emmberk: I think the term "source" is confusing here because we already have https://github.com/codice/ddf/blob/3efc289a995afbddf1170c716233a34ee82eaab8/catalog/core/catalog-core-api/src/main/java/ddf/catalog/source/Source.java. I would like to consider another name, which would mitigate the sources tab issue.

emmberk:
I have addressed some comments, but not all comments have been addressed yet. This PR is still a WIP.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
peterhuffer: I changed this to throw an IllegalArgumentException if the Listener cannot be registered (ie if the Harvester only supports a limited number of Listeners), and to be a no-op if the Listener is already registered.

catalog/harvest/catalog-harvest-metacardadapter/src/main/java/org/codice/ddf/catalog/harvest/adapter/MetacardOnlyAdapter.java
peterhuffer: I think I tried this but there were modification exceptions in other parts of the system that were modifying the request arguments map.

catalog/harvest/catalog-harvest-file/src/main/java/org/codice/ddf/catalog/harvest/file/DirectoryHarvesterListenerAdaptor.java
```diff
@@ -0,0 +1,73 @@
...
+
+  DirectoryHarvesterListenerAdaptor(Set<Listener> listeners) {
+    this.listeners = listeners;
+  }
+
+  @Override
+  public void onFileCreate(final File file) {
+    HarvestedResource harvestedResource = createHarvestedResource(file);
+    if (harvestedResource != null) {
+      listeners.forEach(listener -> listener.onCreate(harvestedResource));
```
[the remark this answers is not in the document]
peterhuffer: The onCreate|Update|Delete methods do not currently throw exceptions - unless the intention was to catch RuntimeExceptions?

peterhuffer:
Closing this PR to get rid of all the comments. Will re-open under another PR.

PR #3164, opened 2018-04-17
DDF-3739 Updated poms to include ddf feature dependencies

emanns95:
#### What does this PR do?
Completes the pom dependencies from #2896.

#### Who is reviewing it? 
(please choose AT LEAST two reviewers that need to approve the PR before it can get merged)
@mackncheesiest 
@oconnormi 
@tbatie 
@Lambeaux
@ryeats

#### Select relevant component teams: 
@codice/test

#### Choose 2 committers to review/merge the PR. 
(please choose ONLY two committers from below, delete the rest)
@rzwiefel 
@figliold
@pklinef

#### How should this be tested? (List steps with links to updated documentation)
All tests pass and the install profiles work

#### Notes on Review Process
Please see [Notes on Review Process](https://codice.atlassian.net/wiki/spaces/DDF/pages/71946981/Pull+Request+Guidelines) for further guidance on requirements for merging and abbreviated reviews. 

#### Review Comment Legend:
- ✏️ (Pencil) This comment is a nitpick or style suggestion, no action required for approval. This comment should provide a suggestion either as an in line code snippet or a gist. 
- ❓ (Question Mark) This comment is to gain a clearer understanding of design or code choices, clarification is required but action may not be necessary for approval.
- ❗ (Exclamation Mark) This comment is critical and requires clarification or action before approval.

PR #3170, opened 2018-04-19
DDF-3739 fixed missing config file dependencies

oconnormi:
#### What does this PR do?
Further changes related to #3164 and #2896 

* Moves all `.cfg` and `.config` files from each feature into ddf-common. This is to avoid an issue that occurs during the integration-test phase where those files haven't yet been installed in the local repo.
* Adds codice-snapshots, codice-releases, and codice-3rd-party repositories to the test-common DistributionOptions to better support artifact resolution during incremental builds

#### Who is reviewing it? 
@mackncheesiest
@emanns95 
@tbatie
@Lambeaux
@ryeats

@codice/test 
@codice/build 

@rzwiefel
@figliold
@pklinef

With an empty local m2 repo it should be possible to build the following modules in this order

`distribution/ddf-common`
`libs/test-common`
`features`

`mvn install` should now work from each of these folders without having pre-populated dependencies locally

[DDF-3739](https://codice.atlassian.net/browse/DDF-3739)

#### Notes on Review Process

#### Review Comment Legend:


oconnormi:
Build Now

libs/test-common/src/main/java/org/codice/ddf/test/common/AbstractComponentTest.java
```diff
@@ -73,7 +74,8 @@
     return options(
         getContainerOptions().get(),
         getApplicationOptions(portFinder).get(),
-        getTestBundleOptions().build());
+        getTestBundleOptions().build(),
+        defaultMavenRepositoriesOption());
```
oconnormi: Does anyone know if there is a reason we wouldn't want this to happen for all component tests. It definitely fixes problems with incremental builds. I still hope to find a way to make it use only the local maven repo since maven should have already pulled down all the required dependencies by the time the container is launched I think

oconnormi:
Build Now

features/utilities/src/main/resources/features.xml
```diff
@@ -77,19 +77,6 @@
         <bundle>mvn:ddf.mime.core/mime-core-configurableresolver/${project.version}</bundle>
         <bundle>mvn:org.apache.tika/tika-core/${tika.version}</bundle>
         <bundle>mvn:org.codice.thirdparty/tika-bundle/${tika.thirdparty.bundle.version}</bundle>
-
-        <configfile finalname="${ddf.etc}/DDF_Custom_Mime_Type_Resolver-csw.config">
```
ryeats: Do the default configurations without a config file work for these features. I dislike the idea of having a feature that needs some configuration which isn't installed with the feature just to work around an issue with Itests. That being said you could install it here and have it already exist in the itests.

oconnormi: Yeah I don't like this either. I'm thinking about creating a new module (or modules) that these configs  live under so that they can still be referenced in the features files. I'll be looking at that again after I get the features tests passing in CI

pklinef: Something to consider is this should make it easier for downstream distributions to remove or override managed service factory configurations.

oconnormi: There is that 😄 
As it stands with these changes all the defaults will be inherited as long as none of the configs are getting excluded from a custom distribution when pulling in ddf-common

oconnormi:
Build Now

features/utilities/src/main/resources/features.xml
tbatie: We could also deploy those config files from the ddf-common and keep the features as is.

oconnormi:
I'm gonna close this one in favor of #3176 since that one is cleaner and looks like it will work

PR #3182, opened 2018-04-24
DDF-3790 Updates alert settings component to use property components

andrewkfiedler:
#### What does this PR do?
 - Updates the alert settings component to use property components so that the styling is consistent with the rest of the UI.
 - Updates property views to turnOnLimitedWidth by default to make development easier.
 - Updates labels for property views to have a better line height for labels that wrap multiple lines.  Also adds a tooltip for labels.

#### Who is reviewing it? 
@brendan-hofmann
@mackncheesiest 
@adimka 
@rzwiefel 

Go to the notification settings view in the user settings and verify things look similar to before.  Verify settings still persist as they should. 

[DDF-3790](https://codice.atlassian.net/browse/DDF-3790)

### DIFF
--- a/ui/packages/catalog-ui-search/src/main/webapp/component/property/property.hbs
+++ b/ui/packages/catalog-ui-search/src/main/webapp/component/property/property.hbs
@@ -23,11 +23,12 @@
         <span class="if-editing fa fa-undo property-revert is-button"></span>
         <span class="property-validation fa fa-exclamation-triangle is-hidden"></span>
     </div>
-    <label for="{{cid}}">
-        {{label}}
-        {{#unless label}}
+    <label for="{{cid}}" title="{{#if label}}{{label}}{{else}}{{id}}{{/if}}">
+        {{#if label}}
+            {{label}}
+        {{else}}
             {{id}}
-        {{/unless}}
+        {{/if}}
     </label>
     <div class="if-block">
         <span class="if-editing fa fa-undo property-revert is-button"></span>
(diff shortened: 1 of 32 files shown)


ui/packages/catalog-ui-search/src/main/webapp/component/property/property.hbs
```diff
@@ -23,7 +23,7 @@
         <span class="if-editing fa fa-undo property-revert is-button"></span>
         <span class="property-validation fa fa-exclamation-triangle is-hidden"></span>
     </div>
-    <label for="{{cid}}">
+    <label for="{{cid}}" title="{{label}}{{#unless label}}{{id}}{{/unless}}">
```
mackncheesiest: ❓ So the title falls back to `id` if `label` is falsey?

andrewkfiedler: Yup.

brendan-hofmann: ✏️ I understand the semantic equivalency, but is there any reason not to use the clearer `{{if label}}{{label}}{{else}}{{id}}{{/if}}` instead?

andrewkfiedler: Just trying to keep you guys on your toes 😉  I'll switch it over.

rzwiefel:
build now!

PR #3223, opened 2018-05-07
DDF-3821 Fix for: OpenSearch sources do not set up their connections when …

rymach:
…configurations change that would impact it

#### What does this PR do?
Changes OpenSearchSource to component-managed from container-managed 
Ensures that a new CXF Client object is created when relevant configuration changes
#### Who is reviewing it? 
@millerw8 @lamhuy @AzGoalie 
@bdeining
@jlcsmith 
Create an OpenSearch source and edit the configuration to disable the CN check. Confirm that the change is reflected in the created connection (through an actual connection or by inspecting in a debugger). 
[DDF-3821](https://codice.atlassian.net/browse/DDF-3821?filter=-1)


catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
```diff
@@ -119,6 +120,28 @@
   @SuppressWarnings("squid:S2068" /*Key for the requestProperties map, not a hardcoded password*/)
   protected static final String PASSWORD_PROPERTY = "password";
 
+  protected static final String ENDPOINT_URL_PROPERTY = "endpointUrl";
```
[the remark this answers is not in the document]
rymach: I followed the style of existing properties like USERNAME_PROPERTY and PASSWORD_PROPERTY. It looks like they were changed to `protected` in a recent commit. @emmberk can you shed some light on why these constants were changed from `private`?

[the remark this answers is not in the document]
emmberk: There's a downstream project extends `OpenSearchSource`, `OpenSearchFilterVisitor`, and `OpenSearchParserImpl`, so I updates some variables and methods to be protected. I'm not opposed to changing them back.

rymach:
Changed the design to something simpler - see updated description above.

rymach:
build now

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
```diff
@@ -123,9 +123,7 @@
 
   protected final EncryptionService encryptionService;
 
-  protected SecureCxfClientFactory<OpenSearch> factory;
-
-  protected boolean isInitialized = false;
+  private SecureCxfClientFactory<OpenSearch> factory;
```
emmberk: ✏️ `@Nullable`

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
```diff
@@ -209,9 +207,15 @@ public OpenSearchSource(
    * called for each property specified in the metatype.xml file.
    */
   public void init() {
-    factory = createClientFactory(endpointUrl.getResolvedString(), username, password);
     configureXmlInputFactory();
-    isInitialized = true;
+  }
+
+  // lazy init
+  protected SecureCxfClientFactory<OpenSearch> getClientFactory() {
```
emmberk: ✏️ Should this be `@VisibleForTesting`? (or, not overridden in the unit tests? -- see comment below)

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
emmberk: ✏️ Should `factory ` be initialized to `null`?

catalog/opensearch/catalog-opensearch-source/src/test/java/org/codice/ddf/opensearch/source/OpenSearchSourceTest.java
```diff
@@ -901,7 +900,12 @@ protected InputTransformer lookupTransformerReference(String namespaceUri)
...
     protected SecureCxfClientFactory createClientFactory(
         String url, String username, String password) {
-      return this.factory;
+      return factory;
+    }
+
+    @Override
+    protected SecureCxfClientFactory getClientFactory() {
+      return factory;
     }
```
emmberk: ❓ Why do you need to override the `getClientFactory` method here? Can you get away with just overriding `createClientFactory`?

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
rymach: Will @Nullable be meaningful for a private field? 

`factory` is initialized to null automatically

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
rymach: Fixed

catalog/opensearch/catalog-opensearch-source/src/test/java/org/codice/ddf/opensearch/source/OpenSearchSourceTest.java
rymach: OpenSearchSource uses two different paths for acquiring a SecureCxfClientFactory instance. One uses the instance member `factory` and the other calls `createClientFactory`. I didn't look close enough to see why it was written that way, but these changes should not affect the original behavior.

catalog/opensearch/catalog-opensearch-source/src/main/java/org/codice/ddf/opensearch/source/OpenSearchSource.java
emmberk: I think @nullable is worthwhile because the IDE will warn you to check for null before invoking calling the `SecureCxfClientFactory` methods.

rymach: Fixed. Thanks for the explanation :)

catalog/opensearch/catalog-opensearch-source/src/test/java/org/codice/ddf/opensearch/source/OpenSearchSourceTest.java
rymach: You are right- thanks for the comment!

gjvera:
🎉 Hero Successful
Rebuilt DDF
Executed new OS query via Intrigue
Verified breakpoint was triggered successfully on factory creation line
Edited admin config to turn off CN check
Verified breakpoint was still triggered successfully!

LG2M!

PR #3031, opened 2018-03-02
DDF-3621 Solr client/factory initial abstraction with common retry mechanism

paouelle:
#### The commits are organized in a way that should make it easier to review all the changes.
--------------------

#### What does this PR do?
Implemented a new SolrClient adapter layer that takes care of retrying in the background to create the real solr client if Solr is not available yet. It also takes care of managing the client's availability state when it becomes unavailable by pinging in the background the Solr server until is becomes reachable and then notifying listeners. This new client is a starting point for when we finalize abstracting the SolrClient API into a pure set of interfaces. The package as therefore been chosen with that in mind.

Moved the different levels of retry mechanisms for Solr that existed in various places (i.e. SolrCache, RemoteSolrCatalogProvider, and the various SolrClient factories) in this new layer.

Modified the HTTP client factory used when connecting to an external Solr server to not retry forever as it did.

Modified the FilteringSolrIndex to reuse the EmbeddedSolrFactory as opposed to duplicating the code.  Fixed an issue where it would modify the ConfigurationStore singleton to mark it in memory; thus affecting all newly created Solr clients from that point on. This singleton should be removed but it turned out to be more work that I was willing to add to this ticket.

More importantly and the initial reason for the ticket, it unblocks the Solr cache initialization which was preventing proper shutdown when Solr was not available.

Fixed Catalog Solr Provider to no longer embed test jars in the bundle.

Removed banana.

#### Who is reviewing it? 
@tbatie 
@emanns95 
@figliold 

The following reviewers are added at the request of Phil:
@bdeining 
@rzwiefel 
@coyotesqrl 
@pklinef 

The following people are tagged to keep them in the loop:
@lessarderic 
@ahoffer 

#### Select relevant component teams: 
https://github.com/orgs/codice/teams/solr

#### Choose 2 committers to review/merge the PR. 
@pklinef
@coyotesqrl
@rzwiefel

#### Any background context you want to provide?
Refer to the `package-info.java` file in the `org.codice.solr.factory.impl` package for a design description.

#### What are the relevant tickets?
[DDF-3621](https://codice.atlassian.net/browse/DDF-3621)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
- [x] Update / Add Integration Tests

#### Notes on Review Process
Please see [Notes on Review Process](https://codice.atlassian.net/wiki/spaces/DDF/pages/71946981/Pull+Request+Guidelines) for further guidance on requirements for merging and abbreviated reviews. 

[...]

PR #3225, opened 2018-05-08
[2.11.x] DDF-3817 Add support for separate external/internal urls

vinamartin:
#### What does this PR do?
Adds support for external host and port properties in DDF. 
This is useful when DDF is behind a proxy and the URL a user sees is different than the local url DDF uses.
I still need to update documentation on this and any feedback on where that should go and what it should be would be helpful!

⚠️ **THIS PR WILL BREAK PROJECTS THAT DEPEND ON DDF**
⚠️ **SHOULD THE INTERNAL PROPERTIES BE RENAMED?**

#### Who is reviewing it? 
@alexaabrd 
@SmithJosh 
@tbatie 
@rzwiefel 
@clockard 

@brendan-hofmann
@lessarderic

Configure a DDF with an internal host of localhost and an external host of the machine's name. Configure them to be on different ports.
Verify that when accessing action providers in Intrigue that the link is the external host and port. 
IDP may need to be disabled to test this as it doesn't work with the external host/port. 

Optionally, use docker and nginx to set up a DDF with a reverse proxy and verify things work as expected. 

With DDF-2.11.6, when a DDF is set up with a reverse proxy, the action providers in Intrigue don't work as expected. 

[DDF-3817](https://codice.atlassian.net/browse/DDF-3817)

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java
```diff
@@ -39,17 +53,51 @@
...
+  }
 
   /**
    * Gets the port number based on the the system protocol
    *
    * @return
    */
-  public static String getPort() {
-    String port = System.getProperty(PORT);
+  public String getPortString() {
```
vinamartin: ❗️ **Currently,** I'm using two different methods for internal and external properties. Should I refactor this so they only use one common method?

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -59,6 +59,13 @@ org.codice.ddf.system.port=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
 org.codice.ddf.system.internalHttpsPort=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
```
brendan-hofmann: ❓ These might be redundant now. Do we need to deconflict the two sets of properties now that `internal` is the default behavior and the full set of `external` properties can be configured externally? I think the comment on line 52 is inaccurate now.

libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java
brendan-hofmann: The two different methods are the singleton methods and legacy static accessors. Eliminating the redundancy would be large refactoring, and perhaps should be a separate piece of tech debt.

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
vinamartin: Essentially reduce it down to this? : 
``` org.codice.ddf.system.protocol=https://
 org.codice.ddf.system.hostname=localhost
 org.codice.ddf.system.httpsPort=8993
 org.codice.ddf.system.httpPort=8181
 org.codice.ddf.system.port=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
 org.codice.ddf.system.rootContext=/services

#useful when DDF is behind a proxy and the URL a user sees is different than the local url DDF uses
org.codice.ddf.external.protocol=https://
org.codice.ddf.external.hostname=localhost
org.codice.ddf.external.httpsPort=8993
org.codice.ddf.external.httpPort=8181
org.codice.ddf.external.port=_DO_NOT_EXPAND_${org.codice.ddf.external.httpsPort}

```

vinamartin:
build now

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -49,16 +49,21 @@ javax.net.ssl.trustStoreType=jks
...
 org.codice.ddf.system.hostname=localhost
 org.codice.ddf.system.httpsPort=8993
 org.codice.ddf.system.httpPort=8181
 org.codice.ddf.system.port=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
-org.codice.ddf.system.internalHttpsPort=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
-org.codice.ddf.system.internalHttpPort=_DO_NOT_EXPAND_${org.codice.ddf.system.httpPort}
 org.codice.ddf.system.rootContext=/services
+# external.httpsPort and external.httpPort are useful when DDF is behind a proxy and
+# the URL a user sees is different than the local url DDF uses
+org.codice.ddf.external.protocol=https://
```
clockard: Why don't we use variables for the values on these?

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -49,16 +49,21 @@ javax.net.ssl.trustStoreType=jks
...
+org.codice.ddf.external.hostname=localhost
+org.codice.ddf.external.httpsPort=8993
```
clockard: Is there a reason we decided to go with external here as opposed to expanding on the pre-existing internal properties? I think most cases care about the external and only a handful would care about the internal. This could make for a much smaller changes set and maybe less confusion in the future.

clockard:
I think there have been some updates that have been missed. There are probably several places in blueprint that are using these variables that need to be updated. This is an example of one https://github.com/codice/ddf/blob/2.11.x/catalog/spatial/csw/spatial-csw-endpoint/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L149

rzwiefel:
Any reason not just make the default calls (SystemBaseUrl.(getBaseUrl,constructUrl,etc) ) map to the EXTERNAL calls so we don't have to change everywhere that uses SystemBaseUrl to use EXTERNAL?

vinamartin:
I'll be refactoring this to go for a more explicit `EXTERNAL` and `INTERNAL` approach

distribution/ddf-common/src/main/resources/etc/ws-security/issuer/encryption.properties
```diff
@@ -3,5 +3,5 @@ org.apache.ws.security.crypto.provider=org.apache.wss4j.common.crypto.Merlin
 org.apache.ws.security.crypto.merlin.keystore.type=jks
 org.apache.ws.security.crypto.merlin.keystore.password=${javax.net.ssl.keyStorePassword}
 org.apache.ws.security.crypto.merlin.keystore.private.password=${javax.net.ssl.keyStorePassword}
-org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.system.hostname}
+org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.external.hostname}
```
brendan-hofmann: ❗️ These four should not be external.

distribution/test/itests/test-itests-common/src/main/resources/issuerencryption.properties
```diff
@@ -3,6 +3,6 @@ org.apache.ws.security.crypto.provider=org.apache.wss4j.common.crypto.Merlin
 org.apache.ws.security.crypto.merlin.keystore.type=jks
 org.apache.ws.security.crypto.merlin.keystore.password=${javax.net.ssl.keyStorePassword}
 org.apache.ws.security.crypto.merlin.keystore.private.password=${javax.net.ssl.keyStorePassword}
-org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.system.hostname}
+org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.external.hostname}
```
brendan-hofmann: ❗️ These four should probably be changed back for the itests.

libs/platform-configuration-impl/src/main/java/org/codice/ddf/configuration/impl/ConfigurationWatcherImpl.java
```diff
@@ -33,40 +33,40 @@ public ConfigurationWatcherImpl() {}
   /**
    * Helper method to get the hostname or IP from the configuration
    *
-   * @return the value associated with {@link SystemBaseUrl#HOST} property name
+   * @return the value associated with {@link SystemBaseUrl#INTERNAL_HOST} property name
```
brendan-hofmann: ❗️ The javadocs in this file are `INTERNAL` but the return values are `EXTERNAL`.

platform/platform-configuration/src/main/java/org/codice/ddf/configuration/ConfigurationManager.java
```diff
@@ -110,9 +110,9 @@
   private static Map<String, String> propertyMapping = new HashMap<>();
 
   static {
-    propertyMapping.put(PROTOCOL, SystemBaseUrl.PROTOCOL);
-    propertyMapping.put(HOST, SystemBaseUrl.HOST);
-    propertyMapping.put(PORT, SystemBaseUrl.PORT);
+    propertyMapping.put(PROTOCOL, SystemBaseUrl.EXTERNAL_PROTOCOL);
```
brendan-hofmann: ❓ Are we sure all the users of platform-config are external?

platform/security/common/src/main/java/org/codice/ddf/security/common/Security.java
```diff
@@ -337,7 +337,7 @@ private PKIAuthenticationTokenFactory createPKITokenFactory() {
   }
 
   private String getCertificateAlias() {
-    return System.getProperty("org.codice.ddf.system.hostname");
+    return System.getProperty("org.codice.ddf.external.hostname");
```
brendan-hofmann: ❗️ I don't think the system cert is aliased to the external hostname.

platform/security/common/src/test/resources/encryption.properties
```diff
@@ -3,5 +3,5 @@ org.apache.ws.security.crypto.provider=org.apache.wss4j.common.crypto.Merlin
 org.apache.ws.security.crypto.merlin.keystore.type=jks
 org.apache.ws.security.crypto.merlin.keystore.password=${javax.net.ssl.keyStorePassword}
 org.apache.ws.security.crypto.merlin.keystore.private.password=${javax.net.ssl.keyStorePassword}
-org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.system.hostname}
+org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.external.hostname}
```
brendan-hofmann: ❗️ Probably need to change these two back for tests.

platform/security/core/security-core-impl/src/test/resources/encryption.properties
```diff
@@ -3,5 +3,5 @@ org.apache.ws.security.crypto.provider=org.apache.wss4j.common.crypto.Merlin
 org.apache.ws.security.crypto.merlin.keystore.type=jks
 org.apache.ws.security.crypto.merlin.keystore.password=${javax.net.ssl.keyStorePassword}
 org.apache.ws.security.crypto.merlin.keystore.private.password=${javax.net.ssl.keyStorePassword}
-org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.system.hostname}
+org.apache.ws.security.crypto.merlin.keystore.alias=${org.codice.ddf.external.hostname}
```
brendan-hofmann: ❗️ These two as well.

platform/security/filter/security-filter-login/src/main/java/org/codice/ddf/security/filter/login/LoginFilter.java
```diff
@@ -555,7 +555,7 @@ private SAMLAuthenticationToken renewSecurityToken(
           new SecurityAssertionImpl(((SecurityToken) savedToken.getCredentials()));
 
       if (savedAssertion.getIssuer() != null
-          && !savedAssertion.getIssuer().equals(SystemBaseUrl.getHost())) {
+          && !savedAssertion.getIssuer().equals(SystemBaseUrl.EXTERNAL.getHost())) {
```
brendan-hofmann: ❗️ Won't the issuer be the internal/system host?

platform/platform-configuration/src/main/java/org/codice/ddf/configuration/ConfigurationManager.java
vinamartin: Good question, I'm not sure.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/AssertionConsumerService.java
```diff
@@ -394,9 +394,9 @@ public Response retrieveMetadata() throws WSSecurityException, CertificateEncodi
     X509Certificate encryptionCert =
         findCertificate(systemCrypto.getEncryptionAlias(), systemCrypto.getEncryptionCrypto());
 
-    String hostname = SystemBaseUrl.getHost();
-    String port = SystemBaseUrl.getPort();
-    String rootContext = SystemBaseUrl.getRootContext();
+    String hostname = SystemBaseUrl.EXTERNAL.getHost();
```
brendan-hofmann: ❗️ I think we should leave the IdP alone for now. We don't have enough independently configurable URLs for it to work behind a reverse proxy today.

platform/security/rest/security-rest-cxfwrapper/src/main/java/org/codice/ddf/cxf/SecureCxfClientFactory.java
```diff
@@ -449,7 +449,7 @@ private void configureConduit(ClientConfiguration clientConfig) {
     } catch (NoSuchAlgorithmException | KeyStoreException e) {
       LOGGER.debug("Unable to initialize TrustManagerFactory.", e);
     }
-    tlsParams.setCertAlias(SystemBaseUrl.getHost());
+    tlsParams.setCertAlias(SystemBaseUrl.EXTERNAL.getHost());
```
brendan-hofmann: ❗️ Need the internal hostname to get the system cert.

distribution/ddf-common/src/main/resources/etc/ws-security/issuer/encryption.properties
oconnormi: I definitely had to change them to external to get things to work, granted this might be more due to how the `CertNew` tool loads the cert into the keystore than anything since it assumes the value provided to the `-cn <fqdn>` option should be used as the keystore alias

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/PropertyPlaceholderWrapper.java
```diff
@@ -46,9 +46,9 @@ public PropertyPlaceholderWrapper(
     stsProperties = properties;
     // set the default values in case there is no configuration
     samlConditionsProvider.setLifetime(DEFAULT_LIFETIME);
-    stsProperties.setSignatureUsername(SystemBaseUrl.getHost());
-    stsProperties.setIssuer(SystemBaseUrl.getHost());
-    stsProperties.setEncryptionUsername(SystemBaseUrl.getHost());
+    stsProperties.setSignatureUsername(SystemBaseUrl.EXTERNAL.getHost());
```
brendan-hofmann: ❗️ Won't the STS be the system user?

brendan-hofmann:
❗️ The docs never mention that you can't have the localhost cert in the keystore if you're using a system cert. This is implied by the instructions telling you to create *new* keystores with your existing certs, but is never explicitly stated or explained.

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -49,16 +49,21 @@ javax.net.ssl.trustStoreType=jks
...
 org.codice.ddf.system.httpsPort=8993
 org.codice.ddf.system.httpPort=8181
 org.codice.ddf.system.port=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
-org.codice.ddf.system.internalHttpsPort=_DO_NOT_EXPAND_${org.codice.ddf.system.httpsPort}
-org.codice.ddf.system.internalHttpPort=_DO_NOT_EXPAND_${org.codice.ddf.system.httpPort}
 org.codice.ddf.system.rootContext=/services
+# external.httpsPort and external.httpPort are useful when DDF is behind a proxy and
+# the URL a user sees is different than the local url DDF uses
+org.codice.ddf.external.protocol=https://
+org.codice.ddf.external.hostname=localhost
```
millerw8: Should this and other `org.codice.ddf.external.*` properties default to system: ```org.codice.ddf.external.hostname=${org.codice.ddf.system.hostname}```
In order to minimize the amount of breakage on downstream products?

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -49,16 +49,21 @@ javax.net.ssl.trustStoreType=jks
 #
 # Global URL Properties
 #
-# The httpsPort, httpPort, and port properties are used to advertise the system externally
-# The internalHttpsPort and internalHttpPort properties are the ports that the system will run on
+# The httpsPort and httpPort are the ports that the system will run on
 org.codice.ddf.system.protocol=https://
 org.codice.ddf.system.hostname=localhost
```
millerw8: What is the expectation on the CN and alias now in the keystore?  Both match the org.codice.ddf.external.hostname?

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
brendan-hofmann: I think that is the intent. We could default it in the properties file or leave them blank and let the base URL resolver fall through. Now that you mention it though, I think setting it in the properties file is more explicit and easier to maintain.

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
brendan-hofmann: Both should match the system/internal hostname. The external properties are for a reverse proxy, load balancer, or other independent system that clients are expected to access DDF through.

distribution/ddf-common/src/main/resources/etc/ws-security/issuer/encryption.properties
brendan-hofmann: I think it was because some other properties got refactored to `EXTERNAL` when they should not have been. 😸

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
vinamartin: I think this makes sense, I can update this.

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
millerw8: Will that cause problems then?  Will that certificate get exposed all the way out to the consumer?  Will they see browser mismatches?  Or are we expecting the the load balancer or reverse proxy to change out the certificate that gets exposed?

brendan-hofmann: In an SSL-terminating configuration the external interface would have its own cert. Trust between the internal interface and DDF would also need to be established.

It is possible to pass certs through in some configurations, which I think could be handled by a SAN for the external interface. Wildcard certs could also work depending on the network topology.

millerw8: DDF does not do wildcard certs well.  And in many environments, we may not have much ownership of the certs.

brendan-hofmann: Yeah, we need to do better with wildcard certs.

oconnormi: Ah so this is primarily geared towards proxies that will terminate and re-encrypt traffic to the backend then? That might explain why I had to change a couple things in order to get this to work. I had tested by simply forwarding one port to another via `socat`

Are there plans to support raw tcp forwarding better in the future as well?

I think I can probably make it work more easily once these properties are added since i can manually set all of them as well as the linkages to the keystore alias to achieve it, but would be nice to have easier support out of the box. I'm thinking specifically of scenarios where ddf may be running in a container that is exposing a potentially randomized port.

Anyway I'm rambling now, its good to see that we'll be able to support ssl-termination more easily going forward since that was very difficult before

brendan-hofmann: I think this would still support TCP forwarding if the certificates were generated correctly. These properties enable DDF to construct the correct URLs, but the certificates still need to be configured correctly.

I think today you have to configure `system.properties` when you randomize the port, so I am not sure how this PR affects that. Are you thinking of picking up the system port more dynamically?

oconnormi: Not sure really, currently I have to change the port the whole system runs on since all the urls are relative to the system port. I think if i left the system port on 8993 and changed the external one to 443 while leaving everything else it would probably work with these changes. I think the most ideal situation would be that the urls were generated relative to the incoming request instead since then we wouldn't have to care so much about knowing the external port and hostname. That would probably require quite a lot of changes compared to how we currently do it though.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/AssertionConsumerService.java
```diff
@@ -394,9 +394,9 @@ public Response retrieveMetadata() throws WSSecurityException, CertificateEncodi
     X509Certificate encryptionCert =
         findCertificate(systemCrypto.getEncryptionAlias(), systemCrypto.getEncryptionCrypto());
 
-    String hostname = SystemBaseUrl.getHost();
-    String port = SystemBaseUrl.getPort();
-    String rootContext = SystemBaseUrl.getRootContext();
+    String hostname = SystemBaseUrl.INTERNAL.getHost();
```
lessarderic: ❓General question for @stustison. Looks like we're exposing the IdP using the internal URL, and the client code is also using the internal URL to access it. Will that cause problems in an HA configuration?

platform/solr/solr-factory/src/main/java/org/codice/solr/factory/impl/HttpSolrClientFactory.java
```diff
@@ -109,7 +109,7 @@
    * @return Solr server secure HTTP address
    */
   public static String getDefaultHttpsAddress() {
-    return SystemBaseUrl.constructUrl("https", SOLR_CONTEXT);
+    return SystemBaseUrl.INTERNAL.constructUrl("https", SOLR_CONTEXT);
```
lessarderic: ❓@paouelle, @pklinef, should we be using the external URL here instead? Does this still even matter?

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -49,16 +49,21 @@ javax.net.ssl.trustStoreType=jks
 #
 # Global URL Properties
 #
-# The httpsPort, httpPort, and port properties are used to advertise the system externally
```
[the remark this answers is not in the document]
brendan-hofmann: We have to change them either way. Since the existing properties had to match the system DDF was running on, those were really more internal than external already.

platform/solr/solr-factory/src/main/java/org/codice/solr/factory/impl/HttpSolrClientFactory.java
paouelle: The default should continue to go to the internal one. Otherwise, the customer has to configure it otherwise. That being said, this code is completely modified by the following PR (https://github.com/codice/ddf/pull/3031) which is scheduled to be merged into master this sprint.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/AssertionConsumerService.java
paouelle: in HA, we expected the IdP to always respond to the external URL which is why that hostname had to be changed to match the external visible URL and not the internal one.

vinamartin:
@paouelle Can you explain why this would break HA? 
It sounds like if you configure the internal and external hostname to be the same but simply change the port we should have the same capability.

platform/platform-commands/src/main/java/org/codice/ddf/commands/platform/DescribeCommand.java
```diff
@@ -28,18 +28,17 @@
...
-    System.out.printf("%s=%s%n", "Root Context", SystemBaseUrl.getRootContext());
-    System.out.printf("%s=%s%n", "External Http Port", SystemBaseUrl.getHttpPort());
-    System.out.printf("%s=%s%n", "External Https Port", SystemBaseUrl.getHttpsPort());
+    System.out.printf("%s=%s%n", "Protocol", SystemBaseUrl.EXTERNAL.getProtocol());
+    System.out.printf("%s=%s%n", "Host", SystemBaseUrl.EXTERNAL.getHost());
+    System.out.printf("%s=%s%n", "Port", SystemBaseUrl.EXTERNAL.getPort());
+    System.out.printf("%s=%s%n", "Root Context", SystemBaseUrl.EXTERNAL.getRootContext());
+    System.out.printf("%s=%s%n", "External Http Port", SystemBaseUrl.EXTERNAL.getHttpPort());
+    System.out.printf("%s=%s%n", "External Https Port", SystemBaseUrl.EXTERNAL.getHttpsPort());
     System.out.printf(
```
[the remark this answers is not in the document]
vinamartin: That makes sense, I can go ahead and add this.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/SystemPropertiesAdmin.java
```diff
@@ -174,10 +174,39 @@ public void writeSystemProperties(Map<String, String> updatedSystemProperties) {
     try {
       Properties systemDotProperties = new Properties(systemPropertiesFile);
 
-      updateProperty(SystemBaseUrl.HOST, updatedSystemProperties, systemDotProperties);
-      updateProperty(SystemBaseUrl.PROTOCOL, updatedSystemProperties, systemDotProperties);
-      updateProperty(SystemBaseUrl.HTTP_PORT, updatedSystemProperties, systemDotProperties);
-      updateProperty(SystemBaseUrl.HTTPS_PORT, updatedSystemProperties, systemDotProperties);
+      // Duplicate properties into the internal properties for legacy functionality
+      if (updatedSystemProperties.containsKey(SystemBaseUrl.EXTERNAL_HOST)) {
```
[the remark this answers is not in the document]
vinamartin: Yeah, that makes a lot more sense, I was writing this quickly so I can go ahead and update this

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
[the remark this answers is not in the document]
paouelle: ❓ Why not keep the properties as is to mean as it did external and internal. And have an optional one for internal that when defined is used and if not, the normal one continues to be used. That way, the system continue to work as it did and yet we have the option to define an internal hostname when needed?

oconnormi:
I think it would still likely work the same way for HA, each node would still have identical external hostnames, and the internal hostnames for each would probably just be `localhost` which should all work. As for the looping back through the proxy, IMO we only did that because there was no easy way to avoid it at the time. requests that a node direct at itself should never in an ideal world have to use an external network interface. I _think_ that in an ssl-passthrough configuration like we had been using for HA we should be able to achieve the same things we were doing in the past, but there might need to be some minor changes to the HA documentation

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/AssertionConsumerService.java
oconnormi: I think it would cause issues in any configuration, depends on what is getting handed back to the user's browser during the IdP redirect. I can't remember where that url comes from though.

vinamartin: We don't expose enough url configurations to allow IdP to work with a reverse proxy today. I think that would be additional follow on work to this task.
For now, I've left all of these as internal.

oconnormi: fyi i _think_ this is determines the url that the user gets redirected to: https://github.com/codice/ddf/blob/master/platform/security/idp/security-idp-client/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L33

libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java
```diff
@@ -39,15 +53,49 @@
...
+
+  private String host;
+
+  private String protocol;
+
+  /* EXTERNAL and INTERNAL singleton objects */
+  public static final SystemBaseUrl EXTERNAL = new SystemBaseUrl(true);
+  public static final SystemBaseUrl INTERNAL = new SystemBaseUrl(false);
+
+  private SystemBaseUrl(boolean external) {
```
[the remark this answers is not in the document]
vinamartin: I think this would work considering how this has changed.

libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java
```diff
@@ -39,15 +53,49 @@
...
+    if (external) externalMode();
```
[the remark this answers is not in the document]
vinamartin: Refactored this out.

libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java
```diff
@@ -39,15 +53,49 @@
...
 
   /**
    * Gets the port number based on the the system protocol
    *
    * @return
    */
-  public static String getPort() {
-    String port = System.getProperty(PORT);
+  public String getPort() {
+    String port = System.getProperty(this.port);
```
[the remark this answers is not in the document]
vinamartin: 👍

libs/common-system/src/test/java/org/codice/ddf/configuration/SystemBaseUrlTest.java
```diff
@@ -31,82 +31,84 @@ public void setUp() throws Exception {
 
   @Test
```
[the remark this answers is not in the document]
vinamartin: Added.

libs/test-common/src/main/java/org/codice/ddf/test/common/configurators/DdfBaseOptions.java
```diff
@@ -162,17 +162,18 @@ private Option getSystemPropertyOptions() {
     return composite(
         editConfigurationFilePut(
             SYSTEM_PROPERTIES_FILE,
-            SystemBaseUrl.PORT,
```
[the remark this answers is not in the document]
vinamartin: Good call

libs/test-common/src/main/java/org/codice/ddf/test/common/configurators/DdfBaseOptions.java
```diff
@@ -182,14 +197,14 @@ private Option getSystemPropertyOptions() {
...
+            portFinder.getPortAsString(SystemBaseUrl.EXTERNAL_HTTP_PORT)),
         editConfigurationFilePut(
             PAX_WEB_CONFIG_FILE,
             OSGI_SERVICE_HTTP_PORT_SECURE_KEY,
-            portFinder.getPortAsString(SystemBaseUrl.HTTPS_PORT)),
+            portFinder.getPortAsString(SystemBaseUrl.EXTERNAL_HTTPS_PORT)),
         editConfigurationFilePut(PAX_WEB_CONFIG_FILE, "org.osgi.service.http.enabled", "true"),
         editConfigurationFilePut(SYSTEM_PROPERTIES_FILE, "ddf.home", "${karaf.home}"),
-        editConfigurationFilePut(SYSTEM_PROPERTIES_FILE, SystemBaseUrl.HOST, "localhost"));
+        editConfigurationFilePut(SYSTEM_PROPERTIES_FILE, SystemBaseUrl.EXTERNAL_HOST, "localhost"));
```
vinamartin: Looks like we're missing internal host

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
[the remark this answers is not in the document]
lessarderic: Talked to @pklinef and he wasn't overly concerned about renaming the system properties. Those are not guaranteed to remain between releases so we're free to change them. It still bugs me a little though that we had a comment saying
```
The httpsPort, httpPort, and port properties are used to advertise the system externally
```
and we've changed their meaning to be internal now and added new `external` ones. Seems to me that keeping the old system property names for external ones and create new `internal` (or `instance`, or `node`) ones would have limited the impact, both for DDF and downstream projects.

distribution/docs/src/main/resources/content/_configuring/global-settings.adoc
```diff
@@ -107,50 +150,32 @@ Does not change the address the system runs on.
...
+|External Default Port
+|org.codice.ddf.external.port
 |String
 |The default port used to advertise the system. This should match either the above http or https port.
 
 NOTE: Does not change the port the system runs on.
 |8993
 |Yes
 
-|Internal HTTPS Port
```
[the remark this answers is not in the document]
lessarderic: Talked to @pklinef and it looks like we're ok removing these.

libs/common-system/src/test/java/org/codice/ddf/configuration/SystemBaseUrlTest.java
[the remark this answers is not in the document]
lessarderic: Thanks!

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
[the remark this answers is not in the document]
brendan-hofmann: I think that comment was always wrong. They go with the system hostname, which is required to match the machine hostname, the certificate DA, and the keystore alias.

External might also be an overloaded term, which doesn't help matters.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/AssertionConsumerService.java
stustison: Yeah this doesn't seem quite right to me for all of the IdP configuration. It looks like all of the IdP stuff is pointing towards the internal properties, but all of it gets proxied through the browser, so will go through the external interfaces. Seems like this is wrong to me. Not just here, but all the stuff below for logging in and logging out.

platform/security/sts/security-sts-server/src/main/java/ddf/security/sts/PropertyPlaceholderWrapper.java
```diff
@@ -46,9 +46,9 @@ public PropertyPlaceholderWrapper(
     stsProperties = properties;
     // set the default values in case there is no configuration
     samlConditionsProvider.setLifetime(DEFAULT_LIFETIME);
-    stsProperties.setSignatureUsername(SystemBaseUrl.getHost());
-    stsProperties.setIssuer(SystemBaseUrl.getHost());
-    stsProperties.setEncryptionUsername(SystemBaseUrl.getHost());
+    stsProperties.setSignatureUsername(SystemBaseUrl.INTERNAL.getHost());
+    stsProperties.setIssuer(SystemBaseUrl.INTERNAL.getHost());
+    stsProperties.setEncryptionUsername(SystemBaseUrl.INTERNAL.getHost());
```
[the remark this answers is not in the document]
stustison: The STS is actually the piece that I would expect to work with internal properties, not external.

platform/security/idp/security-idp-client/src/main/java/org/codice/ddf/security/idp/client/AssertionConsumerService.java
brendan-hofmann: I think the intent was to wait until we make sure it's even possible to configure our IdP to work behind a proxy. My understanding is that that scenario does not work currently.

vinamartin:
There's an OWASP issue that needs to be addressed on this branch.

PR #3263, opened 2018-05-23
DDF-3849 Refactored CsvQueryResponseTransformer into utility jar

brianfelix:
#### What does this PR do?
Pulled out CSV transformation logic in CsvQueryResponseTransformer into a common library so that it is reusable.
#### Who is reviewing it? 
@Bdthomson 
@jrnorth 
@Variadicism 
@glenhein 
@bdeining 
@millerw8 
[DDF-3849](https://codice.atlassian.net/browse/DDF-3849)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+      final List<AttributeDescriptor> sortedAttributeDescriptors,
+      final Map<String, String> aliasMap)
+      throws CatalogTransformerException {
+    StringBuilder stringBuilder = new StringBuilder();
+
+    try {
+      CSVPrinter csvPrinter = new CSVPrinter(stringBuilder, CSVFormat.RFC4180);
+      printColumnHeaders(csvPrinter, sortedAttributeDescriptors, aliasMap);
+
+      metacards.forEach(m -> printMetacard(csvPrinter, m, sortedAttributeDescriptors));
```
Bdthomson: ✏️ Might want to change m to metacard

catalog/transformer/catalog-transformer-csv/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -1,33 +0,0 @@
...
-            <entry key="id" value="csv"/>
-            <entry key="mime-type" >
-                <list>
-                    <value>text/csv</value>
-                </list>
-            </entry>
-        </service-properties>
-    </service>
-
-    <bean id="CsvQueryResponseTransformer"
```
Bdthomson: This blueprint file should have been moved to new CSV QueryResponse bundle, not deleted.

catalog/transformer/catalog-transformer-csv-common/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>transformer</artifactId>
+        <groupId>ddf.catalog.transformer</groupId>
+        <version>2.12.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>catalog-transformer-csv-common</artifactId>
+    <name>DDF :: Catalog :: Transformer :: Common :: CSV</name>
+    <modelVersion>4.0.0</modelVersion>
+    <packaging>bundle</packaging>
```
Bdthomson: We should probably be able to change this from a bundle to a jar with minimal side effects.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+        .map(MetacardType::getAttributeDescriptors)
+        .flatMap(Set::stream)
+        .filter(attributeNotBinary)
+        .filter(attributeNotObject)
+        .filter(attributeNotExcluded)
+        .collect(toSet());
+  }
+
+  /**
+   * Given a list of {@link Metacard}s, a string set of requested attributes and a string set of
```
Variadicism: :pencil2: This three-item list should use [the Oxford comma](https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/)!

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+        new CsvAttributeDescriptorComparator(attributeOrder);
+
+    return attributeSet.stream().sorted(attributeComparator).collect(toList());
+  }
+
+  /**
+   * Given a list of {@link Metacard}s and a string set of hiddenFields, returns a set of {@link
+   * AttributeDescriptor}s that contains all attributes that exist on the given metacard types
+   * except those that should be excluded.
+   */
```
Variadicism: These Javadocs lack important pieces to be complete like `@param` and `@return`. The Javadoc should either be completed or the methods should be self-explanatory enough to allow the Javadocs to be removed. (I think that with a method name change or two where mentioned, the methods in here are pretty self-explanatory and might not require Javadoc.)

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Map;
+import java.util.Scanner;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.junit.Before;
+import org.junit.Test;
+
+public class CsvTransformerTest {
+
+  private static final List<AttributeDescriptor> ATTRIBUTE_DESCRIPTOR_LIST = new ArrayList<>();
```
Variadicism: Line 206 modifies this list. While technically `final` only prevents variable reassignment in Java, it is still very confusing to see a variable marked as `final` in all-caps snake case, which generally indicates a consant... and then see that variable's value change. This would be better off non-`final` with a lowercase camel case name like the Collections below it.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+  private static Map<String, Attribute> metacardDataMap = new HashMap<>();
+
+  private static List<Metacard> metacardList = new ArrayList<>();
+  private static final int METACARD_COUNT = 2;
+
+  private static final String CSV_REGEX = "[\\n\\r,]";
```
Variadicism: This could use a more descritive name; `CSV_REGEX` shoulds like a regex made to match an entire CSV file or record. Perhaps `CSV_ITEM_SEPARATOR_REGEX`?

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+  @Before
+  public void setup() {
+    ATTRIBUTE_DESCRIPTOR_LIST.clear();
+    metacardList.clear();
+    buildMetacardDataMap();
+    buildMetacardList();
+  }
+
+  @Test
+  public void getAllAttributes() {
```
Variadicism: This is just a reminder that if the `getAllAttributes` method in `CsvTransformer.java` is renamed, this should be, too.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Given a list of {@link Metacard}s, a string set of requested attributes and a string set of
+   * excluded fields, returns a set of {@link AttributeDescriptor}s containing the requested
+   * attributes, additionally filtering out excluded attributes.
+   */
+  public static Set<AttributeDescriptor> getOnlyRequestedAttributes(
+      final List<Metacard> metacards,
+      final Set<String> requestedAttributes,
+      final Set<String> excluded) {
```
Variadicism: It seems redundant and confusing to include what amounts to an attribute whitelist _and_ an attribute blacklist. The `excluded` parameter could be removed and `requestedAttributes` can simply not contain any "excluded" attributes.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+  public void getOnlyRequestedAttributes() {
+    Set<String> requestedAttributes = new HashSet<>();
+    requestedAttributes.add("attribute1");
+
+    Set<AttributeDescriptor> onlyRequestedAttributes =
+        CsvTransformer.getOnlyRequestedAttributes(
+            metacardList, requestedAttributes, Collections.emptySet());
+
+    assertThat(onlyRequestedAttributes, hasSize(1));
+    final String name = onlyRequestedAttributes.stream().findFirst().get().getName();
```
Variadicism: This should make an assertion that the result of `findFirst()` is present before calling `get()` and checking the value inside.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+  public void getOnlyRequestedAttributesWithExcludedConfig() {
+    Set<String> requestedAttributes = Sets.newHashSet("attribute1", "attribute2");
+    Set<String> excludedAttributes = Sets.newHashSet("attribute1");
+
+    Set<AttributeDescriptor> onlyRequestedAttributes =
+        CsvTransformer.getOnlyRequestedAttributes(
+            metacardList, requestedAttributes, excludedAttributes);
+
+    assertThat(onlyRequestedAttributes, hasSize(1));
+    final String name = onlyRequestedAttributes.stream().findFirst().get().getName();
```
Variadicism: This should make an assertion that the result of `findFirst()` is present before calling `get()` and checking the value inside.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+    Appendable csvText =
+        CsvTransformer.writeMetacardsToCsv(
+            metacardList, requestedAttributes, Collections.emptyMap());
+
+    Scanner scanner = new Scanner(csvText.toString());
+    scanner.useDelimiter(CSV_REGEX);
+
+    // Only attributes in "attributes" config field will exist in output.
+    // OBJECT types, BINARY types and excluded attributes will be filtered out
+    // excluded attribute take precedence over attributes that appear in requested attribute list.
```
Variadicism: Everything in these comments should be detailed in the implementation rather than in the test (preferably by intuitive method signatures and code, but also by Javadoc if necessary).

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+  private static final Object[][] ATTRIBUTE_DATA = {
+    {"attribute1", "value1", BasicTypes.STRING_TYPE},
+    {"attribute2", "value2", BasicTypes.STRING_TYPE},
+    {"attribute3", 101, BasicTypes.INTEGER_TYPE},
+    {"attribute4", 3.14159, BasicTypes.DOUBLE_TYPE},
+    {"attribute5", "value,5", BasicTypes.STRING_TYPE},
+    {"attribute6", "value6", BasicTypes.STRING_TYPE},
+    {"attribute7", "OBJECT", BasicTypes.OBJECT_TYPE},
+    {"attribute8", "BINARY", BasicTypes.BINARY_TYPE}
+  };
```
Variadicism: Type checking and readability can be improved by replacing these three item arrays with [Apache's `Triple`s](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/tuple/Triple.html).

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+  @Before
+  public void setup() {
+    ATTRIBUTE_DESCRIPTOR_LIST.clear();
+    metacardList.clear();
```
Variadicism: `buildMetacardList()` two lines below clears the `metacardList`; this call is not necessary.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
rzwiefel: Our checkstyle enforces that final attributes must be in all-caps snake case (for consistency).

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+
+  public static BinaryContent createResponse(final Appendable csv) {
+    InputStream inputStream =
+        new ByteArrayInputStream(csv.toString().getBytes(StandardCharsets.UTF_8));
+    return new BinaryContentImpl(inputStream, CSV_MIME_TYPE);
+  }
+
+  public static Appendable writeMetacardsToCsv(
+      final List<Metacard> metacards,
+      final List<AttributeDescriptor> sortedAttributeDescriptors,
```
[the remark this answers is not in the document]
Bdthomson: Whoever calls this method has already sorted the attribute descriptors (in whatever way they want). This method will place them in the CSV text in whatever order you pass them.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
Bdthomson: The `excluded` set was included to provide a final "Don't include this" via configuration, but I think you are right and whoever calls this function can filter out whatever requested attributes they want, and this function should not be required to filter out excluded attributes.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
Bdthomson: Normally I don't check my `get` calls with `findFirst` in tests because if the optional is not present my test should fail anyway.

Variadicism: The main difference is the ability for the test to display more useful, descriptive error messages when assertions are used instead of just waiting for Exceptions.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
Variadicism: That makes sense, but my point is that though it is technically `final`, using `final` in this way is confusing. I'm suggesting the removal of the `final` modifier.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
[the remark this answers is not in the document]
Variadicism: Then the "sorted" designation should be removed as it seems to imply some assumption about the order of the `AttributeDescriptor`s, but all it really means is that the order is respective to the order of the lines output.

catalog/transformer/catalog-transformer-csv-common/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+            <artifactId>catalog-core-api-impl</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-csv</artifactId>
+            <version>1.4</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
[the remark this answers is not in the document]
bdeining: commons-lang3 is probably in dependency management in a higher pom, which eliminates the need from declaring the version here

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
bdeining: I don't think we want to remove the `final` keyword here, either.  Knowledge of the `final` keyword should make it clear enough what the intent is.  Let's also not forget that this is a test class.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Given a list of {@link Metacard}s and a string set of hiddenFields, returns a set of {@link
+   * AttributeDescriptor}s that contains all attributes that exist on the given metacard types
+   * except those that should be excluded.
+   */
+  public static Set<AttributeDescriptor> getAllAttributes(
+      final List<Metacard> metacards, final Set<String> excludedFields) {
+
+    final Predicate<AttributeDescriptor> attributeNotBinary =
```
jrnorth: :pencil2: The `NotBinary` and `NotObject` predicates could be made static members of the class.

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
```diff
@@ -0,0 +1,111 @@
...
+public class CsvQueryResponseTransformer implements QueryResponseTransformer {
+  private static final String HIDDEN_FIELDS_KEY = "hiddenFields";
+
+  private static final String COLUMN_ORDER_KEY = "columnOrder";
+
+  private static final String COLUMN_ALIAS_KEY = "aliases";
+
+  /**
+   * @param upstreamResponse the SourceResponse to be converted.
+   * @param arguments this transformer accepts 3 parameters in the 'arguments' map. 1) key:
```
jrnorth: Might want to use some HTML formatting.

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
```diff
@@ -0,0 +1,111 @@
...
+      throws CatalogTransformerException {
+
+    List<Metacard> metacards =
+        upstreamResponse
+            .getResults()
+            .stream()
+            .map(Result::getMetacard)
+            .collect(Collectors.toList());
+
+    Set<String> hiddenFields =
```
jrnorth: There may be a way to use `getOrDefault()` for this instead.

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
```diff
@@ -0,0 +1,111 @@
...
+        Optional.ofNullable((Map<String, String>) arguments.get(COLUMN_ALIAS_KEY))
+            .orElse(Collections.emptyMap());
+
+    Set<AttributeDescriptor> allAttributeDescriptors = getAllAttributes(metacards, hiddenFields);
+
+    Set<String> requestedFields = new HashSet<>(attributeOrder);
+
+    // If requestedFields is not empty, additionally filter out all non-requested attributes.
+    if (!requestedFields.isEmpty()) {
+      allAttributeDescriptors =
```
jrnorth: ✏️ I think `allAttributeDescriptors.removeIf(ad -> !requestedFields.contains(ad.getName()));` would also work.

Variadicism: That public static method `CsvTransformer.getOnlyRequestedAttributes` might also be useful here.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,173 @@
...
+    Iterator<String> columnHeaderIterator =
+        new ColumnHeaderIterator(orderedAttributeDescriptors, aliasMap);
+    printData(csvPrinter, columnHeaderIterator);
+  }
+
+  private static void printData(final CSVPrinter csvPrinter, final Iterator iterator) {
+    try {
+      csvPrinter.printRecord(() -> iterator);
+    } catch (IOException ioe) {
+      LOGGER.debug("Failed to print the CSV data.", ioe);
```
beyelerb: Would be nice to know which metacard failed - helps track down data-specific issues when it occurs in the field.

lcrosenbu:
build now pretty please

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,231 @@
...
+  @Before
+  public void setup() {
+    ATTRIBUTE_DESCRIPTOR_LIST.clear();
+    buildMetacardDataMap();
+    buildMetacardList();
+  }
+
+  @Test
+  public void getAllCsvAttributeDescriptors() {
+    Set<String> hiddenFields = new HashSet<>();
```
jrnorth: Remove

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,231 @@
...
+  private AttributeDescriptor buildAttributeDescriptor(String name, AttributeType type) {
+    AttributeDescriptor attributeDescriptor = mock(AttributeDescriptor.class);
+    when(attributeDescriptor.getName()).thenReturn(name);
+    when(attributeDescriptor.getType()).thenReturn(type);
+    return attributeDescriptor;
+  }
+
+  private void validate(Scanner scanner, String[] expectedValues) {
+    for (String expectedValue : expectedValues) {
+      assertThat(scanner.hasNext(), is(true));
```
jrnorth: May want to add a failure message to this assertion that includes the expected value. Otherwise the failure message will just say that it was false when it was expected to be true and we'll have no idea which value it failed at.

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private static final String COLUMN_ORDER_KEY = "columnOrder";
+
+  private static final String COLUMN_ALIAS_KEY = "aliases";
+
+  /**
+   * @param upstreamResponse the SourceResponse to be converted.
+   * @param arguments this transformer accepts 2 parameters in the 'arguments' map.
+   *     <ol>
+   *       <li>key: 'attributeOrder' value: a java.utilList containing Attribute name (as Strings)
```
jrnorth: How about "a {@link List} of attribute names (as strings) that specifies the order in which the columns will appear in the output."?

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+   *           to identify the order that the columns will appear in the output.
+   *       <li>key: 'aliases' value: a java.util.Map whose keys are attribute names and values are
+   *           the how that attribute column should be aliased in the output. For example, if the
```
jrnorth: How about "...and values are the corresponding column headers that will be printed in the output."?

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * An implementation of QueryResponseTransformer that produces CSV output.
+ *
+ * @see ddf.catalog.transform.QueryResponseTransformer
+ */
+public class CsvQueryResponseTransformer implements QueryResponseTransformer {
+
+  private static final String COLUMN_ORDER_KEY = "columnOrder";
```
jrnorth: Should these be public if clients are expected to use them in the arguments map?

catalog/transformer/catalog-transformer-csv-queryresponse/src/main/java/ddf/catalog/transformer/csv/CsvQueryResponseTransformer.java
jrnorth: Also, shouldn't this key match the value of `COLUMN_ORDER_KEY`?

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
brianfelix: Added a new method for this

brianfelix:
All comments addressed

jrnorth:
Yeah those methods should just be `private static` and not `final`.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+
+  private static void printMetacardData(
+      final CSVPrinter csvPrinter, final Iterator iterator, Metacard metacard) {
+    try {
+      csvPrinter.printRecord(() -> iterator);
+    } catch (IOException ioe) {
+      LOGGER.debug(
+          "Failed to print the CSV data for metacard with id {}: {}", metacard.getId(), ioe);
```
jrnorth: Remove the second `{}`, otherwise the stack trace won't be printed.

catalog/transformer/catalog-transformer-csv-common/src/main/java/ddf/catalog/transformer/csv/common/CsvTransformer.java
```diff
@@ -0,0 +1,180 @@
...
+      metacards.forEach(
+          metacard -> printMetacard(csvPrinter, metacard, orderedAttributeDescriptors));
+
+      return csvPrinter.getOut();
+    } catch (IOException ioe) {
+      throw new CatalogTransformerException(ioe);
+    }
+  }
+
+  private static final boolean attributeNotBinary(AttributeDescriptor attributeDescriptor) {
```
jrnorth: These methods shouldn't be `final`.

catalog/transformer/catalog-transformer-csv-common/src/test/java/ddf/catalog/transformer/csv/common/CsvTransformerTest.java
```diff
@@ -0,0 +1,221 @@
...
+    AttributeDescriptor attributeDescriptor = mock(AttributeDescriptor.class);
+    when(attributeDescriptor.getName()).thenReturn(name);
+    when(attributeDescriptor.getType()).thenReturn(type);
+    return attributeDescriptor;
+  }
+
+  private void validate(Scanner scanner, String[] expectedValues) {
+    for (String expectedValue : expectedValues) {
+      assertThat(
+          "Expected next value to be " + expectedValue + "but scanner.hasNext() returned false",
```
jrnorth: Space before "but"

brianfelix:
All Comments addressed

brianfelix:
build now

brianfelix:
build now

Bdthomson:
build now

Bdthomson:
Successfully hero'd by verifying that after ingesting `ddf/catalog/spatial/kml/spatial-kml-transformer/src/test/resources/kmlPoint.kml` I was able to open up table view and export to CSV by using `Export all to CSV` and `Export visible to CSV` after hiding columns, getting the expected results.

PR #3280, opened 2018-05-31
DDF-3863 Update RESTEndpoint respect attribute/metacard overrides

clockard:
#### What does this PR do?
Updates the rest endpoint and the catalog standard framework to respect attribute/metacard overrides
#### Who is reviewing it? 
@emmberk @peterhuffer @glenhein 
@lessarderic
@rzwiefel

Ingest a metacard with an id via the rest endpoint and verify it has the same id after ingest.
Ingest a metacard with id and product together and verify the metacard has the same id after ingest

To test general attribute override run the command `curl -k -X POST -i -H "Content-Type: multipart/mixed" -F parse.resource=@/path/to/txt/file -F parse.media.duration=5.0 https://localhost:8993/services/catalog` and verify that the ingested text file has the media.duration attribute set.
[DDF-3863](https://codice.atlassian.net/browse/DDF-3863)

### DIFF
--- a/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OverrideAttributesSupport.java
+++ b/catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OverrideAttributesSupport.java
@@ -16,9 +16,9 @@ package ddf.catalog.impl.operations;
 import ddf.catalog.content.data.ContentItem;
 import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.Metacard;
-import ddf.catalog.data.MetacardType;
 import ddf.catalog.data.impl.AttributeImpl;
 import ddf.catalog.data.impl.MetacardImpl;
+import ddf.catalog.data.impl.MetacardTypeImpl;
 import java.io.Serializable;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -43,23 +43,21 @@ public class OverrideAttributesSupport {
       Metacard currentMetacard = metacardMap.get(contentItem.getId());
       Metacard overrideMetacard = contentItem.getMetacard();
 
-      Metacard updatedMetacard = overrideMetacard(currentMetacard, overrideMetacard, false, false);
+      Metacard updatedMetacard = overrideMetacard(currentMetacard, overrideMetacard, false);
 
       metacardMap.put(contentItem.getId(), updatedMetacard);
     }
   }
 
   public static Metacard overrideMetacard(
-      Metacard currentMetacard,
-      Metacard overrideMetacard,
-      boolean ignoreType,
-      boolean onlyFillNull) {
-    MetacardType updatedMetacardType = currentMetacard.getMetacardType();
-    if (!ignoreType && !MetacardImpl.BASIC_METACARD.equals(overrideMetacard.getMetacardType())) {
-      updatedMetacardType = overrideMetacard.getMetacardType();
-    }
+      Metacard currentMetacard, Metacard overrideMetacard, boolean onlyFillNull) {
 
-    Metacard updatedMetacard = new MetacardImpl(currentMetacard, updatedMetacardType);
+    MetacardImpl updatedMetacard = new MetacardImpl(currentMetacard);
+    updatedMetacard.setType(
+        new MetacardTypeImpl(
+            currentMetacard.getMetacardType().getName(),
+            currentMetacard.getMetacardType(),
+            overrideMetacard.getMetacardType().getAttributeDescriptors()));
 
     addAttributes(updatedMetacard, overrideMetacard, onlyFillNull);
     return updatedMetacard;
@@ -67,7 +65,7 @@ public class OverrideAttributesSupport {
 
   private static void addAttributes(
       Metacard metacard, Metacard otherMetacard, boolean onlyFillNull) {
-    otherMetacard
+    metacard
         .getMetacardType()
         .getAttributeDescriptors()
         .stream()
--- a/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
+++ b/catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
@@ -14,7 +14,6 @@
 package ddf.catalog.impl.operations;
 
 import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.doReturn;
@@ -23,6 +22,7 @@ import static org.mockito.Mockito.mock;
 import com.google.common.collect.ImmutableMap;
 import ddf.catalog.content.data.ContentItem;
 import ddf.catalog.content.data.impl.ContentItemImpl;
+import ddf.catalog.data.AttributeDescriptor;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.MetacardType;
 import ddf.catalog.data.impl.AttributeDescriptorImpl;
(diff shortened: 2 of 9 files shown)


clockard:
@emmberk Added a couple unit tests

clockard:
build now

clockard:
@rzwiefel Looks like it gets overwritten if it is an invalid id (which I think is ok) or if the resource uri is present and it not 'our' content scheme. I didn't realize the content scheme part and am trying to understand why we would change a valid metacard id because the resource uri isn't a certain scheme. Makes no sense to me. I am wondering if the uri check should be removed.

clockard:
build now

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
```diff
@@ -154,28 +157,35 @@ public void testOverrideMetacardIgnoreType() throws URISyntaxException {
...
-        OverrideAttributesSupport.overrideMetacard(metacard, overrideMetacard, false, false);
+        OverrideAttributesSupport.overrideMetacard(metacard, overrideMetacard, false);
 
     assertThat(updatedMetacard.getMetadata(), is("updated"));
     assertThat(updatedMetacard.getTitle(), is("updated"));
     assertThat(updatedMetacard.getResourceURI().toString(), is("content:newstuff"));
     assertThat(updatedMetacard.getId(), is("original"));
-    assertThat(updatedMetacard.getMetacardType().getName(), is("other"));
+    assertThat(updatedMetacard.getMetacardType().getName(), is("ddf.metacard"));
+    assertThat(
```
lessarderic: ❓Should we also assert this?
```
assertThat(
    updatedMetacard.getMetacardType().getAttributeDescriptors(),
    containsInAnyOrder(
        SetUtils.union(
            metacard.getMetacardType().getAttributeDescriptors(),
            updatedMetacard.getMetacardType().getAttributeDescriptors())));
```

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
```diff
@@ -154,28 +157,35 @@ public void testOverrideMetacardIgnoreType() throws URISyntaxException {
...
+        updatedMetacard.getMetacardType().getAttributeDescriptor(Media.DURATION), notNullValue());
+    assertThat(updatedMetacard.getAttribute(Media.DURATION).getValue(), is(1.0));
   }
 
   @Test
```
lessarderic: ❗️We should also add tests for the case where `onlyFillNull` is `true`.

catalog/rest/catalog-rest-endpoint/src/main/java/org/codice/ddf/endpoints/rest/RESTEndpoint.java
```diff
@@ -949,25 +963,38 @@ public Response addDocument(
     if (metacard == null) {
       metacard = new MetacardImpl();
     }
+
+    Set<AttributeDescriptor> missingDescriptors = new HashSet<>();
```
lessarderic: ❓Extract `private Set<AttributeDescriptor> setMetacardAttributes(metacard, attributes)`?
We could also include L977-982 and name extract `private Metacard overrideAttributes(metacard, attributes);`.

catalog/rest/catalog-rest-endpoint/src/test/java/org/codice/ddf/endpoints/rest/RestEndpointTest.java
```diff
@@ -318,6 +421,43 @@ BundleContext getBundleContext() {
     assertThat(response.getMetadata().get(Metacard.ID).get(0).toString(), equalTo(SAMPLE_ID));
   }
 
+  @Test
+  @SuppressWarnings({"unchecked"})
+  public void testAddDocumentWithMetadataMcardId() throws Exception {
```
lessarderic: ✏️ Spell out `Mcard` in method names?

lessarderic:
@clockard, I don't think we did, but if we changed the way attributes are being overridden when a metacard is provided as metadata attachment, we should update the documentation accordingly.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OverrideAttributesSupport.java
```diff
@@ -43,31 +43,29 @@ public static void overrideAttributes(
...
-      boolean ignoreType,
-      boolean onlyFillNull) {
-    MetacardType updatedMetacardType = currentMetacard.getMetacardType();
-    if (!ignoreType && !MetacardImpl.BASIC_METACARD.equals(overrideMetacard.getMetacardType())) {
-      updatedMetacardType = overrideMetacard.getMetacardType();
-    }
+      Metacard currentMetacard, Metacard overrideMetacard, boolean onlyFillNull) {
 
-    Metacard updatedMetacard = new MetacardImpl(currentMetacard, updatedMetacardType);
+    MetacardImpl updatedMetacard = new MetacardImpl(currentMetacard);
```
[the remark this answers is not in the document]
clockard: That constructor creates a new internal map and only includes the attributes from the metacard type descriptors. It could work here but there is still the possibility that some attributes don't have descriptors. I know you can't count on attributes making it through without a descriptor but this way we don't loose them here. Was just thinking this was the safest way but could change it if you want

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/impl/operations/OverrideAttributesSupport.java
```diff
@@ -43,31 +43,29 @@ public static void overrideAttributes(
...
+    updatedMetacard.setType(
+        new MetacardTypeImpl(
+            currentMetacard.getMetacardType().getName(),
```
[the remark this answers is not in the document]
clockard: Shouldn't be. The descriptors should be stored in a set so we won't have duplicates. We keep the original (current) metacard type name because that is based on what the input product was determined to be whereas the override doesn't necessarily respect that.

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
```diff
@@ -125,11 +128,11 @@ public void testOverrideAttributesOther() throws URISyntaxException {
     assertThat(metacardMap.get("original").getTitle(), is("updated"));
     assertThat(metacardMap.get("original").getResourceURI().toString(), is("content:newstuff"));
     assertThat(metacardMap.get("original").getId(), is("original"));
-    assertThat(metacardMap.get("original").getMetacardType().getName(), is("other"));
+    assertThat(metacardMap.get("original").getMetacardType().getName(), is("ddf.metacard"));
```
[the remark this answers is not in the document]
clockard: Changed to `testOverrideAttributes()`

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
clockard: Sure. Updated

catalog/core/catalog-core-standardframework/src/test/java/ddf/catalog/impl/operations/OverrideAttributesSupportTest.java
clockard: Added test for this case

catalog/rest/catalog-rest-endpoint/src/test/java/org/codice/ddf/endpoints/rest/RestEndpointTest.java
clockard: Changed

clockard:
@lessarderic The changes made should make the system better reflect what the documentation currently says.

clockard:
build now

clockard:
build now

clockard:
exam setup failure 
build now

clockard:
Bypassing the sonarqube findings since the critical findings are not in the change set.

PR #3315, opened 2018-06-08
DDF-3878 Allow anyText search attributes to be configured

troymohl:
#### What does this PR do?
Adds optional fields to Solr provider and Solr cache on which attributes to search for anyText queries. This will help alleviate performances problems in cases where the number of attributes expands to hundreds for anyText.

#### Who is reviewing it? 
@bdeining 
@jlcsmith 
@pklinef 

https://github.com/orgs/codice/teams/solr

@coyotesqrl
@millerw8

Full build with tests
Install DDF
Admin set whitelist for Catalog -> Solr Cache (use attribute name. e.g. title)
Set whitelist for Solr -> Solr Provider (use same attribute name above)
Enable DEBUG logging for:
`log:set DEBUG ddf.catalog.source.solr.SolrMetacardClientImpl`
Tail DDF log and perform a anyText query in the UI (e.g. simple search form). You should see only title_txt along with a couple of other fields in the query. Remove the configurations made above. Verify that Basic Metacard fields are now default expanded for anyText search. 

When hundreds of attributes are expanded queries will slow to a crawl on the server as Solr attempts for find matching document in each of the indices. This allows administrators to configure Solr to search a subset of their total attributes or possibly just a single attribute containing relevant metadata.

The default out of box configuration is the same where every STRING attribute type is expanded and searched.

[DDF-3878](https://codice.atlassian.net/browse/DDF-3878)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
+++ b/catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
@@ -126,6 +126,8 @@ public class DynamicSchemaResolver {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DynamicSchemaResolver.class);
 
+  private static final String ANY_ATTRIBUTE = "*";
+
   static {
     ClassLoader tccl = Thread.currentThread().getContextClassLoader();
     try {
@@ -148,6 +150,8 @@ public class DynamicSchemaResolver {
 
   protected Set<String> anyTextFieldsCache = new HashSet<>();
 
+  protected Set<String> anyTextAttributesWhitelist = new HashSet<>();
+
   protected SchemaFields schemaFields;
 
   protected Cache<String, MetacardType> metacardTypesCache =
@@ -156,9 +160,18 @@ public class DynamicSchemaResolver {
   protected Cache<String, byte[]> metacardTypeNameToSerialCache =
       CacheBuilder.newBuilder().maximumSize(4096).initialCapacity(64).build();
 
+  protected List<String> additionalFields = new ArrayList<>();
+
   private Processor processor = new Processor(new Config());
 
-  public DynamicSchemaResolver(List<String> additionalFields) {
+  public DynamicSchemaResolver(
+      List<String> anyTextAttributeWhitelist, List<String> additionalFields) {
+    if (anyTextAttributeWhitelist != null) {
+      this.anyTextAttributesWhitelist.addAll(anyTextAttributeWhitelist);
+    }
+    if (additionalFields != null) {
+      this.additionalFields.addAll(additionalFields);
+    }
     this.schemaFields = new SchemaFields();
 
     fieldsCache.add(Metacard.ID + SchemaFields.TEXT_SUFFIX);
@@ -167,15 +180,19 @@ public class DynamicSchemaResolver {
         Metacard.ID + SchemaFields.TEXT_SUFFIX + SchemaFields.TOKENIZED + SchemaFields.HAS_CASE);
     fieldsCache.add(Metacard.TAGS + SchemaFields.TEXT_SUFFIX);
 
-    anyTextFieldsCache.add(Metacard.METADATA + SchemaFields.TEXT_SUFFIX);
+    if (shouldAddAnyTextAttribute(Metacard.METADATA)) {
+      anyTextFieldsCache.add(Metacard.METADATA + SchemaFields.TEXT_SUFFIX);
+    }
     Set<String> basicTextAttributes =
         MetacardImpl.BASIC_METACARD
             .getAttributeDescriptors()
             .stream()
             .filter(descriptor -> BasicTypes.STRING_TYPE.equals(descriptor.getType()))
+            .filter(descriptor -> shouldAddAnyTextAttribute(descriptor.getName()))
             .map(stringDescriptor -> stringDescriptor.getName() + SchemaFields.TEXT_SUFFIX)
             .collect(Collectors.toSet());
     anyTextFieldsCache.addAll(basicTextAttributes);
+
     fieldsCache.add(Validation.VALIDATION_ERRORS + SchemaFields.TEXT_SUFFIX);
     fieldsCache.add(Validation.VALIDATION_WARNINGS + SchemaFields.TEXT_SUFFIX);
 
@@ -190,7 +207,7 @@ public class DynamicSchemaResolver {
   }
 
   public DynamicSchemaResolver() {
-    this(Collections.emptyList());
+    this(Collections.emptyList(), Collections.emptyList());
   }
 
   /**
@@ -238,7 +255,13 @@ public class DynamicSchemaResolver {
       String key = e.getKey();
       fieldsCache.add(key);
       if (key.endsWith(SchemaFields.TEXT_SUFFIX)) {
-        anyTextFieldsCache.add(key);
+        String attributeName = key;
+        if (key.contains("_")) {
+          attributeName = key.substring(0, key.indexOf('_'));
+        }
+        if (StringUtils.isNotBlank(attributeName) && shouldAddAnyTextAttribute(attributeName)) {
+          anyTextFieldsCache.add(key);
+        }
       }
     }
   }
@@ -698,7 +721,9 @@ public class DynamicSchemaResolver {
                 + schemaFields.getFieldSuffix(format)
                 + getSpecialIndexSuffix(format)
                 + SchemaFields.HAS_CASE);
-        anyTextFieldsCache.add(ad.getName() + schemaFields.getFieldSuffix(format));
+        if (shouldAddAnyTextAttribute(ad.getName())) {
+          anyTextFieldsCache.add(ad.getName() + schemaFields.getFieldSuffix(format));
+        }
       }
 
       if (format.equals(AttributeFormat.XML)) {
@@ -710,6 +735,9 @@ public class DynamicSchemaResolver {
                 + SchemaFields.HAS_CASE);
         fieldsCache.add(
             ad.getName() + schemaFields.getFieldSuffix(format) + getSpecialIndexSuffix(format));
+        if (anyTextAttributesWhitelist.contains(ad.getName())) {
+          anyTextFieldsCache.add(ad.getName() + SchemaFields.TEXT_SUFFIX + SchemaFields.TOKENIZED);
+        }
       }
     }
   }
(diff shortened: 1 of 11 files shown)


coyotesqrl:
❗️ At one level, there is nothing wrong with this PR. 
- It is implemented correctly
- It provides a solution to an identified performance problem
- It allows for backwards compatibility

But at another level, this is completely wrong, symptomatic and endemic to issues that plague DDF.

DDF is a huge, complex beast and there is little that can be done about that; however, in our efforts to deal with its size and complexity we make it _more complex_. Every tweak, every configuration toggle, every dynamic conditional addition to satisfy real users, imagined users, Platonic ideals of users, and users who may never touch the product but might - _if only they had that one extra configurable option!_ - makes it more brittle, more fragile, and harder to use. Even if the thinking is that this is a short-term solution, we all know too well that tech debt of this sort is rarely paid down.

One of its core tenets is that DDF should "Unzip and Run!" but every dial, knob, and switch added to the system pushes it further away from that. It's not easily administered if it requires consulting a thousand pages of documentation, hitting the user groups, tweaking arcane configurations, and chanting incantations over firelight.

I am not rejecting this change, but I can't in good conscience approve it. I leave it to my fellow committers to decide its fate.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
```diff
@@ -698,6 +698,10 @@ public String getIdentifierQuery(String fieldName, List<? extends Serializable>
     return queryBuilder.toString();
   }
 
+  public void setAnyTextAttributes(List<String> anyTextAttributes) {
+    resolver = new DynamicSchemaResolver(anyTextAttributes, resolver.additionalFields);
```
paouelle: ❗️ We shouldn't be exposing `additionalFields` like this. It breaks encapsulation. If we need to create a new DynamicSchemaResolver from a previous one then let's expose either a constructor that receives another resolver and the new text attributes or a method on the original one that creates a new one from itself with new text attributes

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
```diff
@@ -156,9 +160,18 @@
   protected Cache<String, byte[]> metacardTypeNameToSerialCache =
       CacheBuilder.newBuilder().maximumSize(4096).initialCapacity(64).build();
 
+  protected List<String> additionalFields = new ArrayList<>();
```
paouelle: ❓ Can we not expose the internal list for testing? This makes the tests brittle. It also opens up the implementation to unexpected changes.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
```diff
@@ -148,6 +150,8 @@
 
   protected Set<String> anyTextFieldsCache = new HashSet<>();
 
+  protected Set<String> anyTextAttributesWhitelist = new HashSet<>();
```
paouelle: ❓ Does it have to be protected?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/RemoteSolrCatalogProvider.java
```diff
@@ -47,6 +50,16 @@
 
   private static final Properties DESCRIBABLE_PROPERTIES = new Properties();
 
+  private FilterAdapter filterAdapter;
+
+  private SolrClient client;
+
+  private SolrFilterDelegateFactory solrFilterDelegateFactory;
+
+  protected DynamicSchemaResolver resolver;
```
paouelle: ❓ Why does it have to be protected? Again opening up the implementation to outside changes.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/RemoteSolrCatalogProvider.java
```diff
@@ -47,6 +50,16 @@
...
+
+  protected List<String> anyTextAttributes = new ArrayList<>();
```
paouelle: ❓ Same question about being protected

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/RemoteSolrCatalogProvider.java
```diff
@@ -74,15 +87,22 @@ public RemoteSolrCatalogProvider(
...
+    this.client = client;
+    this.solrFilterDelegateFactory = solrFilterDelegateFactory;
+    this.resolver = resolver;
+    init();
     provider.maskId(getId());
   }
 
+  protected void init() {
+    if (resolver == null) {
+      resolver = new DynamicSchemaResolver(anyTextAttributes, Collections.emptyList());
```
paouelle: ❗️ This should be moved to the constructor to avoid potentially leaving `resolver` to `null` and making it clear that it can never be.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/RemoteSolrCatalogProvider.java
```diff
@@ -74,15 +87,22 @@ public RemoteSolrCatalogProvider(
...
-            (resolver == null) ? new DynamicSchemaResolver() : resolver);
+    this.filterAdapter = filterAdapter;
+    this.client = client;
+    this.solrFilterDelegateFactory = solrFilterDelegateFactory;
+    this.resolver = resolver;
+    init();
     provider.maskId(getId());
   }
 
+  protected void init() {
```
paouelle: ❗️ This method is meant to finalize the initialization of the object via its constructor and as such, it must be private. Making it protected allows subclasses to override it and be called before their own constructor is ever called.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/RemoteSolrCatalogProvider.java
```diff
@@ -74,15 +87,22 @@ public RemoteSolrCatalogProvider(
...
+    if (resolver == null) {
+      resolver = new DynamicSchemaResolver(anyTextAttributes, Collections.emptyList());
+    }
+    this.provider =
```
paouelle: ❗️ This should also be done in the ctor to clarify that the `provider` attribute should never be `null` in this class. Plus it would allow it to remain `final` making it clear it doesn't change and making thus avoiding any potential multi-threading issue around it.

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/SolrMetacardClientImpl.java
paouelle: ❓ How do we deal with synchronization issues now that the resolver can be changed from any threads whereas before it was final?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SolrCache.java
```diff
@@ -292,6 +294,16 @@ public void removeById(String[] ids) throws IOException, SolrServerException {
     return getMetacardsFromResponse(response);
   }
 
+  public void setAnyTextAttributes(List<String> anyTextAttributes) {
```
paouelle: ❗️ We are clearing the content if the list to add items in it. What is the though about multi-threading here?

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SolrCache.java
```diff
@@ -292,6 +294,16 @@ public void removeById(String[] ids) throws IOException, SolrServerException {
...
+    this.anyTextAttributes.clear();
+    if (CollectionUtils.isNotEmpty(anyTextAttributes)) {
```
paouelle: ✏️ By design it cannot be `null` so simply call its `.isEmpty()` method

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SolrCache.java
```diff
@@ -292,6 +294,16 @@ public void removeById(String[] ids) throws IOException, SolrServerException {
...
+      this.anyTextAttributes.addAll(anyTextAttributes);
+    }
+    if (metacardClient instanceof CacheSolrMetacardClient) {
+      ((CacheSolrMetacardClient) metacardClient).setAnyTextAttributes(anyTextAttributes);
```
paouelle: 🤔 This seems like a smell to me if we have to check the implementation of the client ...

catalog/solr/catalog-solr-provider/src/main/java/ddf/catalog/solr/provider/SolrCatalogProvider.java
```diff
@@ -30,4 +32,13 @@ public SolrCatalogProvider(
         solrFilterDelegateFactory,
         null);
   }
+
+  public void setAnyTextAttributes(List<String> anyTextAttributes) {
+    super.resolver = null;
+    this.anyTextAttributes.clear();
+    if (CollectionUtils.isNotEmpty(anyTextAttributes)) {
```
paouelle: ❗️ Do we support the list to be `null`? I don't think we should and as such, we should only need to call its `.isEmpty()` method. In any case, you would get a NPE in the above line. In addition. `addAll()` would take care of the case when there is no elements in the list so no need to check.
❗️ I am still concern about multi-threading issue since we first clear the list before adding to it.

catalog/core/catalog-core-standardframework/src/main/java/ddf/catalog/cache/solr/impl/SolrCache.java
dimitry-sherman: Shouldn't the `instanceof` check be done before the attributes are cleared?

catalog/core/catalog-core-solr/src/main/java/ddf/catalog/source/solr/DynamicSchemaResolver.java
```diff
@@ -238,7 +255,10 @@ public void addFieldsFromClient(SolrClient client) throws SolrServerException, I
       String key = e.getKey();
       fieldsCache.add(key);
       if (key.endsWith(SchemaFields.TEXT_SUFFIX)) {
-        anyTextFieldsCache.add(key);
+        String attributeName = key.substring(0, key.indexOf(SchemaFields.TEXT_SUFFIX));
```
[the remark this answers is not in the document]
millerw8: That's an awkward read.  Could do substring with string lengths though

PR #3331, opened 2018-06-14
DDF-3905 add support for MATCHCASE in UI search forms

glenhein:
#### What does this PR do?
Add support for MATCHCASE in UI search forms.

#### Who is reviewing it? 
@Lambeaux 
@Bdthomson 
@brianfelix 
@dimitry-sherman

@bdeining
@millerw8
Create a search form that uses MATCHCASE. Make sure the forms saves and reloads.

[DDF-3905](https://codice.atlassian.net/browse/DDF-3905)
#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


catalog/ui/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/forms/filter/VisitableXmlElementImplTest.java
```diff
@@ -128,8 +128,10 @@ public void testBinaryComparisonOpType() {
 
   @Test
   public void testPropertyIsLikeType() {
-    parameterizedMappingVerification(
-        PropertyIsLikeType.class, (v) -> verify(v).visitPropertyIsLikeType(eq(node)));
+    PropertyIsLikeType propertyIsLikeType = new PropertyIsLikeType();
+    propertyIsLikeType.setMatchCase(true);
+    propertyIsLikeTypeVerification(
+        propertyIsLikeType, (v) -> verify(v).visitPropertyIsLikeType(eq(node)));
```
Lambeaux: ❗️ We should probably have another test for `setMatchCase(false)` and verify the following: 
```
(v) -> verify(v).visitPropertyIsILikeType(eq(node)));
```

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/model/FilterNodeMapImpl.java
```diff
@@ -69,7 +70,13 @@ public String getOperator() {
...
         .flatMap(List::stream)
         .map(Map.class::cast)
-        .map(FilterNodeMapImpl::new)
+        .map(
+            new Function<Map, FilterNodeMapImpl>() {
+              @Override
+              public FilterNodeMapImpl apply(Map json1) {
+                return new FilterNodeMapImpl(json1);
+              }
+            })
```
Lambeaux: ❓ Can this be collapsed down? 
```
.map(json -> new FilterNodeMapImpl(json))
```

Actually, shouldn't the old syntax still work? 
```
.map(FilterNodeMapImpl::new)
```

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/builder/XmlModelBuilder.java
```diff
@@ -180,14 +184,18 @@ public XmlModelBuilder beginBinarySpatialType(String operator) {
...
+        spatialMapping, "Cannot find mapping for binary spatial operator: " + operator);
     supplierInProgress = new TerminalNodeSupplier(spatialMapping);
     return this;
   }
 
+  private void validateOperatorMapping(MultiNodeReducer multiNodeReducer, String message) {
+    if (multiNodeReducer == null) {
+      throw new IllegalArgumentException(message);
+    }
+  }
```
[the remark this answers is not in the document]
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/filter/VisitableXmlElementImpl.java
```diff
@@ -61,24 +61,27 @@
...
+              .put(LiteralType.class, FilterVisitor2::visitLiteralType)
+              .put(FunctionType.class, FilterVisitor2::visitFunctionType)
+              .put(BinaryLogicOpType.class, FilterVisitor2::visitBinaryLogicType)
+              .put(UnaryLogicOpType.class, FilterVisitor2::visitUnaryLogicType)
+              .put(BinaryTemporalOpType.class, FilterVisitor2::visitBinaryTemporalType)
+              .put(BinarySpatialOpType.class, FilterVisitor2::visitBinarySpatialType)
+              .put(DistanceBufferType.class, FilterVisitor2::visitDistanceBufferType)
+              .put(BBOXType.class, FilterVisitor2::visitBoundingBoxType)
+              .put(BinaryComparisonOpType.class, FilterVisitor2::visitBinaryComparisonType)
+              .put(PropertyIsLikeType.class, VisitableXmlElementImpl::acceptPropertyIsLike)
```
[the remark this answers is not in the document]
glenhein: The method itself is documented, and I'm hesitant to add a code comment outside of javadoc.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/filter/VisitableXmlElementImpl.java
```diff
@@ -61,24 +61,27 @@
...
-          .put(BinarySpatialOpType.class, FilterVisitor2::visitBinarySpatialType)
-          .put(DistanceBufferType.class, FilterVisitor2::visitDistanceBufferType)
-          .put(BBOXType.class, FilterVisitor2::visitBoundingBoxType)
-          .put(BinaryComparisonOpType.class, FilterVisitor2::visitBinaryComparisonType)
-          .put(PropertyIsLikeType.class, FilterVisitor2::visitPropertyIsLikeType)
-          .put(PropertyIsNullType.class, FilterVisitor2::visitPropertyIsNullType)
-          .put(PropertyIsNilType.class, FilterVisitor2::visitPropertyIsNilType)
-          .put(PropertyIsBetweenType.class, FilterVisitor2::visitPropertyIsBetweenType)
-          .build();
+  private static final Boolean DEFAULT_MATCH_CASE = Boolean.TRUE;
```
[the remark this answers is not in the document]
glenhein: According to the XML schema, the default is already TRUE. So I was just trying to be consistent with that.

catalog/ui/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/forms/filter/VisitableXmlElementImplTest.java
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/model/FilterNodeMapImpl.java
glenhein: done

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/api/FlatFilterBuilder.java
```diff
@@ -45,7 +45,9 @@
 
   FlatFilterBuilder beginBinaryComparisonType(String operator);
 
-  FlatFilterBuilder beginPropertyIsLikeType(String operator, boolean matchCase);
+  FlatFilterBuilder beginPropertyIsLikeType(String operator);
+
+  FlatFilterBuilder beginPropertyIsILikeType(String operator);
```
AzGoalie: What does the `I` stand for?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/builder/XmlModelBuilder.java
```diff
@@ -142,35 +140,35 @@ public XmlModelBuilder beginBinaryComparisonType(String operator) {
...
-    }
+    validateOperatorMapping(
+        comparisonMapping, "Cannot find mapping for like operator: " + operator);
     supplierInProgress = new TerminalNodeSupplier(comparisonMapping);
     return this;
   }
 
+  @Override
+  public FlatFilterBuilder beginPropertyIsILikeType(String operator) {
+    return beginPropertyIsLikeType(operator);
```
Variadicism: :question: I don't see the part of this code that differentiates between `ILIKE` and `LIKE` comparisons. Are they meant to be equivalent? If so, why? Isn't `ILIKE` case-insensitive while `LIKE` is case-sensitive?

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/filter/VisitableXmlElementImpl.java
```diff
@@ -464,4 +468,49 @@ private LiteralElement(JAXBElement element) {
       return value;
     }
   }
+
+  /**
+   * NB: This should only be called when {@code visitable} is guaranteed to contain a {@link
```
Variadicism: :question: What does "NB:" mean? Is this an established or documented convention? If not, I would like to see it replaced with something more meaningful.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/api/FlatFilterBuilder.java
Lambeaux: "Insensitive" ... i.e. "Case Insensitive LIKE"

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/builder/XmlModelBuilder.java
Lambeaux: There are two directions in play here, JSON to XML and vice versa. At best, we can simply move the "kludgy-ness" around the codebase without removing it completely (that will come later when Geotools gets upgraded, hopefully). 

Normally Filter 2.0 only has a `PropertyIsLikeType` which itself has a boolean attribute for `matchCase`. Since we are mapping to the UI's model, and it distinguishes case sensitivity as two different operations (the CQL `LIKE` and `ILIKE` operators) we decided it best to be declarative until subsequent improvements can be made.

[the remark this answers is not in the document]
Lambeaux: Because at the XML level, the difference is the value of an attribute, not the operator itself.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/filter/VisitableXmlElementImpl.java
glenhein: https://en.wikipedia.org/wiki/Nota_bene

It's been an established convention since the year 1711 :-)

Variadicism: Fair enough.

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/forms/builder/XmlModelBuilder.java
Variadicism: So if I'm understanding correctly, Filter 2.0 can really only do `LIKE`, but we have to support `ILIKE`, so this makes `ILIKE` just work like `LIKE`?

Variadicism: Glen helped explain in person that the operator string "LIKE" or "ILIKE"
 is passed into this method; then, at [this line in `beginPropertyIsLikeType`](https://github.com/codice/ddf/pull/3331/files/154125c740d717194b4078b821d0c874b2c0b01c#diff-9c1e9e448375a9b21e42d9b1e5e109a2R153), the appropriate operation is retrieved using the operator as the key "LIKE" or "ILIKE" from the map `TERMINAL_OPS`. [These two lines](https://github.com/codice/ddf/pull/3331/files/154125c740d717194b4078b821d0c874b2c0b01c#diff-9c1e9e448375a9b21e42d9b1e5e109a2R63) add LIKE and ILIKE to `TERMINAL_OPS` so that either can be found with this map retrieval.

That's how this code differentiates between LIKE and ILIKE even though both are calling `beginPropertyIsLikeType`.

Consider this :pencil2:, but I think that this confusion could be avoided by factoring the body of `beginPropertyIsLikeType` into a new private method, e.g., `beginPropertyIsLikeOrILikeType`, and making `beginPropertyIsLikeType` and `beginPropertyIsILikeType` call that.

glenhein: That's a good idea. Done.

catalog/ui/catalog-ui-search/src/main/java/net/opengis/filter/v_2_0/PropertyIsLikeType.java
```diff
@@ -0,0 +1,756 @@
...
+  public StringBuilder appendFields(
+      ObjectLocator locator, StringBuilder buffer, ToStringStrategy2 strategy) {
+    super.appendFields(locator, buffer, strategy);
+    {
+      List<JAXBElement<?>> theExpression;
+      theExpression = (this.isSetExpression() ? this.getExpression() : null);
+      strategy.appendField(
+          locator, this, "expression", buffer, theExpression, this.isSetExpression());
+    }
+    {
```
GabrielFabian: ❓ Why the curly braces in this method?

glenhein: This class came from another library, in which it was auto-generated with an xml binding tool. We needed to fix it to account for a missing xml attribute that is allowed by the schema. There is a TODO at the top of the class to replace class once a proper xml binding library is found.

Lambeaux: I think in this case the issue is the schema is not wrong. It's correct. The JAXB bindings are wrong. Hence our need for only overriding one of its classes. I think this works as-is. If we need to enhance comments at the top to make it clearer as to what's going on, we should definitely do that.

catalog/ui/catalog-ui-search/src/main/java/net/opengis/filter/v_2_0/PropertyIsLikeType.java
```diff
@@ -0,0 +1,756 @@
...
+        } else {
+          if (wildCardShouldBeCopiedAndSet == Boolean.FALSE) {
+            copy.wildCard = null;
+          }
+        }
+      }
+      {
+        Boolean singleCharShouldBeCopiedAndSet =
+            strategy.shouldBeCopiedAndSet(locator, this.isSetSingleChar());
+        if (singleCharShouldBeCopiedAndSet == Boolean.TRUE) {
```
bdeining: Why `== Boolean.TRUE`?

Lambeaux:
@bdeining I would argue, since `PropertyIsLikeType.java` doesn't belong to us and is a temporary measure due to an issue in the library itself, we keep it "as is" and do not modify it more than we have to.

bdeining:
ah; i didn't realize we are copying this in from somewhere

catalog/ui/catalog-ui-search/src/main/java/net/opengis/filter/v_2_0/PropertyIsLikeType.java
Lambeaux: Wait a minute that doesn't make sense. If these were auto-generated classes from the schema, where did the error occur? Looking at the official schema at http://schemas.opengis.net/filter/2.0/filter.xsd if you search for `PropertyIsLikeType` you'll see the attribute is there: 
```
<xsd:complexType name="PropertyIsLikeType">
  <xsd:complexContent>
    <xsd:extension base="fes:ComparisonOpsType">
      <xsd:sequence>
        <xsd:element ref="fes:expression" minOccurs="2" maxOccurs="2"/>
      </xsd:sequence>
      <xsd:attribute name="wildCard" type="xsd:string" use="required"/>
      <xsd:attribute name="singleChar" type="xsd:string" use="required"/>
      <xsd:attribute name="escapeChar" type="xsd:string" use="required"/>
      <xsd:attribute name="matchCase" type="xsd:boolean" use="optional" default="true"/>
    </xsd:extension>
  </xsd:complexContent>
</xsd:complexType>
```

Lambeaux: They might have used the wrong schemas for code generation.

bdeining:
build now

Bdthomson:
build now (failed fetch of artifact)

hayleynorton:
hero in progress

Bdthomson:
build now (connection reset)

hayleynorton:
hero (quick) build successful. 
created, saved and reloaded a MATCHCASE search form
image
image

bdeining:
@glenhein since we are copying in 3rd party class with sonarqube findings as a temporary measure, is it possible to suppress them?

glenhein:
@bdeining I'll look into suppressing the warnings.

glenhein:
@Bdthomson Could you try adding `@Generated("This code should eventually be replaced by DDF-3926.")` to the PropertyIsLikeType class? We are trying to suppress the sonarqube warnings for the generated code.

Bdthomson:
build now

Bdthomson:
build now

catalog/ui/catalog-ui-search/src/main/java/net/opengis/filter/v_2_0/PropertyIsLikeType.java
millerw8: I don't think that is the right answer.  Checking in generated code seems to be a nightmare.  There are precedents of us generating the code during the build.  If the schema is right, we can work even less hard.  Note the ReadMe here: http://schemas.opengis.net/filter/ReadMe.txt

It seems like the schema was fixed last June to add a missing attribute.  I would prefer not to rely on a library partially, and generated code partially.  And I certainly don't like generated code being checked in

coyotesqrl: There is a plan to pull schema-derived classes out to their own repo. Hopefully that will happen sooner rather than later.

millerw8: @coyotesqrl are you saying you approve of this submission because it will be cleaned up later?

coyotesqrl: We don't generally check in generated code (in fact, I don't think we do anywhere) so I don't think it's a great workaround right here to do it as a one-off fix. I was just providing a little extra background information so if it turned out the workaround _was_ the best solution for the short-term, we would at least be comforted by the fact it would be fixed down the road.

bdeining:
will be OBE in a future ticket

PR #3338, opened 2018-06-14
DDF-3909 Updates paging buttons to use less vars

djblue:
Forward port of #3337


bdeining:
hero'd and reviewed on 2.12.x

PR #3348, opened 2018-06-20
DDF-3919 Solr connectivity alert

sblackmore:
#### What does this PR do?
Adds an alert if solr is not reachable. The alert will appear in the admin console. This PR also updates a unit test to expect an alert if solr is down. This PR relies on the new separate solr process.
#### Who is reviewing it? 
@mcalcote 
@rzwiefel 
@ahoffer 
@clockard 
@rzwiefel

Manually steps to test:
1) Start the DDF
2) Log in to the admin console
3) Manually stop solr using {DDF_HOME}/solr/bin/solr stop -p 8994
4) Wait for at least 30 seconds
5) The solr connectivity alert should eventually display in the admin console (can try refreshing)
6) Manually start solr using {DDF_HOME}/solr/bin/solr start -p 8994
7) Wait at least 30 seconds
8) Dismiss the solr alert, refresh the admin console page, and you should again see the insecure defaults message and not the solr connectivity message
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
+++ b/platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
@@ -30,6 +30,7 @@ import org.codice.ddf.persistence.PersistenceException;
 import org.codice.ddf.persistence.PersistentItem;
 import org.codice.ddf.persistence.PersistentStore;
 import org.codice.ddf.system.alerts.Alert;
+import org.codice.ddf.system.alerts.NoticePriority;
 import org.codice.ddf.system.alerts.SystemNotice;
 import org.osgi.service.event.Event;
 import org.osgi.service.event.EventAdmin;
@@ -75,6 +76,15 @@ public class AdminAlertImpl extends BasicMBean implements AdminAlertMBean {
       }
     } catch (PersistenceException pe) {
       LOGGER.debug("Error retrieving system alert.", pe);
+
+      return Collections.singletonList(
+          new Alert(
+                  "solr",
+                  NoticePriority.CRITICAL,
+                  "Persistent Storage Not Responding. Could Not Retrieve Alerts.",
+                  Collections.singleton(
+                      "Critical alerts may be present, but not displayed because the persistent storage is not responding."))
+              .getProperties());
     }
     alerts.sort(
         (map1, map2) ->
--- a/platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
+++ b/platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
@@ -79,7 +79,7 @@ public class AdminAlertImpl extends BasicMBean implements AdminAlertMBean {
 
       return Collections.singletonList(
           new Alert(
-                  "solr",
+                  "unable_to_retrieve_alerts",
                   NoticePriority.CRITICAL,
                   "Persistent Storage Not Responding. Could Not Retrieve Alerts.",
                   Collections.singleton(
--- a/platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
+++ b/platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
@@ -111,7 +111,7 @@ public class AdminAlertImplTest {
   public void testGetAlertsPersistentStoreError() throws Exception {
     when(persistentStore.get(eq("alerts"), any())).thenThrow(new PersistenceException("error"));
     List<Map<String, Object>> alertList = adminAlert.getAlerts();
-    assertThat(alertList.size(), equalTo(0));
+    assertThat(alertList.size(), equalTo(1));
   }
 
   @Test
--- a/platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
+++ b/platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
@@ -112,7 +112,8 @@ public class AdminAlertImplTest {
     when(persistentStore.get(eq("alerts"), any())).thenThrow(new PersistenceException("error"));
     List<Map<String, Object>> alertList = adminAlert.getAlerts();
     assertThat(alertList.size(), equalTo(1));
-    assertThat(alertList.get(0).get(Alert.SYSTEM_NOTICE_SOURCE_KEY), equalTo("unable_to_retrieve_alerts"));
+    assertThat(
+        alertList.get(0).get(Alert.SYSTEM_NOTICE_SOURCE_KEY), equalTo("unable_to_retrieve_alerts"));
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


ahoffer:
build now

platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
```diff
@@ -111,7 +111,7 @@ public void testGetAlertsWithSingleDetail() throws Exception {
   public void testGetAlertsPersistentStoreError() throws Exception {
     when(persistentStore.get(eq("alerts"), any())).thenThrow(new PersistenceException("error"));
     List<Map<String, Object>> alertList = adminAlert.getAlerts();
-    assertThat(alertList.size(), equalTo(0));
+    assertThat(alertList.size(), equalTo(1));
```
clockard: Could also assert that it is the solr alert here.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -75,6 +76,15 @@ public AdminAlertImpl(PersistentStore persistentStore, EventAdmin eventAdmin)
       }
     } catch (PersistenceException pe) {
       LOGGER.debug("Error retrieving system alert.", pe);
+
+      return Collections.singletonList(
```
vinamartin: ❓  Should we provide the admins with any more information to help them possibly debug the issue, such as checking logs/lock files? 
I think this is fine as is, but I'm wondering if we can make this more helpful.

ahoffer: I thought about that too. My first thought was to mention Solr, but the system might be using a different persistence implementation. I also thought about mentioned to check the logs, but that seemed like an obvious thing to do. Still, I think it would be fine to mention checking the logs.

platform/admin/core/admin-core-impl/src/test/java/org/codice/ddf/admin/core/impl/AdminAlertImplTest.java
sblackmore: I added the assertion that it is the solr alert

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
paouelle: This might also be the wrong place to make that determination as you are really high on the stack and you wouldn't know why Solr is not reachable yet. It could be because we are initializing, because it went down, because there was a communication issue, ...

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -75,6 +76,15 @@ public AdminAlertImpl(PersistentStore persistentStore, EventAdmin eventAdmin)
...
+          new Alert(
```
paouelle: ❓ What is the effect of returning an alert like this? Will it be clearable by the admin or will that fail also? If the later than it will be confusing that we show an alert but they cannot clear it.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
paouelle: ❗️ There is a difference in the collection that is returned normally versus this one. The collection returned here is immutable whereas normally it ain't. This could trigger issues from a client of this API as they would not think they could modify it. We should probably simply add to `alerts` and return that list normally.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
```diff
@@ -75,6 +76,15 @@ public AdminAlertImpl(PersistentStore persistentStore, EventAdmin eventAdmin)
...
+                  "solr",
```
paouelle: ❗️  How do you know it is a solr issue here? What you are receiving is a `PersistenceException`. The fact that the current underlying implementation of this API is using Solr is unknown to you here.

paouelle:
❓ If our goal is simply to detect when Solr is down, wouldn't that be better serve by detecting it at the lowest layer?
This solution only accounts for this situation when alerts are retrieved and not when other parts of the system are trying to access Solr.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
ahoffer: This alert is not specific to Solr. This message is generated only when the alert notification system cannot retrieve information from PersistentStorage. Persistent storage is presently backed by Solr, but is really an implementation detail. 

The current behavior is to display nothing on the Admin UI if alerts cannot be retrieved from persistent storage. This is disadvantageous to system security because severe alerts available, but because persistent storage is unavailable, nothing is displayed in the admin UI. This PR is meant to solve the "silent failure" situation.

ahoffer: What you describe is possible. It would be concerning if callers of the method were relying on the fact that the return object was/was not mutable. It is an assumption I don't think client code is entitled to make.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
ahoffer: As long the alert system cannot contact persistent storage, the alert appears. I think it would be hazardous to allow this alert to be dismissed because critical information may exist, but be unaccessible.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
ahoffer: You are correct. I hope we are not mentioning Solr by name in any of the messages. If so, we should replace references to "Solr" with "persistent storage".

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
paouelle: Since it is not documented in anyway, I don't think we can assume that. Better be safe than sorry.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
paouelle: Agreed. That being said, the whole PR was stating to have a Solr connectivity alert and that is not what we have done here.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
paouelle: That is not how the ticket was written. I would update suggest we re-write the ticket since this PR is not addressing the ticket as written.

platform/admin/core/admin-core-impl/src/main/java/org/codice/ddf/admin/core/impl/AdminAlertImpl.java
paouelle: Let's make sure we rename this to something else than `"solr"`

ahoffer:
build now

oconnormi:
☝️ need to run the formatter: 
```
[ERROR] Found 1 non-complying files, failing build
19:50:11 [ERROR] To fix formatting errors, run "mvn fmt:format"```

ahoffer:
build now

ahoffer:
@brendan-hofmann 
Yes, this is an alert that we can't get any alerts.
Yes, this should be more visible than the Solr errors in the log? It should be visible on admin UI.

PR #3361, opened 2018-06-26
DDF-3928 Add a simple Harvester API

peterhuffer:
#### What does this PR do?
DO NOT MERGE until 2.13.x branch has been created. This PR will not go into the 2.13.0 release.

Adds a simple first-pass Harvester API.

#### Who is reviewing it? 
@emmberk @kcover @brendan-hofmann 

@clockard
@lessarderic 

Ensure DDF starts correctly.

Implementations will come in subsequent PRs. This first iteration will focus on converting the CDM in place monitoring mechanisms to use the framework. Documentation will come in a later PR.

[DDF-3928](https://codice.atlassian.net/browse/DDF-3928)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/catalog-app/src/main/resources/features.xml
+++ b/catalog/catalog-app/src/main/resources/features.xml
@@ -502,6 +502,11 @@
         <bundle>mvn:ddf.catalog.async/catalog-async-processingframework/${project.version}</bundle>
     </feature>
 
+    <feature name="catalog-harvest-api" version="${project.version}"
+        description="API for Harvesting components">
+        <bundle>mvn:ddf.catalog.harvest/catalog-harvest-api/${project.version}</bundle>
+    </feature>
+
     <feature name="catalog-app" version="${project.version}"
              description="The Catalog provides a framework for storing, searching, processing, and transforming information.\nClients typically perform query, create, read, update, and delete (QCRUD) operations against the Catalog.\nAt the core of the Catalog functionality is the Catalog Framework, which routes all requests and responses through the system, invoking additional processing per the system configuration.">
         <feature>admin-app</feature>
--- a/catalog/harvest/catalog-harvest-api/pom.xml
+++ b/catalog/harvest/catalog-harvest-api/pom.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>harvest</artifactId>
+        <groupId>ddf.catalog.harvest</groupId>
+        <version>2.13.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <name>DDF :: Catalog :: Harvest :: API</name>
+    <artifactId>catalog-harvest-api</artifactId>
+    <packaging>bundle</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>ddf.catalog.core</groupId>
+            <artifactId>catalog-core-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
+            <version>${jsr305.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                    </instructions>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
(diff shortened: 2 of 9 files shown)


catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/StorageAdaptor.java
```diff
@@ -0,0 +1,50 @@
...
+package org.codice.ddf.catalog.harvest;
+
+/**
+ * A {@code StorageAdaptor} is responsible for persisting {@link HarvestedResource}s to a data
+ * store.
+ *
+ * <p><b> This code is experimental. While this class is functional and tested, it may change or be
+ * removed in a future version of the library. </b>
+ */
+public interface StorageAdaptor {
```
clockard: Remind me why we need this interface. It seems like a Listener implementation could be a storage adapter without a need for this interface but it has been a while since I looked over this design.

peterhuffer: This piece was to abstract away what was done with the harvested resource. For example, whether or not we store just the metadata of a product, or if we store the product as well as its metadata. The Listener could do that, however it would require special logic to decide what to do based on the configuration.

peterhuffer: There was talk about hiding this interface inside the Listener implementation's module since there will initially only be one implementation.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/HarvestedResource.java
```diff
@@ -0,0 +1,49 @@
...
+   *     values, or an empty {@link Map} if no overrides were specified
+   */
+  Map<String, List<String>> getAttributeOverrides();
+
+  /**
+   * A {@code Map} of properties associated with this {@code HarvestedResource}.
+   *
+   * @return a map of properties
+   */
+  Map<String, Serializable> getProperties();
```
peterhuffer: We can either use this method and pass around attribute overrides similar to how we do it other places by sticking it in this map, or have the above `getAttributeOverrides()` method. Since only the content framework supports attribute overrides, and the first implementations will only be going through the catalog framework, these properties were not being passed into the catalog framework.

catalog/catalog-app/src/main/resources/features.xml
```diff
@@ -502,6 +502,11 @@
         <bundle>mvn:ddf.catalog.async/catalog-async-processingframework/${project.version}</bundle>
     </feature>
 
+    <feature name="catalog-harvest-api" version="${project.version}"
+        description="API for Harvesting components">
+        <bundle>mvn:ddf.catalog.harvest/catalog-harvest-api/${project.version}</bundle>
+    </feature>
```
emmberk: ❓ This feature will not be started by default, right?

catalog/harvest/catalog-harvest-api/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>ddf.catalog.core</groupId>
+            <artifactId>catalog-core-api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.servicemix.bundles</groupId>
+            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
+            <version>${jsr305.version}</version>
```
emmberk: ✏️ Should the scope be `provided` here?

catalog/catalog-app/src/main/resources/features.xml
peterhuffer: Correct.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Registers a new {@link Listener} to receive events from the remote source being harvested from.
+   * Newly-registered {@link Listener}s will only receive new events from this {@code Harvester}. If
+   * a {@link Listener} is already registered, the call to this method will be a no-op.
+   *
+   * @param listener the {@link Listener} to register
+   * @throws IllegalArgumentException if the {@link Listener} cannot be registered
+   */
+  void registerListener(Listener listener);
```
brendan-hofmann: I'd prefer something like the whiteboard pattern personally. What happens if the listener comes up before the harvester?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/StorageAdaptor.java
brendan-hofmann: I favor keeping our APIs explicit while we iterate and deleting them when appropriate, for what it's worth.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Harvester.java
peterhuffer: For the first Listener implementation, Listeners would be notified when a new Harvester was registered and any interested Listeners (based on Harvester's ID) would register themselves to the Harvester. Could certainly invert that and get rid of the register/unregister methods. One thing to note is that if we get rid of these methods the API would assume a registry is available. I don't feel strongly either way though.

brendan-hofmann: I guess my other concern is that we're not specifying all of that in the API contracts. I'd rather let the framework take care of what it can handle already. I think it will be relatively trivial in the future to refactor or replace that once we start splitting things apart.

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/Listener.java
```diff
@@ -0,0 +1,49 @@
...
+ * removed in a future version of the library. </b>
+ */
+public interface Listener {
+
+  /**
+   * Called when a {@link Harvester} receives a newly-created file on a remote source.
+   *
+   * @param resource the {@link HarvestedResource} created on the remote source
+   */
+  void onCreate(HarvestedResource resource);
```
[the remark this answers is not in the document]
lessarderic: ❗️What is the `Harvester` expected to do if one of those methods fails? Should it call them again a certain number of times before giving up? Should the listener be responsible for retrying? Should we even retry?

catalog/harvest/catalog-harvest-api/src/main/java/org/codice/ddf/catalog/harvest/StorageAdaptor.java
lessarderic: Based on our discussion with @pklinef, I think we should remove this interface from the API for now.
If we do, we may want to remove the `HarvestException` class from this API bundle as well.

peterhuffer:
Closing this for now and will re-evaluate when the common ingest service is added.

PR #3351, opened 2018-06-20
DDF-3710 Updates drawing tools to take into account units

andrewkfiedler:
#### What does this PR do?
 - Previously everything in the location model got normalized to meters, but that is no longer the case.  As such the drawing tools needed to be updated to convert things to meters when using the model value, and to convert to the proper units when updating the model value.

#### Who is reviewing it? 
@bdeining
@millerw8 
@djblue 
@mackncheesiest 

#### How should this be tested? (List steps with links to updated documentation)
Open both the 2d and 3d map side by side.  Use the drawing tools to draw a point radius.  Update the units and verify that drawing changes accordingly.  Do the same with the line drawing.  

#### What are the relevant tickets?
[DDF-3710](https://codice.atlassian.net/browse/DDF-3710)

PR #3365, opened 2018-06-27
[2.12.x] DDF-3710 Updates drawing tools to take into account units

andrewkfiedler:
Backport of https://github.com/codice/ddf/pull/3351

#### What does this PR do?
 - Previously everything in the location model got normalized to meters, but that is no longer the case.  As such the drawing tools needed to be updated to convert things to meters when using the model value, and to convert to the proper units when updating the model value.

#### Who is reviewing it? 
@bdeining
@millerw8 
@djblue 
@mackncheesiest 

Open both the 2d and 3d map side by side.  Use the drawing tools to draw a point radius.  Update the units and verify that drawing changes accordingly.  Do the same with the line drawing.  

[DDF-3710](https://codice.atlassian.net/browse/DDF-3710)


bdeining:
build now

PR #3373, opened 2018-06-29
DDF-3936 Adds relative time capability to advanced search

andrewkfiedler:
#### What does this PR do?
 - Refactors the relative time component out of the query time component so it can be instantiated separately.
 - Updates the query time component to use the new relative time component.
 - Removes unused 24hr code from query time component.
 - Updates filter components to take into account the relative time capability.  There are a few gotchas since relative time reuses the `=` comparator.  Basically we need to start checking the attribute type (date, number, geo, etc) to fully know what view we need to show.  This adds a bit of complexity to the filter component but I've tried to leave comments that explain why we need to do certain things.
 - Updates the property model to add a new flag that allows disabling of the transformValue upon initialize.  This was necessary to support using the property model for relative time without making it a full fledged input (where it would implement input.view and live under that input folder).  This avoids possible recursive issues which might arise if we did so (having an input-enum inside the new input-relative-time).  It does however highlight the need to consider refactoring the property model, as it's getting to be gigantic.
 - The query time component supports passing in either a model, a value, or nothing.  If given a model it'll maintain syncing between the value.  If given a value it'll simply use that for initialization (it's up to you to call getViewValue when you decide you need the value that's been adjusted by the user).  If given nothing, uses a default value and similar to passing in a value it will expect you to call getViewValue when you decide you need the value that's been adjusted by the user.
 - Updates QueryResult collection to use the filter utility.  No real code change here, it just looks like we missed switching it over.
 - Updates local filtering to handle relative times

#### Who is reviewing it? 
@bdeining
@mackncheesiest 
@millerw8 
@adimka 

Go to the advanced search form and test out the relative comparator.  Verify it works as expected.  Go to the local filtering on the result view and verify that it works locally as well.

[DDF-3936](https://codice.atlassian.net/browse/DDF-3936)

screen shot 2018-06-28 at 7 54 28 pm


ui/packages/catalog-ui-search/src/main/webapp/js/filter.js
```diff
@@ -20,6 +20,15 @@ var wkx = require('wkx');
 var moment = require('moment');
 var cql = require('js/cql');
 
+// strip extra quotes
+const stripQuotes = (value) => {
+    return value.replace(/^"(.+(?="$))"$/, '$1');
+}
+
+const getDurationFromRelativeValue = (value) => {
+    return value.substring(9, value.length - 1);
```
bellcc: ✏️ Create a variable to describe the integer '9'

ui/packages/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
```diff
@@ -59,9 +59,34 @@ define([
...
+            propertyJSON.transformValue = false;
+        }
+
         return propertyJSON;
     }
 
+    // Relative time view is needed for the relative date, multivalue for all else
+    const determineView = (comparator) => {
+        let necessaryView;
+        switch(comparator) {
```
leo-sakh: ✏️ seems overkill to have a `switch` here since it's only one predicate (plus an inverse) that's tested.

ui/packages/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
```diff
@@ -157,6 +183,14 @@ define([
             }
             return comparator;
         },
+        // With the relative date comparator being the same as =, we need to try and differentiate them this way
+        getComparatorForFilter(filter) {
+            if (metacardDefinitions.metacardTypes[stripQuotes(filter.property)].type === 'DATE' && filter.type === '=') {
```
leo-sakh: ✏️ if `filter.property` is not a valid type this will end badly, no?

ui/packages/catalog-ui-search/src/main/webapp/component/filter/filter.view.js
andrewkfiedler: Yeah, I'm expecting this to have more in the future.  I can make it an if statement or ternary for now if people would prefer.

rymach:
👍 Hero successful

andrewkfiedler:
build now

ui/packages/catalog-ui-search/src/main/webapp/component/relative-time/relative-time.view.js
```diff
@@ -0,0 +1,141 @@
...
+var CustomElements = require('js/CustomElements');
+var PropertyView = require('component/property/property.view');
+var Property = require('component/property/property');
+var CQLUtils = require('js/CQLUtils');
+var Common = require('js/Common');
+
+
+/*
+    For specifying a relative time.  It shows a number field and a units field.
+    Supports passing in a model that will have it's value field auto updated and synced.
```
shaundmorris: ✏️ nit: `its`

shaundmorris:
@andrewkfiedler just checking is there anything blocking this PR at this point?

ui/packages/catalog-ui-search/src/main/webapp/js/filter.js
```diff
@@ -212,6 +221,26 @@ function matchesAFTER(value, filter) {
     return false;
 }
 
+function matchesRelative(value, filter) {
+    const date1 = moment(value);
+    const date2 = moment().subtract(moment.duration(getDurationFromRelativeValue(filter.value)));
+    if (date1 >= date2) {
+        return true;
+    }
+    return false;
```
Kjames5269: ✏️ could instead be written as `return date1 >= date2` to chop off some lines

PR #3378, opened 2018-06-30
DDF-3933 Fix attribute overrides in confluence source

clockard:
#### What does this PR do?
Injects attribute descriptors into confluence metacards and fixes attribute overrides so they are the correct type. Address some sonarqube findings as well.
#### Who is reviewing it? 
@peterhuffer @emmberk @kcover 
@shaundmorris 
@figliold
Install DDF and create a confluence source.
Add overrides that are in the common taxonomy but not in the confluence type. (I used `datetime.start=2018-06-28`)
Perform a query via Intrigue using the confluence source and verify the overrides show up as expected.
[DDF-3933](https://codice.atlassian.net/browse/DDF-3933)

### DIFF
--- a/catalog/confluence/catalog-confluence-source/src/main/java/org/codice/ddf/confluence/source/ConfluenceSource.java
+++ b/catalog/confluence/catalog-confluence-source/src/main/java/org/codice/ddf/confluence/source/ConfluenceSource.java
@@ -13,6 +13,8 @@
  */
 package org.codice.ddf.confluence.source;
 
+import ddf.catalog.data.AttributeDescriptor;
+import ddf.catalog.data.AttributeRegistry;
 import ddf.catalog.data.ContentType;
 import ddf.catalog.data.Metacard;
 import ddf.catalog.data.Result;
@@ -41,17 +43,23 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
 import java.net.URI;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import javax.ws.rs.core.Response;
+import javax.xml.bind.DatatypeConverter;
+import org.apache.commons.collections.MapUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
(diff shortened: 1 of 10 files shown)


catalog/confluence/catalog-confluence-source/src/main/java/org/codice/ddf/confluence/source/ConfluenceSource.java
```diff
@@ -174,18 +184,22 @@ public boolean isAvailable() {
       isAvailable = lastAvailable;
     }
 
-    if (lastAvailable != isAvailable) {
+    updateMonitors(isAvailable);
+
+    lastAvailable = isAvailable;
```
paouelle: ❗️ I see concurrency issues around `lastAvailable` and `lastAvailableDate` which can be accessed modified from many simultaneous threads. In addition because

clockard:
build now

catalog/confluence/catalog-confluence-source/src/main/java/org/codice/ddf/confluence/source/ConfluenceSource.java
```diff
@@ -351,15 +365,74 @@ public void setBodyExpansion(String bodyExpansion) {
...
+            newValue.add(override.getBytes(Charset.forName("UTF-8")));
+            break;
+          case OBJECT:
+          case STRING:
+          case GEOMETRY:
+          case XML:
+            newValue.add(override);
+            break;
+        }
+      } catch (IllegalArgumentException e) {
```
shaundmorris: ❓ Do we need to account for any other type of exception handling for defensive purposes and test accordingly? For example if some reason the attributeFormat is Integer but the override value causes a NumberFormatException? Although worst case here is the single attribute not overridden?

catalog/confluence/catalog-confluence-source/src/main/java/org/codice/ddf/confluence/source/ConfluenceSource.java
```diff
@@ -351,15 +365,74 @@ public void setBodyExpansion(String bodyExpansion) {
...
+            break;
+          case BINARY:
+            newValue.add(override.getBytes(Charset.forName("UTF-8")));
+            break;
+          case OBJECT:
+          case STRING:
+          case GEOMETRY:
+          case XML:
+            newValue.add(override);
+            break;
```
bellcc: ❓ Should we log whenever an unsupported type is encountered?

clockard: I think that could only happen if the enum changes which would probably be a big ddf-3.0.0 change

catalog/confluence/catalog-confluence-source/src/main/java/org/codice/ddf/confluence/source/ConfluenceSource.java
clockard: NumberFormatException is an instance of IllegalArgumentException so that case would be covered here. This is the same logic used in two other places in the code base for attribute overrides so I hope it is ok as is.

JGatley:
Manually verified that performing a query in Intrigue after setting up a confluence source with the specified override displays 

```
datetime.start
2018-06-28T04:00:00.000+00:00
```

on generated metacards.

PR #3379, opened 2018-07-02
DDF-3943 Default timeout option in platform UI config

peterhuffer:
#### What does this PR do?
- Disallows setting platform UI configuration timeout to below 2

#### Who is reviewing it? 
@emmberk @SmithJosh @mackncheesiest 

@clockard
@rzwiefel

Unit tests.
Manually test setting timeout below 2 defaults to 15.

Even though the bean will have the correct timeout value, the actual value saved in the configuration file will be whatever the admin puts in the UI. For example, if the admin set a timeout of 1, the configuration will have a value of 1, but the platform ui configuration endpoint will return 15.

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
+++ b/distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
@@ -19,58 +19,65 @@
 |Enable System Usage Message
 |systemUsageEnabled
 |Boolean
-|Turns on a system usage message, which is shown when the Search Application is opened
+|Turns on a system usage message, which is shown when the Search Application is opened.
 |false
 |true
 
 |System Usage Message Title
 |systemUsageTitle
 |String
-|A title for the system usage Message when the application is opened
-|true
+|A title for the system usage Message when the application is opened.
 |
+|false
 
 |System Usage Message
 |systemUsageMessage
 |String
-|A system usage message to be displayed to the user each time the user opens the application
-|true
+|A system usage message to be displayed to the user each time the user opens the application.
 |
+|false
 
 |Show System Usage Message once per session
 |systemUsageOncePerSession
-|With this selected, the system usage message will be shown once for each browser session. Uncheck this to have the usage message appear every time the search window is opened or refreshed.
 |Boolean
+|With this selected, the system usage message will be shown once for each browser session. Uncheck this to have the usage message appear every time the search window is opened or refreshed.
 |true
 |true
 
 |Header
 |header
-|Specifies the header text to be rendered on all pages.
 |String
-|true
+|Specifies the header text to be rendered on all pages.
 |
+|false
 
 |Footer
 |footer
-|Specifies the footer text to be rendered on all pages.
 |String
-|true
+|Specifies the footer text to be rendered on all pages.
 |
+|false
 
 |Text Color
 |color
-|Specifies the Text Color of the Header and Footer. Use html css colors or `#rrggbb`.
 |String
-|true
+|Specifies the Text Color of the Header and Footer. Use html css colors or `#rrggbb`.
 |
+|false
 
 |Background Color
 |background
-|Specifies the Background Color of the Header and Footer. Use html css colors or `#rrggbb`.
 |String
-|true
+|Specifies the Background Color of the Header and Footer. Use html css colors or `#rrggbb`.
 |
+|false
+
+|Session Timeout
+|timeout
+|Integer
+|Specifies the length of inactivity (in minutes) that will cause a user to be logged out automatically. This value must be 2 minutes or greater, as users are warned when only 1 minute remains. If a value of less than 2 minutes is used, the timeout is set to the default time of 15 minutes.
+|15
+|true
 
 |===
 
--- a/platform/platform-configuration/src/main/java/org/codice/ddf/configuration/PlatformUiConfiguration.java
+++ b/platform/platform-configuration/src/main/java/org/codice/ddf/configuration/PlatformUiConfiguration.java
@@ -20,6 +20,8 @@ import javax.ws.rs.Produces;
 import net.minidev.json.JSONObject;
 import org.codice.ddf.branding.BrandingPlugin;
 import org.codice.ddf.branding.BrandingRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Configuration class for pid=ddf.platform.ui.config.
@@ -28,6 +30,8 @@ import org.codice.ddf.branding.BrandingRegistry;
  */
 @Path("/")
 public class PlatformUiConfiguration {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(PlatformUiConfiguration.class);
 
   public static final String SYSTEM_USAGE_TITLE_CONFIG_KEY = "systemUsageTitle";
 
@@ -55,6 +59,10 @@ public class PlatformUiConfiguration {
 
   public static final String TIMEOUT_CONFIG_KEY = "timeout";
 
+  private static final int DEFAULT_TIMEOUT_MINUTES = 15;
+
+  private static final int MINIMUM_TIMEOUT_MINUTES = 2;
+
   private boolean systemUsageEnabled;
 
   private String systemUsageTitle;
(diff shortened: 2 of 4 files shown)


platform/platform-configuration/src/main/resources/OSGI-INF/metatype/metatype.xml
```diff
@@ -64,7 +64,7 @@
                 default=""
         />
         <AD
-                description="Specifies the length of inactivity (in minutes) that will cause a user to be logged out automatically. This value should be 2 minutes or greater, as users are warned when only 1 minute remains."
+                description="Specifies the length of inactivity (in minutes) that will cause a user to be logged out automatically. This value must be 2 minutes or greater, as users are warned when only 1 minute remains. If a value of less than 2 minutes is used, the timeout is set to the default time of 15 minutes."
```
vinamartin: ✏️ This is an interesting change as I think some users are still going to get into a weird state if they don't read this.
I'd think we'd want to validate and warn users from the Admin UI, but that might be beyond the scope of this change.

peterhuffer: Current Admin UI can't report configuration validation very well ATM. Currently if an Admin sets it to below 2 minutes there will be issues with other timeouts. The other option to defaulting was to throw an IllegalArgumentException. In that case the configuration would not be saved at all and the timeout will be the previous value, which I think is also weird behavior.

vinamartin: I agree, it's kind of awkward. 
I've approved as I think this works for now, but maybe in the future of Admin UI/configuration this could get improved 👍

distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
```diff
@@ -27,50 +27,57 @@
 |systemUsageTitle
 |String
 |A title for the system usage Message when the application is opened
-|true
 |
+|false
 
 |System Usage Message
```
peterhuffer: All these were jumbled around with the column headers so just re-arranged them.

platform/platform-configuration/src/main/resources/OSGI-INF/metatype/metatype.xml
clockard: With the current limitations of the admin ui configurations I think you would have to use the ConfigurationAdmin to look up your own configuration and then save it with the default value in order to keep the ui and this class instant in sync.

distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
```diff
@@ -27,50 +27,57 @@
...
+|With this selected, the system usage message will be shown once for each browser session. Uncheck this to have the usage message appear every time the search window is opened or refreshed.
 |true
 |true
 
 |Header
 |header
-|Specifies the header text to be rendered on all pages.
 |String
-|true
+|Specifies the header text to be rendered on all pages.
```
emmberk: ✏️ This column could be updated to consistently always end with periods or not end with periods.

platform/platform-configuration/src/main/resources/OSGI-INF/metatype/metatype.xml
clockard: FYI I think that would look something like 
```Bundle bundle = FrameworkUtil.getBundle(PlatformUiConfiguration.class);
    if (bundle != null) {
      BundleContext bundleContext = bundle.getBundleContext();
      if (bundleContext != null) {
        ServiceReference<ConfigurationAdmin> serviceReference =
                bundleContext.getServiceReference(ConfigurationAdmin.class);
        if (serviceReference != null) {
          ConfigurationAdmin configAdmin = bundleContext.getService(serviceReference);
          Configuration config = configAdmin.getConfiguration("ddf.platform.ui.config");
          config.getProperties().put("timeout",DEFEAULT_TIMEOUT_MINUTES);
          config.update();
        }
      }
    }

File: platform/platform-configuration/src/main/java/org/codice/ddf/configuration/PlatformUiConfiguration.java [hunk 4]
```diff
@@ -55,6 +55,8 @@
 
   public static final String TIMEOUT_CONFIG_KEY = "timeout";
 
+  private static final int DEFEAULT_TIMEOUT_MINUTES = 15;
```
mackncheesiest: `DEFAULT_TIMEOUT_MINUTES`

platform/platform-configuration/src/main/resources/OSGI-INF/metatype/metatype.xml
paouelle: You probably need to `bundleContext.ungetService(serviceReference)` when done too

peterhuffer:
build now

platform/platform-configuration/src/main/java/org/codice/ddf/configuration/PlatformUiConfiguration.java
```diff
@@ -182,7 +184,7 @@ public int getTimeout() {
   }
 
   public void setTimeout(int timeout) {
-    this.timeout = timeout;
+    this.timeout = timeout >= 2 ? timeout : DEFAULT_TIMEOUT_MINUTES;
```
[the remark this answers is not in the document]
peterhuffer: 👍 Added a warning.

peterhuffer:
build now

platform/platform-configuration/src/main/java/org/codice/ddf/configuration/PlatformUiConfiguration.java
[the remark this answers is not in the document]
shaundmorris: Great! Thanks, I think that will help clarify if there is confusion on setting a timeout less than 2.

distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
```diff
@@ -19,58 +19,65 @@
...
 |String
-|true
+|Specifies the Background Color of the Header and Footer. Use html css colors or `#rrggbb`.
 |
+|false
+
+|Session Timeout
+|timeout
+|Integer
+|Specifies the length of inactivity (in minutes) that will cause a user to be logged out automatically. This value must be 2 minutes or greater, as users are warned when only 1 minute remains. If a value of less than 2 minutes is used, the timeout is set to the default time of 15 minutes.
```
rzwiefel: ❓ They try setting it lower than 2 minutes so we instead set it to 15 minutes instead of 2 minutes?

peterhuffer: I chose the original default of the value. I could be convinced otherwise.

JGatley:
Manually verified that values below 2 default to a 15 minute timeout. Values greater than or equal to 2 set the correct timeout. In some rare cases, one logout is required for the new value to apply.

distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
shaundmorris: That's a fair question @rzwiefel. @peterhuffer did we decide to leave it at 15?

peterhuffer: I'm going to leave it at 15 since it was and is the default value, and that setting the timeout to below 2 is technically a misconfiguration. No matter what we set it to though the Admin UI will not display the correct value.

peterhuffer: I'd like to investigate using the Felix metatype implementation so we can use min and max properties on the metatype attribute descriptors, and that will actually allow us to display an error in this case.

peterhuffer:
@JGatley Currently, after saving the configuration in the Admin UI, a page refresh is required for the new value to kick in. I suspect that's why the value did not immediately apply for you.

distribution/docs/src/main/resources/content/_tables/platform.ui.config.adoc
shaundmorris: @rzwiefel since it was a question and not a ❗️or request for change went ahead. I think it is still something we can change if we need to do so.

PR #3381, opened 2018-07-02
DDF-3945 Added permissions for deleting default users and to executing jaas commands

blen-desta:
#### What does this PR do?
Added permissions for deleting default users and to executing jaas commands

#### Who is reviewing it? 
@vinamartin @brjeter 

@coyotesqrl 
@stustison 

1. Make sure the default users are deleted by the scheduler: ask me if you have questions on how to get that done
2. Make sure jaas commands work:
- Encrypting password
  - In the `etc/org.apache.karaf.jaas.cfg` file, change `encryption.enabled` to true
  - Make sure passwords in the `users.properties` file are encrypted
- Adding users:
  - `jaas:realm-manage --index 1`
  - `jaas:user-add foo bar`
  - `jaas:update`
  - `jaas:realm-manage --index 1`
  - `jaas:user-list` - see foo as a user
- Deleting users:
  - `jaas:realm-manage --index 1`
  - `jaas:user-delete foo`
  - `jaas:update`
  - `jaas:realm-manage --index 1`
  - `jaas:user-list` - foo is deleted as a user

[DDF-3945](https://codice.atlassian.net/browse/DDF-3945)

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources/security/default.policy
+++ b/distribution/ddf-common/src/main/resources/security/default.policy
@@ -173,8 +173,12 @@ grant codeBase "file:/org.apache.karaf.deployer.blueprint/org.apache.karaf.deplo
 }
 
 grant codeBase "file:/org.apache.karaf.jaas.modules" {
-    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read,write";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}keys.properties", "read";
+}
+
+grant codeBase "file:/org.apache.karaf.jaas.command" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read,write";
 }
 
 grant codeBase "file:/pax-logging-log4j2" {
@@ -254,6 +258,10 @@ grant codeBase "file:/admin-core-insecuredefaults/org.apache.felix.scr/org.apach
     permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}org.ops4j.pax.web.cfg", "read";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}artemis.xml", "read";
+}
+
+grant codeBase "file:/admin-core-insecuredefaults/org.apache.karaf.decanter.scheduler.simple/org.apache.felix.scr" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "write";
 }
 
 grant codeBase "file:/admin-core-impl/catalog-admin-poller-service/admin-core-jolokia" {


distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -391,6 +391,14 @@ grant codeBase "file:/security-sts-attributequeryclaimshandler/security-sts-ldap
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}attributeMap.properties", "read";
 }
 
+grant codeBase "file:/org.apache.karaf.jaas.command" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+}
+
+grant codeBase "file:/org.apache.karaf.jaas.modules" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "write";
+}
```
brjeter: ❗️ There's already a grant block for this module that you should just add the write permission to. https://github.com/blen-desta/ddf/blob/5b384c1f7e9a39f00e4b38bb7c5b543f54585e46/distribution/ddf-common/src/main/resources/security/default.policy#L185

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -391,6 +391,14 @@ grant codeBase "file:/security-sts-attributequeryclaimshandler/security-sts-ldap
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}attributeMap.properties", "read";
 }
 
+grant codeBase "file:/org.apache.karaf.jaas.command" {
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+}
```
brjeter: ✏️ We could tack this module onto another similar grant block. The only additional permission it would get is read access to users.attributes, which I don't think is an issue.  @stustison what do you think?

https://github.com/blen-desta/ddf/blob/5b384c1f7e9a39f00e4b38bb7c5b543f54585e46/distribution/ddf-common/src/main/resources/security/default.policy#L229-L232

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -261,7 +265,7 @@ grant codeBase "file:/admin-core-insecuredefaults/org.apache.felix.scr/org.apach
     permission java.io.FilePermission "${ddf.home.perm}etc${/}keystores${/}-", "read";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}issuer${/}-", "read";
-    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read,write";
```
coyotesqrl: ❓ Logging needs permission to _write_ the file?

blen-desta: The default user deletion scheduler deletes the users from the file so it needs write access. That's under `admin-core-insecuredefaults` not sure if that's your question

coyotesqrl: Aw, nuts. Ignore me. I missed the collapsed code block in the diff so when I first looked at this, it looked like it was under `grant codeBase "file:/pax-logging-log4j2" {`. 

Obviously, this change is completely correct.

paouelle: ❓ Do all other bundles need that write permission though?

brjeter: No, they don't. But this is the balance we're trying to strike of minimizing the number of separate grant blocks we have because it affects performance. So if we can use some engineering judgement and decide it's okay to grant the permission to a couple of bundles that don't need it, that saves us making a brand new grant block.

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -182,10 +182,14 @@ grant codeBase "file:/org.apache.karaf.deployer.blueprint/org.apache.karaf.deplo
 }
 
 grant codeBase "file:/org.apache.karaf.jaas.modules" {
-    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read";
+    permission java.io.FilePermission "${ddf.home.perm}etc${/}users.properties", "read,write";
```
paouelle: ❓ Are those required because they are on the stack? If so, could the `admin-core-insecuredefaults` simply extends its privileges when it is performing the deletion?

brjeter: In DDF, we're still operating under the status of adding permissions to bundles that are on the stack. When that changes, this could be revisited.

paouelle: Not necessarily. If your bundle is performing something that requires permissions then in some case this might simply require giving your bundle the permissions and extending your privileges. We shouldn't bleed implementation if we don't have to.

distribution/ddf-common/src/main/resources/security/default.policy
paouelle: ouch! How are we ever going to simplify this file if we keep adding permissions to code that doesn't need it. We shall never remember why a permission was added. This is going to be un-maintainable.

distribution/ddf-common/src/main/resources/security/default.policy
brjeter: Actually, when I look closer, this most likely is directly required for the second half of this PR which is making the jaas commands executable.

blen-desta: Those permissions are added to get the jaas commands working

distribution/ddf-common/src/main/resources/security/default.policy
stustison: For something like writing to the users.properties file, which can grant any access to the entire machine, I'm leaning towards us creating a new block just for things that need write to that file. In general, yes we should try to keep the number of blocks minimized, but we need to balance that with how dangerous a permission might be. Allowing someone to create a classloader, or something like that, wouldn't be nearly as dangerous as this. Plus we already have a block above for the jaas command, just add the insecure defaults bundle to that one and remove write from this block. Bundles can be in 1 - n blocks.

blen-desta:
build now

blen-desta:
`[ERROR] Failed to execute goal on project security-encryption-api: Could not resolve dependencies for project ddf.security.encryption:security-encryption-api:bundle:2.14.0-SNAPSHOT: Could not transfer artifact org.bouncycastle:bcprov-jdk15on:jar:1.54 from/to office (http://nexus.phx.connexta.com:8081/nexus/content/groups/public/): GET request of: org/bouncycastle/bcprov-jdk15on/1.54/bcprov-jdk15on-1.54.jar from office failed: Connection reset -> [Help 1]`

build now

blen-desta:
build now

blen-desta:
build now

blen-desta:
build now

PR #3404, opened 2018-07-09
DDF-3961 Adds developer area to Intrigue

andrewkfiedler:
#### What does this PR do?
 - Adds a developer area to Intrigue.  The code for this area is only included in the bundle if running in developer mode, so it won't be shipped to production or affect production.  The new area can be reached at the #_dev route.
 - Adds initial documentation for buttons, cards, dropdowns, and inputs to the developer area.  These are a WIP but a good starting point.
 - Adds a new behavior to replace the old way of doing dropdowns and deprecates the old way.  The dropdown behavior allows a much easier way to build new dropdowns, and is resilient to rerenders since it is selector based rather than element based.  For the time being I've converted only a few existing dropdowns to this behavior to make sure it works as intended, but eventually all of them should be swapped over.  Until then, I've pulled the common logic for the new and old dropdown into dropdown.behavior.utility.js so the functionality can remain in sync.
 - Adds new dependencies on brace, js-beautify, and react-ace.  These are only used in the dev guide so they aren't included in the bundle that's shipped to production.
 - Updates button behavior to use keydown rather than keyup.  Keydown emulates actual button behavior more closely, and avoids issues where the enter key closes a dropdown and then releasing it reopens it.
 - Updates query-item to use the new dropdown behavior.  This allowed a lot of code deletion with functional parity.
 - Updates dropdown.companion to use the dropdown.behavior.utility so it can remain in sync with the dropdown.behavior during it's deprecation.
 - Removes unused variables in vars.less (@links-visited-color and @links-color).
 - Updates the .is-button mixin to properly handle disabled buttons.
 - Updates the .button-mixin to properly handle disabled buttons.  This was causing buttons to not react to hover / focus.
 - Adds routes-dev which only returns the dev route to be bundled if the mode is development.
 - Updates ApplicationSetup to switch the included css sheet depending on whether in development mode or not.
 - Adds a styles.less for the dev mode which simply uses the production styles.less plus the less for it's components.
 - Updates how we handle the space key in dropdowns and the navigation behavior so it's more in line with how browsers handle it.  Basically, enter key works on keydown, but space key works on keyup.

#### Who is reviewing it? 
@djblue 
@mackncheesiest 
@Bdthomson 
@blen-desta 
@rzwiefel 
@tbatie 

Boot up the dev server and verify that you can reach the new route through the navigator (a developer link should appear).  Or go directly to #_dev.  The guide is a WIP and by no means complete, it's something we should be constantly working on / adding to / revising.

Verify that doing a production build will not include this dev area or any of it's dependencies.  

Verify that the query-item component works as intended since it's been switched to use the new dropdown behavior.  
screen shot 2018-07-09 at 3 26 59 pm

[DDF-3961](https://codice.atlassian.net/browse/DDF-3961)

screen shot 2018-07-09 at 3 30 30 pm

screen shot 2018-07-09 at 3 20 36 pm


andrewkfiedler:
build now

ui/packages/catalog-ui-search/src/main/webapp/dev/component/dropdown-guide/dropdown-guide.hbs
```diff
@@ -0,0 +1,115 @@
...
+            
+            </div>
+            <div class="editor" data-html="exampleOne"></div>
+            <div class="editor" data-html="exampleOneDropdown"></div>
+            <div class="editor" data-js="exampleOneView"></div>
+            <div class="editor" data-js="exampleOneDropdown"></div>
+        </div>
+        <div class="example">
+            <div class="title">
+                Dropdown with it's own dropdown
```
mackncheesiest: ✏️ `its own dropdown`

ui/packages/catalog-ui-search/src/main/webapp/component/list-item/list-item.view.js
```diff
@@ -48,8 +46,32 @@ module.exports = Marionette.LayoutView.extend({
     'click .list-delete': 'triggerDelete',
     'click .list-add': 'triggerAdd'
   },
-  behaviors: {
-    button: {}
+  behaviors() {
```
mackncheesiest: 🤔 ... 😧... 🆒 👍

ui/packages/catalog-ui-search/src/main/webapp/dev/component/base-guide/base-guide.view.js
```diff
@@ -0,0 +1,108 @@
...
+    value,
+    view
+}) => {
+    where.innerHTML = '<div class="is-large-font"><span class="fa fa-refresh fa-spin"></span></div>';
+    view.timeout+=1000;
+    setTimeout(() => {
+        if (view.isDestroyed) {
+            return;
+        }
+        render( <
```
mackncheesiest: ✏️ The formatting in this section is a bit funky. Can we clean it up a bit? (normalize the `{\n value \n}` vs `{ value }` vs `{value}`, etc)

mackncheesiest: Maybe like
```
render(<AceEditor
        mode = { mode }
        theme = "tomorrow_night"
        value = { value }
        ...
        width = '100%'/>, where)
```

ui/packages/catalog-ui-search/src/main/webapp/js/router/routes-dev.js
```diff
@@ -0,0 +1,10 @@
+this._compilation.compiler.options.mode === 'development' ? ({
```
mackncheesiest: ❓ Why not check `process.env.NODE_ENV !== 'production'` like what we do in [ApplicationSetup.js](https://github.com/codice/ddf/blob/4eebf3ce02dc7cc92650e0e71141b971f02488e3/ui/packages/catalog-ui-search/src/main/webapp/js/ApplicationSetup.js#L23-L28)

ui/packages/catalog-ui-search/src/main/webapp/dev/component/base-guide/base-guide.view.js
djblue: I wonder if we should consider using [`reactToMarionette`](https://github.com/codice/ddf/blob/master/ui/packages/catalog-ui-search/src/main/webapp/component/transmute/index.js#L36) here.

ui/packages/catalog-ui-search/src/main/webapp/behaviors/dropdown.behavior.js
```diff
@@ -0,0 +1,255 @@
...
+        }, {}); 
+    },
+    // We can't do this on initialize because view.cid isn't set yet.  OnFirstRender is closest to what we want.
+    onFirstRender() {
+        this.listenForOutsideInteraction();
+        this.listenForResize();
+    },
+    /*
+        buttons don't take action until keyup for space in browsers, try it for yourself
+        https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_button_test
```
austinsteffes: ❓Is this beneficial to have this in with the link?

ui/packages/catalog-ui-search/src/main/webapp/behaviors/dropdown.behavior.js
```diff
@@ -0,0 +1,255 @@
...
+            switch (code) {
+                case 32:
+                    clickHandler(e);
+                    break;  
+            }
+        };
+    },
+    /*
+        buttons take action on keydown for enter in browsers, try it for yourself
+        https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_button_test
```
austinsteffes: ❓same as above with line 40

ui/packages/catalog-ui-search/src/main/webapp/behaviors/dropdown.behavior.js
```diff
@@ -0,0 +1,255 @@
...
+        if (dropdown._region) {
+            dropdown._region.empty();
+            dropdown._region.destroy();
+            dropdown._region.$el.remove();
+        }
+    },  
+    onDestroy(){
+        this.stopListeningForOutsideInteraction();
+        this.stopListeningForResize();
+        //ensure that if a dropdown goes away, it's companions do too
```
austinsteffes: ✏️"its"

ui/packages/catalog-ui-search/src/main/webapp/behaviors/dropdown.behavior.js
mackncheesiest: I like it 🤷‍♂️

ui/packages/catalog-ui-search/src/main/webapp/js/router/routes-dev.js
andrewkfiedler: This happens during a loader and it doesn't have that available unfortunately.

andrewkfiedler:
build now

andrewkfiedler:
hero performed by @djblue

PR #3411, opened 2018-07-10
DDF-3966 Updates cesium map to handle terrain provider issues

andrewkfiedler:
#### What does this PR do?
 - Previously if cesium hit an issue with a terrain provider (bad url, offline, etc.) it would not fail gracefully (fail to load imagery providers, etc.).  Now it will fallback to the default cesium terrain provider.

#### Who is reviewing it? 
@mcalcote 
@rzwiefel 
@djblue 
@ethantmanns 
@clockard 

Configure a bad terrain provider (either with a bad url or a bad type, or without a type) and verify that the UI loads correctly.  There should be warnings in the console that indicate what has gone wrong.

[DDF-3966](https://codice.atlassian.net/browse/DDF-3966)


djblue:
build now

PR #3412, opened 2018-07-10
DDF-3893 The IdP should account for the NameIDPolicy in AuthnRequests

bakejeyner:
#### What does this PR do?
Adds support for `NameIDPolicy`s on the `AuthnRequest` in DDF's IdP.

#### Who is reviewing it? 
@blen-desta
@brjeter

@coyotesqrl
@stustison

1. Build and install DDF
2. Run [this branch of the SAML CTK](https://github.com/bakejeyner/saml-conformance/tree/DDF-3893), and verify that there are no errors.

[DDF-3893](https://codice.atlassian.net/browse/DDF-3893)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>ddf.security.idp</groupId>
+    <artifactId>security-idp-plugin</artifactId>
+    <version>2.13.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>security-idp-plugin-nameidpolicy</artifactId>
+  <name>DDF :: Security :: IdP :: Plugin :: Name ID Policy</name>
+  <packaging>bundle</packaging>
+
+  <dependencies>
+    <dependency>
+      <groupId>ddf.security.idp</groupId>
+      <artifactId>security-idp-plugin-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!--Spock dependencies-->
+    <dependency>
+      <groupId>ddf.lib</groupId>
+      <artifactId>spock-shaded</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+            <Export-Service>org.codice.ddf.security.idp.plugin.SamlPresignPlugin</Export-Service>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-check</id>
+            <goals>
+              <goal>check</goal>
+            </goals>
+            <configuration>
+              <haltOnFailure>true</haltOnFailure>
+              <rules>
+                <rule>
+                  <element>BUNDLE</element>
+                  <limits>
+                    <limit implementation="org.codice.jacoco.LenientLimit">
+                      <counter>INSTRUCTION</counter>
+                      <value>COVEREDRATIO</value>
+                      <minimum>1.00</minimum>
+                    </limit>
+                    <limit implementation="org.codice.jacoco.LenientLimit">
+                      <counter>BRANCH</counter>
+                      <value>COVEREDRATIO</value>
+                      <minimum>0.95</minimum>
+                    </limit>
+                    <limit implementation="org.codice.jacoco.LenientLimit">
+                      <counter>COMPLEXITY</counter>
+                      <value>COVEREDRATIO</value>
+                      <minimum>0.96</minimum>
+                    </limit>
+                  </limits>
+                </rule>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
@@ -50,7 +50,6 @@
         <configuration>
           <instructions>
             <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
-            <Export-Service>org.codice.ddf.security.idp.plugin.SamlPresignPlugin</Export-Service>
           </instructions>
         </configuration>
       </plugin>
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>ddf.security.idp</groupId>
     <artifactId>security-idp-plugin</artifactId>
-    <version>2.13.0-SNAPSHOT</version>
+    <version>2.14.0-SNAPSHOT</version>
   </parent>
   <artifactId>security-idp-plugin-nameidpolicy</artifactId>
   <name>DDF :: Security :: IdP :: Plugin :: Name ID Policy</name>
@@ -39,6 +39,12 @@
       <artifactId>spock-shaded</artifactId>
       <version>${project.version}</version>
       <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ddf.catalog.core</groupId>
+      <artifactId>catalog-core-api</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
     </dependency>
   </dependencies>
 
@@ -71,17 +77,17 @@
                     <limit implementation="org.codice.jacoco.LenientLimit">
                       <counter>INSTRUCTION</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>1.00</minimum>
+                      <minimum>0.92</minimum>
                     </limit>
                     <limit implementation="org.codice.jacoco.LenientLimit">
                       <counter>BRANCH</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>0.95</minimum>
+                      <minimum>0.90</minimum>
                     </limit>
                     <limit implementation="org.codice.jacoco.LenientLimit">
                       <counter>COMPLEXITY</counter>
                       <value>COVEREDRATIO</value>
-                      <minimum>0.96</minimum>
+                      <minimum>0.92</minimum>
                     </limit>
                   </limits>
                 </rule>
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
@@ -48,7 +48,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
 
     if (nameIdPolicy == null
         || (StringUtils.isEmpty(nameIdPolicy.getFormat())
-            && StringUtils.isEmpty(nameIdPolicy.getSPNameQualifier()))) {
+        && StringUtils.isEmpty(nameIdPolicy.getSPNameQualifier()))) {
       return;
     }
 
@@ -77,14 +77,14 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
 
     if (StringUtils.isNotEmpty(nameIdFormatPolicy)) {
       switch (nameIdFormatPolicy) {
-          // supported NameIDFormats
+        // supported NameIDFormats
         case NameID.UNSPECIFIED:
           return; // avoid changing the Format later
         case NameID.PERSISTENT:
           // TODO DDF-3965
           break;
 
-          // partially supported NameIDFormats
+        // partially supported NameIDFormats
         case NameID.X509_SUBJECT:
           LOGGER.warn("Specifying the \"X509Subject\" NameIDPolicy Format is not fully supported.");
           if (!NameID.X509_SUBJECT.equals(assertionNameId.getFormat())) {
@@ -97,7 +97,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
           assertionNameId.setValue(resolveEmail(response));
           break;
 
-          // not supported NameIDFormats
+        // not supported NameIDFormats
         case NameID.TRANSIENT:
         case NameID.WIN_DOMAIN_QUALIFIED:
         case NameID.KERBEROS:
@@ -111,6 +111,13 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
     }
   }
 
+  /**
+   * Attempts to find an email identifier in a given Response. More specifically, this method looks
+   * in a Response's AttributeStatements for the email identifier. Throws an error if not found.
+   *
+   * @param response Response object
+   * @return Email identifier
+   */
   private String resolveEmail(Response response) {
     return response
         .getAssertions()
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
@@ -48,7 +48,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
 
     if (nameIdPolicy == null
         || (StringUtils.isEmpty(nameIdPolicy.getFormat())
-        && StringUtils.isEmpty(nameIdPolicy.getSPNameQualifier()))) {
+            && StringUtils.isEmpty(nameIdPolicy.getSPNameQualifier()))) {
       return;
     }
 
@@ -77,9 +77,9 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
 
     if (StringUtils.isNotEmpty(nameIdFormatPolicy)) {
       switch (nameIdFormatPolicy) {
-        // supported NameIDFormats
+          // supported NameIDFormats
         case NameID.UNSPECIFIED:
-          break;
+          return; // avoid changing the Format later
         case NameID.PERSISTENT:
           // TODO DDF-3965
           break;
@@ -87,7 +87,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
           // TODO DDF-3965
           break;
 
-        // partially supported NameIDFormats
+          // partially supported NameIDFormats
         case NameID.X509_SUBJECT:
           LOGGER.warn("Specifying the \"X509Subject\" NameIDPolicy Format is not fully supported.");
           if (!NameID.X509_SUBJECT.equals(assertionNameId.getFormat())) {
@@ -100,7 +100,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
           assertionNameId.setValue(resolveEmail(response));
           break;
 
-        // not supported NameIDFormats
+          // not supported NameIDFormats
         case NameID.WIN_DOMAIN_QUALIFIED:
         case NameID.KERBEROS:
         case NameID.ENTITY:
@@ -109,9 +109,8 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
               String.format(
                   "The NameIDPolicy Format of value %s is not supported.", nameIdFormatPolicy));
       }
+      assertionNameId.setFormat(nameIdFormatPolicy);
     }
-
-    assertionNameId.setFormat(nameIdFormatPolicy);
   }
 
   private String resolveEmail(Response response) {
@@ -120,9 +119,10 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
         .stream()
         .flatMap(assertion -> assertion.getAttributeStatements().stream())
         .flatMap(statement -> statement.getAttributes().stream())
-        .filter(attribute ->
-            EMAIL_ATTRIBUTE_NAME.equals(attribute.getName())
-                || NameID.EMAIL.equals(attribute.getNameFormat()))
+        .filter(
+            attribute ->
+                EMAIL_ATTRIBUTE_NAME.equals(attribute.getName())
+                    || NameID.EMAIL.equals(attribute.getNameFormat()))
         .flatMap(attribute -> attribute.getAttributeValues().stream())
         .filter(attributeValue -> attributeValue instanceof XSString)
         .map(attributeValue -> (XSString) attributeValue)
@@ -131,6 +131,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
         .findFirst()
         .orElseThrow(
             () ->
-                new UnsupportedOperationException("The \"Email\" NameID could not be retrieved."));
+                new UnsupportedOperationException(
+                    "The \"Email\" NameID could not be retrieved for this principal."));
   }
 }
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
@@ -83,9 +83,6 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
         case NameID.PERSISTENT:
           // TODO DDF-3965
           break;
-        case NameID.TRANSIENT:
-          // TODO DDF-3965
-          break;
 
           // partially supported NameIDFormats
         case NameID.X509_SUBJECT:
@@ -101,6 +98,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
           break;
 
           // not supported NameIDFormats
+        case NameID.TRANSIENT:
         case NameID.WIN_DOMAIN_QUALIFIED:
         case NameID.KERBEROS:
         case NameID.ENTITY:
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
@@ -13,13 +13,17 @@
  */
 package org.codice.ddf.security.idp.plugin.nameidpolicy;
 
+import ddf.security.SubjectUtils;
 import ddf.security.samlp.SamlProtocol.Binding;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import org.apache.commons.lang.StringUtils;
 import org.codice.ddf.security.idp.plugin.SamlPresignPlugin;
 import org.opensaml.core.xml.schema.XSString;
+import org.opensaml.saml.saml2.core.Attribute;
 import org.opensaml.saml.saml2.core.AuthnRequest;
 import org.opensaml.saml.saml2.core.NameID;
 import org.opensaml.saml.saml2.core.NameIDPolicy;
@@ -28,15 +32,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * This plugin is responsible for setting the NameID Format/SPNameQualifier on the outgoing Response
- * according to the values found in the NameIDPolicy of the passed in AuthnRequest.
+ * This plugin is responsible for setting the <@code>NameID</@code>'s <@code>Format</@code> &
+ * <@code>SPNameQualifier</@code> on the outgoing <@code>Response</@code>. It does this by parsing
+ * the <@code>NameIDPolicy</@code> of the passed in <@code>AuthnRequest</@code>.
+ *
+ * <p>See section 3.4.1.1 of the SAML Core specification for more information.
  */
 public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(NameIdPolicyPresignPlugin.class);
-
-  private static final String EMAIL_ATTRIBUTE_NAME =
-      "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress";
 
   @Override
   public void processPresign(
@@ -48,7 +52,7 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
 
     if (nameIdPolicy == null
         || (StringUtils.isEmpty(nameIdPolicy.getFormat())
-        && StringUtils.isEmpty(nameIdPolicy.getSPNameQualifier()))) {
+            && StringUtils.isEmpty(nameIdPolicy.getSPNameQualifier()))) {
       return;
     }
 
@@ -75,16 +79,17 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
       assertionNameId.setSPNameQualifier(spNameQualifierPolicy);
     }
 
+    // the Assertion's NameID Format is set to the NameIDPolicy Format after the switch statement
     if (StringUtils.isNotEmpty(nameIdFormatPolicy)) {
       switch (nameIdFormatPolicy) {
-        // supported NameIDFormats
+          // supported NameIDFormats
         case NameID.UNSPECIFIED:
           return; // avoid changing the Format later
         case NameID.PERSISTENT:
           // TODO DDF-3965
           break;
 
-        // partially supported NameIDFormats
+          // partially supported NameIDFormats
         case NameID.X509_SUBJECT:
           LOGGER.warn("Specifying the \"X509Subject\" NameIDPolicy Format is not fully supported.");
           if (!NameID.X509_SUBJECT.equals(assertionNameId.getFormat())) {
@@ -94,10 +99,10 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
           break;
         case NameID.EMAIL:
           LOGGER.warn("Specifying the \"Email\" NameIDPolicy Format is not fully supported.");
-          assertionNameId.setValue(resolveEmail(response));
+          assertionNameId.setValue(resolveEmail(response, assertionNameId));
           break;
 
-        // not supported NameIDFormats
+          // not supported NameIDFormats
         case NameID.TRANSIENT:
         case NameID.WIN_DOMAIN_QUALIFIED:
         case NameID.KERBEROS:
@@ -112,22 +117,23 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
   }
 
   /**
-   * Attempts to find an email identifier in a given Response. More specifically, this method looks
-   * in a Response's AttributeStatements for the email identifier. Throws an error if not found.
+   * Attempts to find an email identifier in a given <@code>Response</@code>. More specifically,
+   * this method looks in a <@code>Response</@code>'s <@code>AttributeStatements</@code> for the
+   * email identifier. Throws an error if not found.
    *
-   * @param response Response object
+   * @param response <@code>Response</@code> object
    * @return Email identifier
+   * @throws UnsupportedOperationException if an email identifier could not be found
    */
-  private String resolveEmail(Response response) {
-    return response
-        .getAssertions()
-        .stream()
-        .flatMap(assertion -> assertion.getAttributeStatements().stream())
-        .flatMap(statement -> statement.getAttributes().stream())
-        .filter(
-            attribute ->
-                EMAIL_ATTRIBUTE_NAME.equals(attribute.getName())
-                    || NameID.EMAIL.equals(attribute.getNameFormat()))
+  private String resolveEmail(Response response, NameID assertionNameId)
+      throws UnsupportedOperationException {
+    Predicate<Attribute> attributeHasEmailIdentifier =
+        attribute ->
+            SubjectUtils.EMAIL_ADDRESS_CLAIM_URI.equals(attribute.getName())
+                || NameID.EMAIL.equals(attribute.getNameFormat());
+
+    return extractAssertionAttributes(response)
+        .filter(attributeHasEmailIdentifier)
         .flatMap(attribute -> attribute.getAttributeValues().stream())
         .filter(attributeValue -> attributeValue instanceof XSString)
         .map(attributeValue -> (XSString) attributeValue)
@@ -137,6 +143,16 @@ public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
         .orElseThrow(
             () ->
                 new UnsupportedOperationException(
-                    "The \"Email\" NameID could not be retrieved for this principal."));
+                    String.format(
+                        "The \"Email\" NameID could not be retrieved for the principal %s.",
+                        assertionNameId.getValue())));
+  }
+
+  private Stream<Attribute> extractAssertionAttributes(Response response) {
+    return response
+        .getAssertions()
+        .stream()
+        .flatMap(assertion -> assertion.getAttributeStatements().stream())
+        .flatMap(statement -> statement.getAttributes().stream());
   }
 }
--- a/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
+++ b/platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
@@ -1,0 +1,136 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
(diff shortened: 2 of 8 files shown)


platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,136 @@
...
+
+    if (StringUtils.isNotEmpty(nameIdFormatPolicy)) {
+      switch (nameIdFormatPolicy) {
+        // supported NameIDFormats
+        case NameID.UNSPECIFIED:
+          break;
+        case NameID.PERSISTENT:
+          // TODO DDF-3965
+          break;
+        case NameID.TRANSIENT:
```
brjeter: ❗️ We technically don't support this in our metadata, and I don't think STS will ever give us a NameID in this format.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        thrown(UnsupportedOperationException)
+
+        where:
+        policyFormat       | format              | isError
+        NameID.UNSPECIFIED | null                | false
+        NameID.UNSPECIFIED | NameID.UNSPECIFIED  | false
+        NameID.UNSPECIFIED | NameID.PERSISTENT   | false
+        NameID.UNSPECIFIED | NameID.TRANSIENT    | false
+        NameID.UNSPECIFIED | NameID.X509_SUBJECT | false
+        NameID.UNSPECIFIED | NameID.EMAIL        | false
```
brjeter: ✏️ You don't need the `isError` column and variable.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        thrown(UnsupportedOperationException)
+
+        where:
+        policyFormat      | format              | isError
+        NameID.PERSISTENT | null                | false
+        NameID.PERSISTENT | NameID.UNSPECIFIED  | false
+        NameID.PERSISTENT | NameID.PERSISTENT   | false
+        NameID.PERSISTENT | NameID.TRANSIENT    | false
+        NameID.PERSISTENT | NameID.X509_SUBJECT | false
+        NameID.PERSISTENT | NameID.EMAIL        | false
```
brjeter: ✏️ You don't need the `isError` column and variable.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,136 @@
...
+        case NameID.KERBEROS:
+        case NameID.ENTITY:
+        default:
+          throw new UnsupportedOperationException(
+              String.format(
+                  "The NameIDPolicy Format of value %s is not supported.", nameIdFormatPolicy));
+      }
+    }
+
+    assertionNameId.setFormat(nameIdFormatPolicy);
```
brjeter: ❗️ Don't we want to fail if the format value already exists and it's not equal to the policy's format? For example, I think it's incorrect for us to have an email formatted nameID, but to then overwrite the format value to persistent.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        setup:
+        nameIdMock.setFormat(format)
+        nameIdPolicyMock.getFormat() >> policyFormat
+
+        when:
+        plugin.processPresign(responseMock, authnRequestMock, spMetadataMock, supportedBindingsMock)
+
+        then:
+        if (!isError) {
+            assert (nameIdMock.value == EXAMPLE_NAME_ID_VALUE)
```
brjeter: ❗️ I would think we'd want to also verify that the nameId's format value is still the same.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        nameIdMock.setFormat(NameID.UNSPECIFIED)
+
+        emailAttribute =
+                responseMock.assertions.first().attributeStatements.first().attributes.first()
+
+        authnRequestMock.getNameIDPolicy() >> nameIdPolicyMock
+    }
+
+    @Unroll("null NameIDPolicy Format tests for Format #format")
+    def 'null NameIDPolicy Format tests'() {
```
brjeter: ✏️ None of these test names explain what they're doing or verifying. For example, this one could be `test that the nameID is not modified when NameIDPolicy format is null`

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        thrown(UnsupportedOperationException)
+
+        where:
+        policyFormat | format              | isError
+        null         | null                | false
+        null         | NameID.UNSPECIFIED  | false
+        null         | NameID.PERSISTENT   | false
+        null         | NameID.TRANSIENT    | false
+        null         | NameID.X509_SUBJECT | false
+        null         | NameID.EMAIL        | false
```
brjeter: ✏️ I think it's overkill to run this against every format.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        where:
+        policyFormat | format              | isError
+        NameID.EMAIL | null                | false
+        NameID.EMAIL | NameID.UNSPECIFIED  | false
+        NameID.EMAIL | NameID.PERSISTENT   | false
+        NameID.EMAIL | NameID.TRANSIENT    | false
+        NameID.EMAIL | NameID.X509_SUBJECT | false
+        NameID.EMAIL | NameID.EMAIL        | false
+    }
```
brjeter: ❓ Could you just combine all of these identical tests into one with one big table? (excluding the x509 test)

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        where:
+        policyFormat                | format              | isError
+        NameID.WIN_DOMAIN_QUALIFIED | null                | true
+        NameID.WIN_DOMAIN_QUALIFIED | NameID.UNSPECIFIED  | true
+        NameID.WIN_DOMAIN_QUALIFIED | NameID.PERSISTENT   | true
+        NameID.WIN_DOMAIN_QUALIFIED | NameID.TRANSIENT    | true
+        NameID.WIN_DOMAIN_QUALIFIED | NameID.X509_SUBJECT | true
+        NameID.WIN_DOMAIN_QUALIFIED | NameID.EMAIL        | true
+    }
```
brjeter: ✏️ I don't think this test is necessary.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        where:
+        policyFormat    | format              | isError
+        NameID.KERBEROS | null                | true
+        NameID.KERBEROS | NameID.UNSPECIFIED  | true
+        NameID.KERBEROS | NameID.PERSISTENT   | true
+        NameID.KERBEROS | NameID.TRANSIENT    | true
+        NameID.KERBEROS | NameID.X509_SUBJECT | true
+        NameID.KERBEROS | NameID.EMAIL        | true
+    }
```
brjeter: ✏️ I don't think this test is necessary.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        where:
+        policyFormat  | format              | isError
+        NameID.ENTITY | null                | true
+        NameID.ENTITY | NameID.UNSPECIFIED  | true
+        NameID.ENTITY | NameID.PERSISTENT   | true
+        NameID.ENTITY | NameID.TRANSIENT    | true
+        NameID.ENTITY | NameID.X509_SUBJECT | true
+        NameID.ENTITY | NameID.EMAIL        | true
+    }
```
brjeter: ✏️ I don't think this test is necessary.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        when:
+        plugin.processPresign(responseMock, authnRequestMock, spMetadataMock, supportedBindingsMock)
+
+        then:
+        if (!isError) {
+            assert (nameIdMock.value == EXAMPLE_NAME_ID_VALUE)
+            return
+        }
+        thrown(UnsupportedOperationException)
```
brjeter: ✏️ This can be removed.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
```
brjeter: ✏️ This can be removed

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+        setup:
+        nameIdMock.setFormat(format)
+        nameIdPolicyMock.getFormat() >> policyFormat
+
+        when:
+        plugin.processPresign(responseMock, authnRequestMock, spMetadataMock, supportedBindingsMock)
+
+        then:
+        if (!isError) {
+            assert (nameIdMock.value == EXAMPLE_NAME_ID_VALUE)
```
brjeter: ❗️ I would think we'd want to also verify that the nameId's format value is equal to `format`

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        when:
+        plugin.processPresign(responseMock, authnRequestMock, spMetadataMock, supportedBindingsMock)
+
+        then:
+        if (!isError) {
+            assert (nameIdMock.value == EXAMPLE_NAME_ID_VALUE && nameIdMock.format == policyFormat)
+            return
+        }
+        thrown(UnsupportedOperationException)
```
brjeter: ✏️ This can be removed.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
```
brjeter: ✏️ This can be removed.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,417 @@
...
+
+        when:
+        plugin.processPresign(responseMock, authnRequestMock, spMetadataMock, supportedBindingsMock)
+
+        then:
+        if (!isError) {
+            assert (nameIdMock.value == EXAMPLE_EMAIL && nameIdMock.format == policyFormat)
+            return
+        }
+        thrown(UnsupportedOperationException)
```
brjeter: ✏️ This can be removed.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: Should I take out this option, and just return an error if we get the `transient` NameIDPolicy Format?

If we keep it in, it shouldn't cause any problems. It would just be a little more permissive.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,136 @@
...
+          break;
+        case NameID.EMAIL:
+          LOGGER.warn("Specifying the \"Email\" NameIDPolicy Format is not fully supported.");
+          assertionNameId.setValue(resolveEmail(response));
+          break;
+
+        // not supported NameIDFormats
+        case NameID.WIN_DOMAIN_QUALIFIED:
+        case NameID.KERBEROS:
+        case NameID.ENTITY:
```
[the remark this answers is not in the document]
bakejeyner: I see some value for possible future development, but it isn't a lot. I'll opt to keep it in this time.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: In our case, and probably many others, the Subject's NameID is only there to say "by the way, this is how I am going to refer to the user/subject/principal". Depending on how the relying SP decides to use it, it can lead to errors. Luckily for us, our SP only uses the Subject's NameID to store it as the user's "principal".

That being said, our SP parses that NameID based on the Format value, and can error out if we get a NameID whose value and format don't match up. I.e. trying to parse an email as a certificate can get messy. In this PR that is mitigated because we actually do what you said for the `X509SubjectName` NameIDPolicy.

TLDR: It isn't broken, but I can change it if that is our desired behavior.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
bakejeyner: I'll change it to run against the three we support.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
brjeter: I think we're only doing that for if `X509SubjectName` is the format on the NameIDPolicy, but what if `X509SubjectName` is the format of the nameID itself and `persistent` is the format of the policy. We will then change the format to `persistent` even though really the NameID has a `X509SubjectName` value.

bakejeyner: Yes, we would be sending an `X509SubjectName` as a  `persistent` identifier. However, DDF's SP doesn't care what String value it gets for the `persistent` identifier. It's only really picky about what String value is in the `X509SubjectName` identifier.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
bakejeyner: I could group the tests a little more. But if this plugin becomes a little more fleshed out, then the current layout is much better. It allows for custom logic to be put in for each type of `NameIDPolicy` Format. Like the `X509SubjectName` stuff.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
brjeter: I think it should be removed. I don't think this would impact the DDF SP as it is now, but I think it would be surprising and cause odd behavior for any other SP that is getting a "transient" identifier that isn't actually transient.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,135 @@
...
+        default:
+          throw new UnsupportedOperationException(
+              String.format(
+                  "The NameIDPolicy Format of value %s is not supported.", nameIdFormatPolicy));
+      }
+      assertionNameId.setFormat(nameIdFormatPolicy);
+    }
+  }
+
+  private String resolveEmail(Response response) {
```
bellcc: ❓ I find this method very difficult to read. Can we add some javadoc or can it be broken up into smaller/descriptive methods?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,135 @@
...
+    return response
+        .getAssertions()
+        .stream()
+        .flatMap(assertion -> assertion.getAttributeStatements().stream())
+        .flatMap(statement -> statement.getAttributes().stream())
+        .filter(
+            attribute ->
```
coyotesqrl: ✏️ Extracting this test to a named Predicate would make its behavior easier to see while scanning the stream processing.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,135 @@
...
+      }
+      assertionNameId.setFormat(nameIdFormatPolicy);
+    }
+  }
+
+  private String resolveEmail(Response response) {
+    return response
+        .getAssertions()
+        .stream()
+        .flatMap(assertion -> assertion.getAttributeStatements().stream())
```
coyotesqrl: ✏️ Extracting these two `flatMap`s to a method would also help clarify the behavior here. Combined with the Predicate suggestion, you could end up with something like the following:
```java
  private String resolveEmail(Response response) {
    Predicate<Attribute> attributePredicate = attribute ->
        EMAIL_ATTRIBUTE_NAME.equals(attribute.getName())
            || NameID.EMAIL.equals(attribute.getNameFormat());
    
    return extractAssertionAttributes(response.getAssertions().stream())
        .filter(attributePredicate)
        .flatMap(attribute -> attribute.getAttributeValues().stream())
        .filter(attributeValue -> attributeValue instanceof XSString)
        .map(attributeValue -> (XSString) attributeValue)
        .map(XSString::getValue)
        .filter(StringUtils::isNotBlank)
        .findFirst()
        .orElseThrow(
            () ->
                new UnsupportedOperationException(
                    "The \"Email\" NameID could not be retrieved for this principal."));
  }

  private Stream<Attribute> extractAssertionAttributes(Stream<Assertion> stream) {
    return stream
        .flatMap(assertion -> assertion.getAttributeStatements().stream())
        .flatMap(statement -> statement.getAttributes().stream());
  }
```

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: Unfortunately, the email identifier is in a weird part of the `Response`'s data structure, so finding the email identifier is almost always going to feel like spaghetti. Added some javadoc though!

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,135 @@
...
+                "The \"X509Subject\" NameID could not be retrieved.");
+          }
+          break;
+        case NameID.EMAIL:
+          LOGGER.warn("Specifying the \"Email\" NameIDPolicy Format is not fully supported.");
+          assertionNameId.setValue(resolveEmail(response));
+          break;
+
+          // not supported NameIDFormats
+        case NameID.TRANSIENT:
```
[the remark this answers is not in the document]
bakejeyner: I just added them there for possible future development. Gives the layout for where to put logic when needed.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  private void setAssertionNameIdQualifierAndFormat(
+      NameID assertionNameId,
+      @Nullable String nameIdFormatPolicy,
+      @Nullable String spNameQualifierPolicy,
+      Response response) {
+
+    if (StringUtils.isNotEmpty(spNameQualifierPolicy)) {
+      assertionNameId.setSPNameQualifier(spNameQualifierPolicy);
```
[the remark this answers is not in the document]
bakejeyner: According to SAML Core, the `SPNameQualifier`:
```
Further qualifies a name with the name of a service provider or affiliation of providers. This attribute provides an additional means to federate names on the basis of the relying party or parties.
```

It is basically just a string that identifies an SP.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>ddf.security.idp</groupId>
+    <artifactId>security-idp-plugin</artifactId>
+    <version>2.13.0-SNAPSHOT</version>
```
coyotesqrl: ❗️ For master, this version is now `2.14.0-SNAPSHOT`.

You'll also need to backport this to the `2.13.x` branch.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+import org.opensaml.core.xml.schema.XSString;
+import org.opensaml.saml.saml2.core.AuthnRequest;
+import org.opensaml.saml.saml2.core.NameID;
+import org.opensaml.saml.saml2.core.NameIDPolicy;
+import org.opensaml.saml.saml2.core.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This plugin is responsible for setting the NameID Format/SPNameQualifier on the outgoing Response
```
coyotesqrl: ✏️ This javadoc doesn't really tell me anything. 
- What are the values it sets?
- What is the justification (specwise) for setting these values?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+ * according to the values found in the NameIDPolicy of the passed in AuthnRequest.
+ */
+public class NameIdPolicyPresignPlugin implements SamlPresignPlugin {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(NameIdPolicyPresignPlugin.class);
+
+  private static final String EMAIL_ATTRIBUTE_NAME =
```
coyotesqrl: ✏️ _Might_ make sense to replace this with [`SubjectUtils.EMAIL_ADDRESS_CLAIM_URI` ](https://github.com/codice/ddf/blob/master/platform/security/platform-security-core-api/src/main/java/ddf/security/SubjectUtils.java#L53-L54).

I think it's reasonable to refer to `SubjectUtils` from here. Maybe.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+        case NameID.UNSPECIFIED:
+          return; // avoid changing the Format later
+        case NameID.PERSISTENT:
+          // TODO DDF-3965
+          break;
+
+        // partially supported NameIDFormats
+        case NameID.X509_SUBJECT:
+          LOGGER.warn("Specifying the \"X509Subject\" NameIDPolicy Format is not fully supported.");
+          if (!NameID.X509_SUBJECT.equals(assertionNameId.getFormat())) {
```
coyotesqrl: ❓ You attempt to retrieve but do nothing with it?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
coyotesqrl: I'd still like to see this changed to be more compact and readable without level-switching while scanning the code.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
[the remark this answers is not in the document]
blen-desta: I'm still not sure why you are pulling the SPNameQualifier from the AuthnRequest's NameIdPolicy and setting it on the `assertionNameId`.

And if the NameID is persistent it has to be the SP's entityID:
```
8.3.7 Persistent Identifier

The element's SPNameQualifier attribute, if present, MUST contain the unique identifier of the 
service provider or affiliation of providers for whom the identifier was generated.
```

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: I'll see what I can do!

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: This is by design. If the SP forces an `X509Subject` `NameID` `Format`, then we only allow certificate login. And when the user logs in with their certificate, the only thing left to do is make sure the outgoing `NameID` `Format` is `X509Subject` (which it already is), and set the `NameID` value to the certificate DN (which it already is).

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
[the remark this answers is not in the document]
bakejeyner: From section `3.4.1.1 Element <NameIDPolicy>` of the SAML Core specification:
```
if SPNameQualifier is not omitted in <NameIDPolicy>, the SPNameQualifier value of the <NameID> within 
the <Subject> of any <Assertion> MUST be identical to the SPNameQualifier value supplied in the 
<NameIDPolicy>.
```

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
coyotesqrl: Slightly better suggestion...change that `extractAssertionAttributes` method to take the `Response` instead.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+    <dependency>
+      <groupId>ddf.lib</groupId>
+      <artifactId>spock-shaded</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ddf.catalog.core</groupId>
+      <artifactId>catalog-core-api</artifactId>
+      <version>2.14.0-SNAPSHOT</version>
```
coyotesqrl: ❗️ This should be `${project.version}`.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,154 @@
...
+
+  @Override
+  public void processPresign(
+      Response response,
+      AuthnRequest authnRequest,
+      List<String> spMetadata,
+      Set<Binding> supportedBindings) {
+    NameIDPolicy nameIdPolicy = authnRequest.getNameIDPolicy();
+
+    if (nameIdPolicy == null
```
brendan-hofmann: ❓ Why do this check separately instead of just filtering the stream below?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,154 @@
...
+          // supported NameIDFormats
+        case NameID.UNSPECIFIED:
+          return; // avoid changing the Format later
+        case NameID.PERSISTENT:
+          // TODO DDF-3965
+          break;
+
+          // partially supported NameIDFormats
+        case NameID.X509_SUBJECT:
+          LOGGER.warn("Specifying the \"X509Subject\" NameIDPolicy Format is not fully supported.");
```
brendan-hofmann: ❓ Should we log anything to the security log here? Same for the others.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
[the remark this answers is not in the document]
brendan-hofmann: ❓ Should we log anything for these cases?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,154 @@
...
+                  "The NameIDPolicy Format of value %s is not supported.", nameIdFormatPolicy));
+      }
+      assertionNameId.setFormat(nameIdFormatPolicy);
+    }
+  }
+
+  /**
+   * Attempts to find an email identifier in a given <@code>Response</@code>. More specifically,
+   * this method looks in a <@code>Response</@code>'s <@code>AttributeStatements</@code> for the
+   * email identifier. Throws an error if not found.
```
brendan-hofmann: ✏️ This could be an `@throws` to make the signature more clear.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,154 @@
...
+        .filter(attributeHasEmailIdentifier)
+        .flatMap(attribute -> attribute.getAttributeValues().stream())
+        .filter(attributeValue -> attributeValue instanceof XSString)
+        .map(attributeValue -> (XSString) attributeValue)
+        .map(XSString::getValue)
+        .filter(StringUtils::isNotBlank)
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new UnsupportedOperationException(
```
brendan-hofmann: ❓ Should we identify the principal somehow so this can be solved easier?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
```diff
@@ -0,0 +1,276 @@
...
+import org.opensaml.saml.saml2.core.Attribute
+import org.opensaml.saml.saml2.core.AuthnRequest
+import org.opensaml.saml.saml2.core.NameID
+import org.opensaml.saml.saml2.core.NameIDPolicy
+import org.opensaml.saml.saml2.core.Response
+import org.w3c.dom.Document
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class NameIdPolicyPresignPluginSpec extends Specification {
```
brendan-hofmann: ❓ Should we have test coverage for the currently unsupported formats you added cases for in the implementation above?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: It's a nice little short circuit for when we aren't given a `NameIDPolicy` in the `AuthnRequest`.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: I don't think I understand what you are asking. We already have all of the information about the principal, we just need to see if we have an email identifier in that information.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/test/groovy/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPluginSpec.groovy
bakejeyner: I originally did, but they seemed unnecessary and I actually took them out during this review process.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
bakejeyner: I'm not sure what our security logging practices are. This may be a good spot for it.

@blen-desta , @brjeter , @coyotesqrl ?

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
brendan-hofmann: I am suggesting we identify the principal in the exception/log message so that an admin can diagnose the problem easier.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
coyotesqrl: I don't _think_ it's important enough to audit, as it's not a 👍/ 👎 decision and no rights are granted or denied because of this.

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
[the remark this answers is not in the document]
bakejeyner: It is already being logged at the end of the case statements, the `default` case. Should I modify the logging at all?

bakejeyner:
build now

platform/security/idp/security-idp-plugin/security-idp-plugin-nameidpolicy/src/main/java/org/codice/ddf/security/idp/plugin/nameidpolicy/NameIdPolicyPresignPlugin.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    // the Assertion's NameID Format is set to the NameIDPolicy Format after the switch statement
+    if (StringUtils.isNotEmpty(nameIdFormatPolicy)) {
+      switch (nameIdFormatPolicy) {
+          // supported NameIDFormats
+        case NameID.UNSPECIFIED:
+          return; // avoid changing the Format later
+        case NameID.PERSISTENT:
+          // TODO DDF-3965
```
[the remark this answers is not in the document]
bakejeyner: ☝️ Intentional

bakejeyner:
build now

brjeter:
Heroed successfully ✅

PR #3422, opened 2018-07-13
DDF-3975 Update default policy for rest endpoint

clockard:
#### What does this PR do?
Updates rest endpoint policy
#### Who is reviewing it? 
@peterhuffer @emmberk @rzwiefel 

@shaundmorris 
@lessarderic 

CI build with itest should be sufficient 
[DDF-3975](https://codice.atlassian.net/browse/DDF-3975)

### DIFF
--- a/distribution/ddf-common/src/main/resources/security/default.policy
+++ b/distribution/ddf-common/src/main/resources/security/default.policy
@@ -46,7 +46,7 @@ grant
 }
 
 grant
-  codeBase "file:/security-migratable" {
+  codeBase "file:/security-migratable/catalog-rest-endpoint" {
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}encryption.properties", "read";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}signature.properties", "read";
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}issuer${/}encryption.properties", "read";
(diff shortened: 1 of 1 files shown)


shaundmorris:
Wonder if at some point we consider adding comments to help us with maintenance at least the tickets will help a little bit.

clockard:
build now

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -339,9 +339,11 @@ grant codeBase "file:/catalog-core-standardframework/platform-solr-server-standa
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
 }
 
-grant codeBase "file:/catalog-plugin-videothumbnail" {
+grant codeBase "file:/catalog-plugin-videothumbnail/catalog-rest-endpoint/catalog-core-standardframework" {
```
lessarderic: ❓Seems odd to me that we would need to grant those permissions to the `catalog-rest-endpoint` and `catalog-core-standardframework` bundles here. Could/should we use `AccessController.doPrivileged()` where the `ffmpeg` binary is executed instead to avoid having to add those permissions?

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -46,7 +46,7 @@ grant
 }
 
 grant
-  codeBase "file:/security-migratable" {
+  codeBase "file:/security-migratable/catalog-rest-endpoint" {
```
lessarderic: ❓Could we grant those permissions to a WSS4J, CXF or DDF security bundle instead?

distribution/ddf-common/src/main/resources/security/default.policy
```diff
@@ -339,9 +339,11 @@ grant codeBase "file:/catalog-core-standardframework/platform-solr-server-standa
     permission java.io.FilePermission "${ddf.home.perm}etc${/}ws-security${/}server${/}-", "read";
 }
 
-grant codeBase "file:/catalog-plugin-videothumbnail" {
+grant codeBase "file:/catalog-plugin-videothumbnail/catalog-rest-endpoint/catalog-core-standardframework" {
     permission java.io.FilePermission "${ddf.home.perm}bin_third_party", "read,write";
     permission java.io.FilePermission "${ddf.home.perm}bin_third_party${/}-", "read,write,execute";
+    permission java.io.FilePermission "${ddf.home.perm}bin_third_party${/}ffmpeg", "read,write";
```
coyotesqrl: :question: this really shouldn't be necessary. Did you find that the recursive permission on the parent directory wasn't sufficient?

distribution/ddf-common/src/main/resources/security/default.policy
coyotesqrl: This is yet another example of where we need to start creating our own permissions and privileging lower-level code. For now, however, I don't imagine there's anyway around it.

distribution/ddf-common/src/main/resources/security/default.policy
clockard: Probably not. I will remove it.

clockard:
build now

shaundmorris:
Looks like an error downloading CVE information

build now

clockard:
CI build was good except for owasp which failed to download. Bypassing since this pr has no impact on  dependencies.

PR #3448, opened 2018-07-18
[2.13.x] DDF-3855 Updated SystemBaseUrl to return the hostname in lowercase

emmberk:
#### What does this PR do?
This PR fixes an issue that was introduced by the CSRF support in which `503 (Forbidden)` responses were returned when loading the Admin Console when DDF was installed with a hostname than is not lowercase. This PR updates the method that grabs the hostname property to always return lowercase so that the CSRF check does not fail.
#### Who is reviewing it? 
@garrettfreibott @peterhuffer @kcover 
@clockard
@vinamartin
Using a hostname that is not lowercase:
- Install DDF using the `Quick Install of DDF on a remote headless server` section of the documentation. Confirm that the Admin Console is functional after install.
- Install DDF with ?dev=true. Confirm that the Admin Console is functional after install.
[DDF-3855](https://codice.atlassian.net/browse/DDF-3855)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/platform/security/filter/security-filter-csrf/src/main/java/org/codice/ddf/security/filter/csrf/CsrfFilter.java
+++ b/platform/security/filter/security-filter-csrf/src/main/java/org/codice/ddf/security/filter/csrf/CsrfFilter.java
@@ -159,7 +159,7 @@ public class CsrfFilter implements SecurityFilter {
         } else {
           sourceAuthority = url.getAuthority();
         }
-        return (trustedAuthorities.stream().anyMatch(sourceAuthority::equals));
+        return (trustedAuthorities.stream().anyMatch(sourceAuthority::equalsIgnoreCase));
       } catch (MalformedURLException e) {
         LOGGER.debug("Could not extract hostname and port from the request URL", e);
         return false;
--- a/platform/security/filter/security-filter-csrf/src/test/groovy/org/codice/ddf/security/filter/csrf/CsrfFilterSpec.groovy
+++ b/platform/security/filter/security-filter-csrf/src/test/groovy/org/codice/ddf/security/filter/csrf/CsrfFilterSpec.groovy
@@ -39,6 +39,10 @@ class CsrfFilterSpec extends Specification {
     static final String PROXY_HTTPS_NOPORT = "https://" + PROXY_HOST
     static final String EXTERNAL_SITE = "https://example.com"
 
+    static final String DDF_UPPER = "https://" + DDF_HOST.toUpperCase() + ":" + DDF_HTTPS_PORT
+    static final String PROXY_UPPER = "https://" + PROXY_HOST.toUpperCase() + ":" + PROXY_HTTPS_PORT
+    static final String EXTERNAL_UPPER = "https://EXAMPLE.COM"
+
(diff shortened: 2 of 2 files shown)


libs/common-system/src/main/java/org/codice/ddf/configuration/SystemBaseUrl.java
```diff
@@ -240,13 +240,13 @@ public String getHttpsPort() {
   }
 
   /**
-   * Gets the current host name or IP address from the system properties, or {@link #DEFAULT_HOST}
-   * if not set.
+   * Gets the lowercase current hostname or IP address from the system properties, or {@link
+   * #DEFAULT_HOST} if not set.
    *
-   * @return host name, IP address or {@link #DEFAULT_HOST}
+   * @return hostname, IP address, or {@link #DEFAULT_HOST}
```
peterhuffer: This class is exported by this bundle. I'd be curious how this would affect down stream projects. Why didn't we change the comparison in the problem area to ignore case?

brendan-hofmann: I was wondering that as well. Is there a reason not to just do `equalsIgnoreCase` in the CSRF filter? W3 recommends assuming URLs are case-sensitive unless you know otherwise, like in the CSRF filter. https://www.w3.org/TR/WD-html40-970708/htmlweb.html

vinamartin: ❓ Shouldn't hostnames be normalized (lowercased) throughout our system? 
I realize this PR is meant to fix this specific bug, but I can't help but wonder if it's a sign we should have it normalized on the back end when we're passing it around.

brendan-hofmann: This change would normalize on the backend, with the exception of system property injection via blueprint/metatype.

Why can't the hostname just be set as lowercase in `system.properties`? Seems like something we could check and log a warning for.

peterhuffer: If we are using this method to display this value anywhere we are losing the original casing as well.

garrettfreibott: Update added to include case insensitivity within the csrf filter. Since hostnames are case-insensitive, we're keeping the lowercase-format for SystemBaseUrl hostname methods update to keep hostnames consistent and normalized on the backend. Like @brendan-hofmann mentioned, if code needs to have the exact casing that's written in the system.properties file, they can access the property directly instead of asking the system what it's hostname is.

garrettfreibott:
build now

garrettfreibott:
build now

garrettfreibott:
The lowercase hostname change was removed for now and will be addressed more thoroughly in a future PR - this PR is now only for the CSRF Filter case-insensitivity change

garrettfreibott:
build now

rzwiefel:
Hero'ed successfully.

PR #3455, opened 2018-07-19
DDF-3952 Fix backend to allow relative time queries with decimals

kcover:
#### What does this PR do?

Gives GeotoolsFilterAdapterImpl the ability to process relative time
filters that contain decimal values and adds a test for the new
functionality. Also corrects some static analysis findings involving repeated strings.

Note: searches can be inexact when given a year or month decimal value due to the difficulty of accounting for a leap day and varying month lengths in a decimal unit. The actual period parsed
will be longer than the given value by a maximum of 6 hours (when given X.9999.. years)

#### Who is reviewing it? 

@peterhuffer @emmberk @oconnormi 

@andrewkfiedler
@clockard
@tbatie 

Unit tests. Verify that relative time search works as it did before and works when given decimal values. This isn't how we usually test but I also highly suggest putting a break point in GeotoolsFilterAdapterImpl to check that the time parsed from queries containing decimals is acceptable.

[DDF-3952](https://codice.atlassian.net/browse/DDF-3952)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
+++ b/catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
@@ -103,14 +103,24 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
 
   public static final String CQL_KILOMETERS = "kilometers";
 
+  private static final String EXPRESSION_NOT_SUPPORTED =
+      " expression not supported by Filter Adapter.";
+
+  private static final String FILTER_NOT_SUPPORTED = " filter not supported by Filter Adapter.";
+
   private static final double ERROR_THRESHOLD = .000001;
 
   private static final FilterFactory FF = new FilterFactoryImpl();
 
-  // Does not support fractional values
-  private static final Pattern RELATIVE_TEMPORAL_MATCHER =
-      Pattern.compile(
-          "RELATIVE\\(P(?!$)(\\d+Y)?(\\d+M)?(\\d+W)?(\\d+D)?(T(?=\\d)(\\d+H)?(\\d+M)?(\\d+S)?)?\\)");
+  // We shouldn't be using a regex here. This should be replaced by a function like the ones found
+  // in ddf.catalog.impl.filter
+  private static final String DECIMAL_REGEX = "\\\\d*\\\\.\\\\d+|\\\\d+\\\\.?\\\\d*";
+
+  private static final String SHORTENED_RELATIVE_TEMPORAL_REGEX =
+      "RELATIVE\\(P(?!$)(?:(dec)Y)?(?:(dec)M)?(?:(dec)W)?(?:(dec)D)?(?:T(?=dec)(?:(dec)H)?(?:(dec)M)?(?:(dec)S)?)?\\)";
+
+  private static final Pattern RELATIVE_TEMPORAL_REGEX =
+      Pattern.compile(SHORTENED_RELATIVE_TEMPORAL_REGEX.replaceAll("dec", DECIMAL_REGEX));
 
   public <T> T adapt(Filter filter, FilterDelegate<T> filterDelegate)
       throws UnsupportedQueryException {
@@ -128,15 +138,15 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
 
   public Object visit(NilExpression expression, Object delegate) {
     throw new UnsupportedOperationException(
-        NilExpression.class.getSimpleName() + " expression not supported by Filter Adapter.");
+        NilExpression.class.getSimpleName() + EXPRESSION_NOT_SUPPORTED);
   }
 
   public Object visit(Add expression, Object delegate) {
-    throw new UnsupportedOperationException(Add.NAME + " expression not supported.");
+    throw new UnsupportedOperationException(Add.NAME + EXPRESSION_NOT_SUPPORTED);
   }
 
   public Object visit(Divide expression, Object delegate) {
-    throw new UnsupportedOperationException(Divide.NAME + " expression not supported.");
+    throw new UnsupportedOperationException(Divide.NAME + EXPRESSION_NOT_SUPPORTED);
   }
 
   @Override
@@ -211,8 +221,7 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
   }
 
   public Object visit(Id filter, Object delegate) {
-    throw new UnsupportedOperationException(
-        Id.class.getSimpleName() + " filter not supported by Filter Adapter.");
+    throw new UnsupportedOperationException(Id.class.getSimpleName() + FILTER_NOT_SUPPORTED);
   }
 
   @SuppressWarnings("unchecked")
@@ -319,22 +328,11 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
     List<Object> functionArgs = filterValues.functionArgs;
 
     // Special case to handle relative temporal queries
-    if (literal instanceof String
-        && RELATIVE_TEMPORAL_MATCHER.matcher((String) literal).matches()) {
+    if (literal instanceof String && RELATIVE_TEMPORAL_REGEX.matcher((String) literal).matches()) {
       DateTime currentDateTime = new DateTime();
 
-      // Split out the ISO 8601 duration from the ECQL function
-      String duration = (((String) literal).split("[\\(\\)]"))[1];
-
-      org.joda.time.Period period = org.joda.time.Period.parse(duration);
-
-      DateTime pastDateTime =
-          currentDateTime
-              .minusYears(period.getYears())
-              .minusMonths(period.getMonths())
-              .minusDays(period.getDays())
-              .minusHours(period.getHours())
-              .minusMinutes(period.getMinutes());
+      org.joda.time.Period period = PeriodParser.parse((String) literal, RELATIVE_TEMPORAL_REGEX);
+      DateTime pastDateTime = currentDateTime.minus(period);
 
       return ((FilterDelegate<?>) delegate)
           .propertyIsBetween(propertyName, pastDateTime.toDate(), currentDateTime.toDate());
@@ -876,8 +874,7 @@ public class GeotoolsFilterAdapterImpl implements FilterAdapter, FilterVisitor, 
   }
 
   public Object visit(AnyInteracts anyInteracts, Object delegate) {
-    throw new UnsupportedOperationException(
-        AnyInteracts.NAME + " filter not supported by Filter Adapter.");
+    throw new UnsupportedOperationException(AnyInteracts.NAME + FILTER_NOT_SUPPORTED);
   }
 
   public Object visit(Begins begins, Object delegate) {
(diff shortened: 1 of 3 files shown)


catalog/core/catalog-core-impl/filter-proxy/src/test/java/ddf/catalog/filter/proxy/adapter/test/FilterAdapterTest.java
```diff
@@ -624,6 +626,26 @@ public void testTemporalRelativeFilter() {
...
+    BigDecimal hourFraction = new BigDecimal(0.25);
+    int seconds =
+        yearFraction.multiply(secondsPerYear).add(hourFraction.multiply(secondsPerHour)).intValue();
+    org.joda.time.Period period = new org.joda.time.Period(3, 3, 0, 2, 0, 6, 0, 0);
+    period = period.plusSeconds(seconds);
+
+    try {
+      GeotoolsFilterAdapterImpl geotoolsFilterAdapter = new GeotoolsFilterAdapterImpl();
+      assertThat(geotoolsFilterAdapter.parsePeriod(query), is(period));
+    } catch (Exception e) {
```
bellcc: ✏️ Throw a more specific exception here so that we'll be better able to diagnose the problem if this test fails.

catalog/core/catalog-core-impl/filter-proxy/src/test/java/ddf/catalog/filter/proxy/adapter/test/FilterAdapterTest.java
```diff
@@ -603,7 +606,7 @@ public void testTemporalRelativeFilter() {
     doReturn("created").when(mockPropertyName).accept(any(), any());
 
     Literal mockLiteral = mock(Literal.class);
-    doReturn("RELATIVE(P2Y3M2DT1H6M)").when(mockLiteral).accept(any(), any());
+    doReturn("RELATIVE(P3Y3M2DT1H6M)").when(mockLiteral).accept(any(), any());
```
kcover: This got modified while I was playing with the code I'll change it back if there's some other change I need to go in and make.

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -319,23 +326,23 @@ public Object visit(PropertyIsEqualTo filter, Object delegate) {
...
-        && RELATIVE_TEMPORAL_MATCHER.matcher((String) literal).matches()) {
+    if (literal instanceof String && RELATIVE_TEMPORAL_REGEX.matcher((String) literal).matches()) {
       DateTime currentDateTime = new DateTime();
 
       // Split out the ISO 8601 duration from the ECQL function
-      String duration = (((String) literal).split("[\\(\\)]"))[1];
+      String periodText = (((String) literal).split("[\\(\\)]"))[1];
 
-      org.joda.time.Period period = org.joda.time.Period.parse(duration);
+      org.joda.time.Period period = PeriodParser.parse(periodText);
```
ethantmanns: Is there a reason we chose to make our own parser instead of a `PeriodFormatter` to hand to the `Period.parse` method?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -103,14 +103,22 @@
...
   private static final FilterFactory FF = new FilterFactoryImpl();
 
-  // Does not support fractional values
-  private static final Pattern RELATIVE_TEMPORAL_MATCHER =
-      Pattern.compile(
-          "RELATIVE\\(P(?!$)(\\d+Y)?(\\d+M)?(\\d+W)?(\\d+D)?(T(?=\\d)(\\d+H)?(\\d+M)?(\\d+S)?)?\\)");
+  private static final String DECIMAL_REGEX = "\\\\d*\\\\.\\\\d+|\\\\d+\\\\.?\\\\d*";
+
+  private static final String SHORTENED_RELATIVE_TEMPORAL_REGEX =
+      "RELATIVE\\(P(?!$)(?:(dec)Y)?(?:(dec)M)?(?:(dec)W)?(?:(dec)D)?(?:T(?=dec)(?:(dec)H)?(?:(dec)M)?(?:(dec)S)?)?\\)";
```
ethantmanns: I'm reading online and it looks like decimals are only supposed to be supported for the lowest order date and the lowest order time. In other words, if there's a days present, you can't have a decimal on the months, weeks, or years. Same applies to time

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
kcover: We could do that too but wouldn't that essentially do the same thing but hand back a string instead of a period?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
kcover: Yea according to the ISO spec only the seconds field should contain a decimal in a period like this, however, my understanding is the UI component team discussed this and decided they wanted to do unit conversion when the user inputs a decimal value, rather than just deny it. @andrewkfiedler can you chime in here?

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
ethantmanns: Looking at the source code for `Period.parse`, looks like it just calls and returns the value of `PeriodFormatter.parsePeriod`. According to the docs here (https://www.joda.org/joda-time/apidocs/org/joda/time/format/PeriodFormatter.html#parsePeriod-java.lang.String-) looks like that also returns a Period.

I think writing a `PeriodFormatter` and using Joda to call it still looks a little better, but I'm not sure if there are any other benefits to it, really. I meant to put a ✏️

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
ethantmanns: I think the minutes can contain a decimal if it's the lowest order unit specified in the Period string we hand it. I can't remember offhand if we shorten something like 1 year and 1 month to 1y1m or if we fill it in like 1y1m0w0d (what I wrote probably isn't ISO). But I think if we're shortening it then in this case the month can have a decimal.

I think if we really want to support being non-ISO compliant, are we able to update it on the fly in the UI?

andrewkfiedler: The UI will always send over compliant ISO as is.  It only allows one unit at a time for a period, so decimals are always fine.

ethantmanns: Just to double check, is a UI search the only path that will hit this code?

kcover: I want to say yes just because the regex is called RELATIVE_TEMPORAL_REGEX, But I can't be sure and I don't know how to check.

kcover: nvm you can hit it using a cql search from the console so I'm sure there are lot's of other ways to get to it.

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
```diff
@@ -319,22 +326,13 @@ public Object visit(PropertyIsEqualTo filter, Object delegate) {
     List<Object> functionArgs = filterValues.functionArgs;
 
     // Special case to handle relative temporal queries
-    if (literal instanceof String
-        && RELATIVE_TEMPORAL_MATCHER.matcher((String) literal).matches()) {
+    if (literal instanceof String && RELATIVE_TEMPORAL_REGEX.matcher((String) literal).matches()) {
```
ryeats: If you have time I would appreciate if we could find a way to do this as a proper function like ProximityFunction so that it is recognized and parsed as such instead of having to do a regex.

peterhuffer:
Successful hero. Ingested the reuters data set and performed relative searches on created dates and got back expected results. Verified backend time like
RELATIVE(31.21) years = 1987-06-12T19:54:23.583-07:00
RELATIVE(31.2) years = 1987-06-16T11:36:28.738-07:00
Worked out .01 = 3.65 days = 3 days 15 hours 36 minutes
which looks alright to me 👍

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/GeotoolsFilterAdapterImpl.java
kcover: I added a comment at the regex to steer people away from using regexs and I created a ticket to turn this into a function in the future.

kcover:
build now

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/PeriodParser.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.regex.Pattern;
+import javax.validation.constraints.NotNull;
+import org.joda.time.Period;
+
+/* Parses a period similar to ISO8601 format that contains decimal values. ISO8601 doesn't allow
+decimal values so this class converts any decimal values to seconds and adds
+them to the seconds field of the period. For example: "P8.2M" will become "P8MT525601S" */
+public class PeriodParser {
+
+  private static final String DECIMAL_REGEX = "\\\\d*\\\\.\\\\d+|\\\\d+\\\\.?\\\\d*";
```
tbatie: ❗️ We should share the regex here and in the `GeotoolsFilterAdapterImpl` in case on diverges

kcover:
build now

catalog/core/catalog-core-impl/filter-proxy/src/main/java/ddf/catalog/filter/proxy/adapter/PeriodParser.java
```diff
@@ -0,0 +1,126 @@
...
+    HOURS, // 4
+    MINUTES, // 5
+    SECONDS // 6
+  }
+
+  public static Period parse(@NotNull String periodText, Pattern regex) {
+
+    Matcher matcher = regex.matcher(periodText);
+    if (matcher.matches()) {
+      long[] unitDurationsInSeconds = getUnitDurationsInSeconds();
```
clockard: Couldn't this be a static class variable populated once?

PR #3476, opened 2018-07-23
[2.13.x] DDF-3915 Solr script fixes for *nix

ahoffer:
#### What does this PR do?
    Add two new properties to system.properties: START_SOLR; START_KARAF
    Add Jason L.'s fix for managing Solr
    Split ddf script into 3 scripts
    Start Solr with two-way TLS by default

#### Who is reviewing it? 
@mcalcote 
@rzwiefel 
@lamhuy 
@oconnormi 
@JGatley 

@pklinef 
@clockard 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### How to Test:
1. Secure loclahost.
    - Install DDF. 
    - After installation, the admin console should show a warning about insecure defaults (this mean it has a good connection to Solr).
    - Visit https://localhost:8994/solr in a browser.
    - The browser should display a certification error
1. Secure, certs
    - Install DDF with ?dev=true.
    - After installation, the admin console should show a warning about insecure defaults.
    - Visit https://localhost:8994/solr in a browser.
    - The browser should display a certification error
1. Unsecure, localhost
   - Edit system.properties before installation. 
   - Change solr.http.protocol to  "http"
   - Start DDF, but do not go through the installation screens
   - Try http://localhost:8994/solr. It should bring up the Solr admin page
   - Install DDF
   - After installation, the admin console should show a warning about insecure defaults
1. Unsecure, certs
   - Edit system.properties before installation. 
   - Change solr.http.protocol to  "http"
   - Start DDF, but do not go through the installation screens
   - Try http://localhost:8994/solr. It should bring up the Solr admin page
   - Install DDF with **?dev=true**
   - After installation and restart, the admin console should show a warning about insecure defaults

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/distribution/ddf-common/src/main/resources-filtered/etc/system.properties
+++ b/distribution/ddf-common/src/main/resources-filtered/etc/system.properties
@@ -23,6 +23,8 @@
 #
 
 #START DDF SETTINGS
+start.ddf=true
+start.solr=true
 
 # DDF Environment Settings
 ddf.etc=${karaf.etc}
@@ -82,14 +84,14 @@ org.codice.ddf.system.registry-id=
 #
 # Uncomment the properties based on the desired configuration:
 
-# For Solr Server:
+# For Solr Server
+solr.data.dir=_DO_NOT_EXPAND_${karaf.home}/solr/server/solr
 solr.client=HttpSolrClient
 solr.http.port=8994
-# Use "https" in the URL to configure the Solr server for TLS
-# Do NOT update the solr.http.url when quick installing on a remote headless server.
-solr.http.url=http://localhost:_DO_NOT_EXPAND_${solr.http.port}/solr
-solr.data.dir=_DO_NOT_EXPAND_${karaf.home}/solr/server/solr
+solr.http.protocol=https
 
+# Do NOT update the solr.http.url when quick installing on a remote headless server.
+solr.http.url=_DO_NOT_EXPAND_${solr.http.protocol}://_DO_NOT_EXPAND_${org.codice.ddf.system.hostname}:_DO_NOT_EXPAND_${solr.http.port}/solr
 
 # For Solr Cloud Provider:
 # solr.client=CloudSolrClient
--- a/distribution/ddf-common/src/main/resources/bin/ddf
+++ b/distribution/ddf-common/src/main/resources/bin/ddf
@@ -1,186 +1,71 @@
 #!/usr/bin/env bash
 
-# For security reasons, an internally managed Solr configuration
-# should always deny remote requests unless debugging
-# or troubleshooting in a safe environment
-SOLR_DENY_REMOTE_REQUESTS=true
-
 # Store arguments for use in local functions
 FIRST_ARGUMENT=$1
 ALL_ARGUMENTS=$@
+
+# Declare Karaf executable's return code
+KARAF_EXEC_RC=0
 
 # Useful directories
 SCRIPTDIR=$(dirname $0)
 HOME_DIR=$(cd "${SCRIPTDIR}/.."; pwd -P)
 
-# Declare Karaf executable's return code
-KARAF_EXEC_RC=-1
-
-# This variable is automatically set to true if
-# the function set_solr_https_properties() is called
-HTTPS=false
-
 # Useful files
 PROPERTIES_FILE=$HOME_DIR/etc/system.properties
 RESTART_FILE="$SCRIPTDIR/restart.jvm"
-SOLR_EXEC=${HOME_DIR}/solr/bin/solr
+SOLR_EXEC=${SCRIPTDIR}/ddfsolr
+GET=${SCRIPTDIR}/get_property
 
-# Extract the value of a property from a Java properties file
-# This function does not handle multi-line properties
-# If the property does not exist, it will print an error message to stderr and
-# set a flag that it failed to read a property
-get_property() {
-
-    value=""
-    value=`grep ^$1= "$PROPERTIES_FILE" | cut -d = -f2`
-
-    if [ $value ]; then
-        echo $value
-    else
-        echo "ERROR: $1 is not a property of $PROPERTIES_FILE" >&2
-        echo ""
-    fi
-}
+# Flags
+START_KARAF="false"
+START_SOLR="false"
 
 refresh_properties() {
-    # Extract values from Java properties file and construct Solr URL
-    SOLR_CLIENT=$(get_property solr.client)
-    SOLR_PORT=$(get_property solr.http.port)
-
-    # Do not proceed if required properties are missing
-    if [ ! $SOLR_CLIENT ] || [ ! $SOLR_PORT ]; then
-        exit
-    fi
-
-  }
-
-
-
-# Return 0 (success/true) if the Solr lifecycle should be managed by this script
-is_managing_solr() {
-  local RC=1
-  if [ $SOLR_CLIENT = HttpSolrClient  ]; then
-    RC=0
-  fi
-  return $RC
+    # Extract values from Java properties file
+    START_KARAF=$($GET "start.ddf")
+    START_SOLR=$($GET "start.solr")
 }
 
 
 # Return 0 (success/true) if ddf_on_error.sh created a restart file
 is_restarting() {
     local RC=1
-    if [ -f $RESTART_FILE ]; then
+    if [ -f "$RESTART_FILE" ]; then
         RC=0
     fi
     return $RC
 }
-
-
-# Return 0 (success/true) Solr should only accept requests from the same host
-solr_must_deny_remote_requests() {
-    local RC=1
-    if [ $SOLR_DENY_REMOTE_REQUESTS = true ]; then
-        RC=0
-    fi
-    return $RC
-}
-
 
 # Remove the restart file so we can detect later if restart was requested
 clear_restart_flag() {
   rm -f $RESTART_FILE
 }
 
-# Set the appropriate option to prevent Solr from responding to remote requests
-deny_remote_connection_option() {
-    if solr_must_deny_remote_requests;  then
-      if [ $HTTPS = true ]; then
-        echo "-Dsolr.jetty.host=127.0.0.1"
-      else
-        echo "-Djetty.host=127.0.0.1"
-      fi
+# Return 0 (success/true) if this script should start and stop Solr
+managing_solr() {
+    local RC=1
+    if [ "$START_SOLR" = "true" ]; then
+        RC=0
     fi
-}
-
-# Start Solr using Solr's script
-start_solr() {
-
-    # Does not check to see if Solr is running the Solr port; there is no guarantee
-    # the Solr server is for this instance of DDF. It could have been started elsewhere.
-    $SOLR_EXEC start -force -p $SOLR_PORT $(deny_remote_connection_option)
-    local SOLR_START_RC=$?
-        if [ "$SOLR_START_RC" -ne 0 ]; then
-            echo "WARNING. Solr start returned an error. Is Solr already running?"
-	        echo "Please check Solr logs for additional information"
-            echo ""
-        fi
-}
-
-# Set special environment variables used by Solr to configure TLS
-set_solr_https_properties() {
-
-      # Use the same key and trust stores as the DDF uses.
-      export SOLR_SSL_KEY_STORE=$HOME_DIR/$(get_property javax.net.ssl.keyStore)
-      export SOLR_SSL_KEY_STORE_PASSWORD=$(get_property javax.net.ssl.keyStorePassword)
-      export SOLR_SSL_KEY_STORE_TYPE=$(get_property javax.net.ssl.keyStoreType)
-      export SOLR_SSL_TRUST_STORE=$HOME_DIR/$(get_property javax.net.ssl.trustStore)
-      export SOLR_SSL_TRUST_STORE_PASSWORD=$(get_property javax.net.ssl.trustStorePassword)
-      export SOLR_SSL_TRUST_STORE_TYPE=$(get_property javax.net.ssl.trustStoreType)
-
-      # Require two-way TLS. Change this value to false to disable client authentication.
-      export SOLR_SSL_NEED_CLIENT_AUTH=true
-
-      # Not used
-      export SOLR_SSL_WANT_CLIENT_AUTH=false
-
-      # Set flag
-      HTTPS=true
-}
-
-
-# Change these values to alter the memory available to the managed Solr JVM.
-set_solr_memory() {
-    export SOLR_JAVA_MEM="-Xms512m -Xmx2g"
-}
-
-
-stop_solr() {
-    $SOLR_EXEC stop -p $SOLR_PORT
-}
-
-print_messages() {
-    echo ""
-    echo "Home directory is $HOME_DIR"
-    if is_managing_solr; then
-        echo "This script manages starting and stopping Solr"
-    fi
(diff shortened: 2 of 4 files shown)


ahoffer:
build now

distribution/ddf-common/src/main/resources/bin/ddfsolr
```diff
@@ -0,0 +1,121 @@
...
+
+        * )
+            usage
+            ;;
+    esac
+done
+shift $((OPTIND-1))
+
+# -----------------------------------------------------------------------------
+if [ ! $COMMAND ]; then
```
oconnormi: You might want to change this to `if [ -z "$COMMAND" ]; then` which will return true if the variable is either unset or set to an empty string

distribution/ddf-common/src/main/resources/bin/ddfsolr
```diff
@@ -0,0 +1,121 @@
...
+HOME_DIR=$(cd "${SCRIPTDIR}/.."; pwd -P)
+GET=${SCRIPTDIR}/get_property
+SOLR_EXEC=${HOME_DIR}/solr/bin/solr
+SOLR_RC=0
+
+# Read port from system.properties file
+SOLR_PORT=$($GET "solr.http.port")
+
+# Do not proceed if required properties are missing
+if [ ! $SOLR_PORT ]; then
```
oconnormi: Should change this to `if [ -z "$SOLR_PORT" ]; then` to catch both unset and empty string

distribution/ddf-common/src/main/resources/bin/get_property
```diff
@@ -0,0 +1,29 @@
...
+    exit 1
+fi
+
+# Extract the value of a property from a Java properties file
+# This function does not handle multi-line properties
+# If the property does not exist, print  error message to stderr
+value=""
+value=`grep ^$PROPERTY_NAME= "$PROPERTIES_FILE" | cut -d = -f2`
+
+if [ $value ]; then
```
oconnormi: if you are checking that the property is not unset and not equal to an empty string then do `if  [ ! -z "$value" ]; then`

distribution/ddf-common/src/main/resources/bin/ddfsolr
```diff
@@ -0,0 +1,121 @@
...
+    export SOLR_JAVA_MEM="-Xms512m -Xmx2g"
+}
+
+# -----------------------------------------------------------------------------
+stop_solr() {
+    $SOLR_EXEC stop -p $SOLR_PORT
+}
+
+# -----------------------------------------------------------------------------
+if [ $COMMAND = start ]; then
```
oconnormi: Should change this to be `if [ "$COMMAND" = "start" ]; then`

distribution/ddf-common/src/main/resources/bin/ddfsolr
```diff
@@ -0,0 +1,121 @@
...
+    set_solr_memory
+    if [ $SECURE = true ]; then
+        set_solr_https_properties   
+    fi
+    start_solr
+    exit $SOLR_RC
+fi
+
+if [ $COMMAND = stop ]; then
```
oconnormi: Should change this to `if [ "$COMMAND" = "stop" ]; then`

distribution/ddf-common/src/main/resources/bin/ddfsolr
```diff
@@ -0,0 +1,121 @@
...
+
+# -----------------------------------------------------------------------------
+stop_solr() {
+    $SOLR_EXEC stop -p $SOLR_PORT
+}
+
+# -----------------------------------------------------------------------------
+if [ $COMMAND = start ]; then
+    set_solr_memory
+    if [ $SECURE = true ]; then
```
oconnormi: In general it is good practice to put qoutes around variable usages in bash comparisons and around any strings used in the comparison. Here it would just be `"$SECURE"`, `true` will work both quoted and unquoted in this case

distribution/ddf-common/src/main/resources/bin/get_property
```diff
@@ -0,0 +1,29 @@
...
+if [ ! -f $PROPERTIES_FILE ]; then
+    echo Cannot find file $PROPERTIES_FILE >&2
+    exit 1
+fi
+
+# Extract the value of a property from a Java properties file
+# This function does not handle multi-line properties
+# If the property does not exist, print  error message to stderr
+value=""
+value=`grep ^$PROPERTY_NAME= "$PROPERTIES_FILE" | cut -d = -f2`
```
oconnormi: ✏️ Use `$(command | command)` instead of `` `command | command` ``

distribution/ddf-common/src/main/resources/bin/ddf
```diff
@@ -75,112 +37,36 @@ is_restarting() {
...
-    if is_managing_solr; then
-        echo "This script manages starting and stopping Solr"
+attempt_startup() {
+    if managing_solr; then
+        $SOLR_EXEC start
     fi
-    if solr_must_deny_remote_requests; then
-        echo "Solr is configured to reject remote requests"
+    
+    if [ $START_KARAF = true ]; then
```
oconnormi: Should put the `$START_KARAF` in quotes

distribution/ddf-common/src/main/resources/bin/ddf
```diff
@@ -75,112 +37,36 @@ is_restarting() {
...
-    if solr_must_deny_remote_requests;  then
-      if [ $HTTPS = true ]; then
-        echo "-Dsolr.jetty.host=127.0.0.1"
-      else
-        echo "-Djetty.host=127.0.0.1"
-      fi
+# Return 0 (success/true) if this script should start and stop Solr
+managing_solr() {
+    local RC=1
+    if [ $START_SOLR = true ]; then
```
oconnormi: Should put the `$START_SOLR` in quotes

oconnormi:
❓ Just out of curiosity why are we not using the value of the `solr.http.url` to determine whether or not we need to start solr? 

Wouldn't it make more sense to have the solr.http.url default to `https://localhost:_DO_NOT_EXPAND_${solr.http.port}/solr` and the presence of `localhost` could be used as the determining factor. If someone were to then provide an alternate remote url for solr it would simply not start the local solr instance.

distribution/ddf-common/src/main/resources/bin/get_property
oconnormi: Also I think this needs to be `grep "^$PROPERTY_NAME ="` since there is a space after the equals sign typically. If you want to handle both cases then don't search for the equals sign at all and you can later split the result like this:

```
match=$(grep "^$PROPERTY_NAME" "$PROPERTIES_FILE") # find match
IFS='=' read key value <<< "$match" # split match on = character
value=${value##*( )} # Trim whitespace. NOTE: this is bash specific, only suggested it because we are explicitly using bash in this script```

File: distribution/ddf-common/src/main/resources/bin/get_property [hunk 10]
```diff
@@ -0,0 +1,29 @@
...
+# Check file exists where it is supposed to
+if [ ! -f $PROPERTIES_FILE ]; then
+    echo Cannot find file $PROPERTIES_FILE >&2
+    exit 1
+fi
+
+# Extract the value of a property from a Java properties file
+# This function does not handle multi-line properties
+# If the property does not exist, print  error message to stderr
+value=""
```
oconnormi: won't need this if you using the `-z` operator in the `if` bracket

distribution/ddf-common/src/main/resources/bin/get_property
oconnormi: Scratch that there isn't usually a space, however `some.property = "someValue"` is valid in a java properties file but will not be found by this as it currently stands. Definitely recommend the second approach

distribution/ddf-common/src/main/resources/bin/ddfsolr
ahoffer: I'm a minimalist at heart. Extra characters cause me psychic pain. But in the interest of robust software, I will add the quotes.

distribution/ddf-common/src/main/resources/bin/ddfsolr
ahoffer: I'm a minimalist at heart. Extra characters cause me psychic pain. But in the interest of robust software, I will add the quotes.

distribution/ddf-common/src/main/resources/bin/get_property
ahoffer: That equals sign is not supposed to be there. It must be a copy paste error

distribution/ddf-common/src/main/resources/bin/get_property
ahoffer: >Just out of curiosity why are we not using the value of the solr.http.url to determine whether or not we need to start solr?

I don't have an answer. I'm hesitant to do that because every time I think I can predict when to start/stop Solr automatically, reality proves me wrong. Let me tag @mcalcote and @rzwiefel on it so that can weigh in.

distribution/ddf-common/src/main/resources/bin/ddfsolr
LinkMJB: Quoting variables is absolutely critical in shell scripts, especially if you don't have a default value assigned. If the variable is empty and expands, you will end up with an imbalanced comparison at best. At worst, the script will behave in interesting, unexpected, ways. The most common of which, entering into conditional blocks that you did not expect possible in the given situation.

distribution/ddf-common/src/main/resources/bin/get_property
LinkMJB: I would recommend not relying on bash builtins. I love the capabilities that bash provides, but have found that eventually someone will expect to be able to run shell scripts in an environment that only has /bin/sh available. In most modern operating systems /bin/bash can be expected to be available, but isn't a guarantee.

distribution/ddf-common/src/main/resources/bin/ddfsolr
ahoffer: You are right. I tried it yesterday and -z wasn't behaving like it should-- I must have made a mistake. Just tried -z and it did the right thing.

distribution/ddf-common/src/main/resources/bin/get_property
ahoffer: I see. I will drop the equals match. The alternative is to match zero or more whitespace, but I don't' think the extra complexity really adds much value to the reg ex.

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -23,6 +23,8 @@
 #
 
 #START DDF SETTINGS
+start.ddf=true
+start.solr=true
```
ahoffer: Took these two properties from @rzwiefel 's work in progress. 
The directive was to make the components separable for starting/stopping

distribution/ddf-common/src/main/resources-filtered/etc/system.properties
```diff
@@ -85,8 +87,8 @@ org.codice.ddf.system.registry-id=
 # For Solr Server:
 solr.client=HttpSolrClient
 solr.http.port=8994
-# Use "https" in the URL to configure the Solr server for TLS
-solr.http.url=http://localhost:_DO_NOT_EXPAND_${solr.http.port}/solr
+solr.http.protocol=https
```
ahoffer: Added this property so that the scripts can decide to start Solr with or without TLS

pklinef:
❓ Where are the Windows bat file changes?  Will those scripts still work with these *nix changes?

rzwiefel:
They're working on them right now @pklinef

clockard:
Successfully ran the test steps! 👍

distribution/ddf-common/src/main/resources/bin/get_property
```diff
@@ -0,0 +1,28 @@
...
+SCRIPTDIR=$(dirname $0)
+HOME_DIR=$(cd "${SCRIPTDIR}/.."; pwd -P)
+
+# Useful files
+PROPERTIES_FILE=$HOME_DIR/etc/system.properties
+
+# Check file exists where it is supposed to
+if [ ! -f $PROPERTIES_FILE ]; then
+    echo Cannot find file $PROPERTIES_FILE >&2
+    exit 1
```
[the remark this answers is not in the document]
ahoffer: The properties are the defaults. We ship with default values for the properties.

distribution/ddf-common/src/main/resources/bin/get_property
```diff
@@ -0,0 +1,28 @@
...
+# Extract the value of a property from a Java properties file
+# This function does not handle multi-line properties
+# If the property does not exist, print  error message to stderr
+value=$(grep "^$PROPERTY_NAME\s*=\s*" "$PROPERTIES_FILE" | cut -d = -f2)
+
+if [ ! -z "$value" ]; then
+    echo $value
+else
+    echo "ERROR: $PROPERTY_NAME is not a property of $PROPERTIES_FILE" >&2
+    exit 2
```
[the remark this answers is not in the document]
ahoffer: I saw the altered get_property function. The same values were passed into it (as defaults) that were present in the system.properties file. Let's set the default values in the system.properties file and not in the scripts.

distribution/ddf-common/src/main/resources/bin/get_property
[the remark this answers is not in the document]
mcalcote: I think this file just returns the value of the property if found, so the fail can be handled on the callers end and we can put in a default value.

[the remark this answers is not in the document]
ahoffer: The default value is already in the system.properties file.
https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

mcalcote:
@pklinef here are the windows changes: https://github.com/codice/ddf/pull/3507

distribution/ddf-common/src/main/resources/bin/get_property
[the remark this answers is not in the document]
rzwiefel: Not everyone uses this system properties file, which is why it was failing for some downstream projects.

ahoffer:
@brendan-hofmann @pklinef 
Looking for your comments. Right now, these scripts are implemented fail-fast. If the script cannot find a required property in the system.properties file, it fails. This alerts administrators to take action to repair an invalid configuration. I deemed this preferable to an implementation that masked invalid configuration issues because masking issues generally makes debugging more difficult to debug. 

However, I heard from @rzwiefel that this might be causing problems for downstream projects (@brendan-hofmann). Is this the case? If so, how should these scripts behave if properties cannot be retrieved from a system.properties file?

brendan-hofmann:
@ahoffer We've resolved this issue now. I think a solution is out of scope for this PR, as the real issue is that too much configuration involves modifying default files. If you try to manage these files via something like puppet, it is very easy to miss additions, modifications, or removals from the default set, which can result in a formerly valid configuration no longer working. Ideally I would like to be able to configure the system without modifying any included files.

brendan-hofmann:
How does two-way TLS play with hardening using signed certs? Do we need signed certs for both karaf and solr, or will one cert for the machine hostname work on both ends?

ahoffer:
@brendan-hofmann Karaf and Solr use the same keystores, so there is no need for additional certs because both are running on the same host.

clockard:
build now

PR #3489, opened 2018-07-25
DDF-4006,DDF-4007 Adds support for jsx usage in Marionette views, Updates DOM to be React managed, adds region behavior

andrewkfiedler:
#### What does this PR do?
- Adds support for returning jsx in marionette views.
- Adds a dev page for explaining how to use jsx in marionette views.
- Extends the attachElContent method for Views in Marionette so that
     they utilize React's renderer.  This allows us to take advantage of the
     reconciliation provided by the renderer while still using Marionette.
- Building on top of the attachElContent extension comes the new region behavior
     for Marionette Views.  It allows regions (children components) to live
     through subsequent renders.  See the dev guide for more details on
     usage, but suffice to say that it will allow us to render with impunity
     (no more needing to reshow a view every time you want to react to a
     model change).  Going forward we should be using this behavior rather
     than going through Marionette's provided regions (which are destroyed
     upon render).
- Updates the inputs in the UI to no longer use the value attribute.
     With React's render now being utilized, value can't be specified
     without an onchange event, otherwise React will lock down the input and
     prevent the user from changing it (since it's seen as controlled).
     Luckily, the views were already setting the initial values so this was
     an easy change to make (simply remove the value in the hbs).
- Updates the upload.view to handle the new React renderer.  Because of
     the reconciliation, we no longer were recieving a new dom node upon
     clicking new in the upload view.  Dropzone would then complain that it
     was already attached.  There were changes to the upload models to
     ensure that when an upload ends, the models stopped listening to the
     dropzone (since now a new upload would take over the existing
     dropzone).
- Updates the workspace-item view and the workspace-templates view to
     use the new region behavior.
- Updates the content-title view and the query-title view to handle the
     new React renderer.  Both were relying on rerendering and the value
     attribute, so I updated them to use JSX so I could utilize defaultValue
     and the key attribute to ensure React saw them as different when the
     workspace or query changed respectively and things were rerendered.
- Added on to the dev guide a section on how to utilize the new region
     behavior.

#### Who is reviewing it? 
@mackncheesiest 
@djblue 
@adimka 

@codice/ui 

@tbatie 
@vinamartin 

Test all the things!  This is a pretty big change, I went over everything I could with a fine tooth comb and fixed anything that had compatibility issues.

This is to help out new developers who are more familiar with JSX and move us towards React in the future.  

It's also to help new developers avoid some of the potential pitfalls of how Marionette manages children views.

[DDF-4006](https://codice.atlassian.net/browse/DDF-4006)
[DDF-4007](https://codice.atlassian.net/browse/DDF-4007)

#### Review Comment Legend:


andrewkfiedler:
build now

andrewkfiedler:
build now

ui/packages/catalog-ui-search/src/main/webapp/behaviors/region.behavior.js
```diff
@@ -0,0 +1,158 @@
...
+                }
+            })
+    },
+    attachRegion(region) {
+        this.view.$el.find(region.selector).html('');
+        this.view.$el.find(region.selector).append(region._region.currentView.el);
+        region._region.currentView.undelegateEvents();
+        region._region.currentView.delegateEvents();
+    },
+    /* mirror the behavior of marionette (otherwise children views that listen to onAttach will break T_T) */
```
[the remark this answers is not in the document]
andrewkfiedler: No, they don't override anything, this is a behavior so it merely has access to the lifecycle hooks available in the view (one of the reason that behaviors end up being so powerful in Marionette is that they allow a way to compose logic that doesn't run into the issues that mixins or extensions run into).  Put differently, this hooks into the view's lifecycle hooks, but doesn't override that views methods or listeners to the lifecycle hooks at all.  Instead, it adds on to them, composing some behavior on top of them.

This isn't using the default region implementation in Marionette so that we can take advantage of the React renderer to prevent views from being destroyed unnecessarily, but that means we need to imitate some of what the descendant views are expecting so they continue to operate as expected (backwards compatibility).  Basically, a child view might be getting shown deep down in the tree and expect to see an onAttach propagated to it since that's what normal Marionette regions would fire on show.  Yet we're using our own way of attaching regions now which ends up not firing this event.  This goes ahead and fires the event at the correct time for the child views so they can continue to operate without needing to change.

The reason the normal attach doesn't fire when we show the region (see onBeforeReactAttach) is it expects the regions DOM node to be located in the document already, but we're showing and rendering the region long before it reaches the DOM in order to take advantage of React's render in terms of reconciliation.  Thankfully, we can simply have it wait for the parent view's lifecycle events to take place and piggyback off them to fire the events at the right time.

ui/packages/catalog-ui-search/src/main/webapp/component/content-title/content-title.view.js
```diff
@@ -16,47 +16,63 @@
...
-            this.listenTo(this.model, 'change:currentWorkspace', this.handleSaved);
-        },
-        onBeforeShow: function(){
-            this.updateIndicator();
-        },
-        handleSaved: function(){
-            var currentWorkspace = this.model.get('currentWorkspace');
-            this.$el.toggleClass('is-saved', currentWorkspace ? currentWorkspace.isSaved() : false);
-        },
-        updateIndicator: function(workspace){
```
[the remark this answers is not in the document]
andrewkfiedler: This got moved into the region behavior logic.  It's a more advanced usage that takes advantage of the `shouldRegionUpdate` functionality of the region behaviors.  Basically, on subsequent renders this function is called by the behavior to see if the view should be destroyed and reshown.  As a result, all we do when the currentWorkspace changes now is call render.

Notice that all of the definitions for this particular region behavior utilize the function definition rather than a static definition.  As a result it can change across subsequent renders (if and only if `shouldRegionUpdate` is implemented).

ui/packages/catalog-ui-search/src/main/webapp/component/workspace-item/workspace-item.view.js
```diff
@@ -18,26 +18,68 @@ define([
     'marionette',
     'underscore',
     'jquery',
-    './workspace-item.hbs',
```
millerw8: Any reason we moved away from this define in other places (which I am glad to see) but didn't do it here?

ui/packages/catalog-ui-search/src/main/webapp/behaviors/region.behavior.js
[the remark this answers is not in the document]
mackncheesiest: 50% 😖  50% 👍

100% :shipit:

ui/packages/catalog-ui-search/src/main/webapp/dev/component/card-guide/card-guide.view.js
```diff
@@ -39,6 +41,35 @@ module.exports = BaseGuideView.extend({
...
+                    description: 'example',
+                    id: 'example',
+                    title: 'example',
+                    url: 'https://www.google.com'
+                }],
+                distance: null,
+                hasThumbnail: false,
+                isResourceLocal: true,
+                metacard: {
+                    id: 'blah blah blah',
```
jlcsmith: ✏️ this is a nit, but I think we should clean up this example metacard. 
**id** and **validation-warnings** should be more representative
**resource-download-url** we typically use https://example.com for this sort of thing

ui/packages/catalog-ui-search/src/main/webapp/dev/component/card-guide/card-guide.view.js
```diff
@@ -39,6 +41,35 @@ module.exports = BaseGuideView.extend({
...
         this.workspaceExample.currentView.activateGridDisplay();
     },
+    showResult2Example() {
+        this.result2Example.show(new ResultItemView({
+            model: new QueryResultModel({
+                actions: [{
+                    description: 'example',
+                    id: 'example',
+                    title: 'example',
+                    url: 'https://www.google.com'
```
jlcsmith: ✏️ recommend https://example.com

ui/packages/catalog-ui-search/src/main/webapp/component/result-item/result-item.view.js
```diff
@@ -13,210 +13,316 @@
...
+    template(data) {
+        return (
+            <React.Fragment>
+                <div className="result-container" data-metacard-id={data.id}
+                data-query-id={data.metacard.queryId}>
+                <div className="container-indicator">
+                </div>
+                <div className="container-content">
+                    <div className="content-header">
+                        <span className="header-icon fa fa-history" title="Type: Revision" data-help="Indicates the type of result
```
jlcsmith: ❗️ the help text for all of these looks like incorrect - accidental copy/paste?

ui/packages/catalog-ui-search/src/main/webapp/component/result-item/result-item.view.js
```diff
@@ -13,210 +13,316 @@
...
+                                </span> : ''
+                            }
+                            {
+                                !data.hasError && data.hasWarning ?
+                                <span className="fa fa-exclamation-triangle resultWarning" title="Has validation warnings." data-help="Indicates the given result has a validation warning.
+                                See the 'Quality' tab of the result for more details.">
+                                </span> : ''
+                            }
+                        </div>
+                        <button className="result-download fa fa-download is-button is-neutral" title="Downloads the results associated product directly to your machine." data-help="Downloads
```
jlcsmith: ✏️ results s/b **result's**.  Also I thought we were using "resource" instead of "product"

ui/packages/catalog-ui-search/src/main/webapp/component/result-item/result-item.view.js
mackncheesiest: It's at least consistent with how they were before
https://github.com/codice/ddf/pull/3489/files/9c6e46023f51cff34d4234ebd9bfd1b6b18f168f#diff-bf0dcbdee8a22ca2b57876a2fa07e3a0L20

They could probably be better, but it's not a regression

jlcsmith: I'll downgrade my comment to a nit since it was existing

andrewkfiedler:
build now

PR #3550, opened 2018-08-03
DDF-3847 Adjust Jenkinsfile pipeline stages to separate full build stages

mcalcote:
#### What does this PR do?
The Full Build stage has now been split into two separate stages. The first stage runs all tests and analysis except the itests. The second stage just runs the itests. In the case of windows where the agent must be secured, a quick build precedes the itest run.

#### Who is reviewing it? 
@oconnormi @clockard @shaundmorris @mackncheesiest @LinkMJB

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,16 +79,43 @@ pipeline {
                 }
             }
         }
-        // The full build will be run against all regular branches
-        stage('Full Build') {
+        stage('Full Build Except Itests') {
             when { expression { env.CHANGE_ID == null } }
             parallel {
                 stage ('Linux') {
                     steps {
-                        timeout(time: 3, unit: 'HOURS') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -79,17 +79,15 @@ pipeline {
...
         }
-        // The full build will be run against all regular branches
-        stage('Full Build') {
+        stage('Full Build Except Itests') {
             when { expression { env.CHANGE_ID == null } }
             parallel {
                 stage ('Linux') {
                     steps {
-                        timeout(time: 3, unit: 'HOURS') {
+                        timeout(time: 2, unit: 'HOURS') {
```
shaundmorris: ❓ Can we document why we're making this change (down from 3 to 2) somewhere? Not sure if a comment makes the most sense

mcalcote: I can't think of a better place for it, so I will put it it in the file for now.

mcalcote:
This file was validated against the local jenkins validate endpoint.
image

clockard:
Sonarqube and CI builds do not apply to this change. Skipping.

PR #3551, opened 2018-08-03
DDF-4032 Added code to convert single values to arrays.

BernardIgiri:
#### What does this PR do?
Converts single values to arrays thus fixing the bug where flat arrays cause dropmenues to crash DDF.
#### Who is reviewing it? 
@djblue @mackncheesiest @GabrielFabian @troymohl @oconnormi
@andrewkfiedler
@bdeining
[DDF-4032](https://codice.atlassian.net/browse/DDF-4032)
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


ui/packages/catalog-ui-search/src/main/webapp/component/dropdown/dropdown.view.js
```diff
@@ -25,6 +25,11 @@ define([
     'behaviors/button.behavior'
 ], function (Marionette, _, $, CustomElements, DropdownCompanionView, DropdownModel, template, SelectView) {
 
+    function getArrayFromValue(value) {
+      value = value || []
+      return Array.isArray(value) ? value : [value];
+    }
+
```
Kjames5269: ❓ Why not make a js file that exports this function?

ui/packages/catalog-ui-search/src/main/webapp/js/Common.js
```diff
@@ -265,6 +265,15 @@ define([
...
             ]);
+        },
+        getArrayFromValue: function(value) {
+          if (value === undefined || value === '' || value === null) {
+            return []
+          } else if (Array.isArray(value)) {
+            return value
+          } else {
+            return [value]
+          }
```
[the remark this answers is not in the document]
BernardIgiri: This checks if it's an array, if it is an array already then it returns that array. Elsewise it makes sure that values which should evaluate as empty become empty arrays while scalar values are wrapped in an array.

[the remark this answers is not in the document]
samuelechu: Will 0 ever be passed in? My understanding was that 0 would be of value '0' in dropdowns

ui/packages/catalog-ui-search/src/main/webapp/js/Common.js
```diff
@@ -265,6 +265,15 @@ define([
                 this.wrapMapCoordinates(lon, [-180, 180]),
                 this.wrapMapCoordinates(lat, [-90, 90])
             ]);
+        },
+        getArrayFromValue: function(value) {
+          if (value === undefined || value === '' || value === null) {
+            return []
+          } else if (Array.isArray(value)) {
```
leo-sakh: ❓ can't these two conditions be collapsed to something like `return (_.isArray(value) && value) || [value]`?

ui/packages/catalog-ui-search/src/main/webapp/component/dropdown/dropdown.view.js
```diff
@@ -101,7 +102,7 @@ define([
             return this.el;
         },
         determineSelections: function () {
-            var values = this.model.get('value');
+            var values = Common.getArrayFromValue(this.model.get('value'));
```
leo-sakh: ❓ I thought we're trying to use es6 syntax more, no? This might be better served with a `const`

ui/packages/catalog-ui-search/src/main/webapp/js/Common.js
BernardIgiri: Then you could get `[null]` and `[undefined]` both of which would be bad.

ui/packages/catalog-ui-search/src/main/webapp/js/Common.js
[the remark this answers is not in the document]
BernardIgiri: No idea but better safe than sorry.

ui/packages/catalog-ui-search/src/main/webapp/js/Common.js
leo-sakh: ...that would be caught by the first predicate. I'm talking about the last two

Variadicism: Though it does work, I personally find the ternary operator more readable than `&&` and `||` in the context of returning non-boolean values:
```
return _.isArray(value) ? value : [value]
```

I also don't know where we fall on `_.isArray` versus `Array.isArray`.

ui/packages/catalog-ui-search/src/main/webapp/js/Common.js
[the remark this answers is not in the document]
Variadicism: This function is in `Common`; it is not specific to dropdowns. `return value ? [value] : []` would not provide the expected functionality because `0` and `false` are expected to become `[0]` and `[false]`, respectively, while their falsy values would cause the ternary operation above to return `[]` for both.

BernardIgiri:
build now

